PR #29, opened 2023-07-28
Fix: indexed tmp files in unit tests

678098:
Allows to run test cases from this file in parallel


hallfox (approved):
Thanks, looks good!

Issue #27, opened 2023-07-28
Can't download bmqbrkr image from docker


robotnoises:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I'm attempting to start the single-node test project but when I run `docker-compose` I get the following:

```
docker compose -f docker/single-node/docker-compose.yaml up --build -d
[+] Running 0/1
 ⠿ bmqtool Error                                                                                                                                
Error response from daemon: pull access denied for bmqbrkr, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

### Expected Behavior

I expected the docker container(s) to start up.

### Steps To Reproduce

```
> git clone git@github.com:bloomberg/blazingmq.git
> cd blazingmq
> docker login
> docker compose -f docker/single-node/docker-compose.yaml up --build -d
```

### BlazingMQ Version

0.0.0

### Anything else?

I checked out docker hub but I don't see the image `bmqbrkr` anywhere.

jll63:
The example does not download a `bmqbrkr` image - only [a stock ubuntu image](https://github.com/bloomberg/blazingmq/blob/main/docker/Dockerfile#L1). It does, though, create a local `bmqbrkr` image.

Please try again, using the exact command from the documentation:

```
docker-compose -f docker/single-node/docker-compose.yaml up --detach
```

robotnoises:
That worked, thanks!

PR #30, opened 2023-07-28
update to work with docker compose v2

willhoy:
*Issue number of the reported bug or feature request: #27 

**Describe your changes**
add a build: . to prevent `docker compose` (v2) trying to pull the bmqtool image from dockerhub

**Testing performed**

**Additional context**

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -55,6 +55,22 @@ jobs:
           cd cmake.bld/Linux
           ctest -E mwcsys_executil.t --output-on-failure
 
+  docker_single_node:
+    name: “Check whether docker single-node workflow builds correctly.”
+    runs-on: ubuntu-latest
+    steps:
+      - users: actions/checkout@v3
+      - name: Docker Workflow
+        run: docker compose -f docker/single-node/docker-compose.yaml up --build -d
+
+  docker_cluster:
+    name: “Check whether docker cluster workflow builds correctly.”
+    runs-on: ubuntu-latest
+    steps:
+      - users: actions/checkout@v3
+      - name: Docker Workflow
+        run: docker compose -f docker/cluster/docker-compose.yaml up --build -d
+
   Documentation:
     name: "Documentation"
     runs-on: ubuntu-latest
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -55,22 +55,6 @@ jobs:
           cd cmake.bld/Linux
           ctest -E mwcsys_executil.t --output-on-failure
 
-  docker_single_node:
-    name: “Check whether docker single-node workflow builds correctly.”
-    runs-on: ubuntu-latest
-    steps:
-      - users: actions/checkout@v3
-      - name: Docker Workflow
-        run: docker compose -f docker/single-node/docker-compose.yaml up --build -d
-
-  docker_cluster:
-    name: “Check whether docker cluster workflow builds correctly.”
-    runs-on: ubuntu-latest
-    steps:
-      - users: actions/checkout@v3
-      - name: Docker Workflow
-        run: docker compose -f docker/cluster/docker-compose.yaml up --build -d
-
   Documentation:
     name: "Documentation"
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


pniedzielski:
@willhoy, between when this PR was submitted and now, we added a requirement for contributors to agree to the DCO, meaning all commits need to have a `Signed-off-by` line at the bottom.  Unfortunately this prevents the rest of CI from running (ugh).

Could you rebase your patches and commit them with the `-s` flag to add a `Signed-off-by` line to each that you made?

.github/workflows/build.yaml
```diff
@@ -55,6 +55,22 @@ jobs:
           cd cmake.bld/Linux
           ctest -E mwcsys_executil.t --output-on-failure
 
+  docker_single_node:
+    name: “Check whether docker single-node workflow builds correctly.”
+    runs-on: ubuntu-latest
+    steps:
+      - users: actions/checkout@v3
+      - name: Docker Workflow
+        run: docker compose -f docker/single-node/docker-compose.yaml up --build -d
```
678098, line 64: We already added a check for a docker image build via single-node compose file, it's in the bottom of the yaml file:

https://github.com/bloomberg/blazingmq/blob/c8acf1b840fcd9e907d4404f43e5e687edd660f5/.github/workflows/build.yaml#L94-L100

Probably we need to unify it, your variant is more advanced so you can just remove the existing one at the bottom.

678098 (changes requested):
Some comments

.github/workflows/build.yaml
pniedzielski: To move this PR, I've done this modification.

docker/Dockerfile
```diff
@@ -1,10 +1,10 @@
 FROM docker.io/ubuntu:22.04 AS builder
 
 # Set up CA certificates first before installing other dependencies
-RUN apt update && \
-    apt -y install ca-certificates
+RUN apt-get update && \
+    apt-get -y install ca-certificates
```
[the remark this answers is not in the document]
pniedzielski: I'll make a mini-PR with this change separately.

.github/workflows/build.yaml
```diff
@@ -92,9 +92,11 @@ jobs:
         uses: ludeeus/action-shellcheck@master
 
   docker-build-check:
-    name: Docker image build check
+    name: “Check whether docker single-node and cluster workflows build correctly.”
```
678098, line 95: Very unusual quotes here

.github/workflows/build.yaml
```diff
@@ -55,6 +55,22 @@ jobs:
...
+      - name: Docker Workflow
+        run: docker compose -f docker/single-node/docker-compose.yaml up --build -d
+
+  docker_cluster:
+    name: “Check whether docker cluster workflow builds correctly.”
+    runs-on: ubuntu-latest
+    steps:
+      - users: actions/checkout@v3
+      - name: Docker Workflow
+        run: docker compose -f docker/cluster/docker-compose.yaml up --build -d
```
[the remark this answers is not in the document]
pniedzielski: I've pushed a commit that does this, and I'm going to double-check that this indeed does cache a built image between steps.  Once that finishes, I'll rebase to squash some of my commits and bring us up to `main`.  This should also fix one of the CI errors that cropped up by integrating the shell linter fixes.  Once I'm sure CI passes I will merge.

.github/workflows/build.yaml
pniedzielski: I noticed that too, Emacs thinking YAML should have smartquotes....waiting to push the patch to avoid interrupting CI.

.github/workflows/build.yaml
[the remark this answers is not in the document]
pniedzielski: Indeed it does cache the built image, as the cluster step takes only one second: https://github.com/bloomberg/blazingmq/actions/runs/6017586321/job/16324077234?pr=30.  Good idea!

pniedzielski:
@678098 I think we should be safe to merge this now, but GitHub isn't letting me do it automatically.  I'll open a PR with `Dockerfile` change you suggested tomorrow.

678098 (approved):

PR #40, opened 2023-07-29
[HOLD] CI: disable parallel UTs run

678098:


678098:
UTs which use shared resources were fixed, so tmp single-thread run of UTs is not necessary

PR #51, opened 2023-07-31
Add clang formatting checks for src/ files

willhoy:
**Describe your changes**
Adding a formatting action to preserve the clang formatting of the codebase


.github/workflows/build.yaml
```diff
@@ -55,18 +55,30 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
       - name: Set up dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -qy clang-format git doxygen
-      - name: Lint sources
-        run: |
-          git clang-format
+          sudo apt-get install -qy git doxygen
```
willhoy, line 66:
```suggestion
          sudo apt-get install -qy doxygen
```

willhoy: I doubt this is necessary?

hallfox:
I see that it hit a generated file. I think we should use the `exclude-regex` option in the clang-format checks to exclude mqbcfg_messages.(h|cpp)

hallfox:
Going by other generated files in this repo I think we'll need to exclude .*_(messages|values|cpp03).(h|cpp)

Issue #36, opened 2023-07-28
Remove unused header mwcsys_executil.h from the codebase


678098:
### Is there an existing proposal for this?

- [X] I have searched the existing proposals

### Is your feature request related to a problem?

If you search for `ExecUtil` usage, you can find some files with unused `#include <mwcsys_executil.h>`: 
https://github.com/search?q=repo%3Abloomberg%2Fblazingmq%20ExecUtil&type=code

### Describe the solution you'd like

Will be good to remove these unused includes where possible, while keeping files `mwcsys_executil.h`, `mwcsys_executil.cpp`, `mwcsys_executil.t.cpp`. These files might be used by existing library's clients.

parkma99:
I would like to take it.

SaketKaswa20:
I would love to take a try on it.

quarter-note:
Hi @SaketKaswa20, since @parkma99 expressed interest first, I think it would be fair for them to take a shot at it :).  @parkma99 Please let us know if you'd like to proceed.

parkma99:
Hello @quarter-note @SaketKaswa20 , #45  My machine is M2 Mac,  I find it is hard to build this project,so I give up this issue,

SaketKaswa20:
I can surely try, please assign me.

678098:
> I can surely try, please assign me.

Assigned, feel free to ask if you have any questions

SaketKaswa20:
Hey, I searched for `ExecUtil` and found some files. Read the code where `#include<mwcsys_executil.h>`, but I am a bit confused on how to find which is useful or not. It would be great, if you can guide me or refer some documentation. I am eager to solve this issue.

678098:
@SaketKaswa20 
There are 2 base files which are interesting for this task:
https://github.com/bloomberg/blazingmq/blob/main/src/groups/mwc/mwcsys/mwcsys_executil.h
https://github.com/bloomberg/blazingmq/blob/main/src/groups/mwc/mwcsys/mwcsys_executil.cpp

In general, when you want to remove some includes, you need to address several things:
- What entities (classes, functions, variables) declared/defined in interface. From the header at the top you can see that only `struct ExecUtil` declared, so, to the question of "is it useful or not" - you can check if this `ExecUtil` used.
- What includes are added recursively and will be lost. Here you can see the `bsl_string.h` include only. Simply trying to build will show if some includes were lost, but I don't think there will be a problem here.
[...]

PR #53, opened 2023-08-01
fix: Update mwcsys_executil.cpp

SaketKaswa20:
Removed unused header mwcsys_executil.h

*Issue number of the reported bug or feature request: #36 *

**Describe your changes**
Tried to remove the unused <#include mwcsys_executil.h> from the codebase

**Testing performed**

**Additional context**

### DIFF
--- a/src/groups/mwc/mwcsys/mwcsys_executil.cpp
+++ b/src/groups/mwc/mwcsys/mwcsys_executil.cpp
@@ -14,7 +14,6 @@
 // limitations under the License.
 
 // mwcsys_executil.cpp                                                -*-C++-*-
-#include <mwcsys_executil.h>
 
 #include <mwcscm_version.h>
 


678098:
@SaketKaswa20 
Now you can see PR being built with error "/home/runner/work/blazingmq/blazingmq/src/groups/mwc/mwcsys/mwcsys_executil.cpp:45:5: error: ‘ExecUtil’ has not been declared"

This happened because the removed header "linked" together 'mwcsys_executil.h' containing declarations and 'mwcsys_executil.cpp' containing definitions. More strictly, include directive instructs preprocessor to copy contents of the entire 'mwcsys_executil.h' file into the corresponding .cpp file on a preprocessor stage of compilation.
You can read more about here:
https://en.cppreference.com/w/cpp/preprocessor/include

In summary, removing this concrete include is not correct, we want to leave 'mwcsys_executil' files unchanged. You still need to check these files as I mentioned in the issue before removing the header from other files.

There is a link in the issue description with repo search which will help to find other files with this include: https://github.com/search?q=repo%3Abloomberg%2Fblazingmq%20ExecUtil&type=code

PR #34, opened 2023-07-28
build-ubuntu script

jll63:
**Testing performed**
Describe the testing you have performed to ensure that the bug has been addressed, or that the new feature works as planned.

I created a new WSL container from the Ubuntu 22.04.2 LTS image, installed the dependencies as specified at the top of the script, and I was able to build and run the broker.

PR #54, opened 2023-08-01
Add build script

kaikulimu:
Script for building BlazingMQ locally on a MacOS

### DIFF
--- a/bin/build-darwin.sh
+++ b/bin/build-darwin.sh
@@ -1,0 +1,111 @@
+#!/usr/bin/env bash
+
+# This script builds BlazingMQ and all of its dependencies.
+#
+# Required prerequisites:
+# - Clang
+# - CMake
+# - git
+# - Homebrew
+# - ninja
+# - Perl
+# - pkg-config
+# - Python3
+
+
+set -e
+set -u
+[ -z $BASH ] || shopt -s expand_aliases
+
+
+# :: Set some initial constants :::::::::::::::::::::::::::::::::::::::::::::::
+DIR_ROOT="${DIR_ROOT:-`pwd`}"
+
+DIR_THIRDPARTY="${DIR_ROOT}/thirdparty"
+mkdir -p "${DIR_THIRDPARTY}"
+
+DIR_BUILD="${DIR_BUILD:-${DIR_ROOT}/build}"
+mkdir -p "${DIR_BUILD}"
+
+
+# :: Clone dependencies :::::::::::::::::::::::::::::::::::::::::::::::::::::::
+if [ ! -d "${DIR_THIRDPARTY}/bde-tools" ]; then
+    git clone https://github.com/bloomberg/bde-tools "${DIR_THIRDPARTY}/bde-tools"
+fi
+if [ ! -d "${DIR_THIRDPARTY}/bde" ]; then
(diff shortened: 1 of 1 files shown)


quarter-note:
@kaikulimu Can you add some description for this PR.

678098:
This build script works on my M2, `bmqbrkr.tsk` is built successfully and surprisingly fast.

I have a clear Mac OS installation so I had to install `homebrew`, and then `cmake` and `ninja` with it to make it work.

bin/build-darwin.sh
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env bash
+
+# This script buildd BlazingMQ and all of its dependencies.
```
678098, line 3:
```suggestion
# This script builds BlazingMQ and all of its dependencies.
```

bin/build-darwin.sh
```diff
@@ -0,0 +1,82 @@
...
+pushd ${DIR_THIRDPARTY}/ntf-core
+./configure --prefix "${DIR_ROOT}" --output "${DIR_BUILD}/ntf"
+make
+make install
+popd
+
+# Build other dependencies
+brew install flex bison google-benchmark zlib
+
+
```
678098, line 67: Something like this will ensure work with `.local` hostname, but might be good to explicitly ask the user if they want to update `hosts`
```suggestion
SELF_HOST_ADDRESS="127.0.0.1       `hostname`"
if ! (grep -q "$SELF_HOST_ADDRESS" /etc/hosts > /dev/null);
then
    sudo bash -c "echo $SELF_HOST_ADDRESS >> /etc/hosts"
fi
```

quarter-note:
We should also update README with a 'Build Instructions' section like [ntf](https://github.com/bloomberg/ntf-core#build-instructions) when we merge this and the other build PRs.

kaikulimu:
> We should also update README with a 'Build Instructions' section like [ntf](https://github.com/bloomberg/ntf-core#build-instructions) when we merge this and the other build PRs.

There is a README in #34

bin/build-darwin.sh
kaikulimu: d

bin/build-darwin.sh
kaikulimu: This is updating `/etc/hosts` and using `sudo`

Might be better to explicitly ask the user

678098: Let's just propose then, so user decides:

```suggestion

SELF_HOST_ADDRESS="127.0.0.1       `hostname`"
if ! (grep -q "$SELF_HOST_ADDRESS" /etc/hosts > /dev/null);
then
    echo "Warning: self hostname `hostname` not found in /etc/hosts"
    echo "It might be necessary to add it manually to launch bmqbrkr:"
    echo "sudo bash -c \"echo \\\"$SELF_HOST_ADDRESS\\\" >> /etc/hosts\""
fi

```

The output:
```
% ./tmp.sh
Warning: self hostname MYHOST not found in /etc/hosts
It might be necessary to add it manually to launch bmqbrkr:
sudo bash -c "echo \"127.0.0.1       MYHOST\" >> /etc/hosts"
```

jll63 (approved):

PR #59, opened 2023-08-03
Fix: thread race in bmqimp_brokersession.t.cpp

678098:
* Fixed incorrect check for empty write calls after async Configure (must be non-empty after a small timeout)
* Fixed incorrect absolute timestamp for timedPopFront when waiting for events

### DIFF
--- a/src/groups/bmq/bmqimp/bmqimp_brokersession.t.cpp
+++ b/src/groups/bmq/bmqimp/bmqimp_brokersession.t.cpp
@@ -2242,7 +2242,6 @@ bool TestSession::waitForQueueRemoved(
 
     const bsls::TimeInterval expireAfter =
         bsls::SystemTime::nowRealtimeClock() + timeout;
-    mwcsys::Time::nowMonotonicClock() + timeout;
     while (session().lookupQueue(queue->uri()) != 0 &&
            bsls::SystemTime::nowRealtimeClock() < expireAfter) {
         bslmt::ThreadUtil::microSleep(k_TIME_SOURCE_STEP.totalMicroseconds());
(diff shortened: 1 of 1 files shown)


src/groups/bmq/bmqimp/bmqimp_brokersession.t.cpp
```diff
@@ -9795,11 +9794,10 @@ static void test54_distributedTrace()
             // with the given default timeout.
             buffer.resize(expectedNum);
             for (size_t i = 0; i < expectedNum; i++) {
-                ASSERT_EQ(
-                    events.timedPopFront(&buffer[i],
-                                         mwcsys::Time::nowMonotonicClock() +
-                                             bsls::TimeInterval(0.1)),
-                    0);
+                ASSERT_EQ(events.timedPopFront(&buffer[i],
+                                               bdlt::CurrentTime::now() +
```
sgalichkin, line 9798: Maybe it's minor, but better to ask.
`bdlt::CurrentTime::now()` is realtime or monotonic as `mwcsys::Time::nowMonotonicClock()`?
As far as I know, realtime in theory may jump forwards and backwards

sgalichkin (approved):
looks good, one question

src/groups/bmq/bmqimp/bmqimp_brokersession.t.cpp
678098: The default callback provides real time.
https://github.com/bloomberg/bde/blob/ace6c3d37a96ab625d113a32c79e5cc9f07c3122/groups/bdl/bdlt/bdlt_currenttime.cpp#L42C49-L42C49

I don't see usages of the setter `setCurrentTimeCallback` which might alter it.
Also in the depths of implementation of `deque` we can find usage of `bslml::Condition` with a default realtime clock.
https://github.com/bloomberg/bde/blob/ace6c3d37a96ab625d113a32c79e5cc9f07c3122/groups/bsl/bslmt/bslmt_condition.h#L198C45-L198C78

PR #75, opened 2023-08-08
CI: add docker image build check

678098:
Some TDD


678098:
Part of this PR which is already merged: https://github.com/bloomberg/blazingmq/pull/76

PR #80, opened 2023-08-11
Parse until end of line for bmqtool input

hallfox:
**Describe your changes**
For long enough inputs bmqtool would cut off certain lines at a fairly low character limit. Using the `getline` function lets us parse as much as we can allocate for.

### DIFF
--- a/src/applications/bmqtool/m_bmqtool_inpututil.cpp
+++ b/src/applications/bmqtool/m_bmqtool_inpututil.cpp
@@ -24,6 +24,7 @@
 #include <bdlb_tokenizer.h>
 #include <bsl_cstdlib.h>
 #include <bsl_iostream.h>
+#include <bsl_string.h>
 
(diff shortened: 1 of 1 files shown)


src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -34,12 +35,9 @@ namespace m_bmqtool {
 
 bool InputUtil::getLine(bsl::string* out)
 {
-    const int BUFFER_SIZE = 512;
-    char      buffer[BUFFER_SIZE];
     bsl::cout << "> " << bsl::flush;
```
678098, line 38: We could also add some ASSERT statements for args in this file which are missing to ptr arguments, to make it consistent with other parts of the codebase. But this could be done in a separate PR

678098 (approved):
LGTM!

Issue #96, opened 2023-08-23
`bmqa_mocksession`'s `expect_XYZ` functions have different `const` qualifiers from their corresponding `XYZ` functions in `bmqa_abstractsession`


pniedzielski:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The `expect_XYZ` functions in `bmqa_mocksession` (e.g., `expect_configureQueueSync`) are run using the `BMQA_EXPECT_CALL` macro from a mock, and are meant to parallel the functions `XYZ` (e.g., `configureQueueSync`) in `bmqa_abstractsession`.  At the moment, several of these functions `expect_XYZ` differ from their corresponding `XYZ` functions in the `const`ness of their parameters:

  * `expect_closeQueueSync(QueueId*, ...)`, but `closeQueueSync(const QueueId*, ...)`
  * `expect_closeQueueAsync(QueueId*, ...)`, but `closeQueueAsync(const QueueId*, ...)`
  * `expect_configureQueueSync(QueueId*, ...)`, but `configureQueueSync(const QueueId*, ...)`
  
As a result, it is not possible to use `BMQA_EXPECT_CALL` from a mock session that implements the `AbstractSession` interface.

### Expected Behavior

The arguments to `expect_XYZ` functions should match the arguments to `XYZ` functions in their `const`ness.

### Steps To Reproduce

.

### BlazingMQ Version

HEAD

PR #97, opened 2023-08-23
Make `bmqa_mocksession` match `bmqa_abstractsession` in `const`ness

pniedzielski:
The `expect_XYZ` functions in `bmqa_mocksession` (e.g., `expect_configureQueueSync`) are run using the `BMQA_EXPECT_CALL` macro from a mock, and are meant to parallel the functions `XYZ` (e.g., `configureQueueSync`) in `bmqa_abstractsession`.  At the moment, several of these functions `expect_XYZ` differ from their corresponding `XYZ` functions in the `const`ness of their parameters:

  * `expect_closeQueueSync(QueueId*, ...)`, but `closeQueueSync(const QueueId*, ...)`

  * `expect_closeQueueAsync(QueueId*, ...)`, but `closeQueueAsync(const QueueId*, ...)`

  * `expect_configureQueueSync(QueueId*, ...)`, but `configureQueueSync(const QueueId*, ...)`

As a result, it is not possible to use `BMQA_EXPECT_CALL` from a mock session that implements the `AbstractSession` interface.

This patch changes the arguments to each `expect_XYZ` functions to take the same cv-qualifiers as their corresponding arguments in each `XYZ` function.

Fixes: #96

### DIFF
--- a/src/groups/bmq/bmqa/bmqa_mocksession.cpp
+++ b/src/groups/bmq/bmqa/bmqa_mocksession.cpp
@@ -1194,9 +1194,10 @@ MockSession::expect_configureQueue(BSLS_ANNOTATION_UNUSED QueueId* queueId,
 }
 
 MockSession::Call&
-MockSession::expect_configureQueueSync(BSLS_ANNOTATION_UNUSED QueueId* queueId,
-                                       const bmqt::QueueOptions&       options,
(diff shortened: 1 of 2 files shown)


pniedzielski:
We will pursue a different way of resolving this; these changes break existing consumers of the BlazingMQ API.

pniedzielski:
Superseded by #99

PR #100, opened 2023-08-29
Add `DIR_INSTALL` variable to build scripts to ease installation of BlazingMQ and its dependencies

pniedzielski:
**Describe your changes**
The build scripts currently install the build dependencies right within the source tree of BlazingMQ.  Consumers of the BlazingMQ API may wish to install BlazingMQ and its dependencies to a separate source tree (perhaps `/usr/local` or some other local install root) for easier building.  This PR allows users to configure the install root correctly, using a new `DIR_INSTALL` variable.  In the process, it makes `DIR_ROOT` non-configurable, as its remaining uses *must* have been the current source directory.

**Testing performed**
We built BlazingMQ using `./bin/build-darwin.sh` and with `DIR_INSTALL` set to a separate local install tree.  Each of BlazingMQ's dependencies build and install properly (verified manually), and BlazingMQ builds using these installed built dependencies.  Then, when installed using `cmake --install build/blazingmq/` (where `$DIR_BUILD` is `build`), build artifacts from BlazingMQ are properly installed in the `DIR_INSTALL` tree.

**Additional context**
These patches together end up removing most instances of `DIR_ROOT` from the script, as they untangle two unrelated uses of `DIR_ROOT`.

The build scripts still do not install BlazingMQ, though they do install its dependencies; this still must be done manually with `cmake --install`.  This is a strange asymmetry.  This patch maintains the current behavior, but we may want to revisit this.

### DIFF
--- a/bin/build-darwin.sh
+++ b/bin/build-darwin.sh
@@ -25,7 +25,7 @@ if [ ! -f "$script_path" ] || [ "$(realpath "$0")" != "$(realpath "$script_path"
 fi
 
 # :: Set some initial constants :::::::::::::::::::::::::::::::::::::::::::::::
-DIR_ROOT="${DIR_ROOT:-$(pwd)}"
+DIR_ROOT="$(pwd)"
 
 DIR_THIRDPARTY="${DIR_ROOT}/thirdparty"
 mkdir -p "${DIR_THIRDPARTY}"
--- a/bin/build-darwin.sh
+++ b/bin/build-darwin.sh
@@ -33,6 +33,9 @@ mkdir -p "${DIR_THIRDPARTY}"
 DIR_BUILD="${DIR_BUILD:-${DIR_ROOT}/build}"
 mkdir -p "${DIR_BUILD}"
 
+DIR_INSTALL="${DIR_INSTALL:-${DIR_ROOT}}"
+mkdir -p "${DIR_INSTALL}"
+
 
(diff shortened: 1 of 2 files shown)


bin/build-darwin.sh
```diff
@@ -54,10 +57,10 @@ PATH="${DIR_THIRDPARTY}/bde-tools/bin:$PATH"
 
 if [ ! -e "${DIR_BUILD}/bde/.complete" ]; then
     pushd "${DIR_THIRDPARTY}/bde"
-    eval "$(bbs_build_env -u opt_64_cpp17 -b "${DIR_BUILD}/bde")"
-    bbs_build configure --prefix="${DIR_ROOT}"
-    bbs_build build --prefix="${DIR_ROOT}"
-    bbs_build --install_dir="/" --prefix="${DIR_ROOT}" install
+    eval "$(bbs_build_env -u opt_64_cpp17 -b "${DIR_BUILD}/bde" -i "${DIR_INSTALL}")"
```
kaikulimu, line 60: Did you try running the script? Specifically, I want to make sure that the `-i` option works as expected

pniedzielski: Yes---at least on Darwin:

    rm -rf build thirdparty ~/install-root
    export DIR_INSTALL=~/install-root
    ./bin/build-darwin.sh
    ls ~/install-root/lib
    
shows `libbal.a`, `libbbl.a`, `libntc.a` etc.

PR #101, opened 2023-08-30
Collapse two `RUN` layers in the `Dockerfile`

pniedzielski:
**Describe your changes**
We currently have two adjacent `RUN` statements in the `Dockerfile` which both call `apt-get`.  The first of them updates apt’s package index and installs the `ca-certificates` package, while the second installs the BlazingMQ build dependencies and cleans apt’s package lists.  This results in a layer of non-trivial size that we would like to avoid.

This PR makes the minimal change to combine these two layers into a single layer, by unifying the two `RUN` statements.

**Testing performed**
`docker-compose -f docker/single-node/docker-compose.yaml up` and ensure that a single `RUN` statement is executed, and the broker builds and runs.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,9 +2,8 @@ FROM docker.io/ubuntu:22.04 AS builder
 
 # Set up CA certificates first before installing other dependencies
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


678098 (approved):
LGTM

PR #102, opened 2023-08-30
Add a build script for manylinux/musllinux

hallfox:
Meant to be used with the [manylinux containers ](https://github.com/pypa/manylinux) used for building Python extensions. We're targeting a baseline support of manylinux2014 and musllinux_1_1.

### DIFF
--- a/bin/build-manylinux.sh
+++ b/bin/build-manylinux.sh
@@ -1,0 +1,131 @@
+#!/usr/bin/env bash
+
+# This script builds BlazingMQ and all of its dependencies.
+#
+# Before running this script, install following prerequisites, if not present
+# yet, by copy-and-pasting the commands between `<<PREREQUISITES` and
+# `PREREQUISITES` below:
+                                                    # shellcheck disable=SC2188
+<<PREREQUISITES
+sudo yum install -y \
+    gdb \
+    curl \
+    pkgconfig \
+    ninja-build \
+    flex-devel \
+    zlib-devel \
+    m4
+PREREQUISITES
+
+set -e
+set -u
+[ -z "$BASH" ] || shopt -s expand_aliases
+
+# :: Set some initial constants :::::::::::::::::::::::::::::::::::::::::::::::
+DIR_ROOT="${DIR_ROOT:-$(pwd)}"
+
+DIR_THIRDPARTY="${DIR_ROOT}/thirdparty"
+mkdir -p "${DIR_THIRDPARTY}"
+
+DIR_BUILD="${DIR_BUILD:-${DIR_ROOT}/build}"
+mkdir -p "${DIR_BUILD}"
+
+DIR_INSTALL="${DIR_INSTALL:-${DIR_ROOT}}"
+mkdir -p "${DIR_INSTALL}"
+
+script_path="bin/$(basename "$0")"
+
+if [ ! -f "$script_path" ] || [ "$(realpath "$0")" != "$(realpath "$script_path")" ]; then
+    echo 'This script must be run from the root of the BlazingMQ repository.'
+    exit 1
+fi
+
+# :: Clone dependencies :::::::::::::::::::::::::::::::::::::::::::::::::::::::
+
+if [ ! -d "${DIR_THIRDPARTY}/bde-tools" ]; then
+    git clone https://github.com/bloomberg/bde-tools "${DIR_THIRDPARTY}/bde-tools"
+fi
+if [ ! -d "${DIR_THIRDPARTY}/bde" ]; then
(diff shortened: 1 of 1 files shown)


bin/build-manylinux.sh
```diff
@@ -0,0 +1,131 @@
...
+sudo yum install -y \
+    gdb \
+    curl \
+    pkgconfig \
+    ninja-build \
+    flex-devel \
+    zlib-devel \
+    m4
+PREREQUISITES
+                                                    # shellcheck disable=SC2188
```
pniedzielski, line 18: This line should go before the `<<PREREQUISITES` to stop the shellcheck warning.

bin/build-manylinux.sh
```diff
@@ -0,0 +1,131 @@
...
+
+set -e
+set -u
+[ -z "$BASH" ] || shopt -s expand_aliases
+
+# :: Set some initial constants :::::::::::::::::::::::::::::::::::::::::::::::
+DIR_ROOT="${DIR_ROOT:-$(pwd)}"
```
pniedzielski, line 25: Along the lines of [commit `179be2e`](https://github.com/bloomberg/blazingmq/commit/179be2ee766f07c6b0d2ca28f40fb7c0e11e921c), we probably don't want this to be user-configurable.

bin/build-manylinux.sh
```diff
@@ -0,0 +1,131 @@
...
+fi
+if [ ! -d "${DIR_THIRDPARTY}/google-benchmark" ]; then
+    git clone --depth 1 https://github.com/google/benchmark.git "${DIR_THIRDPARTY}/google-benchmark"
+fi
+if [ ! -d "${DIR_BUILD}/bison" ]; then
+    mkdir -p "${DIR_BUILD}/bison"
+    curl https://ftp.gnu.org/gnu/bison/bison-3.8.2.tar.xz | tar -Jx -C "${DIR_BUILD}/bison" --strip-components 1
+fi
+
+mkdir -p "${DIR_THIRDPARTY}/cmake"
```
pniedzielski, line 62: Should this be surrounded by something like line 68 and 76, below?

bin/build-manylinux.sh
```diff
@@ -0,0 +1,131 @@
...
+        -DCMAKE_INSTALL_LIBDIR="lib64" \
+        -DCMAKE_INSTALL_PREFIX="${DIR_INSTALL}" \
+        -DCMAKE_MODULE_PATH="${DIR_ROOT}" \
+        -DCMAKE_PREFIX_PATH="${DIR_THIRDPARTY}/bde-tools/BdeBuildSystem" \
+        -DCMAKE_TOOLCHAIN_FILE="${DIR_THIRDPARTY}/bde-tools/BdeBuildSystem/toolchains/linux/gcc-default.cmake" \
+        -G "Unix Makefiles")
+    PKG_CONFIG_PATH="/usr/lib64/pkgconfig:${DIR_INSTALL}/lib64/pkgconfig" \
+    cmake -B "${DIR_BUILD}/blazingmq" -S "." "${CMAKE_OPTIONS[@]}"
+    cmake --build "${DIR_BUILD}/blazingmq" -j 16 --target bmq
+    cmake --install "${DIR_BUILD}/blazingmq" --component mwc-all
```
pniedzielski, line 128: This is an asymmetry with the `build-ubuntu.sh` and `build-darwin.sh` scripts.  Should those scripts also install BlazingMQ?  Given that this script is being used to build Python extensions rather than intended for end users, perhaps that's fine, but then it's worth making this script less obviously parallel to the other `build-*.sh` scripts (different name?  comment at the top describing the intended usage?).

bin/build-manylinux.sh
```diff
@@ -0,0 +1,131 @@
...
+#
+# Before running this script, install following prerequisites, if not present
+# yet, by copy-and-pasting the commands between `<<PREREQUISITES` and
+# `PREREQUISITES` below:
+<<PREREQUISITES
+sudo yum install -y \
+    gdb \
+    curl \
+    pkgconfig \
+    ninja-build \
```
pniedzielski, line 13: It's not clear to me where we use this dependency.  Is it BDE?

We could potentially use it with Google Benchmark and BlazingMQ itself as well, given both are CMake projects.

pniedzielski (changes requested):
If this is only being used to build the Python extensions, does it belong in this repository, or with the Python SDK?  Regardless, it might be a good idea to differentiate this from `build-ubuntu.sh` and `build-darwin.sh`, which are intended for users of BlazingMQ to use directly.

PR #105, opened 2023-09-06
Enhance bmqt_queueoptions.t subscription test with subscription expression validation

alexander-e1off:
At the moment, `bmqt_queueoptions.t` subscription test (`test4_subscriptionsTest`) tests only valid subscription expression.

This test is enhanced to check that API behaves properly when invalid subscription expression is passed.

Also TEST_EPILOG was changed to `mwctst::TestHelper::e_CHECK_GBL_ALLOC` to avoid test failure due to `mwcu::MemOutStream` usage in `QueueOptions::addOrUpdateSubscription` in case of expression validation failure.

### DIFF
--- a/src/groups/bmq/bmqt/bmqt_queueoptions.t.cpp
+++ b/src/groups/bmq/bmqt/bmqt_queueoptions.t.cpp
@@ -235,6 +235,21 @@ static void test4_subscriptionsTest()
         ASSERT(handles.emplace(handle).second);
     }
 
+    // Subscription expression validation
+    {
+        bmqt::Subscription       subscription;
+        bmqt::SubscriptionHandle handle(bmqt::CorrelationId::autoValue());
+
+        const bmqt::SubscriptionExpression expression(
+            "0invalid",
+            bmqt::SubscriptionExpression::e_VERSION_1);
+
+        subscription.setExpression(expression);
+        bsl::string error;
+        ASSERT(!obj.addOrUpdateSubscription(&error, handle, subscription));
+        ASSERT(!error.empty());
+    }
+
     bmqt::QueueOptions::SubscriptionsSnapshot snapshot(s_allocator_p);
     obj.loadSubscriptions(&snapshot);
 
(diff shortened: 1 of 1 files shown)


src/groups/bmq/bmqt/bmqt_queueoptions.t.cpp
```diff
@@ -235,6 +235,21 @@ static void test4_subscriptionsTest()
...
+    {
+        bmqt::Subscription       subscription;
+        bmqt::SubscriptionHandle handle(bmqt::CorrelationId::autoValue());
+
+        const bmqt::SubscriptionExpression expression(
+            "0invalid",
+            bmqt::SubscriptionExpression::e_VERSION_1);
+
+        subscription.setExpression(expression);
+        bsl::string error;
```
678098, line 248: Could you specify this allocator for string and return the previous check?
```suggestion
        bsl::string error(s_allocator_p);
```

We use this allocator during `bmqt::QueueOptions` creation, later `s_allocator_p` used for `MemOutStream`. And then we assign the memory block allocated with this allocator to this string which use a different allocator.

alexander-e1off: It fixed 1st leaking block, but still there is a problem:
```blazingmq/src/groups/bmq/bmqt/bmqt_queueoptions.t.cpp(294): '_defAlloc.numBlocksTotal()' (1) == '0' (0)    (failed)```
 I noticed that call of `context.lastErrorMessage()` in `QueueOptions::addOrUpdateSubscription` also caused of 1 block leaking.

678098: There is another string in this test without specified allocator:
https://github.com/bloomberg/blazingmq/blob/4567954bfa33cfae9a0d7c11346f94561687e735/src/groups/bmq/bmqt/bmqt_queueoptions.t.cpp#L230

678098: We can probably merge it if it's not working, I will take a look later then

alexander-e1off: Sure, but it was before my changes and it actually does not allocate any memory because no error happens.

678098: I think the allocator check has some "threshold" when to trigger, number of blocks for example.

678098 (approved):

Issue #106, opened 2023-09-07
Instructions for local Mac development of Just the Docs documentation is unclear


pniedzielski:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The `docs/README.md` file describes how to do local development of our Just the Docs documentation website on a Mac system, but has some stray references to tools that most users will not have access to.

### Expected Behavior

Anyone should be able to follow the steps for local Mac development on our Just the Docs documentation, so they can easily submit patches.

### Steps To Reproduce

.

### BlazingMQ Version

HEAD

PR #107, opened 2023-09-07
Fixup README for Just the Docs documentation website

pniedzielski:
*Issue number of the reported bug or feature request:* #106

**Describe your changes**
These two patches improve the Just the Docs documentation website README, by clarifying the license notice for our use of the Just the Docs Jekyll theme and by removing stray references to tools that most users may not have access to.

**Testing performed**
As this is just the `README`, I ensured that the Markdown file renders correctly, and that all tools referenced are available to most end users.


docs/README.md
```diff
@@ -4,15 +4,21 @@ Use [Just the Docs](https://github.com/just-the-docs/just-the-docs) for document
 
 ## License
 
-Uses [Just the Docs](https://github.com/just-the-docs/just-the-docs)
+In order to build the documentation, we use the [Just the
+Docs](https://github.com/just-the-docs/just-the-docs) theme for Jekyll, which
+is licensed under the [MIT license](LICENSE.just-the-docs.txt).  Please see
+that file for the theme’s copyright and permission notices.
```
kaikulimu, line 10: Is it weird that `LICENSE.just-the-docs.txt` is under the `docs` directory instead of the `licenses` directory?

kaikulimu: If we move it to the `licenses` directory, the file should be renamed `LICENSE-just-the-docs.txt` to be consistent with similar files.

kaikulimu (commented):
Minor comments

docs/README.md
pniedzielski: I'll double-check to see whether we link to it from anywhere in the documentation (and thus, whether it needs to be where it is), and if not, I'll move it to `licenses` as you suggest.

pniedzielski: Moved and renamed the license, changed the link to it in the documentation `README`, and removed the old path from the Jekyll excludes.

kaikulimu (approved):
LGTM

PR #111, opened 2023-09-14
Publish a broker image on release

hallfox:
**Describe your changes**
Build and publish an image when a release tag is created.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,41 @@
+name: Publish
+
+on:
+  push:
+    # Publish `v*` tags as releases.
+    tags:
+    - v*
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+    steps:
+      - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,42 @@
+name: Publish
+
+on:
+  push:
+    # Publish `v*` tags as releases.
+    tags:
+    - v*
+  pull_request:
```
pniedzielski, line 8: Should we do this on pull requests?

pniedzielski (dismissed):
One comment above.  If I understand [Github Actions `on`](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow) right, could you change?  Otherwise lgtm

.github/workflows/publish.yaml
hallfox: We may want to consider pushing an image on push to main, but not on every pull request. For now I'm considering tagging only versions we deem releasable, which will be noted by a tag.

pniedzielski (approved):

PR #112, opened 2023-09-14
Add support for building with vcpkg

hallfox:
**Describe your changes**
[vcpkg](https://vcpkg.io/en/) is a useful tool for providing libraries for C++ projects. With a few exceptions for system libraries and bde-tools, this PR adds build presets for CMake that utilize vcpkg as the package provider.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,23 @@ list( APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/etc/cmake" )
 include( BBVendor )
 include( TargetBMQStyleUor )
 
+if(DEFINED VCPKG_TOOLCHAIN)
+  find_package(bsl CONFIG PATH_SUFFIXES share/bde REQUIRED)
+  find_package(bal CONFIG PATH_SUFFIXES share/bde REQUIRED)
+  find_package(bdl CONFIG PATH_SUFFIXES share/bde REQUIRED)
+  find_package(nts REQUIRED)
+  find_package(ntc REQUIRED)
+
+  # vcpkg exports most other targets under a different name than what BdeBuildSystem uses to find
+  # targets. Because of this, if we detect that we're using vcpkg to provide targets, we alias them to the
+  # pkg-config style names BdeBuildSystem is trying to use.
+  find_package(benchmark CONFIG REQUIRED)
+  find_package(ZLIB REQUIRED)
+
+  add_library(benchmark ALIAS benchmark::benchmark)
+  add_library(zlib ALIAS ZLIB::ZLIB)
+endif()
+
 # -----------------------------------------------------------------------------
 #                                INITIALIZATION
 # -----------------------------------------------------------------------------
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,47 @@
+{
+    "version": 6,
+    "configurePresets": [
+        {
+            "name": "base",
+            "hidden": true,
+            "description": "Base configuration for a build of BlazingMQ",
+            "binaryDir": "cmake.bld",
+            "cacheVariables": {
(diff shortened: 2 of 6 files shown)


CMakeLists.txt
```diff
@@ -16,6 +16,24 @@ list( APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/etc/cmake" )
...
+  find_package(bal CONFIG PATH_SUFFIXES share/bde REQUIRED)
+  find_package(bdl CONFIG PATH_SUFFIXES share/bde REQUIRED)
+  find_package(nts REQUIRED)
+  find_package(ntc REQUIRED)
+  find_package(benchmark CONFIG REQUIRED)
+  find_package(GTest CONFIG REQUIRED)
+  find_package(ZLIB REQUIRED)
+
+  add_library(benchmark ALIAS benchmark::benchmark)
+  add_library(zlib ALIAS ZLIB::ZLIB)
```
pniedzielski, line 30: Are the `pkgname::pkgname` target names a VCPKGism?  If so, how come it's just these, and not GTest for instance?  I'd like a short comment right above this line explaining exactly why this is necessary.

CMakePresets.json
```diff
@@ -0,0 +1,47 @@
+{
```
pniedzielski, line 1: Do we need to pull in any of the `bde-tools` toolchain files here, either [by chainloading them](https://learn.microsoft.com/en-us/vcpkg/users/buildsystems/cmake-integration#using-multiple-toolchain-files) or copying some of the configuration into `base`?

vcpkg.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg.schema.json",
+  "name": "bmq",
+  "version": "1.0.0",
```
pniedzielski, line 4: Should this version match our release version?

vcpkg.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg.schema.json",
+  "name": "bmq",
```
pniedzielski, line 3: `bmq` -> `blazingmq` ?

pniedzielski (changes requested):
Small changes

vcpkg.json
```diff
@@ -0,0 +1,13 @@
...
+  "version": "1.0.0",
+  "dependencies": [
+    "gtest",
+    "bde",
+    "ntf-core",
+    "benchmark",
+    "zlib",
+    "readline"
```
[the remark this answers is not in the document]
678098: Nevermind, found in comments

CMakeLists.txt
hallfox: This is how many vcpkg packages choose to export components. It's a common pattern among other uses of CMake. `GTest` as you can tell is not necessary, must have mistakenly added it while I was testing this.

CMakePresets.json
hallfox: No, the toolchains provided by bde-tools haven't been necessary so far. We may consider using chainloading, but what vcpkg's release and override triplet features provide are a fine enough substitute.

vcpkg.json
hallfox: Which version? 😛 I'll set it to the bmq SDK version for now, but we should probably have this version match a number that corresponds to some kind of source release.

CMakeLists.txt
pniedzielski: Resolved in comment.

CMakePresets.json
pniedzielski: Okay, makes sense.

pniedzielski (approved):

pniedzielski (approved):

PR #121, opened 2023-09-29
Two version-related improvements

jll63:
* Add command line switch (`--version` or `-v`) to query broker version.
* Let the broker set `appConfig.brokerVersion` from compiled-in `bmqbrkrscm_version`.

**Testing performed**
* `bmqbrkr.tsk --version` and `bmqbrkr.tsk -v`: displays the version
* `bmqbrkr.tsk`: print previous error message (missing config) and exit - as before
* Execute the `run` script: works as before

### DIFF
--- a/src/applications/bmqbrkr/bmqbrkr.m.cpp
+++ b/src/applications/bmqbrkr/bmqbrkr.m.cpp
@@ -594,6 +594,73 @@ run(bsl::ostream& errorDescription, TaskEnvironment* taskEnv, bool wait = true)
 
 int main(int argc, const char* argv[])
 {
+
+    // Parse command line parameters
+    bsl::string configDir;
+    bsl::string instanceId = "default";
+    bsl::string hostName;
+    bsl::string hostTags;
+    bsl::string hostDataCenter;
+    int         port    = 0;
+    bool        version = false;
+
+    balcl::OptionInfo specTable[] = {
+        {"",
+         "config",
+         "Path to the configuration directory",
+         balcl::TypeInfo(&configDir),
+         balcl::OccurrenceInfo::e_OPTIONAL},
+        {"i|instanceId",
+         "instanceId",
+         "The instance ID ('default' if not provided)",
+         balcl::TypeInfo(&instanceId),
+         balcl::OccurrenceInfo::e_OPTIONAL},
+        {"h|hostName",
+         "hostName",
+         "Override host name",
+         balcl::TypeInfo(&hostName),
+         balcl::OccurrenceInfo::e_OPTIONAL},
+        {"t|hostTags",
+         "hostTags",
+         "Override host tags",
+         balcl::TypeInfo(&hostTags),
+         balcl::OccurrenceInfo::e_OPTIONAL},
+        {"d|hostDataCenter",
+         "hostDataCenter",
+         "Override host data center",
+         balcl::TypeInfo(&hostDataCenter),
+         balcl::OccurrenceInfo::e_OPTIONAL},
+        {"p|port",
+         "port",
+         "Override port",
+         balcl::TypeInfo(&port),
+         balcl::OccurrenceInfo::e_OPTIONAL},
+        {"v|version",
+         "version",
+         "Show version number",
+         balcl::TypeInfo(&version),
+         balcl::OccurrenceInfo::e_OPTIONAL},
+    };
+
+    balcl::CommandLine commandLine(specTable);
+
+    if (commandLine.parse(argc, argv)) {
+        bsl::cerr << "PANIC [STARTUP] Failed to parse command line\n"
+                  << bsl::flush;
+        commandLine.printUsage();
+        return mqbu::ExitCode::e_COMMAND_LINE;  // RETURN
+    }
+
+    if (version) {
+        bsl::cout << bmqbrkrscm::Version::version() << "\n";
+        return 0;
+    }
+
+    if (configDir.empty()) {
+        bsl::cerr << "Error: No value supplied for the non-option argument "
+                     "\"config\".\n";
+    }
+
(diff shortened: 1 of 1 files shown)


pniedzielski:
@jll63, It looks like your git email isn't set up correctly, so the Sanity Check is failing.

src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -299,6 +299,8 @@ static int getConfig(bsl::ostream&      errorDescription,
     }
 
     taskEnv->d_config.appConfig().etcDir() = configDir;
+    taskEnv->d_config.appConfig().brokerVersion() =
+        bmqbrkrscm::Version::versionAsInt();
```
quarter-note, line 303: So, before this, `brokerVersion` field was being populated during config decoding step at line 287 above?

src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -592,50 +594,22 @@ run(bsl::ostream& errorDescription, TaskEnvironment* taskEnv, bool wait = true)
...
+    int         port    = 0;
+    bool        version = false;
 
     balcl::OptionInfo specTable[] = {
         {"",
          "config",
          "Path to the configuration directory",
          balcl::TypeInfo(&configDir),
-         balcl::OccurrenceInfo::e_REQUIRED},
+         balcl::OccurrenceInfo::e_OPTIONAL},
```
quarter-note, line 612: Why is this changed to optional?  So that you can invoke `/path/to/bmqbrkr -v` to check its version, w/o the command line parser complaining about the config dir?

src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -661,6 +635,11 @@ int main(int argc, const char* argv[])
          "Override port",
          balcl::TypeInfo(&port),
          balcl::OccurrenceInfo::e_OPTIONAL},
+        {"v|version",
+         "version",
+         "Show version number",
```
quarter-note, line 640: May change to "Show broker version number"

src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -672,6 +651,46 @@ int main(int argc, const char* argv[])
         return mqbu::ExitCode::e_COMMAND_LINE;  // RETURN
     }
 
+    if (version) {
+        bsl::cout << bmqbrkrscm::Version::version() << "\n";
```
quarter-note, line 655: May be add a string like "BlazingMQ broker version: " before printing the version. Unless, you just want the broker to print the version so that some scripts can rely on the output?

src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -672,6 +651,46 @@ int main(int argc, const char* argv[])
...
+        return 0;
+    }
+
+    if (configDir.empty()) {
+        bsl::cerr << "Error: No value supplied for the non-option argument "
+                     "\"config\".\n";
```
quarter-note, line 661: Broker should exit if no cfg dir is specified, right?

src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -672,6 +651,46 @@ int main(int argc, const char* argv[])
...
+        bsl::cerr << "Failed to install SIGINT handler  (rc: " << rc << ")\n"
+                  << bsl::flush;
+    }
+
+    // If running in a TTY, enable signal handler so that user signals can be
+    // used to stop the task.
+    if (isatty(fileno(stdout))) {
+        ::sigaction(SIGQUIT, &sa, NULL);
+        ::sigaction(SIGTERM, &sa, NULL);
+    }
```
quarter-note, line 689: Any reason to move the signal handling related logic here instead of keeping it earlier?

quarter-note (commented):
Some comments

src/applications/bmqbrkr/bmqbrkr.m.cpp
jll63: Yes.

src/applications/bmqbrkr/bmqbrkr.m.cpp
jll63: Yes. The test for the `config` argument is done "manually" if `-v` is not passed.

src/applications/bmqbrkr/bmqbrkr.m.cpp
jll63: Anyway the more verbose output is not difficult to parse. Adding the string.

src/applications/bmqbrkr/bmqbrkr.m.cpp
jll63: Ah yes.

src/applications/bmqbrkr/bmqbrkr.m.cpp
jll63: We don't need to do that (nor to print "STARTING...")  if we are just printing the version.

jll63:
All comments addressed. Changes in separate commit, to be squashed upon merging.

src/applications/bmqbrkr/bmqbrkr.m.cpp
quarter-note: If the main goal is to let some script invoke `/path/to/bmqbrkr -v` to figure out the version programmatically, then perhaps we should print only the verison w/o any string.  Will leave it up to you to make that call.

quarter-note (approved):
Left a comment.  Feel free to merge once CI is happy.

PR #122, opened 2023-10-05
Fix WConversion warning

banipreetr:
*Issue number of the reported bug or feature request: #87 *

**Describe your changes**
This is a small `good first issue` contribution to fix -Wconversion warnings that pollutes the build log.

This particular PR deals with the following warning:
```
/home/runner/work/blazingmq/blazingmq/src/groups/mwc/mwctst/mwctst_blobtestutil.cpp:62:46: warning: conversion from ‘size_t’ {aka ‘long unsigned int’} to ‘int’ may change value [-Wconversion]
   62 |         bdlbb::BlobBuffer blobBuffer(buffer, length);

```

**Testing performed**
N/A
**Additional context**
Add any other context about your contribution here.

Issue #87, opened 2023-08-17
Fix `-Wconversion` warnings during compilation


678098:
### Is there an existing proposal for this?

- [X] I have searched the existing proposals

### Is your feature request related to a problem?

There are ~2900 `-Wconversion` warnings during compilation (out of ~3000 warnings in total). All or most of them are unprobable to cause any problems in our runtime.

But there are concerns:
- Really interesting warnings could be lost in this noise.
- Logs become larger in general.
- Prevents from zero-warning compilation.

Info to read:
https://gcc.gnu.org/wiki/NewWconversion

### Describe the solution you'd like

Some of these conversion warnings are straightforward to fix. While the scale of 2900 warnings is huge, it might be good to split the work into a smaller chunks: make changes to the simplest cases, on a file or several-files basis, no more than 5-10 changes per PR etc...

### Alternatives you considered

Suppress `-Wconversion` warning during compilation to remove noise - but for me it's better to fix it

### Contribution workflow

1. Go to the Github Actions for this project: https://github.com/bloomberg/blazingmq/actions
2. Find the nearest `main` branch build
3. Open logs for `Build BlazingMQ`
4. Ctrl+F `-Wconversion` to find an intersting place to fix

hallfox:
Agreed here, most of these are due to small comparison, but there doesn't seem to be good tool for helping with applying fixes for these. Personally I think we might want to start by fixing the `-Wconversion` warnings in a package of `mwc` and work our way to `bmq`, `mqb`, and `bmqbrkr` from there.

PR #125, opened 2023-10-09
Fix more Wconversion warnings

banipreetr:

**Describe your changes**
Follow up to https://github.com/bloomberg/blazingmq/pull/122, this PR contribution is to fix more -Wconversion warnings that pollutes the build log.

**Testing performed**
N/A
**Additional context**

### DIFF
--- a/src/groups/mwc/mwctst/mwctst_testhelper.cpp
+++ b/src/groups/mwc/mwctst/mwctst_testhelper.cpp
@@ -47,7 +47,7 @@ void TestHelper::printTestName(bslstl::StringRef value)
     }
 
     bsl::cout << "\n" << value << "\n";
-    int length = value.length();
+    long unsigned int length = value.length();
     while (length--) {
         bsl::cout << "=";
     }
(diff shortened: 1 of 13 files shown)


src/groups/mwc/mwctst/mwctst_testhelper.cpp
```diff
@@ -47,7 +47,7 @@ void TestHelper::printTestName(bslstl::StringRef value)
     }
 
     bsl::cout << "\n" << value << "\n";
-    int length = value.length();
```
678098, line 50: Currently `long unsigned int` is identical to `size_t`, but for the future it's more safe to use `size_t` here, as `length()` returns. The reason is that `size_t` has an explicit contract in its definition `size_t can store the maximum size of a theoretically possible object of any type (including array).`

https://en.cppreference.com/w/c/types/size_t

There are `size_t` used in almost all similar cases in the codebase.

678098 (commented):
One small comment

src/groups/mwc/mwctst/mwctst_testhelper.cpp
banipreetr: Addressed, thanks!

678098 (approved):
LGTM!

678098:
@banipreetr thank you for the contribution!

Issue #133, opened 2023-10-20
Possible to mis-use `QueueOptions` with subscriptions to cause broker crash


678098:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The sample consumer with the problem exposed:

https://github.com/bloomberg/blazingmq/pull/132/files

The crash log looks like this:

```
20OCT2023_13:24:39.771 (6108901376) FATAL mqbblp_queuehandle.cpp:634 Assertion failed: subscription->d_downstreamSubQueueId == downstreamSubId, stack trace:
(0): bmqAssertHandler(char const*, char const*, int)+0xe4 at 0x104461d84 in bmqbrkr.tsk
(1): BloombergLP::bsls::Assert::failOnViolation(BloombergLP::bsls::AssertViolation const&)+0x54 at 0x1053f82f4 in bmqbrkr.tsk
(2): BloombergLP::bsls::Assert::invokeHandler(BloombergLP::bsls::AssertViolation const&)+0x24 at 0x1053f857c in bmqbrkr.tsk
(3): BloombergLP::mqbblp::QueueHandle::registerSubscription(unsigned int, unsigned int, BloombergLP::bmqp_ctrlmsg::ConsumerInfo const&, unsigned int)+0x324 at 0x1048691a8 in bmqbrkr.tsk
(4): BloombergLP::mqbblp::Routers::Consumer::registerSubscriptions(BloombergLP::mqbi::QueueHandle*)+0xb8 at 0x10496acfc in bmqbrkr.tsk
(5): BloombergLP::mqbblp::Routers::AppContext::registerSubscriptions()+0x80 at 0x10496e640 in bmqbrkr.tsk
(6): BloombergLP::mqbblp::RootQueueEngine::configureHandle(BloombergLP::mqbi::QueueHandle*, BloombergLP::bmqp_ctrlmsg::StreamParameters const&, bsl::function<void (BloombergLP::bmqp_ctrlmsg::Status const&, BloombergLP::bmqp_ctrlmsg::StreamParameters const&)> const&)+0xe3c at 0x10494bd24 in bmqbrkr.tsk
[...]

PR #135, opened 2023-10-25
Generating unique Subscription ids for each request

dorjesinpo:
Issue number of the reported bug or feature request: #133

2 Subscription Ids is not enough; we need 3.

In the order of increasing uniqueness:
1. `CorrelationId` - user provided.
2. `SubscriptionHandle` - uniquely identifies a Subscription in `QueueOptions`.  But if reused for different App in the same queue, creates the issue
3.  New unique id generated for each Subscription for each Configure request.

The existing logic involved storing request context associated with subscriptions (like, `CorrelationId`) until response arrives which makes the context effective.  The logic is slightly refactored.

Ideally, the context belongs to `BrokerSession::RequestManagerType::RequestSp` but that calls for a bigger refactoring of the SDK.

### DIFF
--- a/src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
+++ b/src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
@@ -5245,9 +5245,8 @@ BrokerSession::createConfigureQueueContext(const bsl::shared_ptr<Queue>& queue,
 
             bmqp_ctrlmsg::Subscription subscription(d_allocator_p);
 
-            static bsls::AtomicUint s_nextId(0);
-
-            unsigned int internalSubscriptionId = s_nextId.add(1);
+            const unsigned int internalSubscriptionId =
+                ++d_nextInternalSubscriptionId;
 
             subscription.sId() = internalSubscriptionId;
             // Using unique id instead of 'SubscriptionHandle::id()'
@@ -5636,6 +5635,7 @@ BrokerSession::BrokerSession(
 , d_messageExpirationTimeoutHandle()
 , d_nextRequestGroupId(k_NON_BUFFERED_REQUEST_GROUP_ID)
 , d_queueRetransmissionTimeoutMap(allocator)
+, d_nextInternalSubscriptionId(bmqp::Protocol::k_DEFAULT_SUBSCRIPTION_ID)
 {
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(d_scheduler_p->clockType() ==
--- a/src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
+++ b/src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
@@ -3533,10 +3533,9 @@ void BrokerSession::processPushEvent(const bmqp::Event& event)
             bmqt::CorrelationId         correlationId;
             unsigned int                subscriptionHandleId;
             const QueueManager::QueueSp queue =
-                    d_queueManager.observePushEvent(
-                            &correlationId,
-                            &subscriptionHandleId,
-                            *citer);
+                d_queueManager.observePushEvent(&correlationId,
+                                                &subscriptionHandleId,
+                                                *citer);
 
             BSLS_ASSERT(queue);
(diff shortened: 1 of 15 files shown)


quarter-note:
To @678098, assuming that this is ready for review.

src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
```diff
@@ -5239,7 +5245,13 @@ BrokerSession::createConfigureQueueContext(const bsl::shared_ptr<Queue>& queue,
 
             bmqp_ctrlmsg::Subscription subscription(d_allocator_p);
 
-            subscription.sId() = cit->first.id();
+            static bsls::AtomicUint s_nextId(0);
```
678098, line 5248: This static local variable initialization is not thread safe in general. The probability to observe a bug is small though.

Here are some details and advices:

https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/group__bslmt__once.html#_details

And here is how we did a safe initialization for another component:

https://github.com/bloomberg/blazingmq/blob/08a6d16b1b73d1d71113f80f7c22fcdd3e5392d8/src/groups/bmq/bmqt/bmqt_correlationid.cpp#L33-L42

src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
```diff
@@ -5239,7 +5245,13 @@ BrokerSession::createConfigureQueueContext(const bsl::shared_ptr<Queue>& queue,
...
+
+            unsigned int internalSubscriptionId = s_nextId.add(1);
```
678098, line 5250:
```suggestion
            const unsigned int internalSubscriptionId = s_nextId.add(1);
```

src/groups/bmq/bmqimp/bmqimp_queuemanager.h
```diff
@@ -363,8 +352,9 @@ class QueueManager {
     /// `correlationId`, and return a shared pointer to the Queue object (if
     /// found), or an empty shared pointer (if not found).
     QueueSp lookupQueueBySubscriptionId(bmqt::CorrelationId* correlationId,
-                                        int                  queueId,
-                                        unsigned int subscriptionId) const;
+                                        unsigned int* subscriptionHandle,
```
678098, line 355: We have `subscriptionHandle` in some places like here, and `subscriptionHandleId` in other places.

678098 (commented):
Some comments

src/groups/bmq/bmqimp/bmqimp_queue.h
```diff
@@ -528,6 +554,33 @@ inline Queue& Queue::setConfig(const bmqp_ctrlmsg::StreamParameters& value)
...
+    bsl::unordered_map<unsigned int, SubscriptionHandle>::const_iterator cit =
+        d_registeredInternalSubscriptionIds.find(internalSubscriptionId);
+
+    if (cit == d_registeredInternalSubscriptionIds.end()) {
+        return {internalSubscriptionId, bmqt::CorrelationId()};  // RETURN
+    }
+
+    SubscriptionHandle result(cit->second);
+
+    d_registeredInternalSubscriptionIds.erase(cit);
```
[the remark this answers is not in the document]
dorjesinpo: Yes. It is the FSM thread.  So is `createConfigureQueueContext`! So, there is no need for atomic or `Once`.
Will commit new changes.

678098 (approved):
LGTM!

src/groups/bmq/bmqt/bmqt_subscription.cpp
```diff
@@ -43,7 +43,7 @@ const int Subscription::k_DEFAULT_CONSUMER_PRIORITY        = 0;
 
 unsigned int SubscriptionHandle::nextId()
 {
-    static bsls::AtomicUint s_id = 0;
+    static bsls::AtomicUint s_id = k_INVALID_HANDLE_ID;
```
678098, line 46: https://github.com/bloomberg/blazingmq/issues/141

I created an issue about this line. It's out of the scope of this PR and could be fixed in another PR.

PR #139, opened 2023-11-01
[DO NOT MERGE] CMakeLists.txt: Find benchmark and zlib outside of vcpkg

kaikulimu:
This fixes the `ld: cannot find -lbenchmark` error when building unit tests as reported by @jll63 and our internal CI. However, I am not too confident about merging this PR, because I do not understand the significance of the `if(DEFINED VCPKG_TOOLCHAIN)` block.


kaikulimu:
Closing in favor of #145

PR #146, opened 2023-11-07
Fix: slow failover

678098:


src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -286,7 +286,7 @@ void RelayQueueEngine::onHandleConfiguredDispatched(
             << downStreamParameters << ", but assuming success.";
     }
 
-    BALL_LOG_INFO_BLOCK
+    BALL_LOG_TRACE_BLOCK
     {
         mqbcmd::RoundRobinRouter outrr(d_allocator_p);
         context->d_routing_sp->loadInternals(&outrr);
```
678098, line 292: This `loadInternals` might be called for each consumer rapidly and might degrade overall performance when we have thousands of consumers and still continue to add new.

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
```diff
@@ -717,12 +715,13 @@ bool QueueHandle::unregisterSubStream(
...
-                if (d_throttledSubscriptionInfo.requestPermission()) {
-                    BALL_LOG_INFO << "Queue '" << d_queue_sp->description()
-                                  << "' handle " << this
-                                  << " removing Subscription "
-                                  << itSubscription->first;
-                }
+                static const int                k_MAX_INSTANT_MESSAGES = 10;
+                static const bsls::Types::Int64 k_NS_PER_MESSAGE =
+                    bdlt::TimeUnitRatio::k_NANOSECONDS_PER_SECOND;
+                BALL_LOGTHROTTLE_INFO(k_MAX_INSTANT_MESSAGES, k_NS_PER_MESSAGE)
```
678098, line 721: Decided to use in-place directive from `ball` and remove member throttle `d_throttledSubscriptionInfo` to simplify class.
This throttle is thread-safe: https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/group__ball__logthrottle.html#3.2

678098: Side effect: 2 places with throttle previously had a common throttle, and now they are independent. But in my opinion this doesn't matter in this case.

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
hallfox: Could we make this a debug level instead?

dorjesinpo: How about `BALL_LOGTHROTTLE_INFO_BLOCK`?

678098: @hallfox changed trace->debug

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
```diff
@@ -610,12 +606,14 @@ void QueueHandle::registerSubscription(unsigned int downstreamSubId,
                                        const bmqp_ctrlmsg::ConsumerInfo& ci,
                                        unsigned int upstreamId)
 {
-    if (d_throttledSubscriptionInfo.requestPermission()) {
-        BALL_LOG_INFO << "QueueHandle [" << this
-                      << "] registering Subscription [id = " << downstreamId
-                      << ", downstreamSubQueueId = " << downstreamSubId
-                      << ", upstreamId = " << upstreamId << "]";
-    }
+    static const int                k_MAX_INSTANT_MESSAGES = 10;
```
dorjesinpo, line 609: Maybe, move `k_MAX_INSTANT_MESSAGES` and `k_NS_PER_MESSAGE` out in a single place?  They are the same for all occurrences most likely.

dorjesinpo (changes requested):
I guess, there was no `BALL_LOGTHROTTLE_` when we needed it...
Looks good. Couple minor things.

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
678098: @hallfox will use throttle info block instead, so we have something in logs without changing default severity

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -66,6 +67,13 @@ namespace mqbblp {
 
 namespace {
 
+const int k_MAX_INSTANT_MESSAGES = 10;
+// Maximum messages logged with throttling in a short period of time.
+
+const bsls::Types::Int64 k_NS_PER_MESSAGE =
+    bdlt::TimeUnitRatio::k_NANOSECONDS_PER_MINUTE / k_MAX_INSTANT_MESSAGES;
```
678098, line 74: Allow to log at most 10 records in a minute. Should be enough for us.

Currently each cycle of adding a new consumer consumes ~40 ms in total:

```
04NOV2023_14:09:16.926 (140284186806016) INFO mqbblp_relayqueueengine.cpp:824 For queue '...', rebuilt highest priority consumers: [count: 1598]
04NOV2023_14:09:16.885 (140284186806016) INFO mqbblp_relayqueueengine.cpp:824 For queue '...', rebuilt highest priority consumers: [count: 1597]
04NOV2023_14:09:16.842 (140284186806016) INFO mqbblp_relayqueueengine.cpp:824 For queue '...', rebuilt highest priority consumers: [count: 1596]
04NOV2023_14:09:16.802 (140284186806016) INFO mqbblp_relayqueueengine.cpp:824 For queue '...', rebuilt highest priority consumers: [count: 1595]
04NOV2023_14:09:16.761 (140284186806016) INFO mqbblp_relayqueueengine.cpp:824 For queue '...', rebuilt highest priority consumers: [count: 1594]
```

PR #152, opened 2023-11-09
Sending configureStream

dorjesinpo:
Introducing a (temporary) config knob `configureStream` which will replace old `configureQueue` with the new `configureStream` (subscriptions support).

`configureStream` is what the dev uses now.


678098 (approved):

quarter-note:
Should be able to merge this next week.

quarter-note (dismissed):
The `brkrCfg.configureStream()` check has been added at two places.  I guess those are the only two places in the broker which need to be updated?

dorjesinpo:
> The `brkrCfg.configureStream()` check has been added at two places. I guess those are the only two places in the broker which need to be updated?

Correct.  This is about what goes upstream (which is always cluster) so the main place is `CQH`.  The second one is about the trick we did (temporarily) to present Subscription Ids as SubQueue Ids for unaware consumers.
Rebasing...

PR #156, opened 2023-11-17
re-insert queue tracking if needed

dorjesinpo:
The following sequence:

1. Queue  has connected clients
2. OpenQueue request receives a response `[ category = E_REFUSED code = -11 message = "self node is stopping" ]`
3. As the result, the OpenQueue goes into pending contexts collection
4. All existing clients close the queue
5.  As the result, the internal QueueInfo tracking (created upon `upstreamSubQueueId` assignment) gets erased
6.  The state restores (new Leader) and the queue processes pending contexts and sends OpenQueue request.
7. Upon request sending, the queue wants to merge new counters with the existing ones from the tracking, but the tracking is gone

... leads to assertion.

Proposed solution is to re-insert the tracking if it is missing (as if just received the pending OpenQueue request)

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
@@ -906,41 +906,48 @@ void ClusterQueueHelper::processPendingContexts(
 
 void ClusterQueueHelper::assignUpstreamSubqueueId(OpenQueueContext* context)
 {
-    // If needed, generate upstream subQueueId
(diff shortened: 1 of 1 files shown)


dorjesinpo:
6 - mwcio_ntcchannelfactory.t (Failed) - unrelated

678098 (approved):
LGTM

quarter-note (approved):

quarter-note:
Failure of 1 UT is due to known unrelated issue in ntf.  Will merge this PR.

PR #166, opened 2023-12-07
[WIP]Feat: support TLS configuration in broker/client

678098:


src/groups/mqb/mqbcfg/mqbcfg.xsd
```diff
@@ -271,6 +276,27 @@
     </sequence>
   </complexType>
 
+    <complexType name='TlsConfig'>
```
678098, line 279: We already have `Tcp` in this xsd, so I decided to use `Tls` here instead of `TLS`. It's not used in parsing, so we can change these both things to uppercase if we want.

src/groups/mqb/mqbcfg/mqbcfg.xsd
```diff
@@ -239,9 +241,11 @@
         heartbeatIntervalMs..:
             How often (in milliseconds) to check if the channel received data,
             and emit heartbeat.  0 to globally disable.
-       useNtf...............:
+        useNtf...............:
             Use the new NTF based TCP transport library instead of
             the existing one based on BTE
+         tls.................:
```
678098, line 247: Possible to rename this flag to `useTls` similar to `useNtf`.

src/groups/mqb/mqbcfg/mqbcfg.xsd
```diff
@@ -97,6 +98,7 @@
       <element name='plugins'              type='tns:Plugins'/>
       <element name='messagePropertiesV2'  type='tns:MessagePropertiesV2'/>
       <element name='configureStream'      type='boolean' default='false'/>
+      <element name="tls"                  type='tns:TlsConfig' minOccurs='0'/>
```
quarter-note, line 101: May want to call it `tlsConfig` or `tlsCfg`.  

More importantly, should this field go inside `networkInterfaces` or `tcpInterface`, since it is transport-related?

src/groups/mqb/mqbcfg/mqbcfg.xsd
quarter-note: Yeah `useTls` is probably better.  Also, should this field go inside `TlsConfig` complex type?

quarter-note (commented):
Some comments.

src/groups/mqb/mqbcfg/mqbcfg.xsd
678098: > Yeah `useTls` is probably better. Also, should this field go inside `TlsConfig` complex type?

It's a good question.
We have several possibilities.

1. Current structure, `useTls` and `tlsConfig` settings are placed in different paths.
2. `useTls` is a part of `tlsConfig`.
3. `useTls` is not a part of `tlsConfig`, but they both are placed near each other in `TcpInterfaceConfig`.
4. Implicit approach: no `useTls` flag, optional `tlsConfig` presence is a marker that TLS should be used.

2 and 3 are good for me.

src/groups/mqb/mqbcfg/mqbcfg.xsd
```diff
@@ -271,6 +276,30 @@
...
+            A path to the FILE, containing concatenation of known certificates
+            the server can use to reference as its certificate store.
+        certificate..........:
+            A path to the FILE, containing the certificate the broker will use
+            to identify itself to other clients.
+        key..................:
+            A path to the FILE, containing the private key that the broker uses
+            to read the certificate.
+        version..............:
+            Protocol version.
```
willhoy, line 292: would be good to give an example of what's expected here e.g. "TLS1.2" 
from the point of view of the server you might want to allow a set of versions or at least version 1.2 or greater. 
It's good to spec out exactly what this parameter means

678098: @willhoy we plan to use a list of versions indeed

PR #173, opened 2023-12-15
Integration tests

jll63:
* Some Python modules needed to support the integration test framework.
* Integration test framework.
* The integration tests themselves.


.github/workflows/build.yaml
```diff
@@ -41,94 +43,31 @@ jobs:
...
           ctest -E mwcsys_executil.t --output-on-failure
-      - uses: actions/cache@v3
-        with:
-          path: |
-            cmake.bld/Linux
-            deps
-            /opt/bb/include
-          key: cache-${{ github.sha }}
-
-  build_and_test_bmqprometheus_plugin:
```
quarter-note, line 65: Any reason to remove this hook?

quarter-note (commented):
One comment

.github/workflows/build.yaml
jll63: Nope. I don't even recall doing this. Restoring...

PR #174, opened 2023-12-20
fix: mix of strong and weak consistencies

dorjesinpo:
When a partition hosts both weak and strong consistencies and there are recovered messages and the last recovered messages is of weak consistency, replica can issue implicit receipt for the weak consistency GUID.  Primary will ignore that receipt because the collection of unreceipted messages is hashtable, not a map (for performance reasons).  (Primary will process next ones).

Solution is to make sure implicit receipt is for the last strong consistency GUID.

This fix needs manual merge into the future FSM flow.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
@@ -964,6 +964,9 @@ void StorageManager::recoveredQueuesCbImpl(
     mqbs::FileStoreIterator fsIt(fs);
     DataStoreRecordHandles  recordsToPurge;
 
+    bsls::Types::Uint64 lastStrongConsistencySequenceNum    = 0;
+    unsigned int        lastStrongConsistencyPrimaryLeaseId = 0;
+
     while (fsIt.next()) {
         mqbu::StorageKey        appKey;
         mqbu::StorageKey        queueKey;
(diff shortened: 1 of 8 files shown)


src/groups/mqb/mqbs/mqbs_filestore.h
```diff
@@ -329,7 +329,8 @@ class FileStore : public DataStore {
 
     NodeReceiptContexts d_nodes;
 
-    DataStoreRecordKey d_lastRecoveredMessage;
+    bsls::Types::Uint64 d_lastStrongConsistencySequenceNum;
```
kaikulimu, line 332: Why not call it `DataStoreRecordKey d_lastRecoveredStrongConsistencyMessage`?

It shows clearly that the sequence number and the primary lease id are always used in conjunction.

src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
```diff
@@ -1037,6 +1043,8 @@ void StorageManager::recoveredQueuesCbImpl(
         mqbs::ReplicatedStorage* rs = storageMapIt->second;
         BSLS_ASSERT_SAFE(rs);
 
+        const bool isStrongConsistency = !rs->hasReceipt(bmqt::MessageGUID());
```
kaikulimu, line 1046: Looks a rather hacky way to determine whether a message is under strong consistency. But I am fine with it if you don't see any better way.

kaikulimu: If you stick with this way, perhaps add a `isStrongConsistency(const bmqt::MessageGUID& guid)` method somewhere to make the intentions clear.

kaikulimu (commented):
Code looks good. Just some minor comments

src/groups/mqb/mqbs/mqbs_filestore.h
kaikulimu: d

src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
kaikulimu: d

kaikulimu (approved):
LGTM!

PR #176, opened 2023-12-22
Fix: purge inactive queues

678098:
This PR addresses the issue: https://github.com/bloomberg/blazingmq/issues/179

Proposed solution:
- Extend `mqbcmd` with purge commands on a storage level. `StorageQueue -> StorageQueueCommand.purgeAppId` for PURGE QUEUE, `StorageDomain -> StorageDomainCommand.purge` for DOMAIN PURGE.
- PURGE QUEUE: remove PURGE command in `mqbi::Queue`, execute it in `mqbc::StorageUtil` instead, using correct QUEUE DISPATCHER thread.
- PURGE DOMAIN: remove PURGE command in `mqbblp::Domain`, execute it in `mqbc::StorageUtil` in parallel, using correct QUEUE DISPATCHER thread for each `FileStore` corresponding to its partition.


src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -2766,7 +2815,87 @@ int StorageUtil::processCommand(mqbcmd::StorageResult*     result,
...
+                                            purgedQs.end());
+            }
+
+            return 0;  // RETURN
+        }
+    }
+    else if (command.isQueueValue()) {
+        BSLS_ASSERT_SAFE(command.queue().command().isPurgeAppIdValue());
+
+        const bmqt::Uri uri(command.queue().uri(), allocator);
```
678098, line 2850: Maybe rename `uri` in this command to more specific `canonicalUri`.
The idea is, in the canonical uri we also pass a domain name.

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -2766,7 +2815,87 @@ int StorageUtil::processCommand(mqbcmd::StorageResult*     result,
                          command.domain().name(),
                          *fileStores);
             return 0;  // RETURN
+        } else if (command.domain().command().isPurgeValue()) {
+            // todo return value
```
678098, line 2819: Clean up todos

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -2766,7 +2815,87 @@ int StorageUtil::processCommand(mqbcmd::StorageResult*     result,
...
+    else if (command.isQueueValue()) {
+        BSLS_ASSERT_SAFE(command.queue().command().isPurgeAppIdValue());
+
+        const bmqt::Uri uri(command.queue().uri(), allocator);
+
+        BSLS_ASSERT_SAFE(uri.isCanonical());
+
+        mqbi::Storage *queueStorage = NULL;
+        {
+            bslmt::LockGuard<bslmt::Mutex> guard(storagesLock);
```
678098, line 2856: Sent a MSG about `storagesLock`.
Here is an example of its minimal usage.
By the way, it's no harm to access `ReplicatedStorage` without locked `storagesLock`. This lock only sanitaze access to the `vector<map<...>>` container here, and we can safely access `ReplicatedStorage` from the corresponding `FileStore` thread.

src/groups/mqb/mqbblp/mqbblp_queue.cpp
```diff
@@ -852,31 +852,8 @@ int Queue::processCommand(mqbcmd::QueueResult*        result,
...
-                                : command.purgeAppId();
-        dispatcher()->execute(bdlf::BindUtil::bind(&Queue::purge,
-                                                   this,
-                                                   &purgedQueueResult,
-                                                   appId),
-                              this);
-        dispatcher()->synchronize(this);
-        return 0;  // RETURN
+        BSLS_ASSERT_SAFE(false && 
+                         "Should not get here. PURGE QUEUE command must be processed on a storage level.");
```
678098, line 856: Moved out purge queue command from `mqbblp::Queue`.
We should never get here.

And also there is a question, do we want to move 2 other commands from here, so `mqbblp::Queue` does not process commands at all.

TODO: remove purgeAppId from mqbcmd selection

kaikulimu:
Did brief local testing. The purge inactive queue command does work. Starting to review the code now. Sorry for the delays.

kaikulimu (changes requested):
Put some feedback. Additionally:

1. Please add an integration test. Either in this PR or have another PR later.
2. Please add new commits, and do not squash existing commits not rebase. This makes reviewing a lot easier :)

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -2767,6 +2818,86 @@ int StorageUtil::processCommand(mqbcmd::StorageResult*     result,
...
+    else if (command.isQueueValue()) {
+        BSLS_ASSERT_SAFE(command.queue().command().isPurgeAppIdValue());
+
+        const bmqt::Uri uri(command.queue().canonicalUri(), allocator);
+
+        BSLS_ASSERT_SAFE(uri.isCanonical());
+
+        mqbi::Storage* queueStorage = NULL;
+        {
+            bslmt::LockGuard<bslmt::Mutex> guard(storagesLock);
```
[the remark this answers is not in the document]
678098: Yep, I know about the previous code style. When I develop this PR, it was my decision to omit these comments, because they don't make much sense since they are not aligned to the rightmost line and are not visible enough to be useful.
I will change all these places as you ask though.

src/python/blazingmq/dev/it/process/broker.py
```diff
@@ -34,7 +34,7 @@ class Broker(blazingmq.dev.it.process.bmqproc.BMQProcess):
     def __init__(self, config: cfg.Broker, cluster, **kwargs):
         cwd: Path = kwargs["cwd"]
         (cwd / "bmqbrkr.ctl").unlink(missing_ok=True)
-        cwd.joinpath("storage", "archive").mkdir(parents=True)
+        cwd.joinpath("storage", "archive").mkdir(parents=True, exist_ok=True)
```
678098, line 37: @jll63 note that this is needed to restart the cluster with the same configs/work folder/storage.

678098:
>Please add an integration test. Either in this PR or have another PR later.
Hi @kaikulimu, I've added 2 ITs to test purge commands. One IT is a simple breathing test, another is "inactive queue" test.

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -2700,8 +2700,57 @@ void StorageUtil::forceFlushFileStores(FileStores* fileStores)
...
+        (*purgedQueuesVec)[partitionId];
+    fileStores[partitionId]->purgeDomain(&purgedQueues, domainName);
+
+    latch->arrive();
+}
+
+void StorageUtil::purgeQueueDispatched(
+    mqbcmd::PurgeQueueResult* purgedQueue,
+    bslmt::Semaphore*         purgeFinishedSemaphore,
+    mqbs::FileStore*          fileStore,
```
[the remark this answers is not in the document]
678098: I think this arg is still needed, because:
1. `fileStore` contains info about its dispatcher thread, which is necessary for thread safety checks. `mqbi::Storage` doesn't have this info.
2. `fileStore` contains info about its primary node. Again, `mqbi::Storage` doesn't have this information.

src/groups/mqb/mqbcmd/mqbcmd.xsd
```diff
@@ -110,6 +110,7 @@
       <element name="storageContent"        type="tns:StorageContent"/>
       <element name="clusterStorageSummary" type="tns:ClusterStorageSummary"/>
       <element name="replicationResult"     type="tns:ReplicationResult"/>
+      <element name="purgedQueues"          type="tns:PurgedQueues"/>
```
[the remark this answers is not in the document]
678098: The difference is that `InternalResult` represents a call tree in its structure, so every time we do dispatch and process, we add a next node to this call tree.
Final `Result` on the other hand is flattened via `mqbcmd::Util::flatten`, so its structure is plain, without all these intermediate nodes.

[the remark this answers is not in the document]
678098: Added comments

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -2767,6 +2818,86 @@ int StorageUtil::processCommand(mqbcmd::StorageResult*     result,
                          *fileStores);
             return 0;  // RETURN
         }
+        else if (command.domain().command().isPurgeValue()) {
+            bsl::vector<bsl::vector<mqbcmd::PurgeQueueResult> >
+                purgedQueuesVec;
+            purgedQueuesVec.resize(fileStores->size());
+
+            executeForEachPartitions(
```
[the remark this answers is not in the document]
678098: Added comment

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -2701,50 +2701,155 @@ void StorageUtil::forceFlushFileStores(FileStores* fileStores)
...
 }
 
 void StorageUtil::purgeQueueDispatched(
-    mqbcmd::PurgeQueueResult* purgedQueue,
+    mqbcmd::PurgeQueueResult* purgedQueueResult,
     bslmt::Semaphore*         purgeFinishedSemaphore,
-    mqbs::FileStore*          fileStore,
     mqbi::Storage*            storage,
-    const bsl::string*        appId)
+    const mqbs::FileStore*    fileStore,
```
kaikulimu, line 2767: `const mqbs::FileStore*    fileStore` -> `const mqbs::FileStore&    fileStore`

src/integration-tests/test_admin_client.py
```diff
@@ -2,22 +2,74 @@
...
+
+    def append(self, other: "PostRecord") -> None:
+        assert self.domain == other.domain
+        assert self.queue_name == other.queue_name
+        self.num += other.num
+
+
+def post_n_msgs(producer: Client, task: PostRecord, posted: Optional[Dict[str, PostRecord]] = None) -> None:
+    """
+    Open a connection for the specified 'producer' to the specified 'uri'
```
kaikulimu, line 44: The function arguments are `[producer, task, posted]`, but the documentation mentions `[producer, uri, n]`. Please fix the docs.

src/integration-tests/test_admin_client.py
```diff
@@ -41,3 +93,248 @@ def test_admin(single_node: Cluster):
...
+    - Verify that the observed number of active queues in PRIORITY domain is expected.
+    - Purge a subset of PRIORITY queues one by one using queue purge command,
+    some of these queues are inactive.
+    - Purge the skipped queues in the PRIORITY domain using domain purge command.
+    - Cycle through all PRIORITY queues again and verify that we purged everything.
+
+    Stage 3: FANOUT purge
+    - Verify that the observed number of active queues in FANOUT domain is expected.
+    - Purge a subset of FANOUT queues one by one using queue purge command with app_id
+    parameter specified, some of these queues are inactive. For each
```
kaikulimu, line 213: The documentation is incomplete after `For each`

kaikulimu (commented):
Minor comments

src/groups/mqb/mqbc/mqbc_storageutil.cpp
678098: Not good to do this, because this procedure is intended to be used as a dispatched callback. We call this procedure from another thread, while copy-passing arguments. Even const-ref string here is being copied, so I would prefer to roll back it to string pointer again, since we have guaranteed that the string will live longer than the procedure's call.

kaikulimu: Yes, you are correct. Copy-passing pointer is better than copying the entire object. For `appId`, the string size will be smaller so it's not a big deal, but feel free to change.

kaikulimu (approved):
LGTM

PR #177, opened 2023-12-22
Performance[MQB]: inline Put, Push, Ack, Confirm

dorjesinpo:
Messages that have fixed destination (such as PUSH/ACK/Relay PUT/Relay CONFIRM) do not have to go on Cluster thread - it is a bottleneck.
Instead, messages can go directly to mqbnet::Channel which enqueues them in thread-safe manner.
This needs synchronization based on AtomicGate which allows thread-safe, efficient checking of the cluster status and thread-safe opening/closing and draining.

PUSH and ACK require statistics update which we move from Cluster to Queue. For this reason the opening queue context (OpenQueueConfirmationCookie) conveys statistics from Queue to Cluster


678098:
Hi @dorjesinpo, looking

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
```diff
@@ -1173,22 +1249,78 @@ void QueueHandle::onAckMessage(const bmqp::AckMessage& ackMessage)
...
+        // Throttle error log if this is a 'failed Ack': note that we log at
+        // INFO level in order not to overwhelm the dashboard, if a queue is
+        // full, every post will nack, which could be a lot.
+        if (d_throttledFailedAckMessages.requestPermission()) {
+            BALL_LOG_INFO << "Queue '" << d_queue_sp->description()
+                          << "' NACK "
+                          << "[status: " << status
+                          << ", GUID: " << ackMessage.messageGUID()
+                          << ", node '" << d_clientContext_sp->description()
+                          << "']";
```
678098, line 1289: This log line is logged only when we don't have an inline client, is it okay? The same is true for other logs later in this method.

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
```diff
@@ -483,32 +493,67 @@ void QueueHandle::deliverMessageImpl(
...
+                subscription->stats()->onEvent(
+                    mqbstat::QueueStatsClient::EventType::e_PUSH,
+                    message ? message->length() : 0);
+            }
+        }
+    }
+    else if (d_throttledDroppedPutMessages.requestPermission()) {
+        BALL_LOG_WARN << "Queue '" << d_queue_sp->description()
+                      << "' failed to PUSH " << msgGUID
+                      << " with error: " << result;
```
678098, line 555: This is logged only when we don't have an inline client, is it okay to skip that case for logging?

dorjesinpo (commented):
Thank you for reviewing!

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
```diff
@@ -369,6 +345,8 @@ void ClusterProxy::onActiveNodeUp(mqbnet::ClusterNode* activeNode)
         return;  // RETURN
     }
 
+    d_activeNode_p = activeNode;
```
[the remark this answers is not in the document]
678098: Maybe it's worth it to describe this in docs for `d_activeNode_p`

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
```diff
@@ -1322,24 +1293,11 @@ void ClusterProxy::onDispatcherEvent(const mqbi::DispatcherEvent& event)
...
     case mqbi::DispatcherEventType::e_CONFIRM: {
-        const mqbi::DispatcherConfirmEvent* realEvent = event.asConfirmEvent();
-        if (realEvent->isRelay()) {
-            onRelayConfirmEvent(*realEvent);
-        }
-        else {
-            BALL_LOG_ERROR << "#UNEXPECTED_EVENT " << description()
-                           << "CONFIRM event not yet implemented";
-        }
+        BALL_LOG_ERROR << "#UNEXPECTED_EVENT " << description();
```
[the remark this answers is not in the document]
678098: Can be done this way. About `not yet`, we have this log for PUT event log near

678098:
@dorjesinpo thank you for fixes, added some comments in discussions and reassigned. You can assign this to @quarter-note later

pniedzielski:
Assigning back to @dorjesinpo (to put in backlog)

src/groups/mqb/mqbmock/mqbmock_cluster.cpp
```diff
@@ -490,6 +496,29 @@ void Cluster::loadClusterStatus(mqbcmd::ClusterResult* out)
     out->makeClusterStatus();
 }
 
+mqbi::InlineResult::Enum Cluster::sendConfirmInline(
+    BSLS_ANNOTATION_UNUSED int   partitionId,
+    BSLS_ANNOTATION_UNUSED const bmqp::ConfirmMessage& message)
```
678098, line 501:
```suggestion
    BSLA_UNUSED int   partitionId,
    BSLA_UNUSED const bmqp::ConfirmMessage& message)
```
Recent changes

src/groups/mqb/mqbmock/mqbmock_cluster.cpp
```diff
@@ -490,6 +496,29 @@ void Cluster::loadClusterStatus(mqbcmd::ClusterResult* out)
...
+                       const bsl::shared_ptr<bdlbb::Blob>&       options,
+                       const bsl::shared_ptr<bmqu::AtomicState>& state,
+                       bsls::Types::Uint64                       genCount)
+{
+    return d_putFunctor(partitionId,
+                        putHeader,
+                        appData,
+                        options,
+                        state,
+                        genCount);
```
678098, line 519: I think there will be a segfault if we try to call a function that is not set. At least this is how it was in std c++ implementation.

Do you think it's worth handling empty `d_putFunctor` for this mock class?

src/groups/mqb/mqbmock/mqbmock_cluster.cpp
```diff
@@ -240,6 +245,7 @@ Cluster::Cluster(bdlbb::BlobBufferFactory* bufferFactory,
 , d_isLeader(isLeader)
 , d_isRestoringState(false)
 , d_processor()
+, d_putFunctor()
```
678098, line 248:
```suggestion
, d_putFunctor(bsl::allocator_arg, d_allocator_p)
```
Need to handle possibly non-default allocator. This allocator will also be used in `operator=`

src/groups/mqb/mqbmock/mqbmock_cluster.h
```diff
@@ -141,6 +141,18 @@ class Cluster : public mqbi::Cluster {
...
+        int                                       partitionId,
+        const bmqp::PutHeader&                    putHeader,
+        const bsl::shared_ptr<bdlbb::Blob>&       appData,
+        const bsl::shared_ptr<bdlbb::Blob>&       options,
+        const bsl::shared_ptr<bmqu::AtomicState>& state,
+        bsls::Types::Uint64                       genCount)>
+        PutFunctor;
+
+    // Signature of a functor method with one parameter, the processor
+    // handle on which it is being executed.
```
678098, line 154: I think this comment is misleading, since there is more than 1 parameter.
We can remove it or place the updated version in `///` block before `PutFunctor`

src/groups/mqb/mqbi/mqbi_queue.h
```diff
@@ -982,6 +1076,59 @@ class QueueHandleFactory {
...
+}
+
+inline bool InlineResult::isPermanentError(bmqt::AckResult::Enum* ackResult,
+                                           InlineResult::Enum     value)
+{
+    switch (value) {
+    case e_SUCCESS: BSLS_ANNOTATION_FALLTHROUGH;
+    case e_UNAVAILABLE: BSLS_ANNOTATION_FALLTHROUGH;
+    case e_INVALID_PRIMARY: BSLS_ANNOTATION_FALLTHROUGH;
+    case e_INVALID_GEN_COUNT: BSLS_ANNOTATION_FALLTHROUGH;
```
678098, line 1104:
```suggestion
    case e_INVALID_GEN_COUNT: BSLA_FALLTHROUGH;
```
Recent BSLA update

src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
```diff
@@ -1443,20 +1495,38 @@ void RemoteQueue::retransmitPendingMessagesDispatched(
...
+                if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(
+                        mqbi::InlineResult::isPermanentError(&ackResult,
+                                                             rc))) {
+                    BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
+
+                    bmqp::AckMessage ackMessage;
+
+                    ackMessage.setStatus(
+                        bmqp::ProtocolUtil::ackResultToCode(ackResult));
+                    it = nack(it, ackMessage);
```
678098, line 1517: I don't really like the current interface for `mqbi::InlineResult::isPermanentError`. It should be simple yet it's doing several things at once:
1. Checking and returning if an error is permanent.
2. Converting inline result to ack result.
3. Returning ack results via output pointer arg.

I think there is a way to do it simpler:
```suggestion
                if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(
                        mqbi::InlineResult::isPermanentError(rc))) {
                    BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
                    bmqt::AckResult::Enum ackResult = mqbi::InlineResult::toAckResult(rc);

                    bmqp::AckMessage ackMessage;

                    ackMessage.setStatus(
                        bmqp::ProtocolUtil::ackResultToCode(ackResult));
                    it = nack(it, ackMessage);
```

We don't even need to define and pass `ackResult` to a function if there is no error.

src/groups/mqb/mqbi/mqbi_queue.h
```diff
@@ -1173,6 +1324,21 @@ QueueHandleRequesterContext::setRequesterId(RequesterId value)
...
+    const bsl::shared_ptr<bmqst::StatContext>& value)
+{
+    d_statContext_sp = value;
+    return *this;
+}
+
+inline QueueHandleRequesterContext&
+QueueHandleRequesterContext::setInlineClient(InlineClient* inlineClient)
+{
+    d_inlineClient_p = inlineClient;
```
678098, line 1338:
```suggestion
    BSLS_ASSERT_SAFE(inlineClient);

    d_inlineClient_p = inlineClient;
```

From the usage context, `inlineClient` arg is never null. However, we can still add precondition check here just to be sure

src/groups/mqb/mqbi/mqbi_queue.h
```diff
@@ -1134,6 +1283,8 @@ inline QueueHandleRequesterContext::QueueHandleRequesterContext(
 , d_description(original.d_description, allocator)
 , d_isClusterMember(original.d_isClusterMember)
 , d_requesterId(original.d_requesterId)
+, d_statContext_sp(original.d_statContext_sp)
+, d_inlineClient_p(original.d_inlineClient_p)
```
678098, line 1287: As a side note, I am curious why do we even need to define and update this constructor. It's possible that we only store `QueueHandleRequesterContext` in `shared_ptr`

src/groups/mqb/mqbi/mqbi_queue.h
```diff
@@ -982,6 +1076,59 @@ class QueueHandleFactory {
...
+        *ackResult = bmqt::AckResult::e_INVALID_ARGUMENT;
+        break;
+    case e_SELF_PRIMARY: *ackResult = bmqt::AckResult::e_UNKNOWN; break;
+    }
+    return true;
+}
+
+// ----------------------------------
+// struct OpenQueueConfirmationCookie
+// ----------------------------------
```
678098, line 1116:
```suggestion
// -----------------------
// struct OpenQueueContext
// -----------------------
```

src/groups/mqb/mqbi/mqbi_queue.h
```diff
@@ -982,6 +1076,59 @@ class QueueHandleFactory {
...
+, d_stats()
+{
+    // NOTHING
+}
+
+inline OpenQueueContext::OpenQueueContext(QueueHandle* handle)
+: d_handle(handle)
+, d_stats()
+{
+    // NOTHING
```
678098, line 1129: Should we check that `handle` is not null here?
```suggestion
    BSLS_ASSERT_SAFE(handle);
```

src/groups/mqb/mqbi/mqbi_queue.h
```diff
@@ -446,6 +535,8 @@ class QueueHandle {
         unsigned int                   d_upstreamSubQueueId;
         bmqp_ctrlmsg::StreamParameters d_streamParameters;
 
+        bsl::shared_ptr<mqbstat::QueueStatsClient> d_clientStats;
```
678098, line 538:
```suggestion
        bsl::shared_ptr<mqbstat::QueueStatsClient> d_clientStats_sp;
```

src/groups/mqb/mqbi/mqbi_queue.h
```diff
@@ -236,6 +300,31 @@ class QueueHandleRequester {
     handleRequesterContext() const = 0;
 };
 
+// ==================================
+// struct OpenQueueConfirmationCookie
+// ==================================
```
678098, line 305:
```suggestion
// =======================
// struct OpenQueueContext
// =======================
```

src/groups/mqb/mqbi/mqbi_queue.h
```diff
@@ -236,6 +300,31 @@ class QueueHandleRequester {
...
+    // down before the 'openQueue' response got delivered to it.  In this
+    // case, we must rollback upstream state.  This cookie is used for
+    // that: it is initialized to zero (in the 'Cluster' implementation),
+    // and carried over to the original requester of the 'openQueue'.  If
+    // the requester is not able to process the openQueue response, it
+    // needs to set this cookie to the queue handle which it received, so
+    // that the operation can be rolled back.
+
+    QueueHandle*                               d_handle;
+    bsl::shared_ptr<mqbstat::QueueStatsClient> d_stats;
```
678098, line 320:
```suggestion
    bsl::shared_ptr<mqbstat::QueueStatsClient> d_stats_sp;
```

src/groups/mqb/mqbi/mqbi_queue.h
```diff
@@ -169,6 +221,10 @@ class QueueHandleRequesterContext {
     // Unique ID associated with the requester
     // of a queue handle.
 
+    bsl::shared_ptr<bmqst::StatContext> d_statContext_sp;
+
+    InlineClient* d_inlineClient_p;
```
678098, line 226: Can `QueueHandleRequesterContext` outlive `d_inlineClient_p`?

src/groups/mqb/mqbi/mqbi_queue.h
```diff
@@ -115,6 +119,54 @@ class Domain;
...
+        e_CHANNEL_ERROR     = 4,
+        e_INVALID_PARTITION = 5,
+        e_SELF_PRIMARY      = 6
+    };
+    // CLASS METHODS
+    static const char* toAscii(InlineResult::Enum value);
+    static bool        isPermanentError(bmqt::AckResult::Enum* ackResult,
+                                        InlineResult::Enum     value);
+};
+
```
678098, line 140:
```suggestion

// ==================
// class InlineClient
// ==================

```

src/groups/mqb/mqbi/mqbi_queue.h
```diff
@@ -115,6 +119,54 @@ class Domain;
 class Queue;
 class QueueEngine;
 
+// =======================
+// class InlineClient
+// =======================
```
678098, line 124:
```suggestion
// ==================
// class InlineResult
// ==================

```

src/groups/mqb/mqbi/mqbi_queue.h
```diff
@@ -62,6 +62,7 @@
 #include <bmqt_resultcode.h>
 
 #include <bmqc_orderedhashset.h>
+#include <bmqst_statcontext.h>
```
678098, line 65: Can we avoid adding this include by forward-declaring `bmqst::StatContext`?

src/groups/mqb/mqbi/mqbi_queue.cpp
```diff
@@ -48,6 +48,14 @@ bsl::ostream& operator<<(bsl::ostream&                   stream,
...
 bsls::AtomicInt64 QueueHandleRequesterContext::s_previousRequesterId(-1);
 
+// ------------------
+// class InlineClient
+// ------------------
+
+InlineClient::~InlineClient()
+{
+    // NOTHING
+}
```
678098, line 58:
```suggestion
}

```

src/groups/mqb/mqbc/mqbc_clusterstate.h
```diff
@@ -598,9 +602,9 @@ class ClusterState {
     /// `onPartitionPrimaryAssignment()` on each of them, with the
     /// `partitionId` and `node` as parameters.  The bahavior is undefined
     /// unless `partitionId >= 0` and `partitionId < partitionsCount`.
-    ClusterState& setPartitionPrimary(int                  partitionId,
-                                      unsigned int         leaseId,
-                                      mqbnet::ClusterNode* node);
+    ClusterState& setPartitionPrimary(int                 partitionId,
+                                      unsigned int        leaseId,
+                                      ClusterNodeSession* ns);
```
678098, line 607: The comment above still mentions `node`, not the `ns`

src/groups/mqb/mqbc/mqbc_clusterstate.cpp
```diff
@@ -14,6 +14,7 @@
 // limitations under the License.
 
 // mqbc_clusterstate.cpp                                              -*-C++-*-
+#include <mqbc_clusternodesession.h>
```
678098, line 17:
```suggestion
```
Can remove it, it's in the header for this component already

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -345,9 +579,48 @@ inline mqbi::DispatcherClientData& ClusterNodeSession::dispatcherClientData()
...
+                // Do NOT process PUSH in the E_STOPPING state.
+                d_gatePush.close();
+            }
+        }
+        else {
+            // DO process PUTs in the E_STOPPING state.  This is for broadcast
+            // PUTs that "cross" StopRequest.  Since remote 'RemoteQueue' did
+            // not buffer  broadcast PUTs data, "crossed" PUTs will be lost.
+            // But do NOT send to E_STOPPING upstream.
+            d_gatePut.close();
```
678098, line 618: I do not understand this comment.
First, it says "do process puts", but `d_gatePut` is being closed right after.
After this, it says something about broadcast, remote queue, stop request, but there is no code here related to broadcast or any context.
Would be nice to rephrase the comment to show what exactly `d_gatePut` is doing first, or, in other words, plan the comment from the concrete details to higher scope.

Something like this:
```
Closing the PUT gate will prevent from posting PUTs from remote to upstream.
However, this doesn't stop processing PUTs on PRIMARY in Stopping state.
As a side effect, broadcast PUTs in RemoteQueue will be lost, because RemoteQueue is not buffering them.
```

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -296,14 +410,134 @@ class ClusterNodeSession : public mqbi::DispatcherClient,
...
+
+    BSLS_ASSERT_SAFE(result & e_CLOSE);
+    // Do not support more than one writer
+
+    // Spin while locked result > e_CLOSE
+
+    while (result > e_CLOSE) {
+        bslmt::ThreadUtil::yield();
+        result = d_value;
+    }
```
678098, line 441:
```suggestion
    // Spin while locked e_CLOSE < result

    while (e_CLOSE < result) {
        bslmt::ThreadUtil::yield();
        result = d_value;
    }
```
Might be more intuitive use "less" operator, so the lesser arg is on the left

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -296,14 +410,134 @@ class ClusterNodeSession : public mqbi::DispatcherClient,
...
+{
+    const int result = d_value.add(e_ENTER);
+
+    if (result & e_CLOSE) {
+        d_value.subtract(e_ENTER);
+        return false;  // RETURN
+    }
+    else {
+        return true;  // RETURN
+    }
```
678098, line 454:
```suggestion
    return true;
```
Possible to simplify a bit

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -296,14 +410,134 @@ class ClusterNodeSession : public mqbi::DispatcherClient,
...
+
+    (void)result;
+}
+
+inline void AtomicGate::leave()
+{
+    const int result = d_value.subtract(e_ENTER);
+
+    BSLS_ASSERT_SAFE(result >= e_INIT);
+    (void)result;
```
678098, line 472:
```suggestion
    BSLA_MAYBE_UNUSED const int result = d_value.subtract(e_ENTER);
    BSLS_ASSERT_SAFE(e_INIT <= result);
```
Alternative

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -296,14 +410,134 @@ class ClusterNodeSession : public mqbi::DispatcherClient,
...
+}
+
+inline void AtomicGate::open()
+{
+    const int result = d_value.subtract(e_CLOSE);
+
+    BSLS_ASSERT_SAFE(result >= e_INIT);
+    BSLS_ASSERT_SAFE((result & e_CLOSE) == 0);
+
+    (void)result;
```
678098, line 464:
```suggestion
    BSLA_MAYBE_UNUSED const int result = d_value.subtract(e_CLOSE);
    BSLS_ASSERT_SAFE(e_INIT <= result);
    BSLS_ASSERT_SAFE((result & e_CLOSE) == 0);
```
Alternative

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -296,14 +410,134 @@ class ClusterNodeSession : public mqbi::DispatcherClient,
...
 : d_clientStats()
 {
-    d_clientStats.createInplace();
+}
+
+inline void ClusterNodeSession::SubQueueInfo::onEvent(
+    mqbstat::QueueStatsClient::EventType::Enum type,
+    bsls::Types::Int64                         value) const
+{
+    if (d_clientStats) {
```
678098, line 538:
```suggestion
    if (d_clientStats_sp) {
```
It's helpful to add a `_sp` suffix to this field's name, to make `if` condition more clear.

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -245,8 +334,33 @@ class ClusterNodeSession : public mqbi::DispatcherClient,
...
+    sendAck(const bmqp::AckMessage& ackMessage,
+            unsigned int            queueId) BSLS_KEYWORD_OVERRIDE;
+    // Called by the 'queueId' to send the specified 'ackMessage'.
+    //
+    // THREAD: This method is called from the Queue's dispatcher thread.
+
     // ACCESSORS
 
+    GateKeeper& gatePut();
+    GateKeeper& gateConfirm();
```
678098, line 362: These 2 are not accessors because they are not `const`-qualified. Let's move to manipulators

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -245,8 +334,33 @@ class ClusterNodeSession : public mqbi::DispatcherClient,
...
+        BSLS_KEYWORD_OVERRIDE;
+    // Called by the 'queueId' to deliver the specified 'message' with the
+    // specified 'message', 'msgGUID', 'attributes' and 'mps' for the
+    // specified 'subQueueInfos' streams of the queue.
+    //
+    // THREAD: This method is called from the Queue's dispatcher thread.
+
+    mqbi::InlineResult::Enum
+    sendAck(const bmqp::AckMessage& ackMessage,
+            unsigned int            queueId) BSLS_KEYWORD_OVERRIDE;
```
678098, line 354:
```suggestion
            int                     queueId) BSLS_KEYWORD_OVERRIDE;
```

This arg should be `int` the same way as `sendPush` one. Also, we have a special negative value for `queueId`:

```
static const int k_INVALID_QUEUE_ID = -1;
```

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -211,13 +302,11 @@ class ClusterNodeSession : public mqbi::DispatcherClient,
     /// Return a reference to the dispatcherClientData.
     mqbi::DispatcherClientData& dispatcherClientData() BSLS_KEYWORD_OVERRIDE;
 
-    void setNodeStatus(bmqp_ctrlmsg::NodeStatus::Value value);
+    void setNodeStatus(bmqp_ctrlmsg::NodeStatus::Value other,
```
678098, line 305:
```suggestion
    /// @brief Update the status of this node session.
    /// @param other Node status of the node this session is
    ///              connected to.
    /// @param self Node status of the self node in relation to
    ///             this session.
    void setNodeStatus(bmqp_ctrlmsg::NodeStatus::Value other,
```

Might be good to explain better what other/self mean in this context

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -434,10 +712,10 @@ inline int ClusterNodeSession::peerInstanceId() const
     return d_peerInstanceId;
 }
 
-inline const ClusterNodeSession::StatContextSp&
+inline const bsl::shared_ptr<bmqst::StatContext>&
 ClusterNodeSession::statContext() const
 {
-    return d_statContext_sp;
+    return d_queueHandleRequesterContext_sp->statContext();
```
678098, line 718: I understand that you want to keep only 1 source of truth for stat context, so you removed shared_ptr from `ClusterNodeSession`. However, this makes performance slightly worse than it can be.

There is an extra lookup at `d_queueHandleRequesterContext_sp->` where ref counts are checked, and there is an extra jump to another part of the memory. If we call it often, it's better to keep shared_ptr cached in `ClusterNodeSession`. It is a shared pointer after all

src/groups/mqb/mqbc/mqbc_clusternodesession.h
678098: Another thing, might be better to put `queueId` on the first place in args list.
This way, we first answer `where to send?` and after we answer `what to send?`

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -245,8 +334,33 @@ class ClusterNodeSession : public mqbi::DispatcherClient,
...
     void removeAllPartitions();
 
+    //   (virtual: mqbi::QueueHandleClient)
+    mqbi::InlineResult::Enum
+    sendPush(const bmqt::MessageGUID&                  msgGUID,
+             int                                       queueId,
+             const bsl::shared_ptr<bdlbb::Blob>&       message,
+             const mqbi::StorageMessageAttributes&     attributes,
+             const bmqp::MessagePropertiesInfo&        mps,
+             const bmqp::Protocol::SubQueueInfosArray& subQueueInfos)
```
678098, line 344: Might be better to put `queueId` on the first place in args list.
This way, we first answer `where to send?` and after we answer `what to send?`

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -72,13 +157,15 @@ class ClusterNodeSession : public mqbi::DispatcherClient,
...
         /// Stats of this SubQueue, with regards to the client.
-        bsl::shared_ptr<mqbstat::ClusterNodeStats> d_clientStats;
+        bsl::shared_ptr<mqbstat::QueueStatsClient> d_clientStats;
 
         // CREATORS
 
         /// Constructor of a new object, initializes all data members to
         /// default values.
         SubQueueInfo();
+        void onEvent(mqbstat::QueueStatsClient::EventType::Enum type,
```
678098, line 167:
```suggestion

        // ACCESSORS
        void onEvent(mqbstat::QueueStatsClient::EventType::Enum type,
```

Actually, I am also fine with removing `CREATORS` comment as an alternative

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -245,8 +334,33 @@ class ClusterNodeSession : public mqbi::DispatcherClient,
     /// node is the primary.
     void removeAllPartitions();
 
+    //   (virtual: mqbi::QueueHandleClient)
```
678098, line 337:
```suggestion
    //   (virtual: mqbi::InlineClient)
```

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -245,8 +334,33 @@ class ClusterNodeSession : public mqbi::DispatcherClient,
...
+             const mqbi::StorageMessageAttributes&     attributes,
+             const bmqp::MessagePropertiesInfo&        mps,
+             const bmqp::Protocol::SubQueueInfosArray& subQueueInfos)
+        BSLS_KEYWORD_OVERRIDE;
+    // Called by the 'queueId' to deliver the specified 'message' with the
+    // specified 'message', 'msgGUID', 'attributes' and 'mps' for the
+    // specified 'subQueueInfos' streams of the queue.
+    //
+    // THREAD: This method is called from the Queue's dispatcher thread.
+
```
678098, line 351:
```suggestion

    //   (virtual: mqbi::InlineClient)
```

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -51,16 +51,101 @@
 namespace BloombergLP {
 namespace mqbc {
 
-// FORWARD DECLARATION
-class ClusterState;
+// ================
+// class AtomicGate
+// ================
+
+class AtomicGate {
```
678098, line 58: We can move AtomicGate/GateKeeper to its separate source file in the future.
It allows to reduce the size of this header for the reader and make it easier to navigate, and we can also cover it with UTs.

src/groups/mqb/mqbblp/mqbblp_remotequeue.t.cpp
```diff
@@ -216,28 +227,30 @@ bsl::shared_ptr<mqbmock::Queue> TestBench::getQueue()
     return queue_sp;
 }
 
-TestBench::PutEvent::PutEvent(const mqbi::DispatcherPutEvent& event,
-                              mqbi::DispatcherClient*         source)
-: d_header(event.putHeader())
-, d_source(source)
-, d_appData(event.blob())
-, d_options(event.blob())
+TestBench::PutEvent::PutEvent(const bmqp::PutHeader               header,
```
678098, line 230:
```suggestion
TestBench::PutEvent::PutEvent(const bmqp::PutHeader&              header,
```

src/groups/mqb/mqbblp/mqbblp_queuesessionmanager.h
```diff
@@ -345,7 +341,15 @@ inline QueueSessionManager::SubQueueInfo::SubQueueInfo(
...
 , d_queueId(queueId)
 {
-    d_stats.createInplace();
+}
+
+inline void QueueSessionManager::SubQueueInfo::onEvent(
+    mqbstat::QueueStatsClient::EventType::Enum type,
+    bsls::Types::Int64                         value) const
+{
+    if (d_stats) {
```
678098, line 350:
```suggestion
    if (d_stats_sp) {
```
Easier to understand what we implicitly convert to `bool`

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1116,11 +1117,12 @@ mqbi::QueueHandle* RelayQueueEngine::getHandle(
...
-        downstreamInfo,
-        upstreamSubQueueId,
-        mqbi::QueueCounts(handleParameters.readCount(),
-                          handleParameters.writeCount()));
+    mqbi::QueueHandle::SubStreams::const_iterator citSubStream =
+        queueHandle->registerSubStream(
+            downstreamInfo,
+            upstreamSubQueueId,
+            mqbi::QueueCounts(handleParameters.readCount(),
+                              handleParameters.writeCount()));
```
678098, line 1125:
```suggestion
    {
        mqbi::QueueHandle::SubStreams::const_iterator citSubStream =
            queueHandle->registerSubStream(
                downstreamInfo,
                upstreamSubQueueId,
                mqbi::QueueCounts(handleParameters.readCount(),
                                  handleParameters.writeCount()));
        context->d_stats = citSubStream->second.d_clientStats;
     }                          
```
Possible to narrow down the lifetime of this `const_iterator`.

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1135,6 +1137,9 @@ mqbi::QueueHandle* RelayQueueEngine::getHandle(
             &insertResult.first->second.d_handleParameters,
             handleParameters);
     }
+
+    context->d_stats = citSubStream->second.d_clientStats;
```
678098, line 1141:
```suggestion
```

Look at the comment above, possible to assign this right after the iterator is retrieved

src/groups/mqb/mqbblp/mqbblp_queuesessionmanager.h
```diff
@@ -55,9 +56,6 @@ class DomainQualification;
 namespace mqbi {
 class DispatcherClient;
 }
-namespace mqbstat {
-class QueueStatsClient;
-}
 namespace bmqst {
 class StatContext;
```
678098, line 60: Possible to remove this forward declaration if this is included
```
#include <mqbstat_queuestats.h>
```

src/groups/mqb/mqbblp/mqbblp_queuesessionmanager.cpp
```diff
@@ -524,15 +516,16 @@ QueueSessionManager::QueueSessionManager(
 {
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(d_dispatcherClient_p != 0);
-    BSLS_ASSERT_SAFE(d_statContext_p != 0);
+    BSLS_ASSERT_SAFE(statContext);
```
678098, line 519:
```suggestion
    BSLS_ASSERT_SAFE(statContext_sp);
```
Helpful to understand what we implicitly convert to `bool`

src/groups/mqb/mqbblp/mqbblp_queuesessionmanager.cpp
```diff
@@ -524,15 +516,16 @@ QueueSessionManager::QueueSessionManager(
...
     BSLS_ASSERT_SAFE(d_domainFactory_p != 0);
 
     d_requesterContext_sp->setClient(dispatcherClient)
         .setIdentity(clientIdentity)
         .setDescription(dispatcherClient->description())
         .setIsClusterMember(false)
         .setRequesterId(
-            mqbi::QueueHandleRequesterContext ::generateUniqueRequesterId());
+            mqbi::QueueHandleRequesterContext ::generateUniqueRequesterId())
```
678098, line 527:
```suggestion
            mqbi::QueueHandleRequesterContext::generateUniqueRequesterId())
```
Don't know if this is a formatter issue...

src/groups/mqb/mqbblp/mqbblp_queuehandle.h
```diff
@@ -70,6 +70,7 @@ class QueueHandle : public mqbi::QueueHandle {
 
   private:
     // TYPES
+    typedef bsl::shared_ptr<mqbstat::QueueStatsClient> Stats;
```
678098, line 73:
```suggestion
    typedef bsl::shared_ptr<mqbstat::QueueStatsClient> StatsSp;
```

I think it's not good to hide `shared_ptr` from the type name because we can only guess if it is stored by value or is a pointer.

As an example, a few lines below:
```
const Stats d_stats;
```

It is confusing for a reader because they can only guess what is this field. They have to stop and spend time on figuring out that it is actually a shared pointer.

src/groups/mqb/mqbblp/mqbblp_queue.h
```diff
@@ -127,6 +127,7 @@ class Queue BSLS_CPP11_FINAL : public mqbi::Queue {
                              bool          isReconfigure);
 
     void getHandleDispatched(
+        const mqbi::OpenQueueConfirmationCookie& confirmationCookie,
```
678098, line 130: It's a pity that we hide that `mqbi::OpenQueueConfirmationCookie` is actually a shared pointer. This information might improve code readability

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.h
```diff
@@ -689,13 +688,13 @@ class ClusterQueueHelper BSLS_KEYWORD_FINAL
...
-        const bmqp_ctrlmsg::ControlMessage&              request,
-        mqbc::ClusterNodeSession*                        requester,
-        const int                                        peerInstanceId);
+        const bmqp_ctrlmsg::Status&              status,
+        mqbi::QueueHandle*                       queueHandle,
+        const bmqp_ctrlmsg::OpenQueueResponse&   openQueueResponse,
+        const mqbi::OpenQueueConfirmationCookie& confirmationCookie,
+        const bmqp_ctrlmsg::ControlMessage&      request,
+        mqbc::ClusterNodeSession*                requester,
+        const int                                peerInstanceId);
```
678098, line 697:
```suggestion
        int                                      peerInstanceId);
```
Not necessary to const-qualify a basic arithmetic type. This code existed before your PR so we don't need to change it now

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.h
```diff
@@ -648,9 +648,8 @@ class ClusterQueueHelper BSLS_KEYWORD_FINAL
                             const bmqp_ctrlmsg::ControlMessage&   request,
                             mqbc::ClusterNodeSession*             requester);
     void onHandleConfiguredDispatched(
-        const bmqp_ctrlmsg::Status&           status,
+        unsigned int                          qId,
```
678098, line 651:
```suggestion
        int                               queueId,
```

QueueId can be negative

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
```diff
@@ -1352,6 +1244,85 @@ void ClusterProxy::purgeAndGCQueueOnDomain(
...
+    // d_clusterData.electorInfo().electorTerm() if the gate is open
+
+    bsls::Types::Uint64 term = d_clusterData.electorInfo().electorTerm();
+    if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(genCount != term)) {
+        return mqbi::InlineResult::e_INVALID_PRIMARY;  // RETURN
+    }
+
+    bmqt::GenericResult::Enum rc;
+
+    rc = d_activeNode_p->channel().writePut(putHeader, appData, state);
```
678098, line 1315:
```suggestion
    const bmqt::GenericResult::Enum rc = d_activeNode_p->channel().writePut(putHeader, appData, state);
```

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
```diff
@@ -1352,6 +1244,85 @@ void ClusterProxy::purgeAndGCQueueOnDomain(
...
+    }
+
+    return mqbi::InlineResult::e_SUCCESS;
+}
+
+mqbi::InlineResult::Enum ClusterProxy::sendPutInline(
+    BSLS_ANNOTATION_UNUSED int          partitionId,
+    const bmqp::PutHeader&              putHeader,
+    const bsl::shared_ptr<bdlbb::Blob>& appData,
+    BSLS_ANNOTATION_UNUSED const bsl::shared_ptr<bdlbb::Blob>& options,
```
678098, line 1283:
```suggestion
    BSLA_UNUSED int                                 partitionId,
    const bmqp::PutHeader&                          putHeader,
    const bsl::shared_ptr<bdlbb::Blob>&             appData,
    BSLA_UNUSED const bsl::shared_ptr<bdlbb::Blob>& options,
```
BSLA_UNUSED

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
```diff
@@ -1352,6 +1244,85 @@ void ClusterProxy::purgeAndGCQueueOnDomain(
...
+        return mqbi::InlineResult::e_INVALID_PRIMARY;  // RETURN
+    }
+
+    bmqt::GenericResult::Enum rc;
+
+    BSLS_ASSERT_SAFE(d_activeNode_p);
+
+    rc = d_activeNode_p->channel().writeConfirm(message.queueId(),
+                                                message.subQueueId(),
+                                                message.messageGUID());
```
678098, line 1268:
```suggestion
    BSLS_ASSERT_SAFE(d_activeNode_p);
    
    const bmqt::GenericResult::Enum rc = d_activeNode_p->channel().writeConfirm(message.queueId(),
                                                message.subQueueId(),
                                                message.messageGUID());
```

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
```diff
@@ -1352,6 +1244,85 @@ void ClusterProxy::purgeAndGCQueueOnDomain(
     result->makeError().message() = os.str();
 }
 
+mqbi::InlineResult::Enum
+ClusterProxy::sendConfirmInline(BSLS_ANNOTATION_UNUSED int  partitionId,
+                                const bmqp::ConfirmMessage& message)
```
678098, line 1249:
```suggestion
ClusterProxy::sendConfirmInline(BSLA_UNUSED int             partitionId,
                                const bmqp::ConfirmMessage& message)
```

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
```diff
@@ -806,9 +699,7 @@ ClusterProxy::sendRequest(const RequestManagerType::RequestSp& request,
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(dispatcher()->inDispatcherThread(this));
 
-    mqbnet::ClusterNode* activeNode = d_activeNodeManager.activeNode();
-
-    if (!activeNode) {
+    if (!d_activeNode_p) {
```
678098, line 702: Note that `ClusterProxy::processEvent` is called from IO thread.

There is also `ClusterProxy::onActiveNodeDown` that is called from cluster dispatcher thread. It does the following:
```
void ClusterProxy::onActiveNodeDown(...) {
    ...
    d_gateActiveNode.close();
    d_activeNode_p = 0;
}
```

So it's possible that `d_activeNode_p` will change from non-null to null when we run `ClusterProxy::processEvent`. The condition `if (!d_activeNode_p)` might be false, and right after we skip it `d_activeNode_p` becomes NULL. This leads to possible crashes when we dereference it later in this method

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
```diff
@@ -824,28 +715,28 @@ ClusterProxy::sendRequest(const RequestManagerType::RequestSp& request,
...
     }
 
-    request->setGroupId(activeNode->nodeId());
+    request->setGroupId(d_activeNode_p->nodeId());
 
     return d_clusterData.requestManager().sendRequest(
         request,
         bdlf::BindUtil::bind(&mqbnet::ClusterNode::write,
-                             activeNode,
+                             d_activeNode_p,
```
678098, line 736: Note that we pass a value of the pointer that can be changed to NULL in the meantime. How this code should work if there is `void ClusterProxy::onActiveNodeDown` call from cluster dispatcher thread?

src/groups/mqb/mqba/mqba_clientsession.cpp
```diff
@@ -2259,9 +2239,8 @@ void ClientSession::onPutEvent(const mqbi::DispatcherPutEvent& event)
         BSLS_ASSERT_SAFE(queueStatePtr && subQueueInfoPtr);
         BSLS_ASSERT_SAFE(queueStatePtr->d_handle_p);
 
-        subQueueInfoPtr->d_stats->onEvent(
-            mqbstat::QueueStatsClient::EventType::e_PUT,
-            appDataSp->length());
+        subQueueInfoPtr->onEvent(mqbstat::QueueStatsClient::EventType::e_PUT,
```
678098, line 2242: Self note: can remove include of QueueStatsClient if a template function is declared

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -1712,63 +1393,138 @@ void Cluster::onRelayRejectEvent(const mqbi::DispatcherRejectEvent& event)
...
+
+    // This assumes thread-safe access to
+    // d_clusterData.membership().netCluster()->selfNodeId()
+    if (primary->nodeId() ==
+        d_clusterData.membership().netCluster()->selfNodeId()) {
+        return mqbi::InlineResult::e_SELF_PRIMARY;  // RETURN
     }
+
+    bmqt::GenericResult::Enum rc =
+        primary->channel().writeConfirm(id, subId, message.messageGUID());
```
678098, line 1445:
```suggestion
        primary->channel().writeConfirm(message.queueId(), message.subQueueId(), message.messageGUID());
```

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -1712,63 +1393,138 @@ void Cluster::onRelayRejectEvent(const mqbi::DispatcherRejectEvent& event)
...
 
     // This relay-CONFIRM message is enqueued by the RemoteQueue on either
     // cluster (in case of replica) or clusterProxy (in case of proxy).  This
     // is a replica so this node just needs to forward the message to queue's
     // partition's primary node (after appropriate checks).
 
-    const bmqp::ConfirmMessage& confirmMsg = event.confirmMessage();
-    const int                   pid        = event.partitionId();
+    const int id    = message.queueId();
+    const int subId = message.subQueueId();
```
678098, line 1410:
```suggestion
```
Can use directly in 1 place below

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -1712,63 +1393,138 @@ void Cluster::onRelayRejectEvent(const mqbi::DispatcherRejectEvent& event)
...
+    }
+
+    return mqbi::InlineResult::e_SUCCESS;
+}
+
+mqbi::InlineResult::Enum Cluster::sendPutInline(
+    int                                 partitionId,
+    const bmqp::PutHeader&              putHeader,
+    const bsl::shared_ptr<bdlbb::Blob>& appData,
+    BSLS_ANNOTATION_UNUSED const bsl::shared_ptr<bdlbb::Blob>& options,
```
678098, line 1461:
```suggestion
    BSLA_UNUSED const bsl::shared_ptr<bdlbb::Blob>& options,
```

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
```diff
@@ -89,13 +89,13 @@ void allSubstreamsDeconfigured(const CompletionCallback& callback)
 
 // CREATORS
 QueueHandle::Subscription::Subscription(
-    unsigned int                       subId,
+    unsigned int                       downstreamSubQueueId,
```
678098, line 92: Can subQueueId have a special negative value?

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
```diff
@@ -752,6 +821,12 @@ bool QueueHandle::unregisterSubStream(
         }
         d_downstreams[downstreamSubQueueId].reset();
         d_subStreamInfos.erase(infoIter);
+
+        if (subStreamInfo.subId() == bmqp::QueueId::k_DEFAULT_SUBQUEUE_ID) {
+            // cache producer stats to avoid lookup
+            d_producerStats.reset();
```
678098, line 827: The comment says about caching, but the code does the opposite and resets a shared_ptr.

src/groups/mqb/mqbc/mqbc_clusternodesession.cpp
```diff
@@ -186,5 +197,91 @@ void ClusterNodeSession::removeAllPartitions()
...
+    }
+
+    return rc == bmqt::GenericResult::e_SUCCESS
+               ? mqbi::InlineResult::e_SUCCESS
+               : mqbi::InlineResult::e_CHANNEL_ERROR;
+}
+
+mqbi::InlineResult::Enum
+ClusterNodeSession::sendAck(const bmqp::AckMessage& ackMessage,
+                            unsigned int            queueId)
```
678098, line 259:
```suggestion
                           int            queueId)
```
queueId can be negative?

678098 (changes requested):

src/groups/mqb/mqbmock/mqbmock_cluster.cpp
dorjesinpo: `BSLS_ASSERT_SAFE(d_putFunctor);`

src/groups/mqb/mqbi/mqbi_queue.h
dorjesinpo: `d_inlineClient_p` is `ClusterNodeSession` which aggregates `QueueHandleRequesterContext`
so, the answer is no but it is not enforced

src/groups/mqb/mqbc/mqbc_clusternodesession.h
dorjesinpo: this comment better be placed in the `if (other == bmqp_ctrlmsg::NodeStatus::E_AVAILABLE)` clause
Rephrasing like
```
            // Even if self is E_STOPPING, do allow PUTs.
            // Otherwise, broadcast PUTs can be lost in the following scenario:
            // This node sends StopRequest to a downstream while some broadcast
            // PUTs "cross" the StopRequest.  
```

src/groups/mqb/mqbc/mqbc_clusternodesession.h
dorjesinpo: sure
(`(void)result;` in my view does a better job)

src/groups/mqb/mqbc/mqbc_clusternodesession.h
dorjesinpo: but `queueId` is not "where".  It is one of the attributes of the PUSH

src/groups/mqb/mqbblp/mqbblp_queue.h
dorjesinpo: this came as the code transformation.  We can rename it to `OpenQueueConfirmationCookieSp`

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
dorjesinpo: No.  There is `bmqp::Protocol::k_DEFAULT_SUBSCRIPTION_ID` (`0`)

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
dorjesinpo: `ClusterProxy::sendRequest` is dispatcher thread only (`BSLS_ASSERT_SAFE(dispatcher()->inDispatcherThread(this));`)
Sp are `ClusterProxy::onActiveNodeDown`, `ClusterProxy::onActiveNodeUp`

There is `d_activeNode_p` accessing outside of the dispatcher gate nut then it is protected by `d_gateActiveNode`

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
dorjesinpo: this is the dispatcher thread

dorjesinpo:
closing in favor of https://github.com/bloomberg/blazingmq/pull/833

PR #178, opened 2024-01-04
On shutdown, disconnect clients earlier to avoid NACKs

dorjesinpo:
Cluster node graceful shutdown does not work well with SDK connected to the node directly (w/o proxies)
Unlike Proxies, SDK does not receive `StopRequest` and can continue sending PUTs
If PUT arrives after `d_storageManager_mp->processShutdownEvent()`, it will result in `STORAGE_FAILURE NACK`

The solution is to disconnect SDKs before `d_storageManager_mp->processShutdownEvent()` by calling `mwcio::Channel::close()` (which is non-blocking)
The challenge is to make sure the close operation completes (`ClientSession::tearDownImpl` is the last step).  One way of doing that is to change `ClientSession::initiateShutdown` to call the shutdown callback from `tearDownImpl` (unless the session is already teared down).
`Cluster::initiateShutdown` now calls `Cluster::continueShutdown` upon all `StopResponses` _and_ all shutdown callbacks.


quarter-note (commented):
Reviewed.

PR #181, opened 2024-01-09
Performance tune-up

dorjesinpo:
1. Increase Event (batch) size. Keep PUT soft limit
 2. Use bsls::BlockGrowth::BSLS_CONSTANT (instead of geometric) for pools 
 3. Avoid mwcu::MemOutStream while validating in Cluster
 4. StorageManager::processReceiptEvent in IO thread (minus one thread switch)
 5. FileStore generates Receipt at the end of received batch processing
 6. Change FileStore storage builder size adjustemnt logic 
 7. Refactor dispatcherFlush


hallfox:
lgtm, what performance stats does this affect?

dorjesinpo:
> lgtm, what performance stats does this affect?

This affects latencies in the (capmon) tests with multiple queues in the same cluster (cluster thread is a bottleneck)

pniedzielski:
Assigning back to @dorjesinpo (to put in backlog)

PR #183, opened 2024-01-11
mwcu::BlobIterator: remove bsl::iterator inheritance

678098:


src/groups/mwc/mwcu/mwcu_blobiterator.t.cpp
```diff
@@ -274,10 +273,8 @@ static void test3_forwardIterator()
             ASSERT_NE_D("line " << data.d_line, iter2, end);
             const mwcu::BlobPosition p(iter2.position());
             const mwcu::BlobIterator it(iter2++);
-            bsl::iterator_traits<mwcu::BlobIterator>::reference r(*it);
```
678098, line 277: Removed in cpp20 anyway
https://en.cppreference.com/w/cpp/iterator/iterator_traits

src/groups/mwc/mwcu/mwcu_blobiterator.h
```diff
@@ -44,11 +43,16 @@ namespace mwcu {
 // ==================
 
 /// A blob iterator
-class BlobIterator : public bsl::iterator<bsl::forward_iterator_tag,
```
678098, line 47: Deprecated
https://en.cppreference.com/w/cpp/iterator/iterator

The intended way now is to define public iterator typedefs manually in the class declaration

678098: Update: removed iterator typedefs because we don't actually need it

src/groups/mwc/mwcu/mwcu_blobiterator.t.cpp
```diff
@@ -274,10 +273,8 @@ static void test3_forwardIterator()
...
-            char                                                c;
+            char                     c;
             mwcu::BlobUtil::readNBytes(&c, blob, p, 1);
-            ASSERT_EQ_D("line " << data.d_line, c, r);
```
kaikulimu, line 280: You should keep this assertion. Since `r` was initialized as `::reference r(*it);`, please check whether 

```
ASSERT_EQ_D("line " << data.d_line, c, *it);
```

makes sense.

kaikulimu (changes requested):
Minor comments

src/groups/mwc/mwcu/mwcu_blobiterator.t.cpp
678098: To make this change, I had to return iterator typedefs

kaikulimu (approved):
LGTM

Issue #141, opened 2023-11-05
Fix static local variable multi-threaded initialization


678098:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Place where the problem might occur:

https://github.com/bloomberg/blazingmq/blob/a6f119b17930c7aa5fcc2d285048f2affc7dada3/src/groups/bmq/bmqt/bmqt_subscription.cpp#L46

This static local variable initialization is not thread safe in general. The probability to observe a bug is small though.

Here are some details and advices:

https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/group__bslmt__once.html#_details

### Expected Behavior

Here is how we did a safe initialization for another component:

https://github.com/bloomberg/blazingmq/blob/08a6d16b1b73d1d71113f80f7c22fcdd3e5392d8/src/groups/bmq/bmqt/bmqt_correlationid.cpp#L33-L42

Will be good to do the same for the problematic line of code above.

### Steps To Reproduce

Try to build multiple `SubscriptionHandle`s at the same time from different threads just after the app starts.

### BlazingMQ Version

0.90.18

syuzvinsky:
I am taking this

678098:
Fixed

PR #185, opened 2024-01-23
Protect static initialization in SubscriptionHandle

syuzvinsky:
Fix:#141
Before this fix is applied, creating multiple SubscriptionHandles at the same time from different threads just after the app starts may have led to duplication of SubscriptionHandle's Ids.


syuzvinsky:
@678098 can you review this please?

678098 (approved):
LGTM

PR #187, opened 2024-01-25
Do not check for isListening when negotiationComplete

dorjesinpo:
Broker can initiate and negotiate connection before `startListening`.
In this case, cannot close the channel before newly created cluster `startDispatched` (the latter relies on read callback).


dorjesinpo:
> This looks good to me, can we update with the latest?

There is test failure - `test_queue_close.py::test_close_while_reopening[multi_node]` - that seems to be relevant.  On the first glance, timing changes and cluster fails to synchronize.  We need to spend more time looking but this is not the highest priority at the moment.

PR #200, opened 2024-02-13
Feat: JSON printer for admin commands

678098:
There are several different ways to convert `mqbcmd::Result` to structured json.
Keeping in mind that we have more than 20 different nested components for `mqbcmd::Result`, here is an overview of possible methods:

### 1. Manually print to the stream to produce the needed json structure

Pros:
- no temporary objects = no overhead

Cons:
- the largest amount of code to write among all these options (>1500 lines rough estimate, without tests)
- need to take care of both: json data structure (layout) and json formatting (separators, indentation etc)
- because of the above, easy to make mistakes

### 2. Convert `mqbcmd::Result` to `bdld::Datum` and encode it to json-string using `baljsn::DatumUtil::encode`

Pros:
- need to take care of json data structure (layout) only
- json formatting performed automatically, possible to switch between **PRETTY** and **COMPACT** json encodings
- `bdld::Datum` allows to build nodes using string references, which lowers overhead

Cons:
- moderate amount of code to write (~1200 lines)
- overhead of using a temporary representation `bdld::Datum` object
- not convenient `bdld::Datum` API for this task, need to take care of allocators and object construction boilerplate everywhere

### 3. Convert `mqbcmd::Result` to `bdljsn::Json` and encode it to json-string using `bdljsn::JsonUtil::write`

Pros:
- need to take care of json data structure (layout) only
- the most convenient syntax
- json formatting performed automatically, possible to switch between **PRETTY** and **COMPACT** json encodings

Cons:
- moderate amount of code (~1000 lines)
- overhead of using a temporary representation `bdljsn::Json` object

### 4. Directly encode `mqbcmd::Result` to json-string using `baljsn::Encoder::encode`

Pros:
- the cheapest solution in terms of code lines, it's almost free (~10 lines)
- no need to test json conversion, because it is performed by a well-tested tool
- xsd schema accurately represents output json-string structure
- `encode` will automatically pick up any changes made in xsd schema, so no changes needed in this case in `JsonPrinter`
- json formatting performed automatically, possible to switch between **PRETTY** and **COMPACT** json encodings
- no temporary objects = no overhead

Cons:
- we lose control of output json structure entirely on a printer level
- we don't have any generated summary fields in this json anymore (like total bytes summary etc), note: input data is still available to calculate these summaries in this json
- we expose fields which we might not want to expose in this json, for example, `help - isPlumbing`
- we lose data post-processing which might be interesting for us, for example, purged queues sorting

Some of the disadvantages of (4) could be mitigated by changing xsd schema (adding missing fields there) and populating it on a `Result` construction level.

For me, options (3) "good" and (4) "free" are okay. We need to decide, if limitations with option (4) are critical or okay for us.


kaikulimu:
Only Options (3) and (4) make sense to me. For (4), is it possible to generate the summary fields and hide the non-exposed fields first before calling `baljsn::Encoder::encode`? What are the challenges for this? Would you need a new `xsd` schema type?

678098:
> Only Options (3) and (4) make sense to me. For (4), is it possible to generate the summary fields and hide the non-exposed fields first before calling `baljsn::Encoder::encode`? What are the challenges for this? Would you need a new `xsd` schema type?

1. It's possible to generate summary fields before with any postprocessing. To do this, we need:
- Extend xsd schema with new summary fields.
- Remove these summary fields computation from `mqbcmd::HumanPrinter` (maybe, it was not the right place to do it in the first place).
- Instead, compute these summary fields in places where we prepare `mqbcmd::InternalResult`. this means that the code will go to the concrete dispatchers which executes admin commands.

There's still a challenge here. We have several fields in xsd which contains "json configs" encoded as strings. With `baljsn::Encoder` alone, it doesn't look possible to convert these json-strings to just jsons, to make a common json structure of a message.

The explanation is tricky, so here is an example of what I mean:

```
{
    "myResponse": {
        "abc": true,
        "myJsonConfig": "{\"value\":123}"
    }
}
```

Here, look at the field "myJsonConfig", it contains a json-string, but we probably want to have it extended to the outer json structure like this:

```
{
    "myResponse": {
        "abc": true,
        "myJsonConfig": {
            "value":123
        }
    }
}
```

It's not possible with `baljsn::Encoder` (4)-th approach.

2. To hide non-exposed fields, we need something like `transient` keyword from Java, to be used in our xsd schemas. Asked BDE about this.

https://en.wikibooks.org/wiki/Java_Programming/Keywords/transient

678098:
There are at least 3 ways to set up admin command output format. Here are they:

1. Extend `bmqp_ctrlmsg.xsd`:

```
  <simpleType name='EncodingFormat'>
    <annotation>
      <documentation>
        Enumeration of the various encoding format versions.
      </documentation>
    </annotation>
    <restriction base='string' bdem:preserveEnumOrder='1'>
      <enumeration value='E_TEXT'         bdem:id='0'/>
      <enumeration value='E_JSON_COMPACT' bdem:id='1'/>
      <enumeration value='E_JSON_PRETTY'  bdem:id='2'/>
    </restriction>
  </simpleType>

  <complexType name='AdminCommand'>
    <annotation>
      <documentation>
        This request is sent by the admin client to execute the specified
        command on the broker.  Command is expected in a human-friendly text
        format, to get a list of available commands 'help' command can be sent.

        The optionally specified 'outputFormat' specifies the format of the
        output result of this command.
      </documentation>
    </annotation>
    <sequence>
      <element name='command'      type='string'/>
      <element name='outputFormat' type='tns:EncodingFormat' default='E_TEXT'/>
    </sequence>
  </complexType>
```

At first glance, the solution looks beautiful, but in reality only the schema is beautiful, the way to pass this arg to `mqba::Application::processCommand` is difficult. Also, this option doesn't cover encoding of results sent via named pipe.

2. Add admin command to switch global encoding of `mqba::Application`.

```
CMD ENCODING JSON_PRETTY
CMD ENCODING JSON_COMPACT
CMD ENCODING TEXT
```

Other options: `SET_ENCODING`, `SET_OUTPUT_ENCODING`, `OUTPUT_ENCODING`

It will work with both pipe/tcp clients. It is easy to implement, and it will not make interfaces more complicated. But the state is shared, so if 2 admin clients connect at the same time, or if they periodically reconnect with the assumption that the encoding was already set, it will cause troubles.

3. Add optional prefix before any admin command.

```
CMD ENCODING TEXT BROKERCONFIG DUMP
CMD ENCODING JSON_PRETTY CLUSTERS LIST
CMD ENCODING JSON_COMPACT HELP
```

Note: `CMD BROKERCONFIG DUMP` with missed `ENCODING` assumes a default `TEXT` encoding type for backward compatibility.

This solution is easy to implement, local (specific to command), doesn't affect existing interfaces in a bad way, works for both pipe/tcp clients.

Personally I like (2) and (3), and doesn't like what should be done with interfaces with option (1).

kaikulimu:
> There are at least 3 ways to set up admin command output format. Here are they:
> 
> ### 1. Extend `bmqp_ctrlmsg.xsd`, to send admin commands with TCP client with an extra encoding option:
> ```
>   <simpleType name='EncodingFormat'>
>     <annotation>
>       <documentation>
>         Enumeration of the various encoding format versions.
>       </documentation>
>     </annotation>
>     <restriction base='string' bdem:preserveEnumOrder='1'>
>       <enumeration value='E_TEXT'         bdem:id='0'/>
>       <enumeration value='E_JSON_COMPACT' bdem:id='1'/>
>       <enumeration value='E_JSON_PRETTY'  bdem:id='2'/>
>     </restriction>
>   </simpleType>
> 
>   <complexType name='AdminCommand'>
>     <annotation>
>       <documentation>
>         This request is sent by the admin client to execute the specified
>         command on the broker.  Command is expected in a human-friendly text
>         format, to get a list of available commands 'help' command can be sent.
> 
>         The optionally specified 'outputFormat' specifies the format of the
>         output result of this command.
>       </documentation>
>     </annotation>
>     <sequence>
>       <element name='command'      type='string'/>
>       <element name='outputFormat' type='tns:EncodingFormat' default='E_TEXT'/>
>     </sequence>
>   </complexType>
> ```
> 
> At first glance, the solution looks beautiful, but in reality only the schema is beautiful, the way to pass this arg to `mqba::Application::processCommand` is difficult. Also, this option doesn't cover encoding of results sent via named pipe.
> 
> ### 2. Add admin command to switch global encoding of `mqba::Application`.
> ```
> CMD ENCODING JSON_PRETTY
> CMD ENCODING JSON_COMPACT
> CMD ENCODING TEXT
> ```
> 
> Other options: `SET_ENCODING`, `SET_OUTPUT_ENCODING`, `OUTPUT_ENCODING`
> 
> It will work with both pipe/tcp clients. It is easy to implement, and it will not make interfaces more complicated. But the state is shared, so if 2 admin clients connect at the same time, or if they periodically reconnect with the assumption that the encoding was already set, it will cause troubles.
> 
> ### 3. Add optional prefix before any admin command.
> ```
> CMD ENCODING TEXT BROKERCONFIG DUMP
> CMD ENCODING JSON_PRETTY CLUSTERS LIST
> CMD ENCODING JSON_COMPACT HELP
> ```
> 
> Note: `CMD BROKERCONFIG DUMP` with missed `ENCODING` assumes the default `TEXT` encoding type for backward compatibility.
> 
> This solution is easy to implement, local (specific to command), doesn't affect existing interfaces in a bad way, works for both pipe/tcp clients.
> 
> Personally I like (2) and (3), and doesn't like what should be done with interfaces with option (1).

I prefer (3) over (2) for the reason that it is stateless. There won't be any trouble due to reconnection, etc.

kaikulimu:
> > Only Options (3) and (4) make sense to me. For (4), is it possible to generate the summary fields and hide the non-exposed fields first before calling `baljsn::Encoder::encode`? What are the challenges for this? Would you need a new `xsd` schema type?
> 
> 1. It's possible to generate summary fields before with any postprocessing. To do this, we need:
> 
> * Extend xsd schema with new summary fields.
> * Remove these summary fields computation from `mqbcmd::HumanPrinter` (maybe, it was not the right place to do it in the first place).
> * Instead, compute these summary fields in places where we prepare `mqbcmd::InternalResult`. this means that the code will go to the concrete dispatchers which executes admin commands.
> 
> There's still a challenge here. We have several fields in xsd which contains "json configs" encoded as strings. With `baljsn::Encoder` alone, it doesn't look possible to convert these json-strings to just jsons, to make a common json structure of a message.
> 
> The explanation is tricky, so here is an example of what I mean:
> 
> ```
> {
>     "myResponse": {
>         "abc": true,
>         "myJsonConfig": "{\"value\":123}"
>     }
> }
> ```
> 
> Here, look at the field "myJsonConfig", it contains a json-string, but we probably want to have it extended to the outer json structure like this:
> 
> ```
> {
>     "myResponse": {
>         "abc": true,
>         "myJsonConfig": {
>             "value":123
>         }
>     }
> }
> ```
> 
> It's not possible with `baljsn::Encoder` (4)-th approach.
> 
> 2. To hide non-exposed fields, we need something like `transient` keyword from Java, to be used in our xsd schemas. Asked BDE about this.
> 
> https://en.wikibooks.org/wiki/Java_Programming/Keywords/transient

1. Why do we have fields such as `"myJsonConfig": "{\"value\":123}"`? In the current `mqbcmd::Result` output, are these json-strings expanded to jsons before returning? 
2. Regarding non-exposed fields, so are we able to implement this `transient`-like logic trivially in C++?

src/integration-tests/test_admin_client.py
```diff
@@ -96,7 +102,96 @@ def test_breathing(single_node: Cluster):
...
+    # Stage 1: encode as TEXT
+    cmds = [json.dumps({"help": {}, "encoding": "TEXT"}), "ENCODING TEXT HELP"]
+    for cmd in cmds:
+        res = admin.send_admin(cmd)
+        assert "This process responds to the following CMD subcommands:" in res
+
+    # Stage 2: encode as JSON_COMPACT
+    cmds = [
+        json.dumps({"help": {}, "encoding": "JSON_COMPACT"}),
+        "ENCODING JSON_COMPACT HELP",
```
678098, line 155: Is it okay to use underscores in plaintext admin commands?
Why do we avoid using underscores in all cmds, like `BROKERCONFIG DUMP`?

src/groups/mqb/mqbcmd/mqbcmd_parseutil.cpp
```diff
@@ -267,6 +291,41 @@ int parseCommand(Command* command, bsl::string* error, WordGenerator next)
...
+    if (equalCaseless(word, "TEXT")) {
+        *format = EncodingFormat::Value::TEXT;
+        return 0;  // RETURN
+    }
+    else if (equalCaseless(word, "JSON_COMPACT")) {
+        *format = EncodingFormat::Value::JSON_COMPACT;
+        return 0;  // RETURN
+    }
+    else if (equalCaseless(word, "JSON_PRETTY")) {
+        *format = EncodingFormat::Value::JSON_PRETTY;
```
678098, line 320: Possible to use `EncodingFormat::fromString`, but it is case sensitive. Probably we don't want to mix case sensitive `fromString` here with case insensitive plaintext admin commands.

src/groups/mqb/mqbcmd/mqbcmd_parseutil.cpp
```diff
@@ -223,40 +223,64 @@ int parseClearCache(ClearCache*              command,
                     WordGenerator            next,
                     const bslstl::StringRef& commandPrefix);
 
+int parseEncodingFormat(EncodingFormat::Value* format,
+                        bsl::string*           error,
+                        WordGenerator&         next);
```
678098, line 228: Might be worth it to refactor everything in this file to pass all args as references

678098:
The current implementation supports admin commands with encodings passed like this (note that no encoding means default TEXT encoding for backward compatibility):
```
HELP
ENCODING TEXT HELP
ENCODING JSON_COMPACT HELP
ENCODING JSON_PRETTY HELP
{"help": {}}
{"help": {}, "encoding": "TEXT"}
{"help": {}, "encoding": "JSON_COMPACT"}
{"help": {}, "encoding": "JSON_PRETTY"}
```

Note that incorrectly passed encoding option will make command decoding fail:
```
ENCODING
ENCODING TEXT
ENCODING unexpected HELP
{"help": {}, "encoding": "unexpected"}
```

src/integration-tests/test_admin_client.py
```diff
@@ -7,7 +7,11 @@
 import re
 from typing import Dict, Optional, Tuple
 
-from blazingmq.dev.it.fixtures import Cluster, single_node, order  # pylint: disable=unused-import
+from blazingmq.dev.it.fixtures import (
+    Cluster,
+    single_node,
+    order,
```
jll63, line 13: `order` is not used. See example [here](https://github.com/bloomberg/blazingmq/blob/main/src/integration-tests/test_breathing.py#L22).

Also, let's re-order the imports with `isort --profile black`:

```
from blazingmq.dev.it.fixtures import (  # pylint: disable=unused-import
    Cluster,
    order,
    single_node,
)
```

jll63 (changes requested):
Just one small remark.

src/integration-tests/test_admin_client.py
678098: Thanks @jll63!
Set up the order, applied formatting and imports reordering

jll63 (approved):

PR #204, opened 2024-02-16
Feat[MQB]: Support listening on multiple ports

hallfox:
This extends the broker config to support listening to connections on multiple ports. This lays the groundwork for future extensions that will be added to `TcpInterfaceListener`. The broker will behave the same without the listener config, but with the listeners added the broker will open new listen operations for each one, ignoring the `port` field in the `networkInterfaces`. We also added an extra component for validating these new configs.

### DIFF
--- a/src/groups/mqb/mqbcfg/doc/mqbcfg.txt
+++ b/src/groups/mqb/mqbcfg/doc/mqbcfg.txt
@@ -10,11 +10,12 @@ broker, as well as the generated corresponding messages component.
 
 /Hierarchical Synopsis
 /---------------------
- The 'mqbcfg' package currently has 2 components having 2 level of physical
+ The 'mqbcfg' package currently has 3 components having 2 levels of physical
  dependency.  The list below shows the hierarchical ordering of the components.
 ..
   1. mqbcfg_messages
   2. mqbcfg_brokerconfig
+  2. mqbcfg_tcpinterfaceconfigvalidator
 ..
 
 /Component Synopsis
@@ -23,3 +24,5 @@ broker, as well as the generated corresponding messages component.
 :      BMQBroker generated configuration messages.
 : 'mqbcfg_brokerconfig'
 :      Provide global access to broker configuration.
+: 'mqbcfg_tcpinterfaceconfigvalidator'
+:      Provide validation logic for `appConfig/networkInterfaces/tcpInterface/listeners` in the broker configuration.
--- a/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
@@ -19,16 +19,18 @@
 #include <mqbscm_version.h>
 /// Implementation Notes
 ///====================
-// When a channel is being created, the following methods are always called, in
-// order, regardless of the success or failure of the negotiation:
-//: o !channelStateCallback!
-//: o !negotiate!
-//: o !negotiationComplete!
-//
-// When a channel goes down, 'onClose()' is the only method being invoked.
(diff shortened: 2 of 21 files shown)


src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp
```diff
@@ -389,6 +390,7 @@ int TransportManager::start(bsl::ostream& errorDescription)
 
     // Create and start the TCPInterface, if any
     const mqbcfg::AppConfig& brkrCfg = mqbcfg::BrokerConfig::get();
+
```
678098, line 393: This file with its changes should probably be excluded from the PR

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -1269,20 +1395,20 @@ void TCPSessionFactory::stop()
...
+    BSLS_ASSERT_SAFE(d_listenContexts.find(port) == d_listenContexts.cend());
 
     // Maintain ownership of 'OperationContext' instead of passing it to
     // 'ChannelFactory::listen' because it may delete the context
     // (on stopListening) while operation (readCallback / negotiation) is in
     // progress.
-    OperationContext* context = new (*d_allocator_p) OperationContext();
-    d_listenContext_mp.load(context, d_allocator_p);
-    bsl::shared_ptr<OperationContext> contextSp;
-    contextSp.reset(context, bslstl::SharedPtrNilDeleter());
```
dorjesinpo, line 1285: Yeah, this insisting on the `ManagedPtr` looks strange

dorjesinpo (commented):
Looks good and straightforward
Few comments, and probably will need to take another look after tests are fixed

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
hallfox: Yeah I'm not in love with this pattern of trading around `ManagedPtr`s as references wrapped in a `shared_ptr`, either way now there's a reason to make it shared so this doesn't even work anymore 😄

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -241,6 +242,95 @@ void stopChannelFactory(mwcio::ChannelFactory* channelFactory)
...
+                bsl::unique(sorted.begin(), sorted.end());
+            return sorted.end() == newEnd;
+        }
+    };
+
+    // PRIVATE DATA
+    UniqueProperty d_unique;
+
+    // PRIVATE STATIC FUNCTIONS
+    static int portView(const mqbcfg::TcpInterfaceListener& listener)
```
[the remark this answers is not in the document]
hallfox: It could create a view, so technically no. The codegen'd API doesn't give me const ref access to this field, so it's a little unfortunate, but It's sufficient enough for our purposes.

src/groups/mqb/mqbcfg/doc/mqbcfg.txt
```diff
@@ -15,6 +15,7 @@ broker, as well as the generated corresponding messages component.
 ..
   1. mqbcfg_messages
   2. mqbcfg_brokerconfig
+  3. mqbcfg_tcpinterfaceconfigvalidator
```
chrisbeard, line 18: nit: This adds another component/level, the synopsis can be updated

pniedzielski (approved):
Merge conflicts, can review again when you fix them.

pniedzielski (approved):
let's ship it

PR #207, opened 2024-02-28
Mark Prometheus integration test as NOT a pytest

jll63:
* Mark Prometheus integration test as NOT a `pytest`. It was picked by `pytest` when run from the root of the repository.
* Black-format the test.


678098 (approved):
LGTM!

PR #218, opened 2024-03-15
Describe assertion failure

syuzvinsky:
Adding a text describing why assertion failed in `Application::printFinalStats()` in bmqtool


pniedzielski (changes requested):
CI format check is catching the double space before `&&`; otherwise lgtm

678098 (approved):

pniedzielski (approved):

Issue #213, opened 2024-03-14
Possible `int` overflow in `bmqt::UriParser`


678098:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Steps To Reproduce

However, I can see one very unprobable scenario with a possible bug here.

1. Within `UriParse::initialize` we increment `int s_initialized` counter in a thread-safely manner (within the locked mutex).
https://github.com/bloomberg/blazingmq/blob/342b608b1f7e2525d22c4c3fbeb4039880fb63c9/src/groups/bmq/bmqt/bmqt_uri.cpp#L195

2. We early return if this counter is greater than 1, meaning, that some other code has already incremented it and initialized a shared state which is const-read-thread-safe.

3. However, we do `bmqt::UriParse::initialize` every time we call a constructor for `bmqt::Uri`, and do not deinitialize `bmqt::UriParser` until we call a destructor `bmqt::Uri::~Uri`.

4. This means, that if we are able to keep approx 2^31 `bmqt::Uri`s in memory, we'll get `int` overflow.

5a. Because of this possible overflow, the condition for early exit will fail on a new `bmqt::Uri` construction, and we might re-initialize some const-read-thread-safe `RegEx`, when some other thread might use it

5b. It is also possible to just launch `bmqt::UriParser::initialize` 2.2kkk times to get overflow. Even if we do call `bmqt::UriParser::shutdown` the same number of times afterwards (according to contract), there will be a problem.

### BlazingMQ Version

0.90.21

### Anything else?

There are 2 possible solutions.

1. The minimal change (and safe) solution is to change the counter type `int` -> `bsls::Types::Int64`, so it will be impossible to reach overflow.

2. The good solution is to give up on a custom reference counting of `bmqt::UriParser::initialize` and initialize it once and for all as a singleton, using the appropriate tool to initialize: `BSLMT_ONCE_DO` https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/group__bslmt__once.html

The greatest benefit of this approach is that we don't lock the mutex every time we need to build a `bmqt::Uri`. As you can see here (https://github.com/bloomberg/bde/blob/88f20d8057d462e2bdaf27b757ff3bf335555dcd/groups/bsl/bslmt/bslmt_once.cpp#L25), `bslmt::Once` can check the need to do something without locking the internal mutex, when the state of once was already decided. As a result, we'll be able to work with `bmqt::Uri`s much, much faster from different threads.

[...]

PR #222, opened 2024-03-19
Fix: prevent possible int32 overflow in bmqt::UriParser

678098:
https://github.com/bloomberg/blazingmq/issues/213


src/groups/bmq/bmqt/bmqt_uri.cpp
```diff
@@ -193,6 +193,7 @@ void UriParser::initialize(bslma::Allocator* allocator)
     //       not work correctly with versions of IBM xlc12 released following
     //       the 'Dec 2015 PTF'.
     ++s_initialized;
+    BSLS_ASSERT(s_initialized > 0);
```
678098, line 196: This `assert` might be an overkill if we use int64 type, because it's way too difficult to cause an overflow of int64 variable here by incrementing by one.

We have another option to keep a counter as `int` like before or use `AtomicInt` and fail on this `assert` on 2.147kk calls.

src/groups/bmq/bmqt/bmqt_uri.t.cpp
```diff
@@ -657,6 +657,28 @@ static void test7_testLongUri()
...
+    // Disable the default allocator check. When 'bmqt::Uri'
+    // is created from a long uri string error message is
+    // logged via 'BALL_LOG_ERROR' which allocates using
+    // the default allocator.
+
+    mwctst::TestHelper::printTestName("URI PARSER OVERFLOW TEST");
+
+    // Initialize bmqt::UriParser many times, and then deinitialize
+    // it the same number of times according to the usage contract.
+    const bsls::Types::Int64 k_INIT_NUM = 2200000000LL;
```
pniedzielski: Would `static_cast<int64_t>( numeric_limits<int32_t>::max() ) + 1` (or corresponding BDE-ism) be clearer?  It's a somewhat lower number of `initialize`/`shutdown`s, but says more clearly what we're trying to test.

src/groups/bmq/bmqt/bmqt_uri.t.cpp
```diff
@@ -657,6 +657,28 @@ static void test7_testLongUri()
...
+    for (size_t i = 0; i < k_INIT_NUM; i++) {
```
pniedzielski: Does this give a signed/unsigned comparison warning?  I think picking either `size_t` or `Int64` is the right path here.

pniedzielski: Oh nvm a later commit did exactly this change.

src/groups/bmq/bmqt/bmqt_uri.cpp
pniedzielski: Signed integer overflow is undefined behavior.  Propose moving this assert one line prior and changing it to `s_initialized < numeric_limits<int64_t>::max()` (or corresponding BDE-ism).

src/groups/bmq/bmqt/bmqt_uri.t.cpp
```diff
@@ -657,28 +657,6 @@ static void test7_testLongUri()
     bmqt::UriParser::shutdown();
 }
 
-static void test8_testUriParserOverflow()
```
pniedzielski: Why is this removed again?  Does it take too long to run?  It seems like a useful test.

pniedzielski (changes requested):

src/groups/bmq/bmqt/bmqt_uri.t.cpp
678098: I removed this test from this branch, because it was too slow, 43 seconds on my M2.

src/groups/bmq/bmqt/bmqt_uri.t.cpp
678098: 43 seconds, and overall this test doesn't test a useful behaviour and instead it tests if we have a bug or not.

pniedzielski (approved):
👍🏻

PR #223, opened 2024-03-19
CI: print expected formatter output on failure

678098:
https://github.com/bloomberg/blazingmq/actions/runs/8345560716/job/22840749145

The example formatting check failure with the desired output.

As a bonus, we remove an extra dependency and the check takes 6 seconds now instead of 6 minutes.

### DIFF
--- a/.github/workflows/formatting-check.yaml
+++ b/.github/workflows/formatting-check.yaml
@@ -1,0 +1,26 @@
+name: Formatting check
+
+on:
+  pull_request:
+    types:
+      - "opened"
+      - "reopened"
+      - "synchronize"
+      - "labeled"
+      - "unlabeled"
+
+jobs:
+  formatting_check:
+    name: Formatting Check
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        path:
+          - 'src'
+    steps:
+      - uses: actions/checkout@v4
+      - name: clang-format style check
+        run: |
+          export FORMAT_DIFF=`git clang-format-14 --diff -q`
+          echo ${FORMAT_DIFF}
+          if [ -n "${FORMAT_DIFF}" ]; then exit 1; fi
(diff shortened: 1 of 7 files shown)


678098:
https://github.com/bloomberg/blazingmq/actions/runs/8345560716/job/22840749145

The example formatting check failure

.github/workflows/formatting-check.yaml
```diff
@@ -0,0 +1,23 @@
...
+  formatting_check:
+    name: Formatting Check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: clang-format style check
+        run: |
+          git clang-format-14 --diff -q origin/main | tee format_diff.txt
```
pniedzielski, line 22: Do we want to diff against `origin/main` or against the common ancestor of `origin/main` and this branch ([`git merge-base`](https://git-scm.com/docs/git-merge-base))?  Not sure which is right in this context, especially when you take into account merge commits in a PR or `main` getting changes of its own.

pniedzielski (changes requested):
I am so very on-board with this change, there was no need to use a pre-cooked Docker-container action for this.  One question in the `formatting-check.yaml` changes, but if the behavior as submitted is right, rerequest a review and I'll approve.

.github/workflows/formatting-check.yaml
678098: For this, I believe we can keep things simple now.
What matters, if the code going to `origin/main` is formatted in the end. It might not work as intended for complex merges from non-main branch to other non-main branch, but it should be fine because at one point we'll decide to merge that non-main branch to origin/main and all formatter problems will arise to fix.

Personally, I prefer this fix to be able to do small patches more easily, without need to switch branches to run clang-format. It might also help with external contributions.

pniedzielski: Let's keep it as is now then.  I'm still not sure what will happen if `main` gets ahead of the PR branch, but we'll have to do a rebase anyway before merging, so it should be fine.  I'll dismiss my review and make a new approve one.

pniedzielski (approved):

PR #224, opened 2024-03-21
ConfigureStream unconditionally

dorjesinpo:
Retire `ConfigureQueueStream`
SDK sends `ConfigureStream` unconditionally.  This must be in accord with broker's MessageProperties V2 and configureStream configurations.


kaikulimu (dismissed):
LGTM. Is this the change we need to make a subscription-aware libbmq release?

PR #225, opened 2024-03-22
blazingmq.util.logging: store switch value in args namespace

jll63:
Make `bmq.util.logging.Action` store the value. Even though the action sets the log levels appropriately, this change makes it possible to detect whether the switch was specified.


pniedzielski (approved):
👍🏻

PR #228, opened 2024-03-27
Avoid flushList upon initiateShutdown

dorjesinpo:
The existing code takes precautions to avoid adding `ClientSession` to the `d_flushList` when `tearDown`.
But `ClientSession::initiateShutdown` can be called twice (by multiple clusters/proxies) and the second call while detecting the state, re-enables `d_flushList`.

The solution could be to make sure `initiateShutdown` does not re-enable `d_flushList`.


kaikulimu (approved):
LGTM

PR #230, opened 2024-03-29
fix: do not use defaultSubQueueInfoArray()

dorjesinpo:
Any broadcast PUSH for now needs real subscription id.  Cannot use `defaultSubQueueInfoArray()`, it will cause `#CLIENT_INVALID_PUSH` or `Assertion failed: cit != d_subscriptions.end()`


kaikulimu (approved):
LGTM

PR #231, opened 2024-03-29
Wait ACKs for all messages posted by bmqtool

syuzvinsky:
Instead of waiting for Ctrl+C in auto-posting mode, bmqtool will be counting ACKs and terminate after all posted messages are acknowledged.

### DIFF
--- a/src/applications/bmqtool/m_bmqtool_application.cpp
+++ b/src/applications/bmqtool/m_bmqtool_application.cpp
@@ -768,6 +768,12 @@ void Application::onMessageEvent(const bmqa::MessageEvent& event)
 
             // Write to log file
             d_fileLogger.writeAckMessage(message);
+
+            if (d_numExpectedAcks != 0 &&
+                d_numExpectedAcks == ++d_numPostedAcknowledged) {
+                BALL_LOG_INFO << "All posted messages have been acknowledged";
+                d_shutdownSemaphore_p->post();
+            }
         }
         else {
             // Message is a push message
@@ -931,7 +937,7 @@ void Application::producerThread()
             }
 
             eventBuilder.reset();
-            for (int msgId = 0; msgId < d_parameters_p->eventSize();
+            for (bsl::uint64_t msgId = 0; msgId < d_parameters_p->eventSize();
                  ++msgId, ++msgSeqId) {
                 bmqa::Message& msg    = eventBuilder.startMessage();
                 int            length = 0;
@@ -1046,12 +1052,7 @@ void Application::producerThread()
         }
     }
 
-    // Finished posting messages in auto mode?
-    // If shutDownGrace is set, signal to the main thread to exit.
-    if (d_parameters_p->mode() == ParametersMode::e_AUTO &&
-        d_parameters_p->shutdownGrace() != 0) {
-        // We do not need to sleep the grace period, since it is done
-        // by the main thread, in the stop() function.
+    if (!bmqt::QueueFlagsUtil::isAck(d_parameters_p->queueFlags())) {
         d_shutdownSemaphore_p->post();
     }
 }
@@ -1131,6 +1132,8 @@ Application::Application(Parameters*       parameters,
 , d_latencies(allocator)
 , d_autoReadInProgress(false)
 , d_autoReadActivity(false)
+, d_numExpectedAcks(0)
+, d_numPostedAcknowledged(0)
 {
     // NOTHING
 }
@@ -1194,8 +1197,12 @@ int Application::run()
         d_shutdownSemaphore_p->post();
     }
     else {
-        // Start the thread
         if (bmqt::QueueFlagsUtil::isWriter(d_parameters_p->queueFlags())) {
+            d_numExpectedAcks = d_parameters_p->eventsCount() *
+                                d_parameters_p->eventSize();
+            d_numPostedAcknowledged = 0;
+
+            // Start the thread
             rc = bslmt::ThreadUtil::create(
                 &d_runningThread,
                 bdlf::MemFnUtil::memFn(&Application::producerThread, this));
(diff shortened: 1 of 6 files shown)


src/applications/bmqtool/m_bmqtool_application.cpp
```diff
@@ -1194,8 +1197,12 @@ int Application::run()
         d_shutdownSemaphore_p->post();
     }
     else {
-        // Start the thread
         if (bmqt::QueueFlagsUtil::isWriter(d_parameters_p->queueFlags())) {
+            d_numExpectedAcks = d_parameters_p->eventsCount() *
+                                d_parameters_p->eventSize();
```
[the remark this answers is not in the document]
678098: At the same time, I am not sure if we should add more constraints for param values, like:

```
BSLS_ASSERT(eventsCount() >= 0);
BSLS_ASSERT(eventsCount() <= 2^31);
BSLS_ASSERT(eventSize() >= 0);
BSLS_ASSERT(eventSize() <= 2^31);
```

Another way, these constraints might be introduced in `xsd` if possible.

678098 (changes requested):

src/applications/bmqtool/m_bmqtool_application.h
```diff
@@ -167,6 +167,19 @@ class Application : public bmqa::SessionEventHandler {
...
     // grace period.
 
+    int d_numExpectedAcks;
+    // Auto-produce mode only. The total number of messages
+    // the tool will send. After posting is finished
+    // the tool will be waiting for this number of ACK
+    // messages, after which the shutdown semaphore will
+    // be posted.
+
+    int d_numPostedAcknowledged;
```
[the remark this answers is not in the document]
syuzvinsky: Good catch in general!
Changing to bsl::uint64_t though, because:
1. I don't plan to store negative values there, so unsigned type is choosen;
2. ::uint64_t did not seem to be supported by C++03
3. bsls::Types::UInt64 is a legacy type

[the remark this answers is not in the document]
678098: I think `bsl::uint64_t` is a good choice, if you don't want to add a special meaning for negative values here.

Note that for `bmqbrkr` and related libs we preferred to use `bsls::Types::...` at least while we support AIX/Solaris, using `bsl::uint64_t` in `bmqtool` should be fine though

src/applications/bmqtool/m_bmqtool_application.cpp
```diff
@@ -1046,12 +1052,7 @@ void Application::producerThread()
         }
     }
 
-    // Finished posting messages in auto mode?
-    // If shutDownGrace is set, signal to the main thread to exit.
-    if (d_parameters_p->mode() == ParametersMode::e_AUTO &&
-        d_parameters_p->shutdownGrace() != 0) {
-        // We do not need to sleep the grace period, since it is done
-        // by the main thread, in the stop() function.
+    if (!bmqt::QueueFlagsUtil::isAck(d_parameters_p->queueFlags())) {
```
dorjesinpo, line 1055: Is this going to affect auto consumer?

src/applications/bmqtool/m_bmqtool_application.h
dorjesinpo, line 177: Why is this 'posted'? It looks like an ACK counter, regardless of what is posted.  Or does it correlate GUIDs to make sure it is a valid ACK?

src/applications/bmqtool/m_bmqtool_application.cpp
```diff
@@ -768,6 +768,12 @@ void Application::onMessageEvent(const bmqa::MessageEvent& event)
 
             // Write to log file
             d_fileLogger.writeAckMessage(message);
+
+            if (d_numExpectedAcks != 0 &&
+                d_numExpectedAcks == ++d_numPostedAcknowledged) {
+                BALL_LOG_INFO << "All posted messages have been acknowledged";
+                d_shutdownSemaphore_p->post();
```
dorjesinpo, line 775: I assume, the SDK does ACK validation, so it not possible to miscount ACKs?

src/applications/bmqtool/m_bmqtool_application.cpp
dorjesinpo, line 1055: Could you verify the use of `shutdownGrace` now?  Do we still need it?

dorjesinpo (commented):
some questions

src/applications/bmqtool/m_bmqtool_application.cpp
syuzvinsky: Discussed this with Julien Bibollet.
The conclusion was that shutdownGrace was designed, and is still needed in _consuming mode_: some integration tests imply running several instances of the consumer, and there is some nondeterminism in how many messages each particular instance should receive, only the total number of messages is defined.
So, the test code will be waiting for a publisher to complete, and after that it will be waiting until each client stops receiving messages and terminate.
In the mean time, using shutdownGrace in a publisher thread looks rather like a hack, and counting ACKs is a better way to make sure that all messages have reached the cluster.

src/applications/bmqtool/m_bmqtool_application.h
dorjesinpo: There is nothing.  So, the "posted' part does not carry much meaning, does it.  It is just a bit confusing because one can think we have posted ACKs.
Very minor

src/applications/bmqtool/m_bmqtool_application.cpp
dorjesinpo: I don't think we can do both at the same time.  So, this does not seem to affect consumer.

dorjesinpo (approved):

678098 (approved):

PR #235, opened 2024-04-02
Fix: enforce strict TCPEndpoint build

678098:
We have 2 similar routines:

`bool TCPEndpoint::fromUri(const bsl::string& uri)`

`void TCPEndpoint::fromUriRaw(const bsl::string& uri)`

The difference is that `fromUriRaw` does not return a bool result, does not perform as much checks and instead relies on a text usage contract `Expects uri to be valid format.`

In my opinion, the current implementation is error-prone, and we can enforce usage contract on a code level instead, at the price of some additional runtime checks.

Alternatively, I would prefer to remove `fromUriRaw` routine since we use it only within `mwc`, replacing its in-place usage with `fromUri`.


678098:
Also note that we don't have unit tests for this component

pniedzielski:
> Alternatively, I would prefer to remove fromUriRaw routine since we use it only within mwc, replacing its in-place usage with fromUri.

In fact, I only see one usage of `fromUriRaw`, which is in the `assign` member function of the same class.  This does some string concatenation of its parameters (which have no contract) and then passes the result to `fromUriRaw` without any validation.  This is a contract error, and so we'd either need to change contract of `assign` or use `fromUri` to perform validation.  I'm also in support of the latter.

However, I don't know whether we can remove `fromUriRaw` entirely, as there may be external package users of this function.  If we can show that there are no packages using this function, it may be safe to remove.  Maybe best course of action is first the changes in this patch, then next make `assign` use `fromUri`, then finally mark `fromUriRaw` as deprecated?

pniedzielski (approved):

PR #239, opened 2024-04-04
protect ClientSession::initiateShutdownDispatched

dorjesinpo:
When calling `ClientSession::initiateShutdown` multiple times (as in the multi-cluster environment), the first call can result in the session destruction _after_ subsequent call(s) schedule `ClientSession::initiateShutdownDispatched`. The latter results in a seqfault.
As a solution, bind `shared_ptr<ClientSession>` to the `ClientSession::initiateShutdownDispatched`.


kaikulimu (approved):
LGTM

PR #241, opened 2024-04-05
CI: reduce cache size

678098:
The current size of a cache entry is 1.9 GB, with GH limits of 10 GB we have a problem with parallel PRs.

From my research:
300MB of cache is used by object files (`*.o`)
400MB of cache is used by built static libraries (`*.a`)
1000MB of cache is used by built unit test drivers (`*.t.tsk`)

This PR adds cleanup step to remove all these files before caching. Since we need test drivers for unit tests, I placed unit tests validation before the cleanup step.

The plot twist here is that a cache doesn't speed up plugins build. You can see that enabling plugins build causes most (all?) files rebuild.

The final size of a cache decreased by 10x: 1.9GB -> 190MB

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -61,7 +61,6 @@ jobs:
           find . -name "*.o" -type f -delete
           find . -name "*.a" -type f -delete
           find . -name "*.t.tsk" -type f -delete
-          du -h -a
       - uses: actions/cache@v4
         with:
           path: |
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -52,6 +52,9 @@ jobs:
             -DCMAKE_PREFIX_PATH=${{ github.workspace }}/deps/srcs/bde-tools/BdeBuildSystem \
             -DCMAKE_INSTALL_LIBDIR=lib64
           cmake --build build/blazingmq --parallel 8 --target all all.t
+      - name: Clean-up build directories before caching
+        run: |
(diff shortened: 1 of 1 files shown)


pniedzielski (commented):
I'm okay with this change (1.9GB cache is way too large), but I'm surprised that removing all of these files leaves the cache useful at all.  Do we really want to delete the `*.a` files for BDE and NTF and rebuild them each run?

678098:
@pniedzielski 
> Do we really want to delete the `*.a` files for BDE and NTF and rebuild them each run?

Since the original workflow used the latest commit sha, the build files were not shared at all across different PRs. I believe the initial idea of this cache was to speed up the prometheus plugin build, or reuse initialized build folder.

We can cache BDE/NTF build files to be useful across different PR workflows indeed, like it's done here: https://github.com/bloomberg/ntf-core/blob/cecbabbcb1f120a4cf0ae544352bf3cb819521cb/.github/workflows/build.yaml#L25
It will take more effort, but it can possibly save ~7 minutes of CI time.

pniedzielski (approved):

pniedzielski:
> We can cache BDE/NTF build files to be useful across different PR workflows indeed, like it's done here: https://github.com/bloomberg/ntf-core/blob/cecbabbcb1f120a4cf0ae544352bf3cb819521cb/.github/workflows/build.yaml#L25
> It will take more effort, but it can possibly save ~7 minutes of CI time.

This sounds useful but also not the highest priority then.

Issue #118, opened 2023-09-26
Allow API documentation to be built directly from Doxygen


pniedzielski:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Our current API documentation was generated using [`bde-tools`'s `bdedox` executable](https://github.com/bloomberg/bde-tools/tree/main/contrib/bdedox).  Since our comments are no longer in BDE-style, running `doxygen` on the codebase does not pick up several things our current published API docs have, like package group, package, and component-level documentation, some BDE formatting, and `\brief` descriptions automatically generated from the first sentence of a documentation comment.

Additionally, some BDE-style documentation failed to convert, as in the [`bmqa_abstractsession` component](https://github.com/bloomberg/blazingmq/blob/66bfc7fe70f2b12893ee3f1e791dba8981a61db2/src/groups/bmq/bmqa/bmqa_abstractsession.h#L288).

### Expected Behavior

Running `doxygen` on the codebase should produce up-to-date documentation, including all our package group, package, and component documentation, and should support Doxygen-style annotations, such as `\deprecated` and `\par`.

### Steps To Reproduce

1. Run `doxygen` on the codebase.
2. Compare the output to the [public API documentation](https://bloomberg.github.io/blazingmq/docs/apidocs/cpp_apidocs/).

### BlazingMQ Version

HEAD

chrisbeard:
Non-exhaustive to-do list for this effort:

1. convert group and package docs in `mwc` to doxygen format like `bmqa`
2. convert group and package docs in `mqb` to doxygen format (less important)
3. add make doc target that at least generates `bmq` and `mwc` docs
4. make the CI run this on release tags and upload it to the website
5. go through and find places where the doxygen conversion failed to properly convert BDE style to doxygen style
6. use doxygen tags for things like `@deprecated`

PR #242, opened 2024-04-08
Convert package-level and component-level documentation into form Doxygen can understand

pniedzielski:
**Issue number of the reported bug or feature request: #118**

**Describe your changes**
The changes in this PR convert our _package-level_ and _component-level_ documentation into _directory_ and _file_ documentation, respectively, so that Doxygen is able to understand them.  Please see the messages on each commit for further details.

Note that I have not included the _built_ documentation files, which live in `docs/docs/apidocs/cpp_apidocs/`.  We may want to keep these on a separate branch (`gh-pages`?), and the current output of Doxygen is still not optimal (see **Future changes** below for details), so I will not modify these in this PR.

**Testing performed**
Testing the documentation build by running `doxygen` manually from the root of the repository.  Using Doxygen version 1.9.8, the only warnings left are because we use the project's `README.md` as our API docs' starting point.  This should be changed.

**Future changes**
A few more changes need to be made to the documentation so we can automatically generate it on release:

  1. **Catch conversion errors from BDE-style documentation to Doxygen documentation.** As noted in the linked issue, there are some places which confused the automated conversion tool, leaving us with old-style BDE formatting that Doxygen cannot understand.  These should be changed.
  2. **Add a real `\mainpage`, which links to each of the packages for easier discoverability.**  This will also fix the remaining Doxygen warnings.
  3. **Teach CMake to find and run Doxygen.** This will let us call the equivalent of `make doc` to generate documentation, which we can do in our CI.  This will let us catch documentation errors in CI and ultimately publish the documentation automatically on releases.
  4. (As a nice-to-have,) **Convert BDE class member grouping to a form Doxygen understands.** Doxygen provides the ability to divide class members into arbitrary groups with names.  If this change is not too intrusive, we can consider doing this to make our API docs easier to understand.


hallfox:
Looks pretty good! I'm happy with the rendered output, even though we've lost the component level breakdown in packages. A few things I'm noticing:

1. Package level documentation appears to be rendering as README links in the related pages section without any of the contents in the README. Is this expected? (Or am I seeing something weird?)
2. Several "impl" classes are visible in the documentation that weren't visible in the previous one. My guess is there was some rule in bdedox that was excluding them by the nature of that name, or some option we had enabled that was removed that I can't see in this PR. Is it likely we'll be able to hide these again? Or should we find other ways to mark them as internal only? Seems like doxygen's `\internal` command might help us here.

pniedzielski:
> I'm happy with the rendered output, even though we've lost the component level breakdown in packages.

Yeah, IIRC component levelization was not part of the OSS release (see https://github.com/bloomberg/blazingmq/commit/ed9fa8e72e0c84689f4a5fc94a0ea6316e49b6ac for the only other commit to these READMEs), but it would be nice to add it back in.  I believe the currently published levelization is incomplete, but I would need to check again.

> 1. Package level documentation appears to be rendering as README links in the related pages section without any of the contents in the README. Is this expected? (Or am I seeing something weird?)

This is "expected", but I haven't found a good way around it.  See https://github.com/doxygen/doxygen/issues/9437, it's a known issue.  Doesn't help us much though.

> 2. Several "impl" classes are visible in the documentation that weren't visible in the previous one. My guess is there was some rule in bdedox that was excluding them by the nature of that name, or some option we had enabled that was removed that I can't see in this PR. Is it likely we'll be able to hide these again? Or should we find other ways to mark them as internal only? Seems like doxygen's `\internal` command might help us here.

`\internal` is a good solution here.  Since in BDE-style docs, the publically-usable symbols are listed in component-level documentation, it may be that `bdedox` only generates documentation for those symbols listed in the component-level documentation.  Either way, we removed that list in this PR intentionally, so `\internal` is a good idea.

I don't have a good solution to the first issue (I was thinking of it lumped together with our `\mainpage` being a broken version of the project README as a "later" problem...), but I'll do the latter in this PR.

hallfox:
> I don't have a good solution to the first issue (I was thinking of it lumped together with our \mainpage being a broken version of the project README as a "later" problem...), but I'll do the latter in this PR.

I'm okay with this for now.

I think we should keep hiding the internal components.

pniedzielski:
@hallfox Rebased onto `main`, but included two additional commits that address internal components being shown in the generated documentation.

hallfox (approved):
lgtm

PR #245, opened 2024-04-10
Fix: mwcc monitoredqueue and orderedhashmap Wconv warnings

melvinhe:

**Describe your changes**
This PR deals with more ‘size_t‘ and ‘int‘ data type inconsistencies, through casting, adjusting data types, etc. It also catches the following -Wconversion warnings that pollutes the build logs:

```
/home/runner/work/blazingmq/blazingmq/src/groups/mwc/mwcc/mwcc_monitoredqueue.h:593:44: warning: conversion from ‘BloombergLP::bsls::Types::Int64’ {aka ‘long long int’} to ‘int’ may change value [-Wconversion]
  593 |     const int currentQueueLen = numElements();
      |                                 ~~~~~~~~~~~^~
/home/runner/work/blazingmq/blazingmq/src/groups/mwc/mwcc/mwcc_orderedhashmap.h:1389:22: warning: conversion from ‘size_t’ {aka ‘long unsigned int’} to ‘int’ may change value [-Wconversion]
 1389 |     int count = erase(get_key(nodeToErase->value()));
      |                 ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/blazingmq/blazingmq/src/groups/mwc/mwcc/mwcc_orderedhashmap.t.cpp:864:50: warning: conversion from ‘size_t’ {aka ‘long unsigned int’} to ‘int’ may change value [-Wconversion]
  864 |     for (; localIt != localEndIt; ++localIt, key += bucketCount) {
      |                                              ~~~~^~~~~~~~~~~~~~
/home/runner/work/blazingmq/blazingmq/src/groups/mwc/mwcc/mwcc_orderedhashmap.t.cpp: In function ‘void testN1_insertPerformanceOrdered_GoogleBenchmark(benchmark::State&)’:
/home/runner/work/blazingmq/blazingmq/src/groups/mwc/mwcc/mwcc_orderedhashmap.t.cpp:1118:34: warning: conversion from ‘int64_t’ {aka ‘long int’} to ‘int’ may change value [-Wconversion]
 1118 |         MyMapType map(state.range(0), s_allocator_p);
      |                       ~~~~~~~~~~~^~~
/home/runner/work/blazingmq/blazingmq/src/groups/mwc/mwcc/mwcc_orderedhashmap.t.cpp: In function ‘void testN3_profile_GoogleBenchmark(benchmark::State&)’:
/home/runner/work/blazingmq/blazingmq/src/groups/mwc/mwcc/mwcc_orderedhashmap.t.cpp:1193:59: warning: conversion from ‘int64_t’ {aka ‘long int’} to ‘int’ may change value [-Wconversion]
 1193 |     MyMapType                              map(state.range(0), s_allocator_p);
      |                                                ~~~~~~~~~~~^~~
``` 

**Additional context**
There is another separate issue I found with 'reserveCapacity', where it takes in ‘size_t’ values like ’d_bucketArraySize’ or ’d_numElements’ indicating errors similar to the captured log below:

``` 
/home/runner/work/blazingmq/blazingmq/src/groups/mwc/mwcc/mwcc_orderedhashmap.h:1197:58: warning: conversion from ‘size_t’ {aka ‘long unsigned int’} to ‘int’ may change value [-Wconversion]
 1197 |             d_nodePool.reserveCapacity(d_bucketArraySize - d_numElements);
      |                                        ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
``` 

However, when I observe 'reserveCapacity', the function signature seems to expect ‘size_t’ rather than ’int’ as a parameter. Please see ’mwcu_memoutstream.h’

Perhaps the issue is caused by the fact that 'reserveCapacity' in ’mwcu_memoutstream.h’ expects 'bsl::size_t' as a parameter instead of ‘size_t’? Should I adjust ’reserveCapacity’ to try to remedy this, and if so, any ideas/feedback on how to fix this case for ’reserveCapacity’?

Also, fixing this one issue seems to fix/clean at least 100 warns/errors logs in BlazingMQ Build!

### DIFF
--- a/src/groups/mwc/mwcc/mwcc_monitoredqueue.h
+++ b/src/groups/mwc/mwcc/mwcc_monitoredqueue.h
@@ -590,7 +590,7 @@ template <class QUEUE, class QUEUE_TRAITS>
 inline int MonitoredQueue<QUEUE, QUEUE_TRAITS>::pushBack(
     bslmf::MovableRef<ElementType> value)
 {
-    const int currentQueueLen = numElements();
+    const int currentQueueLen = static_cast<int>(numElements());
     if (currentQueueLen + 1 >= capacity()) {
         // We've filled the queue.  Alarm
         if (!Traits::isPushBackDisabled(d_queue) &&
--- a/src/groups/mwc/mwcc/mwcc_monitoredqueue.h
+++ b/src/groups/mwc/mwcc/mwcc_monitoredqueue.h
@@ -590,7 +590,7 @@ template <class QUEUE, class QUEUE_TRAITS>
 inline int MonitoredQueue<QUEUE, QUEUE_TRAITS>::pushBack(
     bslmf::MovableRef<ElementType> value)
 {
-    const int currentQueueLen = static_cast<int>(numElements());
+    const bsls::Types::Int64 currentQueueLen = numElements();
     if (currentQueueLen + 1 >= capacity()) {
         // We've filled the queue.  Alarm
         if (!Traits::isPushBackDisabled(d_queue) &&
--- a/src/groups/mwc/mwcc/mwcc_orderedhashmap.t.cpp
+++ b/src/groups/mwc/mwcc/mwcc_orderedhashmap.t.cpp
@@ -215,7 +215,7 @@ static void test3_insert()
     const int k_NUM_ELEMENTS = 1000 * 1000;  // 1M
 #else
     // Avoid timeout on AIX and Solaris
-    const int k_NUM_ELEMENTS = 100 * 1000;   // 100K
+    const int k_NUM_ELEMENTS = 100 * 1000;  // 100K
 #endif
 
     // Insert 1M elements
@@ -826,7 +826,7 @@ static void test14_localIterator()
 
     const size_t bucketCount = map.bucket_count();
 
-    int key         = bucketCount / 2;
+    int key         = static_cast<int>(bucketCount / 2);
     int originalKey = key;
 
     map.insert(bsl::make_pair(key, key * key));
@@ -853,7 +853,7 @@ static void test14_localIterator()
     // Add keys such that they all map to same bucket in the table, while
     // ensuring that table is not rehashed.
     for (unsigned int i = 0; i < (bucketCount - 2); ++i) {
-        key += bucketCount;
+        key += static_cast<int>(bucketCount);
         map.insert(bsl::make_pair(key, key * key));
     }
 
@@ -861,7 +861,8 @@ static void test14_localIterator()
     localEndIt = map.end(bucket);
 
     key = originalKey;
-    for (; localIt != localEndIt; ++localIt, key += bucketCount) {
+    for (; localIt != localEndIt;
+         ++localIt, key += static_cast<int>(bucketCount)) {
         ASSERT_EQ(localIt->first, key);
         ASSERT_EQ(localIt->second, key * key);
     }
@@ -1115,7 +1116,7 @@ testN1_insertPerformanceOrdered_GoogleBenchmark(benchmark::State& state)
         // OrderedHashMap
         typedef mwcc::OrderedHashMap<int, int> MyMapType;
 
-        MyMapType map(state.range(0), s_allocator_p);
+        MyMapType map(static_cast<int>(state.range(0)), s_allocator_p);
         for (auto _ : state) {
             for (int i = 0; i < state.range(0); ++i) {
                 map.insert(bsl::make_pair(i, i));
@@ -1190,7 +1191,7 @@ static void testN3_profile_GoogleBenchmark(benchmark::State& state)
     // This case can be used to profile the component.
 
     typedef mwcc::OrderedHashMap<int, int> MyMapType;
-    MyMapType                              map(state.range(0), s_allocator_p);
+    MyMapType map(static_cast<int>(state.range(0)), s_allocator_p);
 
     // Insert elements.
     for (auto _ : state) {
--- a/src/groups/mwc/mwcc/mwcc_orderedhashmap.t.cpp
+++ b/src/groups/mwc/mwcc/mwcc_orderedhashmap.t.cpp
@@ -215,7 +215,7 @@ static void test3_insert()
     const int k_NUM_ELEMENTS = 1000 * 1000;  // 1M
 #else
     // Avoid timeout on AIX and Solaris
-    const int k_NUM_ELEMENTS = 100 * 1000;  // 100K
+    const int k_NUM_ELEMENTS = 100 * 1000;   // 100K
 #endif
 
     // Insert 1M elements
@@ -861,8 +861,7 @@ static void test14_localIterator()
     localEndIt = map.end(bucket);
 
     key = originalKey;
-    for (; localIt != localEndIt;
-         ++localIt, key += static_cast<int>(bucketCount)) {
+    for (; localIt != localEndIt; ++localIt, key += static_cast<int>(bucketCount)) {
         ASSERT_EQ(localIt->first, key);
         ASSERT_EQ(localIt->second, key * key);
     }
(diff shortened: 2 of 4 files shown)


pniedzielski:
> Also, fixing this one issue seems to fix/clean at least 100 warns/errors logs in BlazingMQ Build!

🎉 🎉 🎉

pniedzielski:
> However, when I observe 'reserveCapacity', the function signature seems to expect ‘size_t’ rather than ’int’ as a parameter. Please see ’mwcu_memoutstream.h’

@melvinhe With regard to this, it's a good catch, but the `reserveCapacity` function here is that of `bdlma_pool`, which is in our dependency BDE.  [Here are the docs](https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/classbdlma_1_1Pool.html#a2a8ab64992212d4f3abde7f47f2a5799) for that function, and as you can see it takes an `int`, so unfortunately we can't change that.

For parity with the standard library, I think `d_bucketArraySize` and `d_numElements` were made to be `bsl::size_t`/`std::size_t`, but the BDE containers prefer to use `int`, with a precondition that the value be nonnegative (see docs linked above).  I think the minimally invasive change is to insert a cast from `size_t` to `int` in calls to `reserveCapacity`, perhaps with a `BSLS_ASSERT` before it making sure that `d_bucketArraySize - d_numElements` is not greater than `static_cast<bsl::size_t>(bsl::numeric_limits<int>::max())`.

But thinking about it more, in principle this cast is not safe: `d_bucketArraySize` gets its values from `OrderedHashMap_ImpDetails::nextPrime`, which lists out roughly geometrically growing primes.  The last of these is `2719929503u`, which is greater than `2^31`.  In this case, the existing implicit conversion from `size_t` to `int` will give a negative number, breaking `bdlma_pool` component's precondition.  So the existing behavior has a bug; the warning is correct, and adding a cast will just silence the warning, not fix the bug.

I'll make another issue for this particular case and we can mull it over there.

src/groups/mwc/mwcc/mwcc_monitoredqueue.h
```diff
@@ -590,7 +590,7 @@ template <class QUEUE, class QUEUE_TRAITS>
 inline int MonitoredQueue<QUEUE, QUEUE_TRAITS>::pushBack(
     bslmf::MovableRef<ElementType> value)
 {
-    const int currentQueueLen = numElements();
+    const int currentQueueLen = static_cast<int>(numElements());
```
pniedzielski, line 593: Here it looks like we're going from `numElements()` of type `Int64` to `int`, and using it below in a comparison against `capacity()` of type `Int64`.  I think it's best to change the `int` to an `Int64` than lose information between `Int64` and `int`, only to promote the result back to `Int64`.  Like what you did in [mwcc_orderedhashmap.h](https://github.com/bloomberg/blazingmq/pull/245/files#diff-98a2f035d83dbc4efe32843358bbd56f2644179e4afe94ff5304c51e9133e67f).

src/groups/mwc/mwcc/mwcc_orderedhashmap.t.cpp
```diff
@@ -826,7 +826,7 @@ static void test14_localIterator()
 
     const size_t bucketCount = map.bucket_count();
 
-    int key         = bucketCount / 2;
+    int key         = static_cast<int>(bucketCount / 2);
```
pniedzielski, line 829: I wonder if the right approach here is changing `MyMapType` to be an `mwcc::OrderedHashMap<size_t, size_t, IdentityHasher>`.  I think this will also need a change in `IdentityHasher`.  This is a case like the one above, where I think really what we would want is C++11 `auto`---if only we were able to use C++11...  So changing `int`s to `size_t`s might be the next best thing.

Good intuition here for going with the solution that is least invasive.  Here we're in a test file, though, so we don't need to worry about breaking interfaces or even changing behavior subtly---we just want to make sure that we still test all the cases we currently test.  This means we can be a little more aggressive in this file doing the morally right changes (fixing types, avoiding casts).

src/groups/mwc/mwcc/mwcc_orderedhashmap.t.cpp
```diff
@@ -853,15 +853,16 @@ static void test14_localIterator()
     // Add keys such that they all map to same bucket in the table, while
     // ensuring that table is not rehashed.
     for (unsigned int i = 0; i < (bucketCount - 2); ++i) {
-        key += bucketCount;
+        key += static_cast<int>(bucketCount);
```
pniedzielski, line 856: And if we do the above, we don't need any casts here or below!  (Though we probably would need to change the `unsigned int` to `size_t`.)

src/groups/mwc/mwcc/mwcc_orderedhashmap.t.cpp
```diff
@@ -1190,7 +1191,7 @@ static void testN3_profile_GoogleBenchmark(benchmark::State& state)
     // This case can be used to profile the component.
 
     typedef mwcc::OrderedHashMap<int, int> MyMapType;
-    MyMapType                              map(state.range(0), s_allocator_p);
+    MyMapType map(static_cast<int>(state.range(0)), s_allocator_p);
```
pniedzielski, line 1194: Perhaps similar here?  I'm honestly surprised that line 1198 doesn't trigger a signed/unsigned comparison warning though.

This one is probably okay on its own.  Generally an explicit cast means "Trust me, this conversion is safe and I mean it," which is why they have such long and scary names in C++, compared with C-style casts.  That's all well and good, and these casts are indeed safe, you're right!  But I'm not convinced we mean them---I think many of these `int`s are incidental, and really what we meant was `whatever type state.range(0) returns`.  So whenever we can avoid casts (without breaking our public interface, without changing our behavior, and without making very large changes to the codebase), I tend to think it's better to do so.

pniedzielski (changes requested):
Thank you thank you thank you so much for doing these PRs!  We've noticed the drop in warnings day-to-day and it's a big help!

Comments inline, I think we can avoid the casts here and still beat the warnings, but I haven't checked.  Try it, and if there are still warnings, or the necessary changes are more extensive than I've indicated, we'll gladly accept these casts. :)

melvinhe:
> > However, when I observe 'reserveCapacity', the function signature seems to expect ‘size_t’ rather than ’int’ as a parameter. Please see ’mwcu_memoutstream.h’
> 
> @melvinhe With regard to this, it's a good catch, but the `reserveCapacity` function here is that of `bdlma_pool`, which is in our dependency BDE. [Here are the docs](https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/classbdlma_1_1Pool.html#a2a8ab64992212d4f3abde7f47f2a5799) for that function, and as you can see it takes an `int`, so unfortunately we can't change that.
> 
> For parity with the standard library, I think `d_bucketArraySize` and `d_numElements` were made to be `bsl::size_t`/`std::size_t`, but the BDE containers prefer to use `int`, with a precondition that the value be nonnegative (see docs linked above). I think the minimally invasive change is to insert a cast from `size_t` to `int` in calls to `reserveCapacity`, perhaps with a `BSLS_ASSERT` before it making sure that `d_bucketArraySize - d_numElements` is not greater than `static_cast<bsl::size_t>(bsl::numeric_limits<int>::max())`.
> 
> But thinking about it more, in principle this cast is not safe: `d_bucketArraySize` gets its values from `OrderedHashMap_ImpDetails::nextPrime`, which lists out roughly geometrically growing primes. The last of these is `2719929503u`, which is greater than `2^31`. In this case, the existing implicit conversion from `size_t` to `int` will give a negative number, breaking `bdlma_pool` component's precondition. So the existing behavior has a bug; the warning is correct, and adding a cast will just silence the warning, not fix the bug.
> 
> I'll make another issue for this particular case and we can mull it over there.

@pniedzielski Cool, opening a new issue sounds good. I'll look more closely at the docs and maybe mull it over at #247

src/groups/mwc/mwcc/mwcc_orderedhashmap.t.cpp
melvinhe: Sounds good, fixing the data types seems to be a better code practice approach here.

I'll test out your suggested changes (basically see if we can get away with fixing more proper datatypes without conflicts with other tests and implementations) soon tonight to see if they work properly (and pass the CI checks).

pniedzielski: Excellent, thanks!

pniedzielski (approved):
LGTM!

PR #246, opened 2024-04-11
Implement batch posting in CLI mode

syuzvinsky:
Adding `batch-post` CLI command to initiate posting series of messages. The posting code was taken from `Application` class and put into a separate class `PostingContext`. This class is now used in both interactive and automatic modes.

### DIFF
--- a/src/applications/bmqtool/README.md
+++ b/src/applications/bmqtool/README.md
@@ -115,9 +115,9 @@ opened regularly (without specifying a mode).  To run the `bmqtool`, you invoke
 | `open`    | `uri=string [async=true] [maxUnconfirmedMessages=N) (maxUnconfirmedByes=M)` | Open a connection with the queue at the given `uri`. |
 | `close`   | `uri=string [async=true]` | Close connection with the queue at the given `uri`. |
 | `post`    | `uri=string payload=string [, ...] [async=true]` | Post a message (the `payload`) to the queue at the given `uri`. |
+| `batch-post` | `uri=string payload=string [, ...] (msgSize=S) (eventSize=N) (eventsCount=M) (postInterval=P) (postRate=R)` | Post `M` events containing `N` messages containing provided `payload` or auto-generated and containing `S` bytes each to the queue at the given `uri` at a rate of `R/P` (where `P` is expressed in ms). `M=0` means endless posting. |
 | `list`    | N/A            | List the messages that have yet to be ACKed by the tool.     |
 | `confirm` | `guid=string`  | Confirm (ACK) the message matching the given GUID.           |
-| `batch-post` | `uri=string payload=string [, ...] (eventSize=N) (eventsCount=M) (postInterval=P) (postRate=R)` | Post M messages to the queue at the given `uri`. |  
 | `help`    | N/A            | Show the help dialog.                                        |
 | `bye`     | N/A            | Exit the tool.                                               |
 | `quit`    | N/A            | Exit the tool.                                               |
--- a/src/applications/bmqtool/README.md
+++ b/src/applications/bmqtool/README.md
@@ -117,6 +117,7 @@ opened regularly (without specifying a mode).  To run the `bmqtool`, you invoke
 | `post`    | `uri=string payload=string [, ...] [async=true]` | Post a message (the `payload`) to the queue at the given `uri`. |
 | `list`    | N/A            | List the messages that have yet to be ACKed by the tool.     |
 | `confirm` | `guid=string`  | Confirm (ACK) the message matching the given GUID.           |
+| `batch-post` | `uri=string payload=string [, ...] (eventSize=N) (eventsCount=M) (postInterval=P) (postRate=R)` | Post M messages to the queue at the given `uri`. |  
 | `help`    | N/A            | Show the help dialog.                                        |
 | `bye`     | N/A            | Exit the tool.                                               |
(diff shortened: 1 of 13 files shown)


src/applications/bmqtool/README.md
```diff
@@ -117,6 +117,7 @@ opened regularly (without specifying a mode).  To run the `bmqtool`, you invoke
 | `post`    | `uri=string payload=string [, ...] [async=true]` | Post a message (the `payload`) to the queue at the given `uri`. |
 | `list`    | N/A            | List the messages that have yet to be ACKed by the tool.     |
 | `confirm` | `guid=string`  | Confirm (ACK) the message matching the given GUID.           |
+| `batch-post` | `uri=string payload=string [, ...] (eventSize=N) (eventsCount=M) (postInterval=P) (postRate=R)` | Post M messages to the queue at the given `uri`. |
```
678098, line 120: Might be good to place this command just under the similar `post` command

678098, line 120: Also might be good to explain optional parameters.
What are default values if these params are not specified?
What are the special values for these params? What if we set `postInterval=0` or negative?
Is post interval in milliseconds?
What is the `postRate`?

src/applications/bmqtool/m_bmqtool_application.h
```diff
@@ -44,6 +44,7 @@
 #include <mwctsk_consoleobserver.h>
 
 // BDE
+#include "m_bmqtool_poster.h"
```
678098, line 47:
```suggestion
#include <m_bmqtool_poster.h>
```

src/applications/bmqtool/m_bmqtool_interactive.cpp
```diff
@@ -147,6 +151,14 @@ void Interactive::printHelp()
         << bsl::endl
         << "    - 'post' command requires 'uri' and 'payload' arguments, "
         << "parameter 'messageProperties' is optional" << bsl::endl
+        << bsl::endl
+        << "  batch-post uri=\"bmq://bmq.test.persistent.priority/qqq\" "
+           "payload=[\"sample message\"] eventsCount=300 postRate=10 "
+           "postInterval=5000"
```
678098, line 157: `eventSize` is not mentioned in the example, is it ok?

src/applications/bmqtool/m_bmqtool_interactive.cpp
```diff
@@ -147,6 +151,14 @@ void Interactive::printHelp()
         << bsl::endl
         << "    - 'post' command requires 'uri' and 'payload' arguments, "
         << "parameter 'messageProperties' is optional" << bsl::endl
+        << bsl::endl
+        << "  batch-post uri=\"bmq://bmq.test.persistent.priority/qqq\" "
+           "payload=[\"sample message\"] eventsCount=300 postRate=10 "
```
678098, line 156: Might be also good to introduce optional params in the same order as in `--help` docs

src/applications/bmqtool/m_bmqtool_interactive.cpp
```diff
@@ -117,6 +118,9 @@ void Interactive::printHelp()
         << "    (messageProperties=[{\"name\": \"\", \"value\": \"\", "
            "\"type\": \"\"}])"
         << bsl::endl
+        << "  batch-post uri=\"\" payload=[\"\",\"\"] (msgSize=u) "
+           "(eventSize=v) (eventsCount=w) (postInterval=x) (postRate=y)"
+        << bsl::endl
```
678098, line 123: Same as above, might be good to describe optional params more

src/applications/bmqtool/m_bmqtool_interactive.cpp
```diff
@@ -746,13 +793,16 @@ void Interactive::eventHandlerThread()
...
 : d_session_p(0)
 , d_sessionEventHandler_p(0)
 , d_parameters_p(parameters)
 , d_uris(allocator)
 , d_eventHandlerThreads(allocator)
 , d_producerIdProperty("** NONE **", allocator)
+, d_poster_p(poster)
 , d_allocator_p(allocator)
 {
     bdls::ProcessUtil::getProcessName(&d_producerIdProperty);  // ignore rc
```
678098, line 808:
```suggestion
    // PRECONDITIONS
    BSLS_ASSERT_SAFE(parameters);
    BSLS_ASSERT_SAFE(poster);

    bdls::ProcessUtil::getProcessName(&d_producerIdProperty);  // ignore rc
```

src/applications/bmqtool/m_bmqtool_poster.cpp
```diff
@@ -0,0 +1,248 @@
...
+               bslma::Allocator*   allocator)
+: d_allocator_p(allocator)
+, d_bufferFactory(4096, d_allocator_p)
+, d_timeBufferFactory(sizeof(bdlb::BigEndianInt64), d_allocator_p)
+, d_statContext(statContext)
+, d_fileLogger(fileLogger)
+{
+    BSLS_ASSERT_SAFE(fileLogger);
+    BSLS_ASSERT_SAFE(statContext);
+    BSLS_ASSERT_SAFE(allocator);
```
678098, line 230: We don't want to support default arg=0 allocator here, right?
`bslma::Allocator* alloc = bslma::Default::allocator(allocator);`

678098 (changes requested):
Some comments

PR #262, opened 2024-04-24
CI: IT cluster mode matrix

678098:
This PR introduces parallel launch of integration tests on a 2x2 Cartesian product,
It shortens CI checks in about 15 minutes, and also makes it possible to see early if the fastest `single/fsm_mode` IT preset fails

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -93,7 +93,10 @@ jobs:
           pytest src/python
 
   integration_tests_ubuntu:
-    name: BlazingMQ integration tests
+    name: BlazingMQ integration tests [${{ matrix.preset }}]
+    strategy:
+      matrix:
+        preset: ["legacy_mode", "fsm_mode"]
     runs-on: ubuntu-latest
     needs: build_ubuntu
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -142,7 +146,7 @@ jobs:
         run: bin/build-darwin.sh
 
   build_and_test_bmqprometheus_plugin:
-    name: "Check if BMQPrometheus plugin can build and pass tests"
+    name: "Build plugin [ubuntu]"
```
pniedzielski, line 149: Maybe having `Prometheus` here is good, or if this job is meant for all future plugins as well, `Build plugins [ubuntu]`.

pniedzielski (dismissed):
The change I suggested is minor, but otherwise this looks right to me.  Your choice whether to address the comment or merge as is.

.github/workflows/build.yaml
678098, line 149:
```suggestion
    name: "Build Prometheus plugin [ubuntu]"
```

PR #264, opened 2024-04-26
mqbcfg.xsd: Increase failover timeout to 5 minutes for proxies

kaikulimu:
Looking through the code, I realize that proxies does not use the cluster definitions from [`clusters.json`](https://bbgithub.dev.bloomberg.com/BMQ/bmqdomains/blob/main/clusters.json) at all. If you look at the constructor of [`ClusterProxy`](https://github.com/bloomberg/blazingmq/blob/b822147d4a5bfe80fe12139d585876ccc93b6b23/src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp#L1023-L1035), it is passed a default-initialized `mqbcfg::ClusterDefinition(allocator)`. Therefore, it will only use the default values, which is [2 minutes](https://github.com/bloomberg/blazingmq/blob/b822147d4a5bfe80fe12139d585876ccc93b6b23/src/groups/mqb/mqbcfg/mqbcfg.xsd#L622) in the case of failover threshold. Therefore, I am raising the default to 5 minutes to make all proxies have a failover timeout of 5 minutes.

Clusters nodes obey definitions in `clusters.json`, so they should not be affected by this change.

### DIFF
--- a/src/groups/mqb/mqbcfg/mqbcfg_messages.cpp
+++ b/src/groups/mqb/mqbcfg/mqbcfg_messages.cpp
@@ -1,4 +1,4 @@
-// Copyright 2018-2023 Bloomberg Finance L.P.
+// Copyright 2018-2024 Bloomberg Finance L.P.
 // SPDX-License-Identifier: Apache-2.0
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


src/groups/mqb/mqbcfg/mqbcfg_messages.cpp
```diff
@@ -335,15 +272,15 @@ const int ClusterMonitorConfig::DEFAULT_INITIALIZER_MAX_TIME_MASTER = 120;
 
 const int ClusterMonitorConfig::DEFAULT_INITIALIZER_MAX_TIME_NODE = 120;
 
-const int ClusterMonitorConfig::DEFAULT_INITIALIZER_MAX_TIME_FAILOVER = 240;
```
pniedzielski, line 338: This is the meat and potatoes.  Everything else is noise.

src/groups/mqb/mqbcfg/mqbcfg_messages.cpp
```diff
@@ -6330,12 +5804,12 @@ Configuration::print(bsl::ostream& stream, int level, int spacesPerLevel) const
 }  // close package namespace
 }  // close enterprise namespace
 
-// GENERATED BY BLP_BAS_CODEGEN_2023.10.07
+// GENERATED BY @BLP_BAS_CODEGEN_VERSION@
```
pniedzielski, line 5807: This looks suspicious.

pniedzielski (approved):
Change to the XSD file looks good.

I was a little confused by the diffs in `mqbcfg_messages`, but it just looks like the codegen changed where it wants to put definitions/changed to inlining more functions.  That's okay with me.  CI passes.  I just see one place where a version string isn't being outputted; I'm not sure if this is important or not.

src/groups/mqb/mqbcfg/mqbcfg_messages.cpp
kaikulimu: I am using the newest version of BAS codegen. For some reason, this string substitution doesn't work but shouldn't be a big deal.

PR #265, opened 2024-04-29
CI: python -> python3, continue ITs on failure, long trace

678098:


.github/workflows/build.yaml
```diff
@@ -98,6 +98,7 @@ jobs:
       matrix:
         mode: ["legacy_mode", "fsm_mode"]
         cluster: ["single", "multi"]
+      fail-fast: false
```
pniedzielski, line 101: 👍🏻 I don't like how this isn't the default.

CMakeLists.txt
```diff
@@ -87,6 +87,11 @@ if( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU|(Apple)?Clang" )
   endif()
 endif()
 
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
```
pniedzielski, line 90: I think this is a good change, but what led to it?  Was there a specific assert that you wanted to fire that wasn't?

pniedzielski (approved):
👍🏻

CMakeLists.txt
```diff
@@ -87,6 +87,11 @@ if( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU|(Apple)?Clang" )
...
+  # In Debug build, ensure that BSLS_ASSERT_SAFE macro is enabled
+  add_definitions("-DBSLS_ASSERT_LEVEL_ASSERT_SAFE")
```
678098, line 92: When built with `build-darwin.sh`, macro `BSLS_ASSERT_SAFE` is not enabled despite `Debug` build type set.

https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/group__bsls__assert.html

```We can build the our_square component in "safe mode" -- e.g., by incorporating -DBSLS_ASSERT_LEVEL_ASSERT_SAFE on the (Unix) command line.```

CMakeLists.txt
678098: @pniedzielski described in the later comment, the problem with Debug build on Darwin

Since this macro is disabled, we can have green ITs locally and red on GH Actions

678098: By the way, it's possible that it happens because we use outdated version of BDE, but it's easier to just enable this condition explicitly

pniedzielski: Makes sense to me!

PR #266, opened 2024-04-30
[WIP]CI: cache dependencies build

678098:


docker/build_deps.sh
```diff
@@ -37,17 +45,21 @@ configure() {
...
 build_ntf() {
     pushd srcs/ntf-core
-    sed -i s/CMakeLists.txt//g ./configure
-    ./configure --prefix /opt/bb --without-usage-examples --without-applications --ufid opt_64_cpp17
-    make -j8
+    if [ "$install_only" = false ]; then
+        sed -i s/CMakeLists.txt//g ./configure
+        ./configure --prefix /opt/bb --without-usage-examples --without-applications  --ufid opt_64_cpp17
+        make -j8
+    fi
```
678098, line 62: I don't 100% like how obscure all these moments to get the script install only.
There is also an issue with ntf library being rebuild from scratch each time we call this script, even if build files already exist. And this rebuild takes 5 minutes.

PR #269, opened 2024-05-01
Fix: bmqtool on AIX build

678098:
AIX compiler does not optimize and tries to use copy constructor when objects constructed and returned by value


pniedzielski (approved):
It's a shame we have to resort to `shared_ptr` to fix these issues on AIX.  Maybe someday someone will invent C++11.

LGTM.

PR #270, opened 2024-05-01
CI: cache dependencies

678098:


678098:
@pniedzielski this is a POC PR, and I believe, we can refine it more. Not sure if the current solution is elegant. Also it increases CI difficulty

pniedzielski:
> @pniedzielski this is a POC PR, and I believe, we can refine it more. Not sure if the current solution is elegant. Also it increases CI difficulty

Right, I see it only deals with normal Ubuntu builds and not Prometheus or Darwin.  Ubuntu is the most important anyway, and slower than Darwin.  I'll take a deeper look tomorrow but I think more elegant will mean more extensive refactoring.  It makes me want to look at what other big projects do for CI with GH Actions.

docker/build_deps.sh
```diff
@@ -37,17 +45,21 @@ configure() {
...
-    bbs_build build -j8
+    if [ "$install_only" = false ]; then
+        bbs_build configure
+        bbs_build build -j8
+    fi
     bbs_build --install=/opt/bb --prefix=/ install
     popd
 }
 
 build_ntf() {
```
678098, line 56: This `build_ntf` function causes ntf to rebuild from scratch every time this script is called, no matter if incremental build object exist. This rebuild takes 5 minutes.

`build_bde`, on the other hand, doesn't recompile everything.

If we can get ntf not to rebuild everything, we can even omit this `install_only` flag.

pniedzielski: I'm looking into this now, I will add some comments with what I find.

pniedzielski: By default, ntf's `configure` script cleans build artifacts before it does anything.  When it's passed the `--keep` flag it refrains from doing this.

pniedzielski (approved):
@678098 I removed the `--install-only` flag in 7fc253a.  Coupled with the `--keep` flag to NTF, it avoids rebuilding BDE and NTF if they are in the cache, without major modifications to `build-deps.sh`.

You should also review it and make sure it was in line with what you suggested, then squash-merge if you'd like.

PR #257, opened 2024-04-21
Fix: bmqp protocol Wconv warns

melvinhe:
*Issue number of the reported bug or feature request: #87 *

**Describe your changes**
This PR attempts to fix all current -Wconversion warnings caused by faulty error checks, data type inconsistencies, and potential overflow/truncation data loss in ‘bmqp_protocol.cpp‘ as well as associated header and test files. 

It should also clear out over a dozen errors that lead to at least 100 ‘-Wconversion‘ warnings in the BlazingMQ build logs such as, but limited to, the following:

```
In file included from /home/runner/work/blazingmq/blazingmq/src/groups/bmq/bmqp/bmqp_protocol.t.cpp:30:
/home/runner/work/blazingmq/blazingmq/src/groups/bmq/bmqp/bmqp_protocol.t.cpp: In function ‘void test1_breathingTest()’:
/home/runner/work/blazingmq/blazingmq/src/groups/bmq/bmqp/bmqp_protocol.t.cpp:720:31: warning: conversion from ‘int’ to ‘unsigned char’ may change value [-Wconversion]
  720 |                       sh.flags(),
      |                       ~~~~~~~~^~
/home/runner/work/blazingmq/blazingmq/src/groups/mwc/mwctst/mwctst_testhelper.h:360:34: note: in definition of macro ‘ASSERT_EQ’
  360 |         _assertCompareEquals("", X, Y, #X, #Y, __FILE__, __LINE__);           \
      |                                  ^
      
/home/runner/work/blazingmq/blazingmq/src/groups/bmq/bmqp/bmqp_protocol.h: In member function ‘BloombergLP::bmqp::MessagePropertiesInfo::SchemaIdType BloombergLP::bmqp::MessagePropertiesInfo::schemaId() const’:
/home/runner/work/blazingmq/blazingmq/src/groups/bmq/bmqp/bmqp_protocol.h:3518:27: warning: conversion from ‘unsigned int’ to ‘BloombergLP::bmqp::MessagePropertiesInfo::SchemaIdType’ {aka ‘short unsigned int’} may change value [-Wconversion]
 3518 |     return d_schemaWireId >> 1;
      |            ~~~~~~~~~~~~~~~^~~~
/home/runner/work/blazingmq/blazingmq/src/groups/bmq/bmqp/bmqp_protocol.h: In constructor ‘BloombergLP::bmqp::RdaInfo::RdaInfo(unsigned int)’:
[...]

PR #273, opened 2024-05-06
Fix: bmqp messageproperties Wconv warns

melvinhe:

**Describe your changes**
This PR is straightforward, addressing several ‘-Wconversion‘ errors relating to ‘bmqp_messageproperties.cpp‘ as well as associated header and test files with appropriate casting.

**Testing performed**
Double-checked that there's no data loss or overflow for each update. We had either checked sizes with assertions beforehand or the term was originally constrained to be of a relatively small size.


melvinhe:
Still waiting for the CI checks to pass to know for sure, but these files are also highly-referenced so I expect these changes to eventually remove a thousand or so occurrences of '-Wconversion' in Build [ubuntu]. With the Build logs errs down to only 3,682 after #257 gets merged, this means we should be pretty close to clearing issue #87 .

pniedzielski (approved):
👍🏻

PR #277, opened 2024-05-08
Remove unused and incompatible C++11 thread header

hallfox:
This is a C++11 and up header which we don't allow, and was caught by our internal CI checks. Fortunately it doesn't appear to be getting used, so we're just removing it.


678098 (approved):
The best PR I've seen on this week, totally approve!

PR #283, opened 2024-05-10
explicitly call gc when not idle

dorjesinpo:
Busy broker may not have enough idleness/time to `gcHistory`
Forcibly schedule it if needed.  It will intersperse with incoming work


pniedzielski (approved):
Increase GC message count comes in a different PR right?

dorjesinpo:
> Increase GC message count comes in a different PR right?

There should be no need.  Now, the gc reschedules itself when the count is not enough.  Before, it was waiting for idleness,

PR #287, opened 2024-05-13
Log watermark names in MointoredQueue warning logs

chrisbeard:
This PR attempts to improve the log emitted when a MonitoredQueue reaches one of its configured thresholds by annotating the thresholds with their respective names. Without this, it's difficult to understand which threshold value applies.


pniedzielski (approved):
👍🏻

PR #288, opened 2024-05-14
Enhance bmqtool to handle poison pill message dumps

alexander-e1off:
Enhance `bmqtool` to add ability to read and replay the `poison pill` message dumped in the file by the broker, so that consumers can debug their applications.

**Proposed changes**
Format of the dump file is defined in [QueueEngineUtil::dumpMessageInTempfile()](https://github.com/bloomberg/blazingmq/blob/c07f2a19d5c31611b19ee6491adc07ceec5af6c1/src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp#L258) method. Since message properties are dumped as a string, it is impossible to parse them properly. Thus, it is proposed to enhance dump, and add properties hexdump (similar way as payload).

Add new command to `bmqtool`:
`load-post uri=<uri> file=<path-to-dump-file>` 
Message content is loaded from dump file and posted to the specified uri.

**Testing performed**
Unit tests for parsing of dump file are added.
Manually with bmqtool simulated poison pill conditions and created dumps with and w/o message properties. Then tested these dumps with `load-post` commands.


src/applications/bmqtool/m_bmqtool_interactive.cpp
```diff
@@ -474,11 +578,8 @@ void Interactive::processCommand(const PostCommand& command, bool hasMPs)
 
         if (hasMPs) {
             d_session_p->loadMessageProperties(&properties);
-
-            for (size_t j = 0; j < command.messageProperties().size(); ++j) {
```
alexander-e1off, line 478: Not sure I understand what is the purpose of this `for` cycle. Inside InputUtil::populateProperties() there is also `for` cycle which is used to process properties. Removed it.

src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -319,5 +501,57 @@ bool InputUtil::populateSubscriptions(bmqt::QueueOptions*              out,
     return !failed;
 }
 
+bool InputUtil::decodeHexDump(bsl::ostream& out,
+                              bsl::istream& in,
```
678098, line 505: Output args should be passed as pointers according to BDE

src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -319,5 +501,57 @@ bool InputUtil::populateSubscriptions(bmqt::QueueOptions*              out,
...
+                              bsl::string*  error)
+{
```
678098, line 507: Might add preconditions

src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -319,5 +501,57 @@ bool InputUtil::populateSubscriptions(bmqt::QueueOptions*              out,
...
+                           // bdlb::Print::hexDump()
+    bdlde::HexDecoder hexDecoder;
+
+    bsl::string line;
+    int         numOut = 0;
+    int         numIn  = 0;
+
+    while (bsl::getline(in, line)) {
+        if (line.empty())
+            continue;  // skip empty lines
```
678098, line 518: This is discussable, but consider adding brackets `{}` here and in other places.
Motivation: avoid cases like this

```
if (condition)
    smth1();
    smth2();  // smth2() is executed unconditionally despite formatting
```

```
if (condition)
//    smth1();  // commented out, and unexpectedly smth2() now is under condition despite formatting
smth2();
```

src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -319,5 +501,57 @@ bool InputUtil::populateSubscriptions(bmqt::QueueOptions*              out,
...
+
+        // Sanity check
+        if (line.find(' ') == bsl::string::npos) {
+            if (error)
+                error->assign(
+                    "Wrong hexdump format, space delimeter is not detected");
+            return false;
```
678098, line 525:
```suggestion
            return false;  // RETURN
```

src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -319,5 +501,57 @@ bool InputUtil::populateSubscriptions(bmqt::QueueOptions*              out,
...
+        }
+
+        // Convert hexdump to binary, see format in bdlb::Print::hexDump()
+        bdlb::Tokenizer           tokenizer(line, " ");
+        bdlb::Tokenizer::iterator tokenizerIt = tokenizer.begin();
+        // Process tokens skipping first one (address offset)
```
678098, line 531:
```suggestion
        // Process tokens skipping the first one (address offset)
```

src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -319,5 +501,57 @@ bool InputUtil::populateSubscriptions(bmqt::QueueOptions*              out,
...
+        for (++tokenizerIt; tokenizerIt != tokenizer.end(); ++tokenizerIt) {
+            bslstl::StringRef token = *tokenizerIt;
+
+            // Stop when ASCII representation is detected
+            if (token.at(0) == '|')
+                break;
+
+            int status = hexDecoder.convert(outputBuffer,
```
678098, line 539:
```suggestion
            const int status = hexDecoder.convert(outputBuffer,
```

Might also change the name to the expected `rc`

src/applications/bmqtool/m_bmqtool_inpututil.h
```diff
@@ -72,6 +72,14 @@ struct InputUtil {
     verifyProperties(const bmqa::MessageProperties&      in,
                      const bsl::vector<MessageProperty>& properties);
 
+    /// Parse the specified `properties` (produced by
+    /// `MessageProperties::print()`) into the specified `out`. Return true on
+    /// success and false on error in which case load the error description
+    /// into the specified `error`.
```
678098, line 78:
```suggestion
    /// into the optionally specified `error`.
```

src/applications/bmqtool/m_bmqtool_inpututil.h
```diff
@@ -80,6 +88,13 @@ struct InputUtil {
                              bsl::string*       error,
                              const bsl::string& jsonInput);
 
+    /// Decode hexdump produced by bdlb::Print::hexDump() into binary format.
+    /// Read hexdump from the specified `in` and write binary into th specified
+    /// `out`. Return true on success and false on error in which case load the
+    /// error description into the specified `error`.
```
678098, line 94:
```suggestion
    /// error description into the optionally specified `error`.
```

src/applications/bmqtool/m_bmqtool_inpututil.h
```diff
@@ -80,6 +88,13 @@ struct InputUtil {
                              bsl::string*       error,
                              const bsl::string& jsonInput);
 
+    /// Decode hexdump produced by bdlb::Print::hexDump() into binary format.
+    /// Read hexdump from the specified `in` and write binary into th specified
```
678098, line 92:
```suggestion
    /// Read hexdump from the specified `in` and write binary into the specified
```

src/applications/bmqtool/m_bmqtool_interactive.cpp
```diff
@@ -85,6 +87,104 @@ void printMessage(bsl::ostream& out, int index, const bmqa::Message& message)
...
+    fileStream->close();
+}
+
+// Load message content from dump file, created by
+// QueueEngineUtil::dumpMessageInTempfile() method. Return `true` on success or
+// `false` otherwise. In case of `false`, `errorDescription` contains details.
+bool loadMessageContentFromFile(
+    bsl::vector<bsl::string>*     payload,
+    bsl::vector<MessageProperty>* messageProperties,
+    bsl::ostream&                 errorDescription,
```
678098, line 101: In some places, `bsl::string *` is passed to return error messages, here `bsl::ostream`.
Might be good to unify it. I believe passing `bsl::ostream` might be better, since it will be easier to output variable parameters like tokens, strings etc without intermediate `mwcu::MemOutStream`. What do you think?

src/applications/bmqtool/m_bmqtool_interactive.cpp
```diff
@@ -85,6 +87,104 @@ void printMessage(bsl::ostream& out, int index, const bmqa::Message& message)
...
+            }
+        }
+
+        bsl::string parseError;
+        if (!InputUtil::parseProperties(messageProperties,
+                                        propertiesStream.str(),
+                                        &parseError)) {
+            errorDescription << "Message properties parse error: "
+                             << parseError;
+            return false;
```
678098, line 159: Here, and in other places below
```suggestion
            return false;  // RETURN
```

src/applications/bmqtool/m_bmqtool_interactive.cpp
```diff
@@ -900,7 +1027,15 @@ int Interactive::mainLoop()
                     if (mps) {
                         hasMPs = keys.find(mps->name()) != keys.cend();
                     }
-                    processCommand(command, hasMPs);
+
+                    bool commandValid = true;
```
678098, line 1031:
```suggestion
                    bool commandIsValid = true;
```

src/applications/bmqtool/m_bmqtool_interactive.cpp
```diff
@@ -900,7 +1027,15 @@ int Interactive::mainLoop()
...
+                    if (!command.file().empty()) {
+                        commandValid = loadMessageFromFile(command);
+                        hasMPs       = !command.messageProperties().empty();
+                    }
+                    if (commandValid) {
+                        processCommand(command, hasMPs);
+                    }
```
678098, line 1038: Should we log something if the command is not valid?

678098 (changes requested):
A few comments

src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -257,15 +302,152 @@ void InputUtil::verifyProperties(
...
+
+    // Sanity check
+    if (properties.at(0) != '[' || properties.back() != ']') {
+        if (error) {
+            error->assign("Open or close markers is missed");
+        }
+        return false;  // RETURN
+    }
+
+    // Tokenize string by space and check markers
```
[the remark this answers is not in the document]
alexander-e1off: This function relies on data format produced by `MessageProperties::print()`, and no tabulations are used there. Space is used as properties separator.

src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -257,15 +302,152 @@ void InputUtil::verifyProperties(
...
+                return false;  // RETURN
+            }
+            state = VALUE;
+        } break;  // BREAK
+        case VALUE: {
+            // Special handling for binary value
+            if (messageProperty.type() ==
+                MessagePropertyType::Value::E_BINARY) {
+                // Check opening quote mark presense
+                if (token.at(0) != '"') {
```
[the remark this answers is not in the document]
alexander-e1off: Tokenizer is used with `soft` delimiter argument and will skip sequential separators. So, it is assured that there will be no empty tokens or tokens containing space.

src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -257,15 +302,152 @@ void InputUtil::verifyProperties(
...
+                mwcu::MemOutStream resultStream(' ');
+                for (++tokenizerIt; tokenizerIt != tokenizer.end();
+                     ++tokenizerIt) {
+                    token = *tokenizerIt;
+                    // Check closing quote mark after end line
+                    if (token.find("|\n\"") != bsl::string::npos) {
+                        break;  // BREAK
+                    }
+
+                    resultStream << token << ' ';
```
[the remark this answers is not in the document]
alexander-e1off: I agree with you that unit tests are required, probably I will add it in scope of this PR, not separate one. 
Ok, I will consider to improve algo, but again, we expect here not user input but content produced by `MessageProperties::print()` which is stable and have no any variances.

src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -257,15 +302,152 @@ void InputUtil::verifyProperties(
...
+                }
+
+                messageProperty.value() = resultStream.str();
+            }
+            else if (messageProperty.type() ==
+                     MessagePropertyType::Value::E_STRING) {
+                // Special handling for string value: it may contain spaces
+                // Join tokens by space (detokenize) inside surrounding quotes.
```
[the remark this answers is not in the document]
alexander-e1off: Agree to separate this code to the function and cover by UTs.

src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -257,15 +302,152 @@ void InputUtil::verifyProperties(
...
+                mwcu::MemOutStream resultStream;
+                for (; tokenizerIt != tokenizer.end(); ++tokenizerIt) {
+                    token = *tokenizerIt;
+                    resultStream << token;
+                    if (token.back() == '"') {
```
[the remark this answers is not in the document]
alexander-e1off: According to [Tokenizer doc](https://github.com/bloomberg/bde/blob/main/groups/bdl/bdlb/bdlb_tokenizer.h#L30)
```
A *soft* *delimiter* is a maximal (non-empty) sequence of soft-delimiter
characters.  Soft delimiters, typically whitespace characters, are used to
separate (rather than terminate) tokens, and thus never result in an empty
token.
```

src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -149,11 +150,55 @@ void InputUtil::populateProperties(
...
+            bsl::istringstream(value) >> bsl::boolalpha >> boolValue;
+            BSLA_MAYBE_UNUSED int result = out->setPropertyAsBool(name,
+                                                                  boolValue);
+            BSLS_ASSERT_SAFE(0 == result);
+        } break;  // BREAK
+
+        case MessagePropertyType::E_CHAR: {
+            BSLA_MAYBE_UNUSED int result = out->setPropertyAsChar(
+                name,
+                static_cast<char>(bsl::stoi(value, 0, 16)));
```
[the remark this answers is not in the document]
678098: > But since this is a debug tool for developers, I assume they are aware about format and value limits.

For this point, I prefer to reduce cognitive complexity for developers who use the tool. To automate checks since it's easy anyway, and can speed up the development/debug process. There is a difference between 2 paradigms: `the tool might break without telling, so better know how to pass params correctly` and `even if you pass incorrect params, the tool will notify you`.
If you think it's unnecessary for this PR, let's skip these proposals and also other ones which you think might increase the scope of the PR too much.

alexander-e1off:
@678098 What do you think about the following idea. To avoid reinventing the wheel during dump/parse message properties and keep existing functionality (human readable text) we can add properties hexdump using existing functionality:
 - During the message dump to the file, in addition to properties print, serialize properties into blob and hexdump them to the file the same way as payload is dumped;
 - When reading dump from the file, decode hexdump to binary and deserialize message properties;

In this case we keep dump file human readable (just one more hexdump will be added) and no need to create new format of properties dump. Below is simplified prototype
```C++
            // During message dump add properties hexdump
            bdlbb::PooledBlobBufferFactory bufferFactory(1024);
            const bdlbb::Blob& blob = properties.streamOut(&bufferFactory);
            dumpFileStream << bdlbb::BlobUtilHexDumper(&blob); 
        

            // When reading dump from the file
            mwcu::MemOutStream  binaryStream;
            bsl::istringstream inpStream(hexdumpStream.str());
            InputUtil::decodeHexDump(&binaryStream, inpStream)

            bdlbb::Blob resultBlob(&bufferFactory);
            bdlbb::BlobUtil::append(&resultBlob, binaryStream.str().data(), binaryStream.str().length());

            bmqa::MessageProperties properties;
            properties.streamIn(resultBlob);
```

alexander-e1off:
@678098 Would you please look again, I've rolled back previous solution and implemented new one, also added unit tests. Thanks you.

678098:
> @678098 Would you please look again, I've rolled back previous solution and implemented new one, also added unit tests. Thanks you.

Hi @alexander-e1off, thanks for the changes! Will look soon

678098 (approved):
LGTM!

PR #289, opened 2024-05-14
Fix: improve storage tool checks, disable default allocator check, fix mwcsys_executil

678098:
Storage tool:

- remove manual memory management
- add preconditions checks
- list all member fields on construction
- make the code robust to NULL allocator

### DIFF
--- a/src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp
+++ b/src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp
@@ -758,6 +758,10 @@ static void test12_printMessagesDetailsTest()
 {
     mwctst::TestHelper::printTestName("PRINT MESSAGE DETAILS TEST");
 
+    s_ignoreCheckDefAlloc = true;
+    // Disable default allocator check for this test until we can debug
+    // it on AIX/Solaris
+
     // Simulate journal file
     const size_t    k_NUM_RECORDS = 15;
     RecordsListType records(s_allocator_p);
--- a/src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp
+++ b/src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp
@@ -758,9 +758,11 @@ static void test12_printMessagesDetailsTest()
 {
     mwctst::TestHelper::printTestName("PRINT MESSAGE DETAILS TEST");
 
+#if defined(BSLS_PLATFORM_OS_SOLARIS) || defined(BSLS_PLATFORM_OS_AIX)
     s_ignoreCheckDefAlloc = true;
     // Disable default allocator check for this test until we can debug
     // it on AIX/Solaris
+#endif
 
     // Simulate journal file
     const size_t    k_NUM_RECORDS = 15;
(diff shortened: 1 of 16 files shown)


src/groups/mwc/mwcsys/mwcsys_executil.t.cpp
```diff
@@ -130,7 +130,7 @@ static void test2_executeSystemFailure()
 
     PVV("Testing abnormal exit of the command");
     rc = mwcsys::ExecUtil::execute(&output,
-                                   "python -c 'import os,signal; "
+                                   "python3 -c 'import os,signal; "
```
678098, line 133: We cannot use implicit `python2` here, `python3` alias might work on most frequently known systems

src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp
```diff
@@ -758,6 +758,12 @@ static void test12_printMessagesDetailsTest()
 {
     mwctst::TestHelper::printTestName("PRINT MESSAGE DETAILS TEST");
 
+#if defined(BSLS_PLATFORM_OS_SOLARIS) || defined(BSLS_PLATFORM_OS_AIX)
+    s_ignoreCheckDefAlloc = true;
+    // Disable default allocator check for this test until we can debug
+    // it on AIX/Solaris
+#endif
```
678098, line 765: I am not sure what code do allocations wrongly, it is difficult to find without means to build and test it on AIX/Solaris. I reviewed the storage tool and improved many things related to allocations, but maybe we should just disable this check until we debug it on the needed platforms

pniedzielski: This makes sense to me.  It's at least better to do this test by default on Linux, which will give us at least some early indications with GH Actions.

src/groups/mwc/mwcsys/mwcsys_executil.t.cpp
pniedzielski: I'm fine with this test.  If it fails, we can use use `python3.11` etc, but I'd rather not tie ourselves to a specific Python version if we can help it.

src/applications/bmqstoragetool/m_bmqstoragetool_testutils.h
```diff
@@ -102,8 +102,8 @@ class JournalFile {
     // Number of records.
     MappedFileDescriptor d_mfd;
     // Mapped file descriptor.
-    char* d_mem_p;
-    // Pointer to allocated memory for journal file.
+    bsl::vector<char> d_mem;
```
pniedzielski: Vector is fine (exception safe!).  I'm surprised there isn't a clear BDE component for this use case, but I couldn't find one with a quick search.

pniedzielski (approved):
👍🏻 as long as CI passes

src/applications/bmqstoragetool/m_bmqstoragetool_testutils.h
678098: I believe allocating data buffers via vector is a common practice for safe (not a joke, but still sounds funny) C++ programming, just as using smart pointers and other RAII things

678098: From the past days, we used `levmar` lib for optimizations in C++ code, with API like this: 
```
int dlevmar_dif(
void (*func)(double *p, double *hx, int m, int n, void *adata)
```
And we were really happy to use vector buffers on top of this

PR #297, opened 2024-05-21
Improve clarity in error message

hallfox:
The wording of "cannot call ... on an empty queue" makes it sound like this function errored (or failed a contract in this case) because the queue data structure was empty, whereas the meaning here is actually "on an empty queue **URI string**". For clarity's sake, we'll just say what we mean: the queue's URI is invalid, so this function is out of contract.


chrisbeard (approved):
👍 nice catch

PR #298, opened 2024-05-21
Fail reconfigure cmd on invalid auto subscription

dorjesinpo:
Fix.  If auto subscription expression is invalid, the reconfigure admin command should fail


src/groups/mqb/mqba/mqba_domainmanager.cpp
```diff
@@ -686,19 +686,25 @@ int DomainManager::processCommand(mqbcmd::DomainsResult*        result,
             }
         }
 
-        DecodeAndUpsertValue unused;
+        DecodeAndUpsertValue configureResult;
         d_configProvider_p->clearCache(domainSp->name());
         d_configProvider_p->getDomainConfig(
```
678098, line 691: No need to do anything, just my observation here

Note that docs for `getDomainConfig` are incorrect. They state that the method is asynchronous:
```
    /// Asynchronously get the configuration for the specified `domainName`
    /// and invoke the specified `callback` with the result.
```

In fact, this method is synchronous and thread safe, protected by mutes.
It means that you are safe to access `configureResult` just after the `getDomainConfig` call because you are guaranteed that the method was already executed.

I will update the docs later.

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -136,7 +138,40 @@ int validateConfig(bsl::ostream& errorDescription,
...
+                        << expression << ", rc: " << context.lastError()
+                        << ", reason: \"" << context.lastErrorMessage()
+                        << "\" ]";
+                    result = false;
+                }
+            }
+        }
+        else {
+            if (expression.text().length()) {
+                errorDescription << "Invalid Expression: [ expression: "
```
678098, line 167: Would rather leave a more explicit comment, so our users can clearly understand why it fails here

Something like
`
This expression version expects only empty expressions: [ expression: <...>, version: <...> ]
`

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -429,7 +464,10 @@ int Domain::configure(bsl::ostream&           errorDescription,
     }
 
     // Validate config. Return early if the configuration is not valid.
-    if (int rc = validateConfig(errorDescription, d_config, finalConfig)) {
+    if (int rc = validateConfig(errorDescription,
```
678098, line 467:
```suggestion
    if (const int rc = validateConfig(errorDescription,
```

src/integration-tests/test_auto_subscriptions.py
```diff
@@ -408,3 +408,95 @@ def test_auto_subscription_fanout_all_negative(self, cluster: Cluster):
...
+            {
+                "appId": "",
+                "expression": {"version": "E_VERSION_1", "text": "invalid expression"},
+            }
+        ]
+    )
+    def test_configure_invalid(self, cluster: Cluster):
+        """
+        Configure the priority queue to evaluate auto subscription negatively.
+        Make sure the queue does not get the message.
```
678098, line 423: For these both tests, there are no messages posted.
Does it worth it to extend these tests with producer?

src/integration-tests/test_auto_subscriptions.py
```diff
@@ -408,3 +408,95 @@ def test_auto_subscription_fanout_all_negative(self, cluster: Cluster):
...
+        )
+
+        cluster.config.domains[
+            tc.DOMAIN_PRIORITY_SC
+        ].definition.parameters.subscriptions[0]["expression"][
+            "text"
+        ] = "invalid expression"
+
+        cluster.reconfigure_domain(tc.DOMAIN_PRIORITY_SC, succeed=None)
+        # TODO: why not succeed=False
```
678098, line 494: Are there a problem with `reconfigure_domain` not checking a negative result?

dorjesinpo: Yes,  we want to make sure the command fails (so the whoever has issued the command can roll back)
The code insists on "processed successfully" 
```
    def command(self, command, succeed=None, timeout=BLOCK_TIMEOUT):
        """
        Send the specified 'cmd' command to the broker, prefixing it with 'CMD '.
        """

        cmd = "CMD " + command
        self.send(cmd)

        return (
            None
            if succeed is None
            else self.capture(f"'{cmd}' processed successfully", timeout=timeout)
        )
        ```

File: src/groups/mqb/mqbblp/mqbblp_domain.cpp [hunk 6]
```diff
@@ -136,7 +138,40 @@ int validateConfig(bsl::ostream& errorDescription,
         return rc_CHANGED_STORAGE_TYPE;  // RETURN
     }
 
-    return 0;
+    // Validate newConfig.subscriptions()
+
+    bsl::size_t size   = newConfig.subscriptions().size();
+    bool        result = true;
```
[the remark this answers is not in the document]
dorjesinpo: Well, let's be more implicit then, like `allSubscriptionsAreValid`

src/integration-tests/test_auto_subscriptions.py
678098: Then we should probably remove mentions of messages from these tests' docs

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -136,7 +138,40 @@ int validateConfig(bsl::ostream& errorDescription,
...
+
+    for (bsl::size_t i = 0; i < size; ++i) {
+        const mqbconfm::Expression& expression =
+            newConfig.subscriptions()[i].expression();
+
+        if (mqbconfm::ExpressionVersion::E_VERSION_1 == expression.version()) {
+            if (expression.text().length()) {
```
hallfox, line 151:
```suggestion
            if (!expression.text().empty()) {
```

I would prefer this for clarity and also in general for collection types it's usually much more efficient to check if a collection is empty than get its length.

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -136,7 +138,40 @@ int validateConfig(bsl::ostream& errorDescription,
...
+                bmqeval::CompilationContext context(allocator);
+
+                if (!bmqeval::SimpleEvaluator::validate(expression.text(),
+                                                        context)) {
+                    errorDescription
+                        << "Expression validation failed: [ expression: "
+                        << expression << ", rc: " << context.lastError()
+                        << ", reason: \"" << context.lastErrorMessage()
+                        << "\" ]";
+                    result = false;
```
hallfox: Can we just `return rc_INVALID_SUBSCRIPTION` here?

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -136,7 +138,38 @@ int validateConfig(bsl::ostream& errorDescription,
...
+            }
+        }
+        else {
+            errorDescription
+                << "Unsupported version: [ expression: " << expression << " ]";
+            allSubscriptionsAreValid = false;
+        }
+    }
+
+    return allSubscriptionsAreValid ? 0 : rc_INVALID_SUBSCRIPTION;
```
hallfox, line 172: This whole block is pretty complex, can we break it down into something more like this?

```c++
for (Subscriptions::const_iterator it = newConfig.subscriptions().cbegin(),
                                   end = newConfig.subscriptions().cend();
       it != end;
      ++it)
{
    const mqbconfm::Expression& expression = it->expression();
    if (!validateExpression(errorDescription, expression, allocator)) {
        return rc_INVALID_SUBSCRIPTION;
    } 
    return 0;
}

```

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -136,7 +138,40 @@ int validateConfig(bsl::ostream& errorDescription,
...
+            }
+        }
+        else {
+            if (expression.text().length()) {
+                errorDescription << "Invalid Expression: [ expression: "
+                                 << expression;
+                result = false;
+            }
+        }
+    }
```
hallfox: This whole block is pretty complex, can we break it down into something more like this?

```c++
for (Subscriptions::const_iterator it = newConfig.subscriptions().cbegin(),
                                                       end = newConfig.subscriptions().cend();
       it != end;
      ++it)
{
    const mqbconfm::Expression& expression = it->expression();
    if (!validateExpression(errorDescription, expression, allocator)) {
        return rc_INVALID_SUBSCRIPTION;
    } 
    return 0;
}

```

src/groups/mqb/mqbblp/mqbblp_domain.cpp
hallfox: Additionally I may even suggest trying to figure out which expresssions are invalid first before populating `errorDescription` so instead of getting a wall of 

```
"Expression validation failed: [ expression: "
                        << expression << ", rc: " << context.lastError()
                        << ", reason: \"" << context.lastErrorMessage()
                        << "\" ]";
```

for each expression, we get one error message with a list of invalid expressions.

678098: @hallfox I believe with this code we do not evaluate `validateExpression` if the bool flag was set to false before due to short-circuit evaluation with `&&` operator
`allSubscriptionsAreValid && validateExpression(errorDescription, expression, allocator);`

hallfox: Ah duh, good catch. It should be broken up into two lines

```c++
bool isValidExpression = validateExpression(errorDescription, expression, allocator);
allSubscriptionsAreValid = allSubscriptionsAreValid && isValidExpression;
```

678098: I also like the iterators approach since we don't actually need the exact index of a subscription

678098: > for each expression, we get one error message with a list of invalid expressions.

We might have details pointing out why exactly each expression is invalid. For example, if the expression is too complex, or if we have a syntax error. If we print just the expressions, it might make the debug process more difficult

678098: If the number of expressions is huge, though, we might use a LimitedPrinter instead, or print at most N errors

dorjesinpo: This is what we will see
`single           22May2024_13:59:23.696 (139804138272448) ERROR    *gmq.tsk.bmqbrkr.bmqbrkr bmqbrkr.m.cpp:324 Error processing command [rc: -2] error = 'Expression validation failed: [ expression: [ version = E_VERSION_1 text = "invalid expression" ], rc: -100, reason: "syntax error, unexpected property at offset 8" ]' config = '[ name = "bmq.test.mmap.priority.sc" mode = [ priority =  ] storage = [ domainLimits = [ messages = 2000 messagesWatermarkRatio = 0.8 bytes = 2097152 bytesWatermarkRatio = 0.8 ] queueLimits = [ messages = 1000 messagesWatermarkRatio = 0.8 bytes = 1048576 bytesWatermarkRatio = 0.8 ] config = [ fileBacked =  ] ] maxConsumers = 0 maxProducers = 0 maxQueues = 0 msgGroupIdConfig = NULL maxIdleTime = 0 messageTtl = 300 maxDeliveryAttempts = 0 deduplicationTimeMs = 300000 consistency = [ strong =  ] subscriptions = [ [ appId = "" expression = [ version = E_VERSION_1 text = "invalid expression" ] ] ] ]'`

I think, this is sufficient enough and there is no need to limit since this per admin command

src/groups/mqb/mqbblp/mqbblp_domain.cpp
dorjesinpo: No, I would prefer to validate and report everything

src/groups/mqb/mqbblp/mqbblp_domain.cpp
dorjesinpo: (there is no such type as mqbconfm::Subscriptions so involving iterator is not going to look prettier)

678098 (changes requested):
A few comments

And also, what do you think about merging 2 integration tests in one test? My concern here is that launching a cluster for each test takes 10-15 seconds. If we can change the tests scenario so we can launch cluster once, we can save some time.

In this PR, there are 2 scenarios:
_1:
- configure invalid auto subscription
- expect open queue fails
- reconfigure valid auto subscription
- expect queue opens

_2:
- configure valid auto subscription
- expect queue opens
- try reconfigure invalid auto subscription - expect admin command fail
- expect queue opens

These 2 scenarios could be merged into one to save cluster start time:
- configure invalid auto subscription
- expect open queue fails
- reconfigure valid auto subscription
- expect queue opens
- try reconfigure invalid auto subscription - expect admin command fail
- expect queue opens

So basically the test might check that the domain configuration remains in valid state once this state is reached.
If you don't want to increase the scope of this PR, but also agree that this change is good, I can do it myself in a separate PR once this is merged.

src/integration-tests/test_auto_subscriptions.py
```diff
@@ -408,3 +408,73 @@ def test_auto_subscription_fanout_all_negative(self, cluster: Cluster):
...
+    )
+    def test_invalid_configuration(self, cluster: Cluster):
+        """
+        Configure priority domain with invalid auto subscription.
+        Make sure a queue fails to open.
+        Reconfigure the domain with valid auto subscription.
+        Make sure a queue opens successfully.
+        Reconfigure the domain with invalid auto subscription.
+        Make sure the reconfigure command fails.
+        Make sure a queue opens successfully.
```
678098, line 428: Here, the formatter sees extra spaces at the line end
```suggestion
        Make sure a queue opens successfully.
```

PR #299, opened 2024-05-23
Fix: cache MessageEventBuilder properties on WRITE->READ switch

678098:
Bug mechanics:
1. `MessageEventBuilder::messageEvent()` downgrades the built `MessageEvent` from `WRITE` mode to `READ` mode
2. Both `MessageEventBuilder::messageEventSize()` and `MessageEventBuilder::messageCount()` asks `PutEventBuilder` for these properties, which is accessible only in `WRITE` mode
3. It's possible to downgrade to `READ` and then fail on assertion that we are in `WRITE` mode when using accessors from (2)

The fix is to cache these params and return them without asking `PutEventBuilder` in `READ` mode.

### DIFF
--- a/src/groups/bmq/bmqa/bmqa_messageeventbuilder.cpp
+++ b/src/groups/bmq/bmqa/bmqa_messageeventbuilder.cpp
@@ -184,6 +184,9 @@ void MessageEventBuilder::reset()
     typedef bsl::shared_ptr<bmqimp::Event> EventSP;
     EventSP& eventSpRef = reinterpret_cast<EventSP&>(d_impl.d_msgEvent);
 
+    d_impl.d_messageCountFinal     = 0;
+    d_impl.d_messageEventSizeFinal = 0;
+
     eventSpRef->upgradeMessageEventModeToWrite();
     // underlying PutEventBuilder is reset in
     // 'upgradeMessageEventModeToWrite'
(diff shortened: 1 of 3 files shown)


678098:
Hi @dorjesinpo
> We need to reset new data fields in the `reset` method
I believe the way how I initialize and use these fields is safe even without resetting them in `reset`. However, I've added a reset for these fields anyway.

> Why don't we put this offending sequence (`messageEvent()`, `messageCount()`/`messageEventSize()` )in some UT?
Added a UT with some usage scenarios of `MessageEventBuilder`, the offending sequence is a part of this UT, but also I make sure to test some other things.

> Just double checking, all the following ones are safe to assume the `MessageEventMode::e_WRITE`?
Looking into this now

678098:
> all the following ones are safe to assume the `MessageEventMode::e_WRITE`?

1.
```
bmqa::Message::clearPropertiesRef()
bmqa::Message::setCompressionAlgorithmType(enum bmqt::CompressionAlgorithmType::Enum)
bmqa::Message::setDataRef(const Blob *)
bmqa::Message::setDataRef(const char *, size_t)
bmqa::Message::setPropertiesRef(const MessageProperties *)
```
All these functions imply modifying `bmqa::Message`, so there is no much sense in calling them in `READ` mode. `READ` mode means that we built everything and intend to send this `MessageEvent` to the broker. However, I found that the existing docs don't clearly state this, and don't forbid accessing the `bmqa::Message`.

So, if we try to modify the `Message` in `READ` mode, we'll fail:
```
// MessageEventBuilder switches from WRITE mode to READ
builder.messageEvent();
bmqa::Message& msg = builder.currentMessage();  // get ref OK
msg.clearPropertiesRef();  // FAIL: assert WRITE mode expected
```

2.
```
bmqa::MessageEventBuilder::messageCount()
bmqa::MessageEventBuilder::messageEventSize()
```
For these 2 accessors, my fix allows to call them in `READ` mode now

3.
```
bmqa::MessageEventBuilder::packMessage(const QueueId &)
```
For this, we expect WRITE mode according to contract in the docs for this function

4.
```
bmqa::MessageEventBuilder::startMessage()
```
For this, we expect calling a `reset()` before, according to contract.

PR #306, opened 2024-05-30
CI: fix path to build_deps.sh script

678098:
Remove the upper `blazingmq` folder from the path since we are already in this folder
```
Run echo "deps_hash=`cat blazingmq/docker/build_deps.sh | shasum`" >> $GITHUB_OUTPUT
  echo "deps_hash=`cat blazingmq/docker/build_deps.sh | shasum`" >> $GITHUB_OUTPUT
  shell: /usr/bin/bash -e {0}
cat: blazingmq/docker/build_deps.sh: No such file or directory
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Get dependencies hash
         id: get-hash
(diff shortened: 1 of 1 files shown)


678098:
replaced by https://github.com/bloomberg/blazingmq/pull/331

PR #309, opened 2024-05-30
Lookup by appId

dorjesinpo:
Avoid iteration in `VirtualStorageCatalog::hasVirtualStorage(const bsl::string& appId` by switching to `TwoKeyHashMap`
Preparatory step for Reliable Broadcast

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.cpp
+++ b/src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.cpp
@@ -61,13 +61,13 @@ VirtualStorageCatalog::put(const bmqt::MessageGUID& msgGUID,
                            const mqbu::StorageKey&  appKey)
 {
     if (!appKey.isNull()) {
-        VirtualStoragesIter it = d_virtualStorages.find(appKey);
+        VirtualStoragesIter it = d_virtualStorages.findByKey2(appKey);
         BSLS_ASSERT_SAFE(it != d_virtualStorages.end());
 
-        return it->second->put(msgGUID,
-                               msgSize,
-                               rdaInfo,
-                               subScriptionId);  // RETURN
+        return it->value()->put(msgGUID,
+                                msgSize,
+                                rdaInfo,
+                                subScriptionId);  // RETURN
     }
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.h
```diff
@@ -58,8 +61,9 @@ class VirtualStorageCatalog {
     typedef bsl::shared_ptr<VirtualStorage> VirtualStorageSp;
 
     /// appKey -> virtualStorage
```
678098, line 63: Need to update the comment here.
`// Any(appId, appKey) -> virtualStorage`

src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.cpp
```diff
@@ -61,21 +61,21 @@ VirtualStorageCatalog::put(const bmqt::MessageGUID& msgGUID,
...
+                                subScriptionId);  // RETURN
     }
 
     // Add guid to all virtual storages.
 
     for (VirtualStoragesIter it = d_virtualStorages.begin();
          it != d_virtualStorages.end();
          ++it) {
-        it->second->put(msgGUID, msgSize, rdaInfo, subScriptionId);
+        it->value()->put(msgGUID, msgSize, rdaInfo, subScriptionId);
```
678098, line 78: We don't check for return codes here, what if one or few `put` fails?
Currently, `VirtualStorage::put` might fail if GUID is not unique
On comparison, we might return `e_GUID_NOT_UNIQUE` on line 67

678098 (dismissed):

src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.cpp
dorjesinpo: Looks like this codepath (`appKey.isNull()`) is for unique GUIDs only.  The callers do not event check the return code.
We can add error code without affecting anything, I guess

678098 (approved):

PR #312, opened 2024-06-04
Out-of-order PUSH

dorjesinpo:
Preparing for the Reliable Broadcast

To get rid of the second VSC in Replica and Proxy (`RelayQueueEngine::d_subStreamMessages_p`):
-  introduce OOO Push
- add OOO to RDL _and_ the (only) VSC in the storage
        
Today, proxy and replica add PUSH to both storage and the additional VSC
If we add a special treatment for OOO Push, then just the storage is enough for in-order PUSH
In-order guarantees first time delivery.
    
The next step would be to refactor VSC to rely on order and to batch Apps

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
@@ -338,7 +338,7 @@ void RemoteQueue::pushMessage(
     const bsl::shared_ptr<bdlbb::Blob>&  options,
     const bmqp::MessagePropertiesInfo&   messagePropertiesInfo,
     bmqt::CompressionAlgorithmType::Enum compressionAlgorithmType,
-    bool                                 isReplayPush)
+    bool                                 isOutOfOrder)
 {
     // executed by the *QUEUE DISPATCHER* thread
 
--- a/src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
@@ -337,7 +337,8 @@ void RemoteQueue::pushMessage(
     const bsl::shared_ptr<bdlbb::Blob>&  appData,
     const bsl::shared_ptr<bdlbb::Blob>&  options,
     const bmqp::MessagePropertiesInfo&   messagePropertiesInfo,
-    bmqt::CompressionAlgorithmType::Enum compressionAlgorithmType)
+    bmqt::CompressionAlgorithmType::Enum compressionAlgorithmType,
+    bool                                 isReplayPush)
 {
     // executed by the *QUEUE DISPATCHER* thread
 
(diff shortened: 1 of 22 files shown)


src/groups/mqb/mqbi/mqbi_dispatcher.h
```diff
@@ -1048,6 +1054,7 @@ class DispatcherEvent : public DispatcherDispatcherEvent,
     messagePropertiesInfo() const BSLS_KEYWORD_OVERRIDE;
     bmqt::CompressionAlgorithmType::Enum
                         compressionAlgorithmType() const BSLS_KEYWORD_OVERRIDE;
+    bool                isOutOfOrderPush() const BSLS_KEYWORD_OVERRIDE;
```
pniedzielski, line 1057: I know the semantics are obvious, but docs?

src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
```diff
@@ -337,7 +337,8 @@ void RemoteQueue::pushMessage(
     const bsl::shared_ptr<bdlbb::Blob>&  appData,
     const bsl::shared_ptr<bdlbb::Blob>&  options,
     const bmqp::MessagePropertiesInfo&   messagePropertiesInfo,
-    bmqt::CompressionAlgorithmType::Enum compressionAlgorithmType)
+    bmqt::CompressionAlgorithmType::Enum compressionAlgorithmType,
+    bool                                 isReplayPush)
```
pniedzielski, line 341: Here the name doesn't differ between the header and source, but still inconsistent with the rest of the call-chain.  This is eventually to be called on put-aside messages as well, yes?  In that case out-of-order is probably a better name.

pniedzielski (changes requested):
Minor changes

src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
dorjesinpo: Exactly.  Put-Aside is the reason for not calling this `isReplay`.  I guess, my Eclipse did not do a good job of renaming...

src/groups/mqb/mqbi/mqbi_dispatcher.h
dorjesinpo: These accessors seem to share the same doc
```
    // Return the value of the corresponding member.  Refer to the various
    // DispatcherEvent view interfaces for more specific information.
```

pniedzielski: As yes---I was seeing Doxygen docs where there are none (Doxygen won't copy docs of this way like BDE does).  That's fine.

pniedzielski (approved):

678098 (approved):

PR #313, opened 2024-06-04
ElectorStateMachine: Do not always enable with initial term of 0

kaikulimu:


src/groups/mqb/mqbnet/mqbnet_elector.cpp
```diff
@@ -1608,9 +1608,10 @@ void ElectorStateMachine::enable(int selfId,
         return;  // RETURN
     }
 
+    // DO NOT set `d_term` to 0 because the Elector could have started with a
+    // non-zero initial term.
     d_state                    = ElectorState::e_FOLLOWER;
     d_reason                   = ElectorTransitionReason::e_STARTED;
-    d_term                     = 0;
```
hallfox, line 1613: With this change are we guaranteed `d_term` is initialized during the other times it's accessed?

kaikulimu: This method is only called during [`mqbnet::Elector::start()`](https://github.com/bloomberg/blazingmq/blob/51ad3c587c38d60318c1543199b4d1359ca4d793/src/groups/mqb/mqbnet/mqbnet_elector.cpp#L2328-L2332), which currently overrides the [`initialTerm`](https://github.com/bloomberg/blazingmq/blob/51ad3c587c38d60318c1543199b4d1359ca4d793/src/groups/mqb/mqbnet/mqbnet_elector.cpp#L2233) set in the ctor. Given that `start` is called after the ctor, we should not be overriding the initial term value.

hallfox (approved):

PR #318, opened 2024-06-07
FSM mode: Fix error related to missing virtual storage [175227886]

kaikulimu:
When we tested switching cluster from non-FSM to FSM mode, we see errors such as `*.mqbblp.rootqueueengine mqbblp_rootqueueengine.cpp:398 #QUEUE_STORAGE_NOTFOUND Virtual storage does not exist for AppId 'foo', queue: 'bmq://bmq.test.mmap.fanout/qqq'`. This is because we used to not populate app Id infos in `QueueAssignmentAdviosry` for non-FSM mode; then, when we switch to FSM mode and attempt to recreate storages, we do not see those expected appIds and thus emit above error.

The fix is to always populate app Id infos in `QueueAssignmentAdviosry` regardless of mode.

**Caveat**: This PR does not retroactively change the storage files of any existing BlazingMQ broker. So, we would still run into issue if we restart an existing cluster from non-FSM mode to FSM mode.


678098 (approved):
lgtm

PR #226, opened 2024-03-22
Add AppId tag to CONFIRM_TIME_MAX and QUEUE_TIME_MAX metrics in fanout mode

waldgange:

PR #319, opened 2024-06-10
Refactor: better support for per-appId queue metrics

678098:
This PR is a preparation PR for introducing per-AppId queue depth metric, it rolls back some changes done in this PR https://github.com/bloomberg/blazingmq/pull/226 and also makes is possible to extend queue metrics more easily with per-AppId metrics.

Changes:
- Fixed metrics reporting duplication
- Simplified `QueueEngineUtil` interface by removing `reportQueueTimeMetric` which is used only locally
- Simplified `QueueStatsDomain` interface by removing complex functions `reportConfirmTime`, `reportQueueTime`, and introduce the second flavour of `onEvent` which expects `appId` parameter to report parameters to the nested SubContext
- So now we have both `onEvent(type,value)` to report global queue metrics and `onEvent(type,value,appId)` to report metrics specific for `appId`

As a result, it is easy to extend `onEvent(type,value,appId)` to log more per-AppId metrics.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
@@ -93,6 +93,15 @@ void releaseHandleAndInvoke(bdlmt::EventSchedulerEventHandle* handle,
     }
 }
 
+// Return delivered message's queue-time from the specified 'attributes'
+bsls::Types::Int64
+getMessageQueueTime(const mqbi::StorageMessageAttributes& attributes)
+{
+    bsls::Types::Int64 timeDelta;
+    mqbs::StorageUtil::loadArrivalTimeDelta(&timeDelta, attributes);
+    return timeDelta;
+}
+
 /// Callback to use in `QueueEngineUtil_AppState::tryDeliverOneMessage`
 struct Visitor {
     mqbi::QueueHandle* d_handle;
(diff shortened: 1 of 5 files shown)


src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -801,13 +798,22 @@ void QueueEngineUtil_AppsDeliveryContext::deliverMessage()
...
+            // First report 'queue time' metric for entire queue
+            d_queue_p->stats()->onEvent(
+                mqbstat::QueueStatsDomain::EventType::e_QUEUE_TIME,
+                timeDelta);
+
+            // Then report 'queue time' metric for all active appIds
             bsl::vector<bslstl::StringRef>::const_iterator it =
                 d_activeAppIds.begin();
             for (; it != d_activeAppIds.cend(); ++it) {
-                QueueEngineUtil::reportQueueTimeMetric(d_queue_p->stats(),
```
678098, line 808: Note that with the previous implementation, with N appIds we logged queue time 2*N times:
- N times we logged duplicating metric for the entire queue
- 1*N times we logged this metric per appId

My change will remove duplications here, so I log the metric N + 1 times:
- 1 time for the entire queue
- 1*N times per appId

hallfox (approved):
lgtm

PR #320, opened 2024-06-12
Feat[MQB]: track queue depth per appId

678098:
* Set queue stats to `VirtualStorageCatalog` on Primary and update per-appId queue metrics on these calls, if necessary
* Do `sort` json stats on admin command
* Add all the remaining queue metrics to json output
* Regenerate queue stats IT expected data


src/groups/mqb/mqbstat/mqbstat_queuestats.cpp
```diff
@@ -651,6 +661,23 @@ void QueueStatsDomain::updateDomainAppIds(
...
 
+QueueStatsDomain::EventCb
+QueueStatsDomain::buildEventCallback(EventType::Enum type)
+{
+    // disambiguate mqbstat::QueueStatsDomain::onEvent
+    bdlf::MemFn<void (mqbstat::QueueStatsDomain::*)(
+        mqbstat::QueueStatsDomain::EventType::Enum,
+        bsls::Types::Int64,
+        const bsl::string&)>
+        f(&mqbstat::QueueStatsDomain::onEvent);
```
678098, line 672: This disambiguate step could be removed if I merge 2 overloads `onEvent(type,value)` and `onEvent(type,value,appId)` into one where `appId` is optionally specified: `onEvent(type,value,*appId = nullptr)`. In this case, null `appId` ptr means that we want to update "the entire queue" metrics, non-null means that we want to update metric per specific appId.

It is simple to merge these `onEvent` functions, but the downside of this merge is the fact that not all the metrics which make sense for the entire queue make sense per appId.

src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.h
```diff
@@ -113,11 +119,13 @@ class VirtualStorageCatalog {
...
-                                  const bmqp::RdaInfo&     rdaInfo,
-                                  unsigned int             subScriptionId,
-                                  const mqbu::StorageKey&  appKey);
+    mqbi::StorageResult::Enum
+    put(const bmqt::MessageGUID& msgGUID,
+        int                      msgSize,
+        const bmqp::RdaInfo&     rdaInfo,
+        unsigned int             subScriptionId,
+        const mqbu::StorageKey&  appKey,
+        const OnStorageUpdateCb& putCb = OnStorageUpdateCb());
```
678098, line 128: Note that these `put`, `remove`, `removeAll` might also be used in the `RelayQueueEngine`, but we don't track these queue depth metrics there.

src/groups/mqb/mqbstat/mqbstat_jsonprinter.cpp
```diff
@@ -107,10 +107,14 @@ class JsonPrinterImpl {
...
-, d_opsPretty(bdljsn::WriteOptions().setSpacesPerLevel(4).setStyle(
-      bdljsn::WriteStyle::e_PRETTY))
+: d_opsCompact(bdljsn::WriteOptions()
+                   .setSpacesPerLevel(0)
+                   .setStyle(bdljsn::WriteStyle::e_COMPACT)
+                   .setSortMembers(true))
+, d_opsPretty(bdljsn::WriteOptions()
+                  .setSpacesPerLevel(4)
+                  .setStyle(bdljsn::WriteStyle::e_PRETTY)
+                  .setSortMembers(true))
```
678098, line 117: `sort` is necessary to have less diff when we update sample json data in ITs

src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.cpp
```diff
@@ -270,8 +271,15 @@ VirtualStorageCatalog::confirm(const bmqt::MessageGUID& msgGUID,
...
+    const mqbi::StorageResult::Enum rc = it->value()->confirm(&data->second);
+    if (queue() && mqbi::StorageResult::Enum::e_SUCCESS == rc) {
+        queue()->stats()->onEvent(
+            mqbstat::QueueStatsDomain::EventType::e_DEL_MESSAGE,
+            data->second.d_size,
+            it->key1());
+    }
 
-    return it->value()->confirm(&data->second);
+    return rc;
```
678098, line 282: @dorjesinpo I checked your proposal to do this update in `RootQueueEngine`. The problem is that we don't expose message size (`data->second.d_size`) from this level (`VirtualStorage`/`VirtualStorageCatalog`) to the upper level of `RootQueueEngine`.

src/groups/mqb/mqbstat/mqbstat_queuestats.cpp
```diff
@@ -544,6 +544,15 @@ void QueueStatsDomain::onEvent(EventType::Enum type, bsls::Types::Int64 value)
     case EventType::e_ADD_MESSAGE: {
         d_statContext_mp->adjustValue(DomainQueueStats::e_STAT_BYTES, value);
         d_statContext_mp->adjustValue(DomainQueueStats::e_STAT_MESSAGES, 1);
+        if (!d_subContextsHolder.empty()) {
+            bsl::list<StatSubContextMp>::iterator it =
+                d_subContextsHolder.begin();
+            while (it != d_subContextsHolder.end()) {
```
dorjesinpo, line 550: We got rid of the loop in the `put` method and here it is back again.
At the moment, I do not see a good way to utilize `VirtualStorageCatalog::numMessages` so let's keep the loop.
Provided we won't maintain statistics for all Dynamic Apps.

dorjesinpo (approved):

src/groups/mqb/mqbstat/mqbstat_queuestats.cpp
678098: Check for empty `d_subContextsHolder` should be fast if we don't have subcontexts for dynamic apps.
We might utilize `numMessages` on the time when we do snapshot, but this change is more complex, since we don't have a good way to ensure that each and every virtual storage is alive, and this change is intrusive.
I would rather merge this PR for now, since it helps to test and compare queue metrics. We can think of this in the future, when we have `prof` outputs for the scenario in question.

678098: Actually, I would also like to move metrics update code to the header and mark it `inline`, with hope that it will have less impact on performance. We have inline functions defined for `StatContext`, but we break the inline chain here when we define in `.cpp`. I would also like to do this change after we merge metrics update.

PR #321, opened 2024-06-12
Add support for BDE 4.9 in unit tests

alexander-e1off:
Starting from BDE 4.9 it is required to access the metrics registry default instance before assign the global allocator. Otherwise unit tests will fail at the global allocator check in TEST_EPILOG.


678098 (approved):
LGTM!

hallfox (approved):
Seems like this API will be stable enough in BDE for us to use

PR #324, opened 2024-06-17
bmqstoragetool UTs/mqbs_filestore/mqbs_filestoreprotocol: fix uninitialized memory

alexander-e1off:


src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp
```diff
@@ -57,7 +57,14 @@ using namespace TestUtils;
 //                                    TESTS
 // ----------------------------------------------------------------------------
 
-static void test1_breathingTest()
+// This test has MemorySanitizer false positive, disable uninitialized checks.
```
678098, line 60: Are you sure that this is a false positive?

678098 (commented):
Just one question

src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp
alexander-e1off: `MemorySanitizer` complains:
```
==42275==WARNING: MemorySanitizer: use-of-uninitialized-value
#0 0x55a742e52178 in test1_breathingTest() src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp:102:13
```
It is  [here](https://github.com/bloomberg/blazingmq/blob/main/src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp#L102), and `records` list is initialized [here](https://github.com/bloomberg/blazingmq/blob/main/src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp#L78). So, it seems that  `recordIter->first` cannot be uninitialized.

678098 (approved):
LGTM

PR #326, opened 2024-06-18
Fix data race in UT bmqimp_brokersession.t

alexander-e1off:


src/groups/bmq/bmqimp/bmqimp_brokersession.t.cpp
```diff
@@ -9927,8 +9910,7 @@ static void test54_distributedTrace()
                                                pQueue->options(),
                                                timeout);
     ASSERT_EQ(rc, bmqt::ConfigureQueueResult::e_SUCCESS);
-    ASSERT(localFns::waitForWriteCall(obj.channel().writeCalls(),
-                                      bsls::TimeInterval(0.1)));
+    ASSERT(obj.channel().waitFor(1, false, bsls::TimeInterval(1)));
```
678098, line 9913: I think the reason why we used `waitForWriteCall` instead of `waitFor` is because the implementation of the second contains a flaw:

https://github.com/bloomberg/blazingmq/blob/0225429cc3399f2703c30c9d2c42bd591cd2d6ea/src/groups/mwc/mwcio/mwcio_testchannel.cpp#L216-L222

Here, 2 details together create a problem:

1. We refresh `bsls::TimeInterval when` on every loop iteration, possibly moving our time limit into the future more and more.

2. From the `int bslmt::Condition::timedWait` docs:
`Note that spurious wakeups are rare but possible, i.e., this method may succeed (return 0) and return control to the thread without the condition object being signaled `

So, with these 2 things together, the loop might be never ending if rare conditions are met again and again.

The situation is unprobable, so we can ignore this maybe

678098 (approved):

src/groups/bmq/bmqimp/bmqimp_brokersession.t.cpp
alexander-e1off: Agree regarding potential problem, but `waitFor()` method is used in this test suite in many times.

678098: Let's merge then

PR #327, opened 2024-06-18
bmqstoragetool: split TestUtils into separate components

678098:


src/applications/bmqstoragetool/m_bmqstoragetool_testutils.h
```diff
@@ -79,206 +79,208 @@ namespace m_bmqstoragetool {
...
-                      const bmqt::MessageGUID& messageGUID,
-                      const bool               addNewLine = true);
-
-/// Value semantic type representing data message parameters.
-struct DataMessage {
-    int         d_line;
-    const char* d_appData_p;
-    const char* d_options_p;
-};
+struct TestUtils {
```
hallfox, line 82: For the purpose of code organization I think we should avoid putting all these classes in a struct. These helper Util structs should exist to support classes. I think we should just split `JournalFile`, `DataMessage`, and `FileManagerMock` to their own components in the bmqstoragetool package.

If we want them to be clearly marked for testing purposes only, then I think we should explain that in the component level documentation.

src/applications/bmqstoragetool/m_bmqstoragetool_testutils.h
```diff
@@ -79,206 +79,208 @@ namespace m_bmqstoragetool {
...
+    // =================
+    // class JournalFile
+    // =================
+
+    class JournalFile {
+      private:
+        // PRIVATE DATA
+
+        size_t d_numRecords;
+        // Number of records.
```
hallfox, line 101:
```suggestion
        // Number of records.
        size_t d_numRecords;
```
We should avoid doing this documentation below the item style

src/applications/bmqstoragetool/m_bmqstoragetool_testutils.h
678098: Extracted `JournalFile` and `FileManagerMock` and removed `m_bmqstoragetool_testutils`.
`DataMessage` struct and 2 smaller free functions from `TestUtils` could be local for one unit test, so I moved it there.

hallfox (approved):

PR #330, opened 2024-06-20
Refactor, CI: automatic license check, fix licenses

678098:
CI job output example:
```
Run python3 /home/runner/work/blazingmq/blazingmq/.github/workflows/ext/check_license.py --path=/home/runner/work/blazingmq/blazingmq | tee missing_licenses.txt
/home/runner/work/blazingmq/blazingmq/src/groups/mqb/mqbc/mqbc_clusterdata.cpp:
// BAD LICENSE EXAMPLE
```

### DIFF
--- a/.github/workflows/ext/check_license.py
+++ b/.github/workflows/ext/check_license.py
@@ -1,0 +1,91 @@
+# Copyright 2024 Bloomberg Finance L.P.
+# SPDX-License-Identifier: Apache-2.0
+#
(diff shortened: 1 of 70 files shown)


chrisbeard (approved):
Looks good to me. We should probably do similar for the client library repos?

678098:
> Looks good to me. We should probably do similar for the client library repos?

Yep, for client libs it's a good thing to do. But we should also extend this for other languages like Java.

PR #334, opened 2024-06-24
Feat[MQB]: Monolithic Virtual Storage

dorjesinpo:
Replacing VirtualStorage collections of GUIDs, one for each app,  with one collection per queue.  The new collection allocates consecutive memory (vector) if needed to keep apps states for each GUID.  Each VirtualStorage representing an App, gets assigned an ordinal to use as an index to access the state.
This allows us to:
- get rid of insert operations for each App upon receiving each PUT in Primary, or PUSH in Replica or Proxy.
- get rid of remove operations for each App upon message deletion
- align the delivery (PUSHing); QueueEngine can access all Apps states.  This lays the ground for Apps _batching_.  Reliable Broadcast will rely on App batching (as in the case of synthetic Apps) to address scaling concerns.
- get rid of maintaining Storage Iterators for each App (calling `beforeMessageRemoved` per App)
- get rid of most Apps iterations.  There are still some in `evaluateAutoSubscriptions`, initialization, and purging.

Replacing the extra `VirtualStorageCatalog` in `RelayQueueEngine` with new type of storage `PushStream`.  It is short-lived, keeping GUIDs only until they get PUSHed.  

CONFIRM logic is now inverted.  CONFIRM is not a delete operation now but a change of the state.


src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -259,6 +262,13 @@ int Application::start(bsl::ostream& errorDescription)
         }
     }
 
+    mqbi::ClusterResources resources;
+
+    resources.d_blobSpPool_p       = &d_blobSpPool;
+    resources.d_bufferFactory_p    = &d_bufferFactory;
+    resources.d_scheduler_p        = d_scheduler_p;
+    resources.d_pushElementsPool_p = &d_pushElementsPool;
```
678098, line 270: What about passing these fields to the constructor instead? My concern is about adding a new field in resources, with a field access like here it's easy to forget a new field somewhere. With a constructor extended with a new field, the compiler will point to all places where a new field is missing

src/groups/mqb/mqbblp/mqbblp_queueenginetester.cpp
```diff
@@ -935,6 +928,24 @@ void QueueEngineTester::post(const bslstl::StringRef& messages)
...
     parseMessages(&msgs, messages);
 
+    bmqp::Protocol::SubQueueInfosArray subscriptions(d_allocator_p);
+
+    if (d_subIds.empty()) {
+        subscriptions.push_back(
+            bmqp::SubQueueInfo(bmqp::Protocol::k_DEFAULT_SUBSCRIPTION_ID));
+    }
+    else {
+        // Assume, RelayQueueEngien will use upstreramSubQueueIds as the
```
678098, line 938: typo
```suggestion
        // Assume, RelayQueueEngine will use upstreramSubQueueIds as the
```

src/groups/mqb/mqbblp/mqbblp_queueenginetester.cpp
```diff
@@ -954,21 +965,32 @@ void QueueEngineTester::post(const bslstl::StringRef& messages)
...
         }
 
+        //        if (d_subStreamMessages_mp) {
+        //            d_subStreamMessages_mp->put(
+        //                msgGUID,
+        //                appData->length(),
+        //                d_rdaInfo,
+        //                bmqp::Protocol::k_DEFAULT_SUBSCRIPTION_ID,
+        //                mqbu::StorageKey::k_NULL_KEY);
+        //        }
```
678098, line 992: This code is not needed anymore, right?

src/groups/mqb/mqbblp/mqbblp_queueenginetester.h
```diff
@@ -368,7 +363,8 @@ class QueueEngineTester {
     /// format above.  The behavior is undefined unless `messages` is
     /// formatted as above and each message is unique (across the lifetime
     /// of this object), or if `createQueueEngine()` was not called.
-    void post(const bslstl::StringRef& messages);
+    void post(const bslstl::StringRef& messages,
+              RelayQueueEngine*        downstream = 0);
```
678098, line 367: The optionally specified `downstream` is not mentioned in docs

src/groups/mqb/mqbblp/mqbblp_queueenginetester.h
```diff
@@ -529,10 +526,10 @@ class QueueEngineTesterGuard {
 
     // ACCESSORS
 
-    /// Return the address of the modifiable QueueEngineTester object under
+    /// Return the address of the modifiable QueueEngine object under
     /// management by this proctor, or 0 if no QueueEngineTester is
```
678098, line 530:
```suggestion
    /// management by this proctor, or 0 if no QueueEngine is
```

src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -397,6 +404,14 @@ void QueueEngineUtil::logRejectMessage(
         << "deleted." << MWCTSK_ALARMLOG_END;
 }
 
+bsls::Types::Int64 QueueEngineUtil::getMessageQueueTime(
```
678098, line 407: In my previous PR, I turned this `getMessageQueueTime` to a local function, because it's  not used outside the `QueueEngineUtil` cpp file. If it's still not used outside, let's keep it hidden from `QueueEngineUtil` interface.

src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -873,75 +870,78 @@ QueueEngineUtil_AppState::~QueueEngineUtil_AppState()
...
 
 size_t
-QueueEngineUtil_AppState::deliverMessages(bsls::TimeInterval*     delay,
-                                          const mqbu::StorageKey& appKey,
-                                          mqbi::Storage&          storage,
-                                          const bsl::string&      appId)
+QueueEngineUtil_AppState::deliverMessages(bsls::TimeInterval*    delay,
+        mqbi::StorageIterator *reader,
+                                          mqbi::StorageIterator* start,
+                                          const mqbi::StorageIterator* end)
```
678098, line 876: Preconditions for delay, reader, start, end

src/groups/mqb/mqbblp/mqbblp_routers.t.cpp
```diff
@@ -72,18 +72,16 @@ struct TestStorage {
                 1,
                 d_domainCfg,
                 &d_capacityMeter,
-                bmqp::RdaInfo(),
                 allocator)
     , d_iterator(d_storage.getIterator(mqbu::StorageKey()))
     , d_bufferFactory(32, allocator)
-    , d_queue_sp(new (*allocator) mqbmock::Queue(0, allocator), allocator)
+    , d_queue_sp(new(*allocator) mqbmock::Queue(0, allocator), allocator)
```
678098, line 78: This might be a bug in the local clang-format
```suggestion
    , d_queue_sp(new (*allocator) mqbmock::Queue(0, allocator), allocator)
```

src/groups/mqb/mqbblp/mqbblp_routers.t.cpp
```diff
@@ -519,8 +549,12 @@ int main(int argc, char* argv[])
     // Initialize Crc32c
     bmqp::Crc32c::initialize();
 
+    // bslma::Default::setDefaultAllocator(&ma);
+
     TEST_PROLOG(mwctst::TestHelper::e_DEFAULT);
 
+    // bslma::DefaultAllocatorGuard _defAllocGuard(&ma);
```
678098, line 556: Changes in this file are not needed, right?

678098 (commented):
Some surface level comments

src/groups/mqb/mqba/mqba_application.cpp
dorjesinpo: Yeah, that was the old way and the constructor signature kept growing.  I thought, it would help to encapsulate everything in `Resources` so when we need something new from `mqba::Application` far away from it, we don't have to change signatures across the entire chain of constructors.
The concern about missing field is addressed by the fact that if a component needs the field, the compilation will fail.
The concern about missing field value can be addressed by an assert

src/groups/mqb/mqbblp/mqbblp_routers.t.cpp
dorjesinpo: The clang-format does without the space but the C++ Formatting Check does not pass so I had to manually insert the space.

src/groups/mqb/mqbblp/mqbblp_routers.t.cpp
dorjesinpo: Thanks!

src/groups/mqb/mqbblp/mqbblp_queueenginetester.cpp
```diff
@@ -954,19 +965,21 @@ void QueueEngineTester::post(const bslstl::StringRef& messages)
                                 msgs[i].data(),
                                 msgs[i].length());
 
+        // Consider this non-Proxy.  Imitate replication or Primary PUT
```
pniedzielski, line 968: The new comments in this file really helped. 👍🏻

src/groups/mqb/mqbblp/mqbblp_queueenginetester.h
```diff
@@ -642,37 +641,21 @@ QueueEngineTester::createQueueEngineHelper(mqbi::QueueEngine* engine)
...
-inline mqbi::QueueEngine* QueueEngineTester::createQueueEngine()
+inline T* QueueEngineTester::createQueueEngine()
 {
     // PRECONDITIONS
     BSLS_ASSERT_OPT(!d_queueEngine_mp && "'createQueueEngine()' was called");
-
+    T* result;
     // Create and configure Queue Engine
-    d_queueEngine_mp.load(new (*d_allocator_p)
+    d_queueEngine_mp.load(result = new (*d_allocator_p)
```
pniedzielski, line 650: Could we put this `result =` up into the initializer above?  Just to make it 100% clear later on that the assignment here is intentional?

src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -619,142 +635,122 @@ QueueEngineUtil_ReleaseHandleProctor::~QueueEngineUtil_ReleaseHandleProctor()
...
-    d_doRepeat       = false;
-    d_currentMessage = 0;
+    d_doRepeat = false;
     d_consumers.clear();
-    d_activeAppIds.clear();
 }
 
 bool QueueEngineUtil_AppsDeliveryContext::processApp(
-    QueueEngineUtil_AppState& app)
+    QueueEngineUtil_AppState& app,
```
pniedzielski, line 657: This logic somehow seems clearer to me than the previous version.  👍🏻

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1682,17 +1676,472 @@ bool RelayQueueEngine::subscriptionId2upstreamSubQueueId(
...
+
+            if (app == 0) {
+                BMQ_LOGTHROTTLE_ERROR()
+                    << "#QUEUE_UNKNOWN_SUBSCRIPTION_ID "
+                    << "Remote queue: " << d_queueState_p->uri()
+                    << " (id: " << d_queueState_p->id()
+                    << ") discarding a PUSH message for guid " << msgGUID
+                    << ", with unknown App Id " << upstreamSubQueueId;
+
+                return 0;  // RETURN
```
pniedzielski, line 1732: The return values of this function are a little opaque to me, and don't seem to be documented in the header file.

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1682,17 +1676,472 @@ bool RelayQueueEngine::subscriptionId2upstreamSubQueueId(
...
+                << ", with unconfigured Subscription Id " << subscriptionId;
+
             return false;
         }
         *subQueueId = itId->value().upstreamSubQueueId();
     }
     return true;
 }
 
+unsigned int
```
pniedzielski, line 1706: This function is the actual logic, and it is clear as day what's going on.  This is the major beneficiary of the refactor, code clarity-wise.

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1682,17 +1676,472 @@ bool RelayQueueEngine::subscriptionId2upstreamSubQueueId(
...
+        d_realStorageIter_mp->reset(msgGUID);
+        if (!d_realStorageIter_mp->atEnd()) {
+            mqbi::AppMessage& appState = d_realStorageIter_mp->appMessageState(
+                app->ordinal());
+
+            if (!appState.isPushing()) {
+                appState.onPush();
+            }
+            else {
+                BMQ_LOGTHROTTLE_INFO()
```
pniedzielski, line 1833: These might be good metrics to have down the road.  We can scrape them from logs, but it would be interesting to see, in production, how frequent duplicate PUSHs are.

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1682,17 +1676,472 @@ bool RelayQueueEngine::subscriptionId2upstreamSubQueueId(
...
+    return d_iterator->first;
+}
+
+PushStream::Element*
+RelayQueueEngine_PushStorageIterator::element(unsigned int appOrdinal) const
+{
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(!atEnd());
+
+    if (d_currentOrdinal > appOrdinal) {
```
pniedzielski, line 2017: Is this logic okay if `appOrdinal` is the invalid ordinal?  It seems like that will end up iterating `d_currentOrdinal` potentially until we hit an `ASSERT(d_currentElement)`.  Is this what we want, and if so, does it make sense to add a precondition/`ASSERT` that `appOrdinal` is not the invalid ordinal?

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1682,17 +1676,472 @@ bool RelayQueueEngine::subscriptionId2upstreamSubQueueId(
...
+}
+
+bool RelayQueueEngine_PushStorageIterator::atEnd() const
+{
+    return (d_iterator == d_owner_p->d_stream.end());
+}
+
+bool RelayQueueEngine_PushStorageIterator::hasReceipt() const
+{
+    if (atEnd()) {
```
pniedzielski, line 2074: `return !atEnd();` clearer?

src/groups/mqb/mqbi/mqbi_storage.h
```diff
@@ -379,74 +415,94 @@ class Storage {
...
     virtual void close() = 0;
 
-    /// Get an iterator for items stored in the virtual storage identified
-    /// by the specified `appKey`.  Iterator will point to point to the
-    /// oldest item, if any, or to the end of the collection if empty.  Note
-    /// that if `appKey` is null, an iterator over the underlying physical
-    /// storage will be returned.  Also note that because `Storage` and
-    /// `StorageIterator` are interfaces, the implementation of this method
-    /// will allocate, so it's recommended to keep the iterator.
+    /// Get an iterator for data stored in the virtual storage identified by
```
pniedzielski, line 418: Single tick -> backtick in all Doxygen comments in this file.

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.h
```diff
@@ -127,16 +127,368 @@ struct RelayQueueEngine_AppState : QueueEngineUtil_AppState {
...
-        bslma::Allocator*                        allocator = 0);
+        unsigned int                  upstreamSubQueueId,
+        const bsl::string&            appId,
+        mqbi::Queue*                  queue,
+        bdlmt::EventScheduler*        scheduler,
+        Routers::QueueRoutingContext& queueContext,
+        bslma::Allocator*             allocator = 0);
 };
 
+struct PushStream {
```
pniedzielski, line 138: This documentation comment is quite terse, and seems to put explaining why this data structure is shaped the way it is before what properties the data structure has.  I could eventually parse that information out, but a sentence at the start of "This data structure stores contexts of PUSHs, allowing for efficient sequential access by GUID and by App, and random-access removal for GUID."  All the information here is good, but I could only make sense of it since you've explained the design before.

pniedzielski: Drawing out the shape of the data structure also helped me here.  I'm not sure if ASCII art is truly worth it, but it's complex to see what this data structure is just from the code (even though it's not terribly complex at a high level).

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.h
```diff
@@ -127,16 +127,368 @@ struct RelayQueueEngine_AppState : QueueEngineUtil_AppState {
...
+        Element* nextInApp() const;
+    };
+
+    Stream d_stream;
+
+    Apps d_apps;
+
+    bdlma::ConcurrentPool* d_pushElementsPool_p;
+
+    PushStream(bdlma::ConcurrentPool* pushElementsPool);
```
pniedzielski, line 262: Precondition on `pushElementsPool` is v. important.

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.h
```diff
@@ -127,16 +127,368 @@ struct RelayQueueEngine_AppState : QueueEngineUtil_AppState {
...
+
+    /// Destroy the specified 'element'
+    void destroy(Element* element, bool doKeepGuid);
+};
+
+// ==========================================
+// class RelayQueueEngine_PushStorageIterator
+// ==========================================
+
+class RelayQueueEngine_PushStorageIterator : public mqbi::StorageIterator {
```
pniedzielski, line 306: As mentioned above, should this really be a `StorageIterator`?  Or rather, it's not a `StorageIterator`; it has different semantics.  Is it one for convenience or is there a deeper need for it to be one?

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.h
```diff
@@ -127,16 +127,368 @@ struct RelayQueueEngine_AppState : QueueEngineUtil_AppState {
...
+    /// 'itApp'.    Erase each corresponding GUIDs from the  PushStream with no
+    /// remaining Elements.
+    /// Erase the corresponding App.
+    unsigned int removeApp(Apps::iterator itApp);
+
+    /// Remove all Elements, Apps, and GUIDs.
+    unsigned int removeAll();
+
+    /// Create new Element associated with the specified 'info',
+    // 'upstreamSubQueueId', and 'iterator'.
```
pniedzielski, line 293: Triple slash this line

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.h
```diff
@@ -496,13 +858,39 @@ class RelayQueueEngine : public mqbi::QueueEngine {
...
     mqbi::StorageResult::Enum evaluateAutoSubscriptions(
         const bmqp::PutHeader&              putHeader,
         const bsl::shared_ptr<bdlbb::Blob>& appData,
         const bmqp::MessagePropertiesInfo&  mpi,
         bsls::Types::Uint64                 timestamp) BSLS_KEYWORD_OVERRIDE;
 
+    // MANIPULATORS
+
+    /// If the specified 'isOutOfOrder' is 'true', insert the specified
+    // 'msgGUID' into the corresponding App redelivery list.  Otherwise, insert
```
pniedzielski, line 885: Triple slash this line.

pniedzielski (changes requested):
Overall comments:

1. The logic for `RelayQueueEngine` becomes **much** clearer with this refactor.  Even more so than in making way for reliable broadcast, I find this to be the most important change in this PR.
2. Maybe as the cost of that, I find the code for `PushStream` quite opaque.  The code feels strongly coupled to the routing logic, whereas really it's a generic data structure with a few indices.  I'm not sure if you hadn't explained the design in person to me that I would have been able to reconstruct it from this code.  I don't need it to be pulled out into an `mwcc`-style component, but making the shape of the data structure clearer, in docs and in code, would go a long way towards making the routing logic more transparent.
3. I have many comments below about `ClusterResources`.  Not because it's an important change, but because it's a change that (necessarily) touches so many places.  It's a good change, rather than adding more and more parameters to all these constructors.  However, I'm very concerned that each of the fields is `nullptr` in different places, and when I see an instance of `ClusterResources`, I have no good way of knowing what should and should not be `nullptr`.  Ideally, none of them could be `nullptr`, and in the few places we keep them null we bite the bullet and construct something for them to point to.  If some of them truly are optional, I'd like that to be documented.  If all of them are optional, I think that needs to be reconsidered.
4. In Doxygen comments (triple slashes), backticks are used, not single ticks.  I started noting those down, but the changes are extensive enough that I missed many of them.

I haven't taken a very deep look at the `InMemoryStorage` + `FileBackedStorage` changes.

pniedzielski:
(I think perhaps having multiple commits could have made this review go faster.)

src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.cpp
```diff
@@ -34,11 +36,70 @@ namespace mqbs {
...
+//
+// (1) QE::beforeMessageRemoved
+//
+
 // CREATORS
 VirtualStorageCatalog::VirtualStorageCatalog(mqbi::Storage*    storage,
                                              bslma::Allocator* allocator)
 : d_storage_p(storage)
 , d_virtualStorages(allocator)
+, d_avaialbleOrdinals(allocator)
```
pniedzielski, line 97: Typo: `available`

src/groups/mqb/mqbblp/mqbblp_remotequeue.h
```diff
@@ -467,17 +461,15 @@ class RemoteQueue {
     /// THREAD: This method is called from the Queue's dispatcher thread.
     void onOpenFailure(unsigned int upstreamSubQueueId);
 
+    /// Return the event scheduler associated with this remote queue.
```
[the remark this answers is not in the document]
kaikulimu: I generally consider any function with a `const` at the end be be an `// ACCESSOR`.

kaikulimu (changes requested):
Some comments

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.h
```diff
@@ -127,16 +127,368 @@ struct RelayQueueEngine_AppState : QueueEngineUtil_AppState {
...
+
+    Stream d_stream;
+
+    Apps d_apps;
+
+    bdlma::ConcurrentPool* d_pushElementsPool_p;
+
+    PushStream(bdlma::ConcurrentPool* pushElementsPool);
+
+    iterator add(const bmqt::MessageGUID& guid);
```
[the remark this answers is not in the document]
dorjesinpo: Good idea.  Renaming it to `findOrAppendMessage`

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.h
```diff
@@ -127,16 +127,368 @@ struct RelayQueueEngine_AppState : QueueEngineUtil_AppState {
...
+        void remove(Element* element, eElementList where);
+
+        /// Return the first Element in the list
+        Element*     first() const;
+        unsigned int numElements() const;
+    };
+
+    struct App {
+        Elements                                   d_elements;
+        bsl::shared_ptr<RelayQueueEngine_AppState> d_app;
```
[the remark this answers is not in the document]
dorjesinpo: What we could try to do is to forward declare `RelayQueueEngine_AppState` and move `PushStream`, `RelayQueueEngine_PushStorageIterator`, and `RelayQueueEngine_VirtualPushStorageIterator` into `mqbblp_pushstream`.
And rename the latter two to `PushStreamIterator` and `VirtualPushStreamIterator`

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.h
```diff
@@ -127,16 +127,368 @@ struct RelayQueueEngine_AppState : QueueEngineUtil_AppState {
...
+struct PushStream {
+    // A Storage for one-time PUSH delivery (non the Reliable Broadcast mode).
+    // When PUSH is a result of round-robin, the number of App ids in the PUSH
+    // is not equal to the number of Apps known to the RelayQueueEngine.
+    // An efficient iteration requires records of variable size per each GUID.
+    // On the other side, there is only sequential access - either for each
+    // GUID or for each App.  An 'Element' holding PUSH context for a GUID and
+    // an App is in two lists - for the GUID for the App.  Removal can be
+    // random for the GUID list and always sequential for the App list.
+    // The storage still supports 'mqbi::StorageIterator' interface but the
```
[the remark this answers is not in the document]
dorjesinpo: Adding comment
```
// The 'mqbi::StorageIterator' interface support is important for QueueEngines
// which access data or delivery exclusively using 'mqbi::StorageIterator' 
// interface for both data in "real" storage or 'PushStream'.
```

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.h
```diff
@@ -625,10 +1008,311 @@ inline void RelayQueueEngine::ConfigureContext::initializeRouting(
...
+{
+    stream.erase(d_iteratorGuid);
+}
+
+inline void PushStream::Element::eraseApp(PushStream::Apps& apps)
+{
+    apps.erase(d_iteratorApp);
+}
+
+inline mqbi::AppMessage* PushStream::Element::appState()
```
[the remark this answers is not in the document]
dorjesinpo: The difference between `appState` and `appView ` is more than a `const`.  The latter will not allocate any state and the former will

src/groups/mqb/mqbblp/mqbblp_cluster.h
```diff
@@ -501,20 +501,18 @@ class Cluster : public mqbi::Cluster,
     // CREATORS
 
     /// Create a new object representing a cluster having the specified
-    /// `name`, `clusterConfig` and `statContexts`, associated to the
-    /// specified `netCluster` and using the specified `domainFactory`,
-    /// `scheduler`, `dispatcher`, `blobSpPool` and `bufferFactory`.  Use
-    /// the specified `allocator` for any memory allocation.
+    /// 'name', 'clusterConfig' and 'statContexts', associated to the
```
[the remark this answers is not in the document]
dorjesinpo: Thanks!  Will replace

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
```diff
@@ -375,9 +366,10 @@ ClusterCatalog::ClusterCatalog(bdlmt::EventScheduler*    scheduler,
 , d_reversedClusterConnections(d_allocator_p)
 , d_clusters(d_allocator_p)
 , d_statContexts(statContexts)
+, d_resources(resources)
 {
     // PRECONDITIONS
-    BSLS_ASSERT_SAFE(scheduler->clockType() ==
+    BSLS_ASSERT_SAFE(d_resources.d_scheduler_p->clockType() ==
```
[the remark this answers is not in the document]
dorjesinpo: Enforcing non-null private data in `ClusterResources` ctor

src/groups/mqb/mqbblp/mqbblp_remotequeue.h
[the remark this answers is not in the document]
dorjesinpo: > I generally consider any function with a `const` at the end be be an `// ACCESSOR`.

same here

src/groups/mqb/mqbblp/mqbblp_remotequeue.t.cpp
```diff
@@ -179,6 +181,8 @@ TestBench::TestBench(bslma::Allocator* allocator_p)
 , d_timeSource(&d_scheduler)
 , d_testClock(d_timeSource)
 , d_stateSpPool(8192, allocator_p)
+, d_pushElementsPool(sizeof(mqbblp::PushStream::Element), allocator_p)
+, d_resources()
```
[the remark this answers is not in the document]
dorjesinpo: refactoring

src/groups/mqb/mqbblp/mqbblp_remotequeue.t.cpp
```diff
@@ -195,6 +199,11 @@ TestBench::TestBench(bslma::Allocator* allocator_p)
         bdlf::BindUtil::bind(&TestClock::monotonicClock, &d_testClock),
         bdlf::BindUtil::bind(&TestClock::highResTimer, &d_testClock));
 
+    d_resources.d_blobSpPool_p       = 0;
```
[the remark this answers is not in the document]
dorjesinpo: If we want to be strict about `ClusterResources`, we need to add `d_blobSpPool`

src/groups/mqb/mqbblp/mqbblp_routers.cpp
```diff
@@ -381,9 +381,9 @@ void Routers::AppContext::load(
     }
 }
 
-size_t Routers::AppContext::finalize()
+unsigned int Routers::AppContext::finalize()
```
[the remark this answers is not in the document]
dorjesinpo: Yes.  Since it does not require `size_t`, `unsigned int` is easier.  Strictly speaking, we should use `bsl::size_t` instead of `size_t`

src/groups/mqb/mqbi/mqbi_cluster.h
```diff
@@ -414,6 +414,33 @@ class Cluster : public DispatcherClient {
     clusterProxyConfig() const = 0;
 };
 
+struct ClusterResources {
```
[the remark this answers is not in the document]
dorjesinpo: In some cases, yes but replacing this to a strict version of never null values

src/groups/mqb/mqbi/mqbi_storage.h
```diff
@@ -261,6 +261,37 @@ bool operator==(const StorageMessageAttributes& lhs,
...
+    enum State { e_NONE = 0, e_PUT = 1, e_PUSH = 2, e_CONFIRM = 3 };
+
+    bmqp::RdaInfo d_rdaInfo;
+    unsigned int  d_subscriptionId;
+    State         d_state;
+
+    AppMessage(const bmqp::RdaInfo& rdaInfo);
+
+    /// Set this object state to indicate a delivery by (Replica or Proxy).
+    void onPush();
```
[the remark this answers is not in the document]
dorjesinpo: I am changing this to `setPushState()`, `setConfirmState()`, `setRemovedState()` but let's see if we can find better names.
(it is not unusual for observers (`onSmth`) to change their state)

src/groups/mqb/mqbi/mqbi_storage.h
```diff
@@ -352,6 +386,8 @@ class Storage {
                         bmqp::Protocol::k_SUBID_ARRAY_STATIC_LEN>
         StorageKeys;
 
+    static const size_t k_INVALID_ORDINAL = 999999;
```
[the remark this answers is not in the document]
dorjesinpo: The former, we need to come back and accord it with `FileStoreProtocol::k_MAX_MSG_REF_COUNT_HARD` or its successor in one of subsequent steps.
Not sure about `std::numeric_limits<std::size_t>::max()` as we may need more than one special value.

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.h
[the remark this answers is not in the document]
pniedzielski: I like this strategy.  It should tame this file quite a bit.

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
dorjesinpo: Adding
```
/// Return number of inserted PushStream Elements.
```

src/groups/mqb/mqbblp/mqbblp_queueenginetester.h
dorjesinpo: It should be refactored, of course

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
dorjesinpo: This will assert if `appOrdinal` is not valid (not just `k_INVALID_ORDINAL`) at  ASSERT(d_currentElement)
Adding assert does not hurt

src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -619,142 +635,122 @@ QueueEngineUtil_ReleaseHandleProctor::~QueueEngineUtil_ReleaseHandleProctor()
...
-            !app.d_storageIter_mp->hasReceipt())) {
-        // An unregistered App Id has void iterator that always points
-        // at the end.
-        // Or, still waiting for Receipt
-        BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
+    BSLS_ASSERT_SAFE(d_currentMessage->hasReceipt());
+
+    const bool isBroadcast = d_queue_p->isDeliverAll();
+
+    if (isBroadcast) {
```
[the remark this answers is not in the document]
dorjesinpo: Up until reliable broadcast, `isDeliverAll()` and `isAtMostOnce()` are synonyms.  Reliable broadcast can be `isDeliverAll() && !isAtMostOnce()`.  Let's return to this, when we introduce this queue mode

src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -796,63 +794,63 @@ void QueueEngineUtil_AppsDeliveryContext::deliverMessage()
...
+    return d_consumers.empty();
+}
+
+bsls::Types::Int64 QueueEngineUtil_AppsDeliveryContext::timeDelta()
+{
+    if (!d_timeDelta.has_value()) {
+        d_timeDelta = getMessageQueueTime(d_currentMessage->attributes());
+    }
+    return d_timeDelta.value();
+}
```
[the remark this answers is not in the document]
dorjesinpo: > Do we really need this method?

We need the cache to avoid extra work.

Yes, we need to reset it in `reset`

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -93,60 +90,84 @@ size_t RootQueueEngine::deliverMessages(AppState*               app,
...
+    }
+    else {
+        start = d_storageIter_mp.get();
+    }
+
     bsls::TimeInterval delay;
-    size_t             numMessages =
-        app->deliverMessages(&delay, key, *d_queueState_p->storage(), appId);
+    size_t             numMessages = app->deliverMessages(&delay,
+            d_realStorageIter_mp.get(),
```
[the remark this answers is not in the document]
dorjesinpo: `reader` is the interface to read date out of whatever storage we use.
Adding 
```
/// Use the specified `reader` to read data for delivery.
```

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.h
```diff
@@ -140,6 +140,12 @@ class RootQueueEngine BSLS_KEYWORD_FINAL : public mqbi::QueueEngine {
     // Throttler for when reject messages
     // are dumped into temp files.
 
+    bslma::ManagedPtr<mqbi::StorageIterator> d_storageIter_mp;
```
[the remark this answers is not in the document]
dorjesinpo: Added
```
    // Storage iterator to the logical stream of messages.  
    // Queue Engine iterates this one sequentially.
```
and
```
    // Storage iterator to access storage state.
    // Queue Engine uses this one to access random message (as in the case of
    // redelivery).
```

src/groups/mqb/mqbblp/mqbblp_queueengineutil.h
```diff
@@ -403,24 +404,24 @@ struct QueueEngineUtil_AppState {
...
+    /// App to make progress: any configuration change, capacity availability.
     /// Note that depending upon queue's mode, messages are delivered either
     /// to all consumers (broadcast mode), or in a round-robin manner (every
     /// other mode).
-    size_t deliverMessages(bsls::TimeInterval*     delay,
-                           const mqbu::StorageKey& appKey,
-                           mqbi::Storage&          storage,
-                           const bsl::string&      appId);
+    size_t deliverMessages(bsls::TimeInterval*          delay,
+                           mqbi::StorageIterator*       reader,
```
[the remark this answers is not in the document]
dorjesinpo:
```
/// Use the specified `reader` to read data for delivery.
```

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -93,60 +90,84 @@ size_t RootQueueEngine::deliverMessages(AppState*               app,
...
+            storageIter_mp = d_queueState_p->storage()->getIterator(
+                app->appKey());
+        }
+        start = storageIter_mp.get();
+    }
+    else {
+        start = d_storageIter_mp.get();
+    }
+
     bsls::TimeInterval delay;
```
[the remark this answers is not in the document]
dorjesinpo: the `delay` is an out parameter for `deliverMessages`
the `numMessages` is `const`, indeed

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -93,60 +90,84 @@ size_t RootQueueEngine::deliverMessages(AppState*               app,
...
     }
 
-    return numMessages;
+    if (app->isReadyForDelivery()) {
+        // If the 'app' has caught up with the queue data stream, need to
+        // continue the delivery from the queue position in the stream.
+        // Cannot rely on 'LocalQueue' calling 'afterNewMessage' since it turns
+        // off 'd_hasNewMessages'.  Just call it explicitly.
+        const bmqt::MessageGUID dummy;
+        afterNewMessage(dummy, 0);
```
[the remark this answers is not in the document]
dorjesinpo: This is something historical.  The interface has then because of potential use (although there is none at the moment).

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -93,60 +90,84 @@ size_t RootQueueEngine::deliverMessages(AppState*               app,
...
+            // This code relies on TTL per Queue (Domain), not per message - if
+            // 'resumePoint()' has exceeded the TTL, everything before that had
+            // as well.
+            storageIter_mp = d_queueState_p->storage()->getIterator(
+                app->appKey());
+        }
+        start = storageIter_mp.get();
+    }
+    else {
+        start = d_storageIter_mp.get();
```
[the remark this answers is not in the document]
dorjesinpo: Added
```
        // 'start' points at either the resume point (if found) or the first
        // unconfirmed message of the 'app' (if not found).
```
and
```
        // 'start' points at the next message in the logical stream (common
        // for all apps).
```

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1240,20 +1271,38 @@ void RootQueueEngine::afterNewMessage(
...
+                // Report queue time metric per App
+                // Report 'queue time' metric for all active appIds
+                d_queueState_p->queue()->stats()->onEvent(
+                    mqbstat::QueueStatsDomain::EventType::e_QUEUE_TIME,
+                    context.timeDelta(),
+                    app->appId());
             }
         }
+        if (!context.isEmpty()) {
+            // Report 'queue time' metric for the entire queue
```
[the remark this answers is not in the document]
dorjesinpo: Yes, it was reported in the `QueueEngineUtil_AppState`/`QueueEngineUtil_AppsDeliveryContext` upon checking if this is Primary (`RootQueueEngine`).
Moved it to. the `RootQueueEngine`

src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -265,12 +265,11 @@ int Application::start(bsl::ostream& errorDescription)
         }
     }
 
-    mqbi::ClusterResources resources;
-
-    resources.d_blobSpPool_p       = &d_blobSpPool;
-    resources.d_bufferFactory_p    = &d_bufferFactory;
-    resources.d_scheduler_p        = d_scheduler_p;
-    resources.d_pushElementsPool_p = &d_pushElementsPool;
+    mqbi::ClusterResources resources(
```
pniedzielski: 👍🏻 

Recording for posterity: one of the things we thought about and could implement in the future if needed is to make the pools lazy-initialized, so they're only created when first needed.  We're not doing this now, but it could be worthwhile.

src/groups/mqb/mqbi/mqbi_storage.h
```diff
@@ -669,17 +670,17 @@ inline AppMessage::AppMessage(const bmqp::RdaInfo& rdaInfo)
     // NOTHING
 }
 
-inline void AppMessage::onPush()
+inline void AppMessage::setPushState()
```
pniedzielski, line 673: To me, at least, this change makes the code a little clearer.

src/groups/mqb/mqbi/mqbi_cluster.h
```diff
@@ -455,15 +486,73 @@ inline bool Cluster::isFSMWorkflow() const
...
+: d_scheduler_p(scheduler)
+, d_bufferFactory_p(bufferFactory)
+, d_blobSpPool_p(blobSpPool)
+, d_pushElementsPool(pushElementsPool)
+{
+    BSLS_ASSERT_SAFE(d_scheduler_p);
+    BSLS_ASSERT_SAFE(d_bufferFactory_p);
+    BSLS_ASSERT_SAFE(d_blobSpPool_p);
+    BSLS_ASSERT_SAFE(d_pushElementsPool);
+}
```
pniedzielski, line 517: I like this solution.

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -0,0 +1,697 @@
+// Copyright 2024 Bloomberg Finance L.P.
```
pniedzielski, line 1: 💯

pniedzielski (dismissed):

kaikulimu (approved):
lgtm

src/groups/mqb/mqbblp/mqbblp_remotequeue.t.cpp
```diff
@@ -195,14 +190,16 @@ TestBench::TestBench(bslma::Allocator* allocator_p)
         bdlf::BindUtil::bind(&TestClock::monotonicClock, &d_testClock),
         bdlf::BindUtil::bind(&TestClock::highResTimer, &d_testClock));
 
-    d_scheduler.start();
+    mwcu::MemOutStream errorDescription(allocator_p);
+
+    int rc = d_cluster.start(errorDescription);
```
678098, line 195: `rc` not used

src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -152,6 +153,8 @@ Application::Application(bdlmt::EventScheduler* scheduler,
                                     bdlf::PlaceHolders::_2),  // allocator
                k_BLOB_POOL_GROWTH_STRATEGY,
                d_allocators.get("BlobSpPool"))
+, d_pushElementsPool(sizeof(mqbblp::PushStream::Element),
+                     d_allocators.get("PushElementsPool"))
```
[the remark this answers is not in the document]
dorjesinpo: We do not expect large volume here as it is highly volatile.  Once message is pushed, it gets removed

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -0,0 +1,696 @@
...
+namespace mqbblp {
+
+// FORWARD DECLARATION
+class RelayQueueEngine_AppState;
+
+struct PushStream {
+    // forward declaration
+    struct Element;
+
+    enum eElementList { e_GUID = 0, e_APP = 1, e_TOTAL = 2 };
```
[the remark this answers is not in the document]
dorjesinpo: removing

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -0,0 +1,696 @@
...
+    /// or return an iterator pointing to the `guid`.
+    iterator findOrAppendMessage(const bmqt::MessageGUID& guid);
+
+    /// Add the specified `element` to both GUID and App corresponding to the
+    /// `element` (and specified when constructing the `element`).
+    void add(Element* element);
+
+    /// Remove the specified `element` from both GUID and App corresponding to
+    /// the `element` (and specified when constructing the `element`).
+    /// Erase the corresponding App from the PushStream if there is no
```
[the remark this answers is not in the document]
dorjesinpo: We do not, the comment is outdated.  Apps get removed explicitly by calling `removeApp`

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -0,0 +1,696 @@
...
+    mutable mqbi::StorageMessageAttributes d_attributes;
+
+    mutable bsl::shared_ptr<bdlbb::Blob> d_appData_sp;
+    // If this variable is empty, it is
+    // assumed that attributes, message,
+    // and options have not been loaded in
+    // this iteration (see also
+    // `loadMessageAndAttributes` impl).
+
+    mutable bsl::shared_ptr<bdlbb::Blob> d_options_sp;
```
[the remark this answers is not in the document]
dorjesinpo: This is the same for all `mqbi::StorageIterator` implementations.  The `mutable` fields is cached data.  I guess the idea was to enforce constness of the underlying storage

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -0,0 +1,696 @@
...
+        BSLS_ASSERT_SAFE(d_first_p == 0);
+        BSLS_ASSERT_SAFE(d_last_p == 0);
+    }
+    else {
+        BSLS_ASSERT_SAFE(d_first_p);
+        BSLS_ASSERT_SAFE(d_last_p);
+    }
+}
+
+inline void PushStream::Elements::remove(Element* element, eElementList where)
```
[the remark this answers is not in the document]
dorjesinpo: We can make a note and return that later,  I would prefer not to address this immediately (the benefit is small)

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -0,0 +1,696 @@
...
+    else {
+        BSLS_ASSERT_SAFE(element->d_base[where].d_next_p);
+
+        element->d_base[where].d_next_p->d_base[where].d_previous_p =
+            element->d_base[where].d_previous_p;
+    }
+
+    onRemove();
+
+    element->d_base[where].d_previous_p = element->d_base[where].d_next_p = 0;
```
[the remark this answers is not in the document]
dorjesinpo: that is `PushStream::destroy`

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -0,0 +1,696 @@
...
+    BSLS_ASSERT_SAFE(it != d_stream.end());
+
+    Element* element = new (d_pushElementsPool_sp->allocate())
+        Element(subscription, it, iteratorApp);
+    return element;
+}
+
+inline void PushStream::destroy(Element* element, bool doKeepGuid)
+{
+    if (element->app().d_elements.numElements() == 0) {
```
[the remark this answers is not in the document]
dorjesinpo: hmm... not sure.  These means "there is not another PUSH for this App at the moment" and "this GUID has only only one App to push"

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -0,0 +1,696 @@
...
+// -----------------
+
+inline PushStream::Element*
+PushStream::create(const bmqp::SubQueueInfo& subscription,
+                   const iterator&           it,
+                   const Apps::iterator&     iteratorApp)
+{
+    BSLS_ASSERT_SAFE(it != d_stream.end());
+
+    Element* element = new (d_pushElementsPool_sp->allocate())
```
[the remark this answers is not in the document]
dorjesinpo: good idea.  Will do that

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -0,0 +1,696 @@
...
+    return numMessages;
+}
+
+inline unsigned int PushStream::removeApp(Apps::iterator itApp)
+{
+    unsigned int numElements = itApp->second.d_elements.numElements();
+    for (unsigned int count = 0; count < numElements; ++count) {
+        Element* element = itApp->second.d_elements.first();
+
+        remove(element);
```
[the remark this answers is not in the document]
dorjesinpo: hmm... This is random.  Elements for the App are located randomly in the  PushStream.
Also, this is on queue purge or handle release.

src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -1085,68 +1085,63 @@ bool QueueEngineUtil_AppState::visitBroadcast(
...
+            BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
+
+            // The message got gc'ed or purged
+            BMQ_LOGTHROTTLE_INFO()
+                << "#STORAGE_UNKNOWN_MESSAGE " << "Queue: '"
+                << d_queue_p->description() << "', app: '" << appId()
+                << "' could not redeliver GUID: '" << *it
+                << "' (not in the storage)";
+        }
+        else if (!reader->appMessageView(ordinal()).isPending()) {
```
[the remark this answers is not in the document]
dorjesinpo: hmm.  I think, this would be a confirmed or gc'ed message. `BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY` would not hurt

src/groups/mqb/mqbblp/mqbblp_pushstream.cpp
```diff
@@ -0,0 +1,320 @@
...
+    // still keep the same ordinal numbering
+    d_currentElement = d_currentElement->next();
+    ++d_currentOrdinal;
+
+    d_owner_p->remove(del);
+    d_owner_p->destroy(del, true);
+    // doKeepGuid because of the d_iterator
+
+    if (d_iterator->second.numElements() == 0) {
+        // d_currentElement->eraseFromStream(d_owner_p->d_stream);
```
678098, line 132: Is this code needed?

src/groups/mqb/mqbblp/mqbblp_pushstream.t.cpp
```diff
@@ -0,0 +1,214 @@
...
+// mqbblp_pushstream.t.cpp                                            -*-C++-*-
+#include <mqbblp_pushstream.h>
+
+// BMQ
+#include <bmqp_messageguidgenerator.h>
+
+#include <mqbcfg_messages.h>
+#include <mqbs_inmemorystorage.h>
+
+#include <mwcu_memoutstream.h>
```
678098, line 25:
```suggestion
// MQB
#include <mqbcfg_messages.h>
#include <mqbs_inmemorystorage.h>

// MWC
#include <mwcu_memoutstream.h>
```

src/groups/mqb/mqbblp/mqbblp_pushstream.t.cpp
```diff
@@ -0,0 +1,214 @@
...
+int main(int argc, char* argv[])
+{
+    TEST_PROLOG(mwctst::TestHelper::e_DEFAULT);
+
+    bmqt::UriParser::initialize(s_allocator_p);
+
+    switch (_testCase) {
+    case 0:
+    case 1: test1_basic(); break;
+    case 2: test2_iterations(); break;
```
678098, line 204: Typically we have a descending order, not sure if UTs runner relies on this
```suggestion
    case 2: test2_iterations(); break;
    case 1: test1_basic(); break;
```

dorjesinpo: hmm.  I thought otherwise.  For example, `mqbblp_remotequeue.t`

src/groups/mqb/mqbblp/mqbblp_pushstream.cpp
```diff
@@ -0,0 +1,320 @@
...
+// ----------------------------
+
+// PRIVATE MANIPULATORS
+void PushStreamIterator::clear()
+{
+    // Clear previous state, if any.  This is required so that new state can be
+    // loaded in 'appData', 'options' or 'attributes' routines.
+    d_appData_sp.reset();
+    d_options_sp.reset();
+    d_attributes.reset();
```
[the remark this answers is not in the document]
dorjesinpo: We explicitly call `loadMessageAndAttributes` when accessing `d_attributes` and `d_options_sp`, so we could leave them.  On the other hand, an instance of `PushStreamIterator` would be in an inconsistent state.  Is the benefit worth it?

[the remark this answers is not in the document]
678098: This inconsistency only means that some fields are not reseted between calls, but we do have the `d_appData_sp` pointer which indicates that these fields are not reseted. So we can guarantee safe usage via class API. The non-initialized fields will only be visible if someone looks at the binary representation of this class with debugger.

678098 (approved):

678098 (approved):

678098 (dismissed):

src/groups/mqb/mqbblp/mqbblp_routers.t.cpp
```diff
@@ -224,6 +219,41 @@ static void test2_priority()
...
+
+    virtual ~MyAllocator() {}
+
+    // MANIPULATORS
+    virtual void* allocate(size_type size) { return new char[size]; }
+
+    virtual void deallocate(void* address) { delete[] address; }
+};
+
+MyAllocator ma;
```
678098, line 255: This class and variable declaration should be removed, right? @dorjesinpo

678098 (dismissed):

678098 (approved):

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -2175,6 +2175,10 @@ bsl::shared_ptr<mqbi::Queue> ClusterQueueHelper::createQueueFactory(
                 context.d_domain_p);
         }
 
+        BALL_LOG_ERROR << d_cluster_p->description()
```
kaikulimu, line 2178: Why is this log at ERROR level? Should be at INFO level.

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.h
```diff
@@ -510,19 +510,23 @@ class RelayQueueEngine : public mqbi::QueueEngine {
     virtual void
     onTimer(bsls::Types::Int64 currentTimer) BSLS_KEYWORD_OVERRIDE;
 
-    /// Called after the specified `appIdKeyPair` has been dynamically
-    /// registered.
+    /// Called after creation of a new storage for the  specified
```
kaikulimu, line 513: `the  specified` -> `the specified`

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1631,36 +1631,58 @@ void RelayQueueEngine::loadInternals(mqbcmd::QueueEngine* out) const
...
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(d_queueState_p->queue()->dispatcher()->inDispatcherThread(
         d_queueState_p->queue()));
 
-    const bsl::string& appId = appIdKeyPair.first;
-
     AppIds::iterator iter = d_appIds.find(appId);
-    mqbu::StorageKey key;
 
     if (iter == d_appIds.end()) {
```
kaikulimu, line 1646: You can remove the `if` block as it is no-op.

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -798,6 +797,16 @@ mqbi::QueueHandle* RootQueueEngine::getHandle(
 
         d_queueState_p->adopt(iter->value());
         upstreamSubQueueId = iter->value()->upstreamSubQueueId();
+
+        if (!iter->value()->isAuthorized()) {
+            if (iter->value()->authorize()) {
```
kaikulimu, line 802: If `authorize()` call returns false here, is this a reason for concern?

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -495,6 +495,12 @@ int StorageUtil::addVirtualStoragesInternal(
 
                 return rc_VIRTUAL_STORAGE_CREATION_FAILURE;  // RETURN
             }
+
+            BALL_LOG_INFO << clusterDescription << " PartitionId ["
```
kaikulimu, line 499: `PartitionId` -> `Partition` as we agree to uniform the naming.

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -546,6 +552,10 @@ int StorageUtil::removeVirtualStorageInternal(mqbs::ReplicatedStorage* storage,
 
         return rc_APP_KEY_NOT_PRESENT;  // RETURN
     }
+    BALL_LOG_INFO << " PartitionId [" << partitionId
```
kaikulimu, line 555: `PartitionId` -> `Partition`

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -2518,6 +2528,10 @@ void StorageUtil::registerQueue(
         return;  // RETURN
     }
 
+    BALL_LOG_INFO << clusterDescription << " PartitionId [" << partitionId
```
kaikulimu, line 2531: `PartitionId` -> `Partition`

kaikulimu: This results in double logging of queue registration together with Line 2427. Ideally, we print such log once in each code path.

src/groups/mqb/mqbs/mqbs_filebackedstorage.h
```diff
@@ -256,6 +252,9 @@ class FileBackedStorage BSLS_KEYWORD_FINAL : public ReplicatedStorage {
     /// otherwise return false.
     virtual bool isPersistent() const BSLS_KEYWORD_OVERRIDE;
 
+    /// Return the queue this storage is associated with.
+    virtual mqbi::Queue* queue() const BSLS_KEYWORD_OVERRIDE;
```
kaikulimu, line 256: Can the queue still be null? If so, please document.

src/groups/mqb/mqbs/mqbs_inmemorystorage.h
```diff
@@ -425,6 +421,9 @@ class InMemoryStorage BSLS_KEYWORD_FINAL : public ReplicatedStorage {
     // ACCESSORS
     //   (virtual mqbi::Storage)
 
+    /// Return the queue this storage is associated with.
+    virtual mqbi::Queue* queue() const BSLS_KEYWORD_OVERRIDE;
```
kaikulimu, line 425: Can the queue be null? If so, please document.

src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.h
```diff
@@ -289,6 +295,8 @@ class VirtualStorageCatalog {
 
     /// Return the default App state.
     const mqbi::AppMessage& defaultAppMessage() const;
+
+    mqbi::Queue* queue() const;
```
kaikulimu, line 299: Document, especially if the queue can be null.

kaikulimu (changes requested):
Some comments

src/groups/mqb/mqbs/mqbs_inmemorystorage.h
dorjesinpo: > Can the queue be null? If so, please document.

Always been a case.  Storage exists without a queue before `setQueue` call.

src/groups/mqb/mqbs/mqbs_filebackedstorage.h
dorjesinpo: Storage exists without a queue before setQueue call.

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
dorjesinpo: Consider the race: Domain is reconfigured but VirtualStorage is not added yet.
`setQueue` is not called yet.
`isAuthorized()` is `false` and we need to `authorize()`

Another case is when App is not registered.  

To be refactored when we add Dynamic Apps

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
dorjesinpo: Good catch, thank you.  Removing this leftover

678098 (dismissed):
Approve for sanitizer

678098 (approved):

kaikulimu (approved):
LGTM!

PR #335, opened 2024-06-25
MQB: structured stats support, out of order stats snapshot

678098:
Changes:
- `mqbstat`: introduce a new component `JsonPrinter` responsible for printing stats from the registered stat context snapshots as a pretty or compact json. This component is used only during admin command processing from the `mqbstat::StatController` scheduler thread.
- `mqbstat::StatController`: split `snapshot()` into 2 separate functions: `snapshot()` (which just makes snapshot but doesn't notify the registered stat consumers) and `snapshotAndNotify()` (which does both). Now we schedule `snapshotAndNotify()` as a periodic event, not the `snapshot()`.
- `mqbstat::StatController`: remove unnecessary `captureStats` and extend `captureStatsAndSemaphorePost` instead. This function supports the old unchanged code path with text stats request (which doesn't make an unscheduled snapshot) and also a new code path to request the json with the recent stats (it makes an out-of-order snapshot).
- `mqbstat::StatController`: rework the too-often snapshot safeguard to use `bslmt::Throttle`.
- `mqbstat::StatController::snapshot`: add a bool return code to notify if the snapshot attempt was successful or the snapshot was throttled due to too often invocations.
- `mqba::Application`: pass the encoding options to the `mqbstat::StatController` on admin command.
- `mqbstat::QueueStats::getValue`: add the ability to access the oldest snapshot by passing a negative `snapshotId`. This makes the API easier and closer to python data containers where you can specify something like `data[-1]` to access the last element even without knowing the exact size of a container.
- `mqbstat::QueueStatsDomain`: add `toString` converter with values synced with the prometheus plugin metric names.
- ITs: add missing broker property `appIdTagDomains` which allows per-appId metrics and regenerate tweaks.
- ITs: add a new integration test `test_stats` which tests per-appId metrics from the broker and also checks the safeguard mechanism.
- Docs: fix comments and documentation, cleanup.

### DIFF
--- a/src/groups/mqb/mqbstat/mqbstat_jsonprinter.cpp
+++ b/src/groups/mqb/mqbstat/mqbstat_jsonprinter.cpp
@@ -40,7 +40,7 @@ namespace {
 // ---------------------
 
 /// The implementation class for JsonPrinter, containing all the cached options
-/// for printing statistics as json.  This implementation exists and is hidden
+/// for printing statistics as JSON.  This implementation exists and is hidden
 /// from the package include for the following reasons:
 /// - Don't want to expose `bdljsn` names and symbols to the outer scope.
 /// - Member fields and functions defined for this implementation are used only
@@ -52,20 +52,14 @@ class JsonPrinterImpl {
 
   private:
     // PRIVATE TYPES
-    typedef bsl::unordered_map<bsl::string, mwcst::StatContext*>
-        StatContextsMap;
-
-    typedef mqbstat::QueueStatsDomain::Stat Stat;
+    typedef JsonPrinter::StatContextsMap StatContextsMap;
 
   private:
     // DATA
-    /// Config to use
-    const mqbcfg::StatsConfig& d_config;
-
-    /// Options for printing a compact json
+    /// Options for printing a compact JSON
     const bdljsn::WriteOptions d_opsCompact;
 
-    /// Options for printing a pretty json
+    /// Options for printing a pretty JSON
     const bdljsn::WriteOptions d_opsPretty;
 
     /// StatContext-s map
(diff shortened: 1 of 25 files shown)


src/groups/mqb/mqbstat/mqbstat_jsonprinter.cpp
```diff
@@ -0,0 +1,268 @@
...
+                                                 Stat::e_PUT_BYTES_DELTA,
+                                                 Stat::e_PUSH_MESSAGES_DELTA,
+                                                 Stat::e_PUSH_BYTES_DELTA,
+                                                 Stat::e_ACK_DELTA,
+                                                 Stat::e_ACK_TIME_AVG,
+                                                 Stat::e_ACK_TIME_MAX,
+                                                 Stat::e_NACK_DELTA,
+                                                 Stat::e_CONFIRM_DELTA,
+                                                 Stat::e_CONFIRM_TIME_AVG,
+                                                 Stat::e_CONFIRM_TIME_MAX};
```
678098, line 146: Todo: safe static variable initialization in a multithreaded env. Cache as a const class field instead.

src/groups/mwc/mwcst/mwcst_statvalue.h
```diff
@@ -339,7 +339,7 @@ class StatValue {
     /// Return the snapshot referred to by the specified `location`.  The
     /// behavior is undefined unless
     /// `location.level() < numLevels()` and
-    /// `location.index() <= historySize(location.level())`
+    /// `location.index() < historySize(location.level())`
```
678098, line 342: The actual condition is more strict:
https://github.com/bloomberg/blazingmq/blob/1387dfab4bc21f1df279ff84c617b0089e90fd68/src/groups/mwc/mwcst/mwcst_statvalue.h#L700

src/integration-tests/test_admin_client.py
```diff
@@ -70,6 +71,53 @@ def post_n_msgs(
...
+
+
+class GreaterThan:
+    def __init__(self, value):
+        self.value = value
+
+
+def expect_same_structure(obj, expected) -> None:
+    """
+    Check if the specified 'obj' dictionary has the same structure as the specified 'expected' dictionary.
```
678098, line 95: Todo: rename `dictionary` -> `entity` and use `Union` type hint

src/integration-tests/test_admin_client.py
```diff
@@ -108,6 +156,234 @@ def test_breathing(single_node: Cluster) -> None:
...
+            "queue_confirm_msgs": 0,
+            "queue_push_bytes": 0,
+            "queue_consumers_count": 0,
+            "queue_producers_count": 0,
+            "queue_push_msgs": 0,
+            "queue_ack_time_avg": GreaterThan(0),
+            "queue_put_bytes": 96,
+            "queue_put_msgs": 32,
+        },
+    }
```
678098, line 276: These sample stats are massive, but it's the most convenient way to check expected metrics. It is much easier than adding `assert` for each path within a dictionary, and much easier to modify if we change the metrics generation.

kaikulimu:
Sorry didn't get to the PR yet. Will review tmr.

src/integration-tests/test_admin_client.py
```diff
@@ -70,6 +71,53 @@ def post_n_msgs(
...
+    - The outer layer is a 'str' response from the admin session
+    - Next layer is a 'dict' containing "stats" field with 'str' text
+    - The last layer is another 'dict' containing the stats itself
+    """
+    d1 = json.loads(admin_response)
+    d2 = json.loads(d1["stats"])
+    return d2
+
+
+class GreaterThan:
```
kaikulimu, line 88: Add a small comment explaining the purpose of the `GreaterThan` class.

The class itself is not self-explanatory.

src/integration-tests/test_admin_client.py
kaikulimu: Agreed. You can move the test data into a separate file.

src/integration-tests/test_admin_client.py
```diff
@@ -108,6 +156,234 @@ def test_breathing(single_node: Cluster) -> None:
...
+    expect_same_structure(queue_stats, expected_stats_after_confirm)
+
+    consumer_foo.close(f"{task.uri}?id=foo")
+    consumer_bar.close(f"{task.uri}?id=bar")
+    consumer_baz.close(f"{task.uri}?id=baz")
+
+    # Stage 3: check too-often stats safeguard
+    for i in range(5):
+        admin.send_admin("encoding json_pretty stat show")
+    res = admin.send_admin("encoding json_pretty stat show")
```
kaikulimu, line 374: Do you test "encoding json_compact" anywhere?

src/python/blazingmq/dev/it/tweaks/generated.py
```diff
@@ -407,6 +407,12 @@ def __call__(
             dispatcher_config = DispatcherConfig()
 
             class Stats(metaclass=TweakMetaclass):
+                class AppIdTagDomains(metaclass=TweakMetaclass):
+                    def __call__(self, value: None) -> Callable:
+                        ...
```
kaikulimu, line 412: Does `...` just mean `pass` here?

src/groups/mqb/mqbstat/mqbstat_statcontroller.h
```diff
@@ -301,8 +312,9 @@ class StatController {
     /// Process the specified `command`, and write the result to the
     /// `result`' object.  Return zero on success or a nonzero value
     /// otherwise.
```
kaikulimu, line 314: Update comment to mention `encoding`

src/groups/mqb/mqbstat/mqbstat_statcontroller.h
```diff
@@ -154,82 +156,85 @@ class StatController {
...
 
     /// Initialize all the stat contexts and associated Tables and TIPs.
     void initializeStats();
 
-    /// Capture the stats and store the stats in the specified `result`
-    /// object.
-    void captureStats(mqbcmd::StatResult* result);
-
     /// Capture the stats to the specified `result`' object and post on the
     /// specified `semaphore` once done.
```
kaikulimu, line 233: Update comment to mention `encoding`

src/groups/mqb/mqbstat/mqbstat_statcontroller.cpp
```diff
@@ -233,26 +241,51 @@ void StatController::initializeStats()
...
+        mwcu::MemOutStream os;
+        d_printer_mp->printStats(os);
+        result->makeStats() = os.str();
+    } break;  // BREAK
+
+    case mqbcmd::EncodingFormat::JSON_COMPACT: BSLS_ANNOTATION_FALLTHROUGH;
+    case mqbcmd::EncodingFormat::JSON_PRETTY: {
+        // Make an unscheduled snapshot, but do not notify stats consumers
+        // since it's not necessary
+        const bool savedNextSnapshot = snapshot();
```
kaikulimu, line 267: Why is it necessary to make a snapshot before printing as JSON?

kaikulimu (changes requested):
Some comments

src/groups/mqb/mqbstat/mqbstat_statcontroller.cpp
678098: We specifically want to use this code path in integration tests, to check metrics generation. So we want to get the latest stats. To do this, we either `sleep` in IT with hope that it's enough time to make a new snapshot, or add a sync point like I did.
Note that this code path is for stats requested by JSON only, so the previous cmd `STAT SHOW` will not be affected.

src/groups/mqb/mqbstat/mqbstat_jsonprinter.cpp
```diff
@@ -0,0 +1,268 @@
...
+}  // close unnamed namespace
+
+// -----------------
+// class JsonPrinter
+// -----------------
+
+JsonPrinter::JsonPrinter(const mqbcfg::StatsConfig& config,
+                         const StatContextsMap&     statContextsMap,
+                         bslma::Allocator*          allocator)
+: d_allocator_p(bslma::Default::allocator(allocator))
```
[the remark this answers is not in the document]
678098: I need to guard against the passed `0` allocator in the next line, where I allocate `JsonPrinterImpl`:

Here, if I simply try to allocate with 0 allocator, there will be a segfault:
```
d_impl_mp(new (*d_allocator_p)
                JsonPrinterImpl(config, statContextsMap, d_allocator_p),
            d_allocator_p)
```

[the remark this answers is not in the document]
678098: Removed the `const` qualifier instead

src/groups/mqb/mqbstat/mqbstat_queuestats.h
```diff
@@ -198,7 +202,8 @@ class QueueStatsDomain {
     /// represented by its associated specified `context` as the difference
     /// between the latest snapshot-ed value (i.e., `snapshotId == 0`) and
     /// the value that was recorded at the specified `snapshotId` snapshots
-    /// ago.
+    /// ago.  Any negative `snapshotId` means that the oldest available
```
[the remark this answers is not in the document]
678098: Updated the docs to support only `-1`, we are not interested in other negative cases actually

src/python/blazingmq/dev/it/tweaks/generated.py
678098: Yes, `ellipsis` literal does the same nothing as `pass` in this context

src/groups/mqb/mqbstat/mqbstat_statcontroller.cpp
678098: Extended the comment

678098:
Thanks for the review @kaikulimu
The fixes are in the 2nd commit of this branch

src/python/blazingmq/dev/it/data/data_metrics.py
```diff
@@ -0,0 +1,200 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Additional data for metrics-related integration tests.
+
+Note: currently these tests are placed in 'test_admin.py' tests block, but might be moved
```
kaikulimu, line 19: `test_admin.py` -> `test_admin_client.py`

kaikulimu (commented):
Looks good! Just one minor comment.

src/python/blazingmq/dev/it/data/data_metrics.py
678098: Fixed

kaikulimu (approved):
LGTM

PR #336, opened 2024-06-25
MQB: wait for unconfirmed before buffering confirms

dorjesinpo:
FIx: do not start buffering CONFIRMs in (remote) queue upon deconfigure response (upon StopRequest processing) because that is too early when waiting for unconfirmed.

Early buffering increases chances of duplicate PUSH messages - those received by RemoteQueue between deconfigure response and finishing waiting for unconfirmed.

Still need to start buffering broadcast PUTs early, otherwise broadcast PUT can get lost.

Still, duplicates are possible in few ITs because of the scenario when a CONFIRM had passed Proxy but did not reach the replication.  New Primary then redelivers and the Proxy cannot detect the duplicate because it had removed the GUID upon the first CONFIRM


pniedzielski (approved):
👍🏻

PR #338, opened 2024-06-27
mwcc_orderedhashmap.t: limit cycles number for memory sanitizer

alexander-e1off:


678098 (changes requested):
1. Let's make the code block defining `k_NUM_ELEMENTS` identical for 2 UTs. There is no need to keep these 2 different inverted conditions.
2. It is not enough to just check if `__has_feature` is defined. This macro can exist but all sanitizers might be disabled.
3. I think we should still check for `__SANITIZE_MEMORY__` aka GCC-supported macro. https://github.com/torvalds/linux/blob/afcd48134c58d6af45fb3fdb648f1260b20f2326/include/linux/compiler-clang.h#L59
4. Overall, I think it's not necessary to try to fit everything in one `if` condition. You can make many branches for each specific case with very basic conditions for each branch. It will be more code lines, but it's more maintainable than one complex condition.
5. Not necessary to put `__has_feature(memory_sanitizer)` check to the preprocessor `#if`, you can use a ternary assignment for `k_NUM_ELEMENTS` instead

alexander-e1off:
Addressed review comments

678098 (approved):
Looks good!

PR #341, opened 2024-06-28
mwcio_ntcchannelfactory.t: skip test under sanitizers

alexander-e1off:
`test4_cancelHandleTest()` is already skipped for UBSan, but it also sometimes fails when running under MSan and TSan due to changed timings. Enhanced existing condition to skip this test for MSan and TSan too.


src/groups/mwc/mwcio/mwcio_ntcchannelfactory.t.cpp
```diff
@@ -1121,17 +1115,33 @@ static void test4_cancelHandleTest()
...
-    // worked on if the test starts failing in non-instrumented builds.
-    // Additionally, we could try to enable this test in UBSan build once all
-    // UBSan reports have been fixed to see if that helps (see `ubsansup.txt`).
-    return;
+#if defined(__has_feature)  // Clang-supported method for checking sanitizers.
+    const bool skipTestForSanitizers = __has_feature(memory_sanitizer) ||
+                                       __has_feature(thread_sanitizer) ||
+                                       __has_feature(
+                                           undefined_behavior_sanitizer);
+#elif defined(__SANITIZE_MEMORY__) || defined(__SANITIZE_THREAD__)
```
678098, line 1123: Also might handle the case with `__SANITIZE_UNDEFINED__`

678098 (approved):

PR #343, opened 2024-06-28
CI: build and run C++ UTs in a separate step

678098:
CI:
- Fix incorrect path to file causing hash evaluation error.
- Pin `ntf-core` version in `build-deps.sh` to prevent hash collisions causing `ntf-core` rebuild even on cache hit.
- Extract UTs build from the main broker build step, to speed up the CI. Some later steps are depend only on the broker/tool, so it's not necessary to wait for all UTs to link and run to launch these tests

Future work: extract packages install and project configuration to reusable workflow and reference it instead of copy-pasting the same.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Get dependencies hash
         id: get-hash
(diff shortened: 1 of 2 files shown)


pniedzielski (approved):
Changes look good, and future work also looks like a good direction.

PR #346, opened 2024-06-28
Admin API Routing

lukedigiovanna:
Enhances the admin command API to automatically route certain commands to the proper node(s) for execution. For example, `DOMAINS DOMAIN <name> PURGE` should only be executed by the primary node, so if a user sends this command to a replica node, then the replica node will route it to the primary and report the result back to the user. In another case the user may want to reconfigure the cluster with `DOMAINS RECONFIGURE <domain>`. With this PR when the user sends this command to any node that node will then propagate the command to every other node in the cluster and collect the responses from each.

**Testing performed**
This includes an integration test, `test_admin_command_routing.py`, that ensures a command sent to an improper node is routed and a response is sent back. This testing does not in any way verify the correctness of the command, merely that it has been routed.

**Additional context**
Part of summer 2024 internship project.

### DIFF
--- a/src/groups/mqb/mqba/mqba_application.cpp
+++ b/src/groups/mqb/mqba/mqba_application.cpp
@@ -551,7 +551,9 @@ int Application::processCommand(const bslstl::StringRef& source,
     }
     else if (command.isStatValue()) {
         mqbcmd::StatResult statResult;
-        d_statController_mp->processCommand(&statResult, command.stat());
+        d_statController_mp->processCommand(&statResult,
+                                            command.stat(),
+                                            commandWithOptions.encoding());
         if (statResult.isErrorValue()) {
             cmdResult.makeError(statResult.error());
         }
--- a/src/groups/mqb/mqbstat/mqbstat_jsonprinter.cpp
+++ b/src/groups/mqb/mqbstat/mqbstat_jsonprinter.cpp
@@ -1,0 +1,270 @@
+// Copyright 2024 Bloomberg Finance L.P.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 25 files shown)


lukedigiovanna:
What type of routing, if any, do we want for the following commands:

`CACHE_CLEAR` commands:

```
DOMAINS RESOLVER CACHE_CLEAR <domain>
CONFIGPROVIDER CACHE_CLEAR <domain>
```

Would it be useful to route this command to all nodes in the cluster?

`SET/GET` commands:

```
CLUSTERS CLUSTER <name> STORAGE REPLICATION SET <parameter> <value>
CLUSTERS CLUSTER <name> STORAGE REPLICATION GET <parameter>
CLUSTERS CLUSTER <name> STATE ELECTOR SET <parameter> <value>
CLUSTERS CLUSTER <name> STATE ELECTOR GET <parameter>
```

Is there a use case for wanting to set/get a parameter on all nodes?

A potential proposal is modifying these commands to optionally set/get a parameter for just one node or for all nodes.

lukedigiovanna:
How do we want to display the result of routing to multiple nodes, particularly in various encoding formats?

When a node routes to the entire cluster, we get several responses back. The current way I have implemented displaying these results as JSON is very similar to the existing way of printing a singular command result (using `baljsn::Encoder`). 

For example, a response of `DOMAINS RECONFIGURE <domain>` looks like:

```
{
    "responses" : [
        {
            "source" : "east\/1",
            "response" : "{\n    \"success\" : {\n\n    }\n}\n"
        },
        {
            "source" : "east\/2",
            "response" : "{\n    \"success\" : {\n\n    }\n}\n"
        },
        {
            "source" : "west\/1",
            "response" : "{\n    \"success\" : {\n\n    }\n}\n"
        },
        {
            "source" : "west\/2",
            "response" : "{\n    \"success\" : {\n\n    }\n}\n"
        }
    ]
}
```

There is a current limitation that the response of a reroute is transmitted as a string, so the individual response is given as a string representation of the encoded JSON, as opposed to the JSON itself. This could be overcome by manually writing the code to display this list of responses as JSON, though this increases complexity and is not consistent with the existing method of encoding (namely using `baljsn::Encoder`)

Is this method sufficient for our purposes, or would we like another way of representing the result of a command routed to several nodes?

dorjesinpo:
> Is there a use case for wanting to set/get a parameter on all nodes?

`STORAGE REPLICATION SET` and `STATE ELECTOR SET`  we probably need on individual nodes, there may be IT(s) relying on that.
Not sure if we need all nodes at once but if we do, we would need an (optional) parameter to still support individual node case

PR #352, opened 2024-07-09
Admin API Routing

lukedigiovanna:
Enhances the admin command API to automatically route certain commands to the proper node(s) for execution. For example, `DOMAINS DOMAIN <name> PURGE` should only be executed by the primary node, so if a user sends this command to a replica node, then the replica node will route it to the primary and report the result back to the user. In another case the user may want to reconfigure the cluster with `DOMAINS RECONFIGURE <domain>`. With this PR when the user sends this command to any node that node will then propagate the command to every other node in the cluster and collect the responses from each.

Testing performed
This includes an integration test, `test_admin_command_routing.py`, that ensures a command sent to an improper node is routed and a response is sent back. This testing does not in any way verify the correctness of the command, merely that it has been routed.

Additional context
Part of summer 2024 internship project.


lukedigiovanna:
What type of routing, if any, do we want for the following commands:

`CACHE_CLEAR` commands:

```
DOMAINS RESOLVER CACHE_CLEAR <domain>
CONFIGPROVIDER CACHE_CLEAR <domain>
```

Would it be useful to route this command to all nodes in the cluster?

`SET/GET` commands:

```
CLUSTERS CLUSTER <name> STORAGE REPLICATION SET <parameter> <value>
CLUSTERS CLUSTER <name> STORAGE REPLICATION GET <parameter>
CLUSTERS CLUSTER <name> STATE ELECTOR SET <parameter> <value>
CLUSTERS CLUSTER <name> STATE ELECTOR GET <parameter>
```

Is there a use case for wanting to set/get a parameter on all nodes?

A potential proposal is modifying these commands to optionally set/get a parameter for just one node or for all nodes.

lukedigiovanna:
How do we want to display the result of routing to multiple nodes, particularly in various encoding formats?

When a node routes to the entire cluster, we get several responses back. The current way I have implemented displaying these results as JSON is very similar to the existing way of printing a singular command result (using `baljsn::Encoder`).

For example, a response of `DOMAINS RECONFIGURE <domain>` looks like:

```
{
    "responses" : [
        {
            "source" : "east\/1",
            "response" : "{\n    \"success\" : {\n\n    }\n}\n"
        },
        {
            "source" : "east\/2",
            "response" : "{\n    \"success\" : {\n\n    }\n}\n"
        },
        {
            "source" : "west\/1",
            "response" : "{\n    \"success\" : {\n\n    }\n}\n"
        },
        {
            "source" : "west\/2",
            "response" : "{\n    \"success\" : {\n\n    }\n}\n"
        }
    ]
}
```

There is a current limitation that the response of a reroute is transmitted as a string, so the individual response is given as a string representation of the encoded JSON, as opposed to the JSON itself. This could be overcome by manually writing the code to display this list of responses as JSON, though this increases complexity and is not consistent with the existing method of encoding (namely using `baljsn::Encoder`)

Is this method sufficient for our purposes, or would we like another way of representing the result of a command routed to several nodes?

kaikulimu:
> What type of routing, if any, do we want for the following commands:
> 
> `CACHE_CLEAR` commands:
> 
> ```
> DOMAINS RESOLVER CACHE_CLEAR <domain>
> CONFIGPROVIDER CACHE_CLEAR <domain>
> ```
> 
> Would it be useful to route this command to all nodes in the cluster?
> 
> `SET/GET` commands:
> 
> ```
> CLUSTERS CLUSTER <name> STORAGE REPLICATION SET <parameter> <value>
> CLUSTERS CLUSTER <name> STORAGE REPLICATION GET <parameter>
> CLUSTERS CLUSTER <name> STATE ELECTOR SET <parameter> <value>
> CLUSTERS CLUSTER <name> STATE ELECTOR GET <parameter>
> ```
> 
> Is there a use case for wanting to set/get a parameter on all nodes?
> 
> A potential proposal is modifying these commands to optionally set/get a parameter for just one node or for all nodes.

@jll63 Are you familiar with the `CACHE_CLEAR <domain>` set of commands?

kaikulimu:
> What type of routing, if any, do we want for the following commands:
> 
> `CACHE_CLEAR` commands:
> 
> ```
> DOMAINS RESOLVER CACHE_CLEAR <domain>
> CONFIGPROVIDER CACHE_CLEAR <domain>
> ```
> 
> Would it be useful to route this command to all nodes in the cluster?
> 
> `SET/GET` commands:
> 
> ```
> CLUSTERS CLUSTER <name> STORAGE REPLICATION SET <parameter> <value>
> CLUSTERS CLUSTER <name> STORAGE REPLICATION GET <parameter>
> CLUSTERS CLUSTER <name> STATE ELECTOR SET <parameter> <value>
> CLUSTERS CLUSTER <name> STATE ELECTOR GET <parameter>
> ```
> 
> Is there a use case for wanting to set/get a parameter on all nodes?
> 
> A potential proposal is modifying these commands to optionally set/get a parameter for just one node or for all nodes.

For the `SET/GET` commands, the current proposal is to add a `SET <parameter> <value>` flavor for only sending to the current node, and `SET_ALL <parameter> <value>` flavor to send to all cluster nodes.

kaikulimu:
> How do we want to display the result of routing to multiple nodes, particularly in various encoding formats?
> 
> When a node routes to the entire cluster, we get several responses back. The current way I have implemented displaying these results as JSON is very similar to the existing way of printing a singular command result (using `baljsn::Encoder`).
> 
> For example, a response of `DOMAINS RECONFIGURE <domain>` looks like:
> 
> ```
> {
>     "responses" : [
>         {
>             "source" : "east\/1",
>             "response" : "{\n    \"success\" : {\n\n    }\n}\n"
>         },
>         {
>             "source" : "east\/2",
>             "response" : "{\n    \"success\" : {\n\n    }\n}\n"
>         },
>         {
>             "source" : "west\/1",
>             "response" : "{\n    \"success\" : {\n\n    }\n}\n"
>         },
>         {
>             "source" : "west\/2",
>             "response" : "{\n    \"success\" : {\n\n    }\n}\n"
>         }
>     ]
> }
> ```
> 
> There is a current limitation that the response of a reroute is transmitted as a string, so the individual response is given as a string representation of the encoded JSON, as opposed to the JSON itself. This could be overcome by manually writing the code to display this list of responses as JSON, though this increases complexity and is not consistent with the existing method of encoding (namely using `baljsn::Encoder`)
> 
> Is this method sufficient for our purposes, or would we like another way of representing the result of a command routed to several nodes?

@678098 Does this response formatting look sound to you?

678098:
> How do we want to display the result of routing to multiple nodes, particularly in various encoding formats?
> 
> When a node routes to the entire cluster, we get several responses back. The current way I have implemented displaying these results as JSON is very similar to the existing way of printing a singular command result (using `baljsn::Encoder`).
> 
> For example, a response of `DOMAINS RECONFIGURE <domain>` looks like:
> 
> ```
> {
>     "responses" : [
>         {
>             "source" : "east\/1",
>             "response" : "{\n    \"success\" : {\n\n    }\n}\n"
>         },
>         {
>             "source" : "east\/2",
>             "response" : "{\n    \"success\" : {\n\n    }\n}\n"
>         },
>         {
>             "source" : "west\/1",
>             "response" : "{\n    \"success\" : {\n\n    }\n}\n"
>         },
>         {
>             "source" : "west\/2",
>             "response" : "{\n    \"success\" : {\n\n    }\n}\n"
>         }
>     ]
> }
> ```
> 
> There is a current limitation that the response of a reroute is transmitted as a string, so the individual response is given as a string representation of the encoded JSON, as opposed to the JSON itself. This could be overcome by manually writing the code to display this list of responses as JSON, though this increases complexity and is not consistent with the existing method of encoding (namely using `baljsn::Encoder`)
> 
> Is this method sufficient for our purposes, or would we like another way of representing the result of a command routed to several nodes?

@lukedigiovanna @kaikulimu When designing admin commands, we should keep in mind several goals I think:
- Keep the legacy text format unchanged.
- Provide an easily parseable (json) format.
- At the same time, keep the number of variations in the admin results output low.
 
The last point means, that we should not ask admin clients to be able to parse 3-5 different forms of the output for the same admin command. We might still need to parse 2 forms: error and success response.

As far as I understood, in the current PR we introduced another form of output, so there are at least 3 different forms:
- Error (local exec)
- Success (local exec)
- Routing results (routed)

This introduces complexity on a client side, since client needs to parse another different case.


Another point, about easily parseable json. For responses, I would like to keep the requested encoding whenever it's possible.

Note that we are able to encode the admin command result in JSON already: https://github.com/bloomberg/blazingmq/blob/346b9c305e74f56841eee93c3562dfae91089b91/src/groups/mqb/mqba/mqba_application.cpp#L628
It is done automatically since `baljsn` supports automatic conversion of schema objects. So we are also able to decode it back the same way. It will be a few more lines, but the changes are local to the place where we aggregate routing results.


Do you pass admin prefix "encoding <format>" when you route commands? Examples of such commands are here: https://github.com/bloomberg/blazingmq/blob/346b9c305e74f56841eee93c3562dfae91089b91/src/integration-tests/test_admin_client.py#L161

lukedigiovanna:
> * Keep the legacy text format unchanged.

I think with the proposal to add a SET_ALL/GET_ALL flavor for STORAGE REPLICATION and STATE_ELECTOR commands we maintain this since a client can still use SET/GET just as before -> its backwards compatible.

> As far as I understood, in the current PR we introduced another form of output, so there are at least 3 different forms:
> 
> * Error (local exec)
> * Success (local exec)
> * Routing results (routed)

Yes, this seems right. I'm not sure how to go about reporting the results in any other way. In the case of a command routed to just one other node (e.g. to the primary), the response is reported back to the client as is (no extraneous "responses" list or details about the node it was routed to). In the other case where a command is routed to >1 nodes, then a response like above is what gets outputted.

Where I think this approach gets inconvenient is in the possible [future] case of multiple primaries for a cluster. Then sometimes we would get a single response as is expected now (when 1 primary) or a list of responses (when multiple). I'm not sure what would be a good solution client-side, but I want to note that now.

> Do you pass admin prefix "encoding " when you route commands?

Yes, when I route a command I just pass the exact command string that was sent to the original node, so the encoding is preserved.

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
```diff
@@ -749,7 +749,7 @@ void ClusterProxy::processEvent(const bmqp::Event&   event,
                        << "Received unexpected event: " << event;
         BSLS_ASSERT_SAFE(false && "Unexpected event received");
         return;  // RETURN
-    }            // break;
+    }  // break;
```
dorjesinpo, line 752: There seem be only formatting changes in this file

src/groups/mqb/mqbblp/mqbblp_cluster.h
```diff
@@ -311,6 +311,8 @@ class Cluster : public mqbi::Cluster,
     // responses from proxies and nodes,
     // and the cluster's shutdown callback.
 
+    mqbnet::Session::AdminCommandEnqueueCb d_adminCb;
```
dorjesinpo, line 314: I wonder if this is necessary.  Cluster receives an admin command (is this always the rerouted one?).  Do we need to enqueue the command to the Application, which then looks up cluster and ends up calling the (same) cluster?

Do we want to support cluster receiving commands for another cluster?

This may worth a discussion.  Let's zoom one day?

src/groups/bmq/bmqp/bmqp_ctrlmsg_messages.cpp
```diff
@@ -561,49 +578,16 @@ const bdlat_AttributeInfo* CloseQueueResponse::lookupAttributeInfo(int id)
 
 // CREATORS
 
-CloseQueueResponse::CloseQueueResponse()
```
dorjesinpo, line 564: What is the reason for this big change in formatting.  `bas_codegen.pl` version/arguments?

src/groups/mqb/mqba/mqba_commandrouter.cpp
```diff
@@ -0,0 +1,367 @@
...
+        }
+        d_responses.responses().push_back(routeResponse);
+    }
+
+    countDownLatch();
+}
+
+RouteCommandManager::RoutingModeMp RouteCommandManager::getCommandRoutingMode()
+{
+    RouteCommandManager::RoutingMode* routingMode = nullptr;
```
dorjesinpo, line 254: `ManagedPtr`

dorjesinpo (changes requested):
I did a first look resulting in few comments.  Let's rebase before the next one?

src/groups/mqb/mqba/mqba_commandrouter.h
```diff
@@ -0,0 +1,177 @@
...
+class ClusterNode;
+}
+namespace mqbnet {
+template <class REQUEST, class RESPONSE, class TARGET>
+class MultiRequestManagerRequestContext;
+}
+
+namespace mqba {
+
+class RouteCommandManager {
```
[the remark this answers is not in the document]
lukedigiovanna: I originally called it CommandRouter, but then designed it to be a single use class (constructed per command and destroyed when the command has been processed). I did this since it has no persistent state. I tried renaming it to make that clearer (that it's disposable), though I don't think `RouteCommandManager` is much better.

Just calling it `CommandRouter` may be best.

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
lukedigiovanna: Yeah looks so, clang-format must have picked it up at some point. Should I just discard these changes?

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -3674,5 +3715,99 @@ void Cluster::processResponse(const bmqp_ctrlmsg::ControlMessage& response)
...
+            }
+            // Check for self
+            if (d_state.isSelfActivePrimary(pit->partitionId())) {
+                *outIsSelfPrimary = true;
+                continue;
+            }
+            outNodes->push_back(primary);
+        }
+        else {
+            // TODO: handle this case
```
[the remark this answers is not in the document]
lukedigiovanna: Right now the response is empty in this case, the idea is that this would be fixed when I implement the buffer and retry (or error if retry fails). But that's not implemented in this PR, so I can go ahead and implement just an error message in this case for now.

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -2929,6 +2933,24 @@ void Cluster::processControlMessage(
             << description() << ": unexpected clusterMessage:" << message
             << MWCTSK_ALARMLOG_END;
     } break;  // BREAK
+    case MsgChoice::SELECTION_ID_ADMIN_COMMAND: {
```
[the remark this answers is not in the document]
lukedigiovanna: I did not implement this with proxies in mind. Is that something we want to support?

src/groups/mqb/mqbblp/mqbblp_cluster.h
lukedigiovanna: The idea is that all commands get processed through the application because that is where we have the necessary dependencies to execute a command (`d_domainManager_mp`, `d_configProvider_mp`, etc.).

A command that is rerouted should always invoke that callback, `d_adminCb`, when the command is received, and that callback enqueues the command to the application instance where it will be executed. A command that has been rerouted never performs additional routing, it just gets executed by the application.

> Do we want to support cluster receiving commands for another cluster?

I'm not sure what exactly you mean by this.

I think a zoom is a good idea

src/groups/bmq/bmqp/bmqp_ctrlmsg_messages.cpp
lukedigiovanna: I am not sure. I cloned the latest version of [https://bbgithub.dev.bloomberg.com/bas/bas-all](url) and used `bas_codegen.pl` from that. When I run `bas_codegen.pl --version` I get `@BLP_BAS_CODEGEN_VERSION@`, which feels like a bug or maybe an error with my dev environment.

The arguments I copied from the prior generated file:
`bas_codegen.pl -m msg --noAggregateConversion --noExternalization --noIdent --package bmqp_ctrlmsg --msgComponent messages bmqp_ctrlmsg.xsd`

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
678098: The CI formatting check complains about these changes, but we might use an older version of clang-format which doesn't handle these cases well. I will look how if we can update it.

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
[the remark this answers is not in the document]
dorjesinpo: Possibly but not in this PR.  Maybe, a comment stating that `ClusterProxy` cannot be a 
"route member"

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
dorjesinpo: Yes, there is a [mismatch](There is a [mismatch](https://github.com/bloomberg/blazingmq/actions/runs/9998715055/job/27660062596?pr=352#step:3:14) between Jenkins formatter and the one in the )

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -2929,6 +2933,24 @@ void Cluster::processControlMessage(
...
+        const bmqp_ctrlmsg::AdminCommand& adminCommand =
+            message.choice().adminCommand();
+        const bsl::string& cmd      = adminCommand.command();
+        const bool         rerouted = adminCommand.rerouted();
```
dorjesinpo, line 2940: Is there a case when `rerouted` is `false`?

src/groups/mqb/mqbblp/mqbblp_cluster.h
dorjesinpo: > > Do we want to support cluster receiving commands for another cluster?
> 
> I'm not sure what exactly you mean by this.

I meant, can cluster "X" receives a (cluster) command (in `Cluster::processControlMessage`) that ends up being processed by cluster "Y"?

What commands are "re-routable"?  From `CommandRouter::getCommandRoutingMode()`: `isDomainsValue()` and `isClustersValue()`.

I am just pointing out that an object of our our routing is a cluster (`getRelevantCluster`) and that is the cluster receiving  the rerouted command (in `Cluster::processControlMessage`).  I wonder if `Application` code can be simplified by not dealing with rerouted commands if the cluster executes rerouted commands when it receives them.   Provided, clusters can parse the commands.

src/groups/mqb/mqba/mqba_commandrouter.cpp
```diff
@@ -0,0 +1,369 @@
...
+    cluster->dispatcher()->execute(
+        bdlf::BindUtil::bind(&mqbi::Cluster::getPartitionPrimaryNode,
+                             cluster,
+                             &node,
+                             &isSelfPrimary,
+                             result,
+                             d_partitionId),
+        cluster);
+
+    cluster->dispatcher()->synchronize(cluster);
```
dorjesinpo, line 110: `getRouteMembers` could be made non-blocking, but it is not necessary.

src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -632,17 +688,123 @@ int Application::processCommand(const bslstl::StringRef& source,
...
+        }
+    }
+
+    if (shouldSelfExecute) {
+        if (0 != executeCommand(commandWithOptions, command, &cmdResult)) {
+            return 0;  // early exit (caused by "dangerous" command)
+        }
+    }
+
+    routeCommandManager.waitForResponses();
```
dorjesinpo, line 781: Worth a comment saying it is blocking (all other commands)

dorjesinpo (changes requested):
`mqbcmd_parseutil.t.cpp CASE 1` needs an update

src/groups/mqb/mqbblp/mqbblp_cluster.h
lukedigiovanna: Oh I see! No, we do not support a cluster executing a command for another cluster. So we may be able to move some of that logic to the cluster code.

The problem, however, is that a lot of the command processing code lives in the `Application` currently (parsing a command, executing it, and formatting results). Some of this could be abstracted to another file/class and be reused by the application and cluster code, but the execution logic needs to live in the `Application` since it utilizes many high-level dependencies. Doing this might make the Application code simpler but would introduce some complexity elsewhere.

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
lukedigiovanna: I was wondering this too. I don't think there is, which would mean the `rerouted` field is essentially useless and can be removed. The only reason for keeping this might be to "future-proof" it. I'm not sure if there is any possible use case for sending admin commands between nodes other than in the context of "rerouting". It seems not.

src/groups/mqb/mqba/mqba_domainmanager.cpp
```diff
@@ -663,27 +688,11 @@ int DomainManager::processCommand(mqbcmd::DomainsResult*        result,
...
-            // To return a result from command execution, we need to
-            // synchronize with the domain creation attempt
-            latch.wait();
-
-            if (0 != locateDomain(&domainSp, name)) {
-                mwcu::MemOutStream os;
-                os << "Domain '" << name << "' doesn't exist";
-                result->makeError().message() = os.str();
-                return -1;  // RETURN
-            }
```
678098, line 686: Hi @lukedigiovanna this refactor with extracting code to `locateOrCreateDomain` might be done in a separate branch. If you do so, we will be able to merge that small branch very fast, and you will be able to trigger GH CI automatically without anyone's approval.
Do you want to try it?

lukedigiovanna: Sure! To clarify - I would just make this change and make a PR for it? Is there a reason to do this as opposed to just leaving it in this PR? Is it because this change is not exactly related to the objective of this PR?

678098: >I would just make this change and make a PR for it?
Yes, you need to copy 2 files with changes from this PR: `src/groups/mqb/mqba/mqba_domainmanager.cpp` and `src/groups/mqb/mqba/mqba_domainmanager.h`, and move these files to another branch. Then open a second PR with another branch.

> Is there a reason to do this as opposed to just leaving it in this PR?
The main reason why I proposed this - to unblock you from using CI. Currently, if you want to check CI messages, someone else from the team has to press a button to approve CI run. If you contribute something to the repo, approval would not be needed, so CI will start automatically if you push anything to the branch.
Also, the reason could be to make this PR smaller. Typically it's better to split huge PRs into smaller chunks, if it's possible. It will make the review process easier, since a reviewer will need to keep in mind less things at the same time while looking at the branch. However, it's not strictly necessary.

>Is it because this change is not exactly related to the objective of this PR?
We don't have a strict requirement to keep not relevant changes in separate PRs. Moreover, the good pattern is to fix a few things around your code when you make changes, if you see something you can easily fix. Typos, wrong formatting, outdated comments, warnings etc, it's too much to create a separate branch for any of these, but adding a few of this changes in your feature branch might be okay to improve overall code quality.

src/groups/mqb/mqbblp/mqbblp_cluster.h
dorjesinpo: I think, there is a possibility of routing commands which are _not_ cluster related, in which case it the `Application` must process the command.  So, calling `Application` callback is the preferred way after all.

src/integration-tests/test_admin_command_routing.py
```diff
@@ -0,0 +1,277 @@
...
+    Test: commands intended only for primary node are automatically routed to
+          primary node and response it sent back when executed from non-primary
+
+    Stage 1: Find non-primary
+    - Start admin client
+    - Find node which is not known leader
+    - Connect to that node
+    - Open FANOUT domain
+
+    Stage 3: Try all primary-only commands:
```
kaikulimu, line 47:
```suggestion
    Stage 2: Try all primary-only commands:
```

src/integration-tests/test_admin_command_routing.py
```diff
@@ -0,0 +1,277 @@
...
+    # Try CLUSTERS CLUSTER <name> STORAGE REPLICATION SET_ALL <param> <value>
+    res = admin.send_admin(
+        f"CLUSTERS CLUSTER {multi_node.name} STORAGE REPLICATION SET_ALL quorum 2"
+    )
+    success_count = res.split().count("Quorum")
+    assert success_count == num_nodes
+
+    # GET_ALL
+    res = admin.send_admin(
+        f"CLUSTERS CLUSTER {multi_node.name} STORAGE REPLICATION SET_ALL quorum 2"
```
kaikulimu, line 151:
```suggestion
        f"CLUSTERS CLUSTER {multi_node.name} STORAGE REPLICATION GET_ALL quorum 2"
```

src/integration-tests/test_admin_command_routing.py
```diff
@@ -0,0 +1,277 @@
...
+    # Try CLUSTERS CLUSTER <name> STATE ELECTOR SET_ALL <param> <value>
+    res = admin.send_admin(
+        f"CLUSTERS CLUSTER {multi_node.name} STATE ELECTOR SET_ALL quorum 2"
+    )
+    success_count = res.split().count("Quorum")
+    assert success_count == num_nodes
+
+    # GET_ALL
+    res = admin.send_admin(
+        f"CLUSTERS CLUSTER {multi_node.name} STATE ELECTOR SET_ALL quorum 2"
```
kaikulimu, line 165:
```suggestion
        f"CLUSTERS CLUSTER {multi_node.name} STATE ELECTOR GET_ALL quorum 2"
```

kaikulimu (changes requested):
Some comments. Overall great job!

src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -632,17 +701,129 @@ int Application::processCommand(const bslstl::StringRef& source,
...
+    // Note that this logic is important to prevent a "deadlock" scenario
+    // where two nodes are waiting on a response from each other to continue.
+    // Currently command from reroutes are executed on their own dedicated
+    // thread.
+    if (fromReroute) {
+        if (0 != executeCommand(commandWithOptions, command, &cmdResult)) {
+            return 0;  // early exit (caused by "dangerous" command)
+        }
+        printCommandResult(cmdResult, commandWithOptions.encoding(), os);
+        return cmdResult.isErrorValue() ? -2 : 0;
```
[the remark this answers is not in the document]
lukedigiovanna: There doesn't appear to be an existing enum to draw from, should I just make my own within the function scope like so:

```
enum RcEnum {
        rc_SUCCESS = 0,
        rc_EARLY_EXIT = -1,
        rc_ERROR = -2 
};
```

As is done in `Application::start`, for example

src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -497,56 +521,94 @@ void Application::stop()
...
 
-    mqbcmd::InternalResult cmdResult;
-    int                    rc = 0;
+int Application::executeCommand(const mqbcmd::Command& commandWithOptions,
+                                const mqbcmd::CommandChoice& command,
+                                mqbcmd::InternalResult*      cmdResult)
+{
+    BSLS_ASSERT_SAFE(cmdResult);
+
+    int rc;
```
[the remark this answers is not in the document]
lukedigiovanna: If these types of enums are local just to the function, is the point just for better code readability?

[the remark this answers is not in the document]
kaikulimu: Yes, the main point is for readability.

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -3555,6 +3578,25 @@ void Cluster::onFailoverThreshold()
     MWCTSK_ALARMLOG_PANIC("CLUSTER") << os.str() << MWCTSK_ALARMLOG_END;
 }
 
+void Cluster::onProcessedAdminCommand(
+    mqbnet::ClusterNode*                source,
+    const bmqp_ctrlmsg::ControlMessage& adminCommandCtrlMsg,
+    int                                 rc,
```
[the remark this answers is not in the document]
lukedigiovanna: Good catch, I just added a `BALL_LOG_ERROR` in this case and it still sends the error string back to the node that routed the command.

src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -632,17 +701,129 @@ int Application::processCommand(const bslstl::StringRef& source,
...
+                  << "[source: " << source << "]";
+
+    mqbcmd::Command commandWithOptions;
+    bsl::string     parseError;
+    if (const int rc = mqbcmd::ParseUtil::parse(&commandWithOptions,
+                                                &parseError,
+                                                cmd)) {
+        os << "Unable to decode command " << "(rc: " << rc << ", error: '"
+           << parseError << "')";
+        return rc;  // RETURN
```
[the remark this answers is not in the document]
lukedigiovanna: Ah, I was worried about error code collisions in this case. I've noticed this technique in other parts of the code. Makes sense 👍

src/groups/mqb/mqba/mqba_commandrouter.cpp
```diff
@@ -0,0 +1,377 @@
...
+        .d_self =
+            true,  // Cluster routing always requires original node to exec.
+    };
+}
+
+bool CommandRouter::route(mqbcmd::InternalResult* result,
+                          mqbi::Cluster*          relevantCluster)
+{
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(d_routingMode);
```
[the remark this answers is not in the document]
lukedigiovanna: Yes, it would fire and `getCommandRoutingMode()` gives a nullptr when the command does not require routing. This function should not run unless the command requires routing.

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -5861,7 +5861,8 @@ void ClusterQueueHelper::onCloseQueueResponse(
                   << contextSp->d_peer->nodeDescription();
 }
 
-void ClusterQueueHelper::gcExpiredQueues(bool immediate)
+void ClusterQueueHelper::gcExpiredQueues(mqbcmd::ClusterResult* result,
```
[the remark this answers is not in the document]
lukedigiovanna: The purpose of this was to provide an error response instead of silently dropping when running `CLUSTERS CLUSTER <name> FORCE_GC_QUEUES` on a non-primary. However, with routing this shouldn't be necessary anymore, so I'll get rid of it.

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -3674,5 +3716,118 @@ void Cluster::processResponse(const bmqp_ctrlmsg::ControlMessage& response)
...
+void Cluster::getPartitionPrimaryNode(mqbnet::ClusterNode**   node,
+                                      bool*                   isSelfPrimary,
+                                      mqbcmd::InternalResult* result,
+                                      int partitionId) const
+{
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(node);
+    BSLS_ASSERT_SAFE(isSelfPrimary);
+    BSLS_ASSERT_SAFE(result);
+    BSLS_ASSERT_SAFE(dispatcher()->inDispatcherThread(this));
```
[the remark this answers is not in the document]
lukedigiovanna: Since the `partitionId` can originate from someone executing `CLUSTERS CLUSTER <name> STORAGE PARTITION <partitionId> ENABLE`, for example, I prefer to send an error message in this case instead of using an assert.

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -3674,5 +3716,118 @@ void Cluster::processResponse(const bmqp_ctrlmsg::ControlMessage& response)
...
+
+            // No primary node
+            result->makeError().message() =
+                "No primary node for partition id " +
+                bsl::to_string(partitionId);
+            return;  // RETURN
+        }
+    }
+
+    // Didn't find a corresponding partition for the given partitionId
```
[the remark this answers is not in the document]
lukedigiovanna: See previous

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -5870,12 +5871,18 @@ void ClusterQueueHelper::gcExpiredQueues(bool immediate)
         d_cluster_p->dispatcher()->inDispatcherThread(d_cluster_p));
 
     if (d_cluster_p->isStopping()) {
+        if (result) {
+            result->makeError().message() = "Cluster is stopping";
+        }
         return;  // RETURN
```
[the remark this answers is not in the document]
lukedigiovanna: This function is `void`

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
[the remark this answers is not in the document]
kaikulimu: I think it's a good practice to return false when something doesn't go right, regardless of admin routing and what not. Now it's silencing errors which is unideal. Better yet, make the function return an error code.

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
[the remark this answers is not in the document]
kaikulimu: I was assuming if you changed the return value of non-void.

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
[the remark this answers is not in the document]
lukedigiovanna: I'll make it return an error code

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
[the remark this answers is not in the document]
kaikulimu: Agree

src/groups/mqb/mqba/mqba_commandrouter.cpp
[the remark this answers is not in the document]
kaikulimu: I see you have code like:
```
if (routeCommandManager.isRoutingNeeded()) {
        ...
        shouldSelfExecute = routeCommandManager.route(&cmdResult, cluster);
}
```
so method access if protected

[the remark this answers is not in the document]
kaikulimu: Good

kaikulimu:
@lukedigiovanna Don't forget to add license banners back to `mqbcfg_messages.*`

kaikulimu (dismissed):
lgtm

kaikulimu (approved):
lgtm

678098:
Congratulations @lukedigiovanna!

PR #360, opened 2024-07-16
Rework configurator and IT dependencies (more fine-grained)

jll63:


src/python/blazingmq/dev/reserveport.py
```diff
@@ -127,4 +127,4 @@ def reserve_port() -> typing.Iterator[TcpAddress]:
 
         address = tcp_address(sockname[0], sockname[1])
 
-    yield address
+        yield address
```
pniedzielski, line 130: 👍🏻

pniedzielski (approved):
LGTM!

src/python/blazingmq/dev/reserveport.py
678098: This line was correct before this PR. I already fixed this, since the port allocation logics were broken on fast machines like M Macs.

678098: https://github.com/bloomberg/blazingmq/pull/377

Issue #314, opened 2024-06-05
Does this mq supports Windows?


portsip:
### Is there an existing proposal for this?

- [X] I have searched the existing proposals

### Is your feature request related to a problem?

No, note related to the problem.

### Describe the solution you'd like

Hi, we would like to use this message queue in our system, my question is: does it support the Windows platform?

Thanks

678098:
Hi @portsip, are you interested in setting up the broker/cluster itself on Windows or writing a Windows client?

If you need the broker work in Windows:
- We haven't tested BlazingMQ build on Windows, nor we tested if it works on this platform.
- It's possible that the broker build will be successful when build with vcpkg: https://github.com/bloomberg/blazingmq?tab=readme-ov-file#with-vcpkg
- However, even if the broker is built on Windows, we cannot guarantee now that it will launch. The main concern is work with filesystem. How the broker searches for files, how it expands directories, how it rolls journal files used to store data. Testing the broker work on Windows was not a priority for us, so we haven't checked this before.
- However, partly because of this, we introduced a Docker based BlazingMQ. So you can build a Docker image with the broker, prepare configuration and launch a container from a Windows host machine.

If you need a Windows client:
- The core (this) repository contains C++ SDK usable for writing a client. The same challenges from above are applied here, but the work with filesystem is not. It might work, but there is a risk that something is missing.
- There is also a Python SDK, but it is based on C++ SDK, and we also haven't tested it on Windows. https://github.com/bloomberg/blazingmq-sdk-python
- There is a Java SDK, and it is a pure Java implementation of BlazingMQ protocol. I believe that it might be build on Windows and also it will work. https://github.com/bloomberg/blazingmq-sdk-java

portsip:
Thanks, do you have any plan to support running the broker/cluster on a Windows(no user docker.)
Our product currently uses NATS, but we would like to switch to Blazing, since our product is cross-platform, we need to support Windows.

678098:
Supporting BlazingMQ on Windows was not a priority for us, so we didn't put effort into this. We tried to build it on Windows as an experiment a few days ago, but encountered some problems with our dependencies. We'll try to bump dependencies versions in vcpkg registry, and it might help to build on Windows at least. Still, even if the build is ok, extensive testing is required.

PR #361, opened 2024-07-16
MWC: make Dummy function placeholders callable to fix build with MSVC

678098:
Simpler to observe what's going on in this source: https://github.com/bloomberg/blazingmq/blob/49fe8ab987bdaeef1298bb16538b373922b8d73f/src/groups/mwc/mwcex/mwcex_job.h

Note that in the code we instantiate some template classes with `Dummy` objects, like this:
```
mwcu::ObjectPlaceHolder<sizeof(Job_Target<Dummy>)> d_target;
```

We don't actually use these `Job_Target<Dummy>` objects, but let's look closely what do we want to instantiate:

```
bslmf::Util::moveIfSupported(d_function.object())();
```
Here:
https://github.com/bloomberg/blazingmq/blob/49fe8ab987bdaeef1298bb16538b373922b8d73f/src/groups/mwc/mwcex/mwcex_job.h#L217-L222

So basically in this code we are trying to "call" a `Dummy` structure here: bslmf::Util::moveIfSupported(d_function.object())**()**;

I am curious why it even compiled before.

Note that we already have this fix in some, but not all the places where it's needed. For example, here: https://github.com/bloomberg/blazingmq/blob/7f817c436d2e444969ad6e673032aaa9a09412b5/src/groups/mwc/mwcu/mwcu_operationchain.h#L390-L392

**UPD:** for executor it's even more complicated, since we have to implement the common interface of `Executor_TargetBase` which is expected as a template parameter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,9 @@ settings.json
 .pytest_cache/
 .cache/
 .venv/
+
+# Symlink from 'src/applications/bmqbrkr/run'
+src/applications/bmqbrkr/etc/etc
+
+# 'sim_cpp11_features.pl' backups
+*.bak
--- a/src/groups/mwc/mwcex/mwcex_executor.h
+++ b/src/groups/mwc/mwcex/mwcex_executor.h
@@ -105,7 +105,6 @@
 //..
 
 // MWC
-
 #include <mwcex_executortraits.h>
 
 // BDE
@@ -295,6 +294,16 @@ class Executor_Box_SboImp {
     /// size of the on-stack buffer used to store the executor target.
     struct Dummy {
         void* d_padding[4];
+
+        bool operator==(const Dummy&) const BSLS_KEYWORD_NOEXCEPT
+        {
+            return false;
+        }
+
+        void post(const bsl::function<void()>&) const
+        {
+            // NOTHING
+        }
     };
 
   private:
--- a/src/groups/mwc/mwcex/mwcex_future.h
+++ b/src/groups/mwc/mwcex/mwcex_future.h
@@ -113,11 +113,11 @@
 // result-supplier thread.
 
 // MWC
-
 #include <mwcu_objectplaceholder.h>
 
 // BDE
 #include <bdlb_nullablevalue.h>
+#include <bdlf_noop.h>
 #include <bsl_algorithm.h>  // bsl::swap
 #include <bsl_exception.h>
 #include <bsl_functional.h>  // bsl::reference_wrapper
@@ -150,7 +150,7 @@
 
 #if BSLS_COMPILERFEATURES_SIMULATE_CPP11_FEATURES
 // Include version that can be compiled with C++03
-// Generated on Wed Jun 29 04:17:13 2022
+// Generated on Wed Jul 17 12:44:45 2024
 // Command line: sim_cpp11_features.pl mwcex_future.h
 #define COMPILING_MWCEX_FUTURE_H
 #include <mwcex_future_cpp03.h>
@@ -232,7 +232,7 @@ class Future_Exception {
 
     /// Provides a "small" dummy object which size is used to calculate the
     /// size of the on-stack buffer used for optimization.
-    struct Dummy {
+    struct Dummy : public bdlf::NoOp {
         void* d_padding[3];
     };
 
@@ -359,7 +359,7 @@ class Future_Callback {
 
     /// Provides a "small" dummy object which size is used to calculate the
     /// size of the on-stack buffer used for optimization.
-    struct Dummy {
+    struct Dummy : public bdlf::NoOp {
         void* d_padding[3];
     };
 
--- a/src/groups/mwc/mwcex/mwcex_future_cpp03.h
+++ b/src/groups/mwc/mwcex/mwcex_future_cpp03.h
@@ -1,4 +1,4 @@
-// Copyright 2022-2023 Bloomberg Finance L.P.
+// Copyright 2018-2023 Bloomberg Finance L.P.
 // SPDX-License-Identifier: Apache-2.0
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -36,17 +36,12 @@
 // regions of C++11 code, then this header contains no code and is not
 // '#include'd in the original header.
 //
-// Generated on Wed Jun 29 04:17:13 2022
+// Generated on Wed Jul 17 12:44:45 2024
 // Command line: sim_cpp11_features.pl mwcex_future.h
 
 #ifdef COMPILING_MWCEX_FUTURE_H
 
 namespace BloombergLP {
-
-// FORWARD DECLARATION
-namespace bslma {
-class Allocator;
-}
 
 namespace mwcex {
 
@@ -59,36 +54,36 @@ class FutureSharedState;
 // class Future_Exception
 // ======================
 
+/// Provides a polymorphic exception wrapper that can hold and emit any type
+/// of exception.
 class Future_Exception {
-    // Provides a polymorphic exception wrapper that can hold and emit any type
-    // of exception.
-
   private:
     // PRIVATE TYPES
-    class TargetBase {
-        // Provides an interface used to implement the type erasure technique.
-        // When creating a polymorphic exception with a target of type 'E', an
-        // instance of derived class template 'Target<E>' is instantiated and
-        // stored via a pointer to its base class (this one). Then, calls to
-        // 'mwcex::Future_Exception's public methods are forwarded to this
-        // class.
 
+    /// Provides an interface used to implement the type erasure technique.
+    /// When creating a polymorphic exception with a target of type `E`, an
+    /// instance of derived class template `Target<E>` is instantiated and
+    /// stored via a pointer to its base class (this one). Then, calls to
+    /// `mwcex::Future_Exception`s public methods are forwarded to this
+    /// class.
+    class TargetBase {
       public:
         // CREATORS
+
+        /// Destroy this object and the contained exception target with it.
         virtual ~TargetBase();
-        // Destroy this object and the contained exception target with it.
 
       public:
         // ACCESSORS
+
+        /// Throw a copy of the contained exception object.
         BSLS_ANNOTATION_NORETURN virtual void emit() const = 0;
-        // Throw a copy of the contained exception object.
     };
 
+    /// Provides an implementation of the `TargetBase` interface containing
+    /// the exception target.
     template <class EXCEPTION>
     class Target : public TargetBase {
-        // Provides an implementation of the 'TargetBase' interface containing
-        // the exception target.
-
       private:
         // PRIVATE DATA
         bslalg::ConstructorProxy<EXCEPTION> d_exception;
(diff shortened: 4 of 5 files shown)


hallfox:
Why was MSVC fixed by making these types callable? I'm a little concerned we're taking a PoD type and unmaking it PoD.

678098:
@hallfox added an explanation for this to the title of the PR

ddragan-bloomberg:
I'll review this.

ddragan-bloomberg:
Template code is not instantiated unless used explicitly, which I believe is the reason this code compiles perfectly in GCC/Clang. MSVC might by buggy. Does `libmwc` even have to compile on Windows? If so, what is the minimum compiler version is has to compile with?

src/groups/mwc/mwcex/mwcex_future.h
```diff
@@ -232,7 +232,7 @@ class Future_Exception {
 
     /// Provides a "small" dummy object which size is used to calculate the
     /// size of the on-stack buffer used for optimization.
-    struct Dummy {
+    struct Dummy : public mwcu::NoOp {
```
ddragan-bloomberg, line 235: I kinda get the rational behind changes in `mwcex_job` -- the `Dummy` is involved with a functor, but what is the rational here? Does the code not compile if you leave this dummy (and the one below) as is?

ddragan-bloomberg:
Side note. Evgenii, while you're at it, might as well replace usage of `mwcu::NoOp` with `bdlf::NoOp`

.gitignore
```diff
@@ -26,3 +26,15 @@ settings.json
 .pytest_cache/
 .cache/
 .venv/
+
+# Symlink from 'src/applications/bmqbrkr/run'
+src/applications/bmqbrkr/etc/etc
+
+# 'sim_cpp11_features.pl' backups
+src/groups/mwc/mwcex/mwcex_bindutil.h.bak
```
ddragan-bloomberg, line 34: wouldn't it be simpler to remove those backup files in that same script that calls sim_cpp11_features.pl ?

ddragan-bloomberg:
Evgenii, since you decided to replace `mwcu::NoOp` in all `mwc` and `BMQ` code, please do this in a separate PR, then rebase this PR on top of the other one once I merge it.

And DO NOT remove the `mwcu_noop` component! Other libraries and apps depend on `libmwc` and might be still using `mwcu::NoOp`.

.gitignore
678098: We don't have access to the script that launches the `sim_cpp11_features.pl`, since it's done within `bde-tools`. I updated `.gitignore` with wildcard backup pattern which covers all the changes, and this is the simplest way.

678098:
> And DO NOT remove the `mwcu_noop` component! Other libraries and apps depend on `libmwc` and might be still using `mwcu::NoOp`.

I already checked that no other libraries depend on `mwcu::NoOp`

ddragan-bloomberg:
> I already checked that no other libraries depend on mwcu::NoOp

Okay, It that case we are good to remove it.

ddragan-bloomberg:
Also note than, unlike `mwc`, BDE provides s static instance of `bdlf::NoOp` -- `bdlf::noOp`, should use it where appropriate.

678098:
> Also note than, unlike `mwc`, BDE provides s static constexpr instance of `bdlf::NoOp` -- `bdlf::noOp`, should use it where appropriate.

This is a good point. We might save a few ticks in some places.

> Evgenii, since you decided to replace `mwcu::NoOp` in all `mwc` and `BMQ` code, please do this in a separate PR, then rebase this PR on top of the other one once I merge it.

Moved `mwcu::NoOp` -> `bdlf::NoOp` transition from this branch.

> Does `libmwc` even have to compile on Windows?

We can get Windows support for `mwc` almost for free, a bit more complicate for lib `bmq` and even more complicated for `mqb`. So, we can pretty easily make buildable Windows SDK (depends on `mwc` + `bmq`), and it's more difficult to build a Windows Broker (depends on `mwc` + `bmq` + `mqb`).

For `mwc`, just fix these few templates and that's it. We don't have requirement to build BlazingMQ on Windows, but we can at least gradually extend support for this platform, by fixing a few non-portable parts of the codebase. There are external users who might benefit from using it on Windows https://github.com/bloomberg/blazingmq/issues/314

> If so, what is the minimum compiler version is has to compile with?

No strict requirements. On this stage, it's worth it to solve problems once they occur, until we have a better vision.

ddragan-bloomberg:
How do you check the code compiles on Windows. Which compiler do you use? Can you use a newer compiler?

678098:
> How do you check the code compiles on Windows. Which compiler do you use? Can you use a newer compiler?

I have a configured Windows preset which I can build. Also, I use the latest MSVC compiler available.

```
$ "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.40.33807\bin\HostX64\x64\CL.exe"
Microsoft (R) C/C++ Optimizing Compiler Version 19.40.33812 for x64
```

If I roll back the proposed changes, I will get errors from the compiler:

```
D:\blazingmq\src\groups\mwc\mwcex\mwcex_job.h(224,33): error C2064: term does not evaluate to a function taking 0 arguments [D:\blazingmq\cmake.bld\src\groups\mwc\mwcex-iface.vcxproj]
  mwcex_promise.cpp
  mwcex_promise_cpp03.cpp
  mwcex_sequentialcontext.cpp
D:\blazingmq\src\groups\mwc\mwcex\mwcex_job.h(224,33): error C2064: term does not evaluate to a function taking 0 arguments [D:\blazingmq\cmake.bld\src\groups\mwc\mwcex-iface.vcxproj]
  mwcex_strand.cpp
D:\blazingmq\src\groups\mwc\mwcex\mwcex_job.h(224,33): error C2064: term does not evaluate to a function taking 0 arguments [D:\blazingmq\cmake.bld\src\groups\mwc\mwcex-iface.vcxproj]
  mwcex_systemexecutor.cpp
  Generating Code...
```

678098:
> How do you check the code compiles on Windows. Which compiler do you use? Can you use a newer compiler?

I have a configured Windows preset which I can build. Also, I use the latest MSVC compiler available:

```
$ "Tools\MSVC\14.40.33807\bin\HostX64\x64\CL.exe"
Microsoft (R) C/C++ Optimizing Compiler Version 19.40.33812 for x64
```

If I roll back the proposed changes, I will get errors from the compiler:

```
D:\blazingmq\src\groups\mwc\mwcex\mwcex_job.h(224,33): error C2064: term does not evaluate to a function taking 0 arguments [D:\blazingmq\cmake.bld\src\groups\mwc\mwcex-iface.vcxproj]
  mwcex_promise.cpp
  mwcex_promise_cpp03.cpp
  mwcex_sequentialcontext.cpp
D:\blazingmq\src\groups\mwc\mwcex\mwcex_job.h(224,33): error C2064: term does not evaluate to a function taking 0 arguments [D:\blazingmq\cmake.bld\src\groups\mwc\mwcex-iface.vcxproj]
  mwcex_strand.cpp
D:\blazingmq\src\groups\mwc\mwcex\mwcex_job.h(224,33): error C2064: term does not evaluate to a function taking 0 arguments [D:\blazingmq\cmake.bld\src\groups\mwc\mwcex-iface.vcxproj]
  mwcex_systemexecutor.cpp
  Generating Code...
```

ddragan-bloomberg:
There is a question left unanswered -- what is the rational behind changes in `mwcex_future.h`? Does the code compiles without them?

678098:
> There is a question left unanswered -- what is the rational behind changes in `mwcex_future.h`? Does the code compiles without them?

It doesn't

ddragan-bloomberg:
I assume this is because `Future_Callback` is functor-related. What about `Future_Exception`? Are you able to remove the use of `NoOp` there?

678098:
> I assume this is because `Future_Callback` is functor-related. What about `Future_Exception`? Are you able to remove the use of `NoOp` there?

No

If I remove inheritance from `NoOp` in either `Future_Callback` or `Future_Exception`, it will cause the same compilation error in `Job_Target::invoke()`:

```
// MANIPULATORS
template <class FUNCTION>
inline void Job_Target<FUNCTION>::invoke()
{
    bslmf::Util::moveIfSupported(d_function.object())(); // blazingmq\src\groups\mwc\mwcex\mwcex_job.h(224,33): error C2064: term does not evaluate to a function taking 0 arguments [blazingmq\cmake.bld\src\groups\mwc\mwcex-iface.vcxproj]

}
```

And regarding the `Executor_Box_SboImp::Dummy`, if left as-is, the compilation errors would be:

```
blazingmq\src\groups\mwc\mwcex\mwcex_executor.h(603,52): error C2676: binary '==': 'const EXECUTOR' does not define this operator or a conversion to a type acceptable to the predefined operator [blazingmq\cmake.bld\src\groups\mwc\mwcex-iface.vcxproj]
blazingmq\src\groups\mwc\mwcex\mwcex_executor.h(603,52): error C2676:         with [blazingmq\cmake.bld\src\groups\mwc\mwcex-iface.vcxproj]
blazingmq\src\groups\mwc\mwcex\mwcex_executor.h(603,52): error C2676:         [ [blazingmq\cmake.bld\src\groups\mwc\mwcex-iface.vcxproj]
blazingmq\src\groups\mwc\mwcex\mwcex_executor.h(603,52): error C2676:             EXECUTOR=BloombergLP::mwcex::Executor_Box_SboImp::Dummy [blazingmq\cmake.bld\src\groups\mwc\mwcex-iface.vcxproj]
blazingmq\src\groups\mwc\mwcex\mwcex_executor.h(603,52): error C2676:         ] [blazingmq\cmake.bld\src\groups\mwc\mwcex-iface.vcxproj]
```

I found out that it's not necessary to inherit `Executor_Box_SboImp::Dummy` from `Executor_TargetBase`, but I had to provide implementations of two methods which causes compilation errors:

```
blazingmq\src\groups\mwc\mwcex\mwcex_executor.h(639,52): error C2676: binary '==': 'const EXECUTOR' does not define this operator or a conversion to a type acceptable to the predefined operator [blazingmq\cmake.bld\src\groups\mwc\mwcex-iface.vcxproj]

blazingmq\src\groups\mwc\mwcex\mwcex_executortraits.h(206,14): error C2039: 'post': is not a member of 'BloombergLP::mwcex::Executor_Box_SboImp::Dummy' [blazingmq\cmake.bld\src\groups\mwc\mwcex-iface.vcxproj]
```

.gitignore
```diff
@@ -26,3 +26,10 @@ settings.json
 .pytest_cache/
 .cache/
 .venv/
+
+# Symlink from 'src/applications/bmqbrkr/run'
+src/applications/bmqbrkr/etc/etc
+
+# 'sim_cpp11_features.pl' backups
+src/groups/mwc/mwcex/*.bak
```
ddragan-bloomberg, line 34: can we generalize that for all packages?

src/groups/mwc/mwcex/mwcex_executor.h
```diff
@@ -295,6 +294,18 @@ class Executor_Box_SboImp {
     /// size of the on-stack buffer used to store the executor target.
     struct Dummy {
         void* d_padding[4];
+
+        bool
+        operator==(BSLS_ANNOTATION_UNUSED const Executor_Box_SboImp::Dummy&)
```
ddragan-bloomberg, line 299: I don't think you'll get warning while using `BSLS_ANNOTATION_UNUSED` if you remove the argument name (like done in this declaration).

src/groups/mwc/mwcex/mwcex_executor.h
```diff
@@ -295,6 +294,18 @@ class Executor_Box_SboImp {
...
+            const BSLS_KEYWORD_NOEXCEPT
+        {
+            return false;
+        }
+
+        void post(BSLS_ANNOTATION_UNUSED const bsl::function<void()>& f) const
```
ddragan-bloomberg, line 305: same as above

src/groups/mwc/mwcex/mwcex_executor.h
ddragan-bloomberg: Also, the argument type can just be `const Dummy&`

ddragan-bloomberg (approved):
Looks good

pniedzielski (approved):

PR #363, opened 2024-07-16
fix: update queue stats on response (not only on request)

dorjesinpo:
In the presence of another client, the reported stats are one step behind because they get updated to early - before consumer fully closes.


pniedzielski (approved):
👍🏻

PR #366, opened 2024-07-18
dev: Optimize counting of unconfirmed

dorjesinpo:
First step in optimizing shutdown logic.  The goal of which is to avoid iterating subStreams (preparing for Reliable Broadcast).

The `ClusterQueueHelper::d_counterOfUnconfirmed` is all we will need for shutdown in the next step.
All other counters will (gradually) retire.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
@@ -2106,6 +2106,7 @@ bsl::shared_ptr<mqbi::Queue> ClusterQueueHelper::createQueueFactory(
                                    d_clusterData_p->scheduler(),
                                    d_clusterData_p->miscWorkThreadPool(),
                                    openQueueResponse.routingConfiguration(),
+                                   &d_counterOfUnconfirmed,
                                    d_allocator_p),
         d_allocator_p);
 
@@ -4485,6 +4486,7 @@ ClusterQueueHelper::ClusterQueueHelper(
 , d_numPendingReopenQueueRequests(0)
 , d_primaryNotLeaderAlarmRaised(false)
 , d_stopContexts(allocator)
+, d_counterOfUnconfirmed(0)
 {
     BSLS_ASSERT(
         d_clusterData_p->clusterConfig()
--- a/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
@@ -2106,7 +2106,7 @@ bsl::shared_ptr<mqbi::Queue> ClusterQueueHelper::createQueueFactory(
                                    d_clusterData_p->scheduler(),
                                    d_clusterData_p->miscWorkThreadPool(),
                                    openQueueResponse.routingConfiguration(),
-                                   &d_counterOfUnconfirmed,
+                                   &d_unconfirmedCounter,
                                    d_allocator_p),
         d_allocator_p);
 
@@ -4486,7 +4486,7 @@ ClusterQueueHelper::ClusterQueueHelper(
 , d_numPendingReopenQueueRequests(0)
 , d_primaryNotLeaderAlarmRaised(false)
 , d_stopContexts(allocator)
-, d_counterOfUnconfirmed(0)
+, d_unconfirmedCounter(NULL)
 {
     BSLS_ASSERT(
         d_clusterData_p->clusterConfig()
--- a/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.h
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.h
@@ -486,6 +486,9 @@ class ClusterQueueHelper : public mqbc::ClusterStateObserver,
 
     StopContexts d_stopContexts;
 
+    mqbu::SingleCounter d_counterOfUnconfirmed;
+    // Top level in the hierarchy: Cluster ->> QueueState ->> QueueHandle
+
   private:
     // PRIVATE MANIPULATORS
 
--- a/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.h
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.h
@@ -486,8 +486,8 @@ class ClusterQueueHelper : public mqbc::ClusterStateObserver,
 
     StopContexts d_stopContexts;
 
-    mqbu::SingleCounter d_counterOfUnconfirmed;
-    // Top level in the hierarchy: Cluster ->> QueueState ->> QueueHandle
+    mqbu::SingleCounter d_unconfirmedCounter;
+    /// Top level in the hierarchy: Cluster ->> QueueState ->> QueueHandle
 
   private:
     // PRIVATE MANIPULATORS
--- a/src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
@@ -151,8 +151,8 @@ Test::Test()
                d_storageKey,
                d_partitionId,
                &d_domain,
-               0,
-               // not counting unconfirmed
+               NULL,
+               // no parent counter of unconfirmed
                s_allocator_p)
 , d_monitor(&d_queueState, s_allocator_p)
 , d_storage(d_queue.uri(),
--- a/src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
@@ -151,6 +151,8 @@ Test::Test()
                d_storageKey,
                d_partitionId,
                &d_domain,
+               0,
+               // not counting unconfirmed
                s_allocator_p)
 , d_monitor(&d_queueState, s_allocator_p)
 , d_storage(d_queue.uri(),
--- a/src/groups/mqb/mqbblp/mqbblp_queueenginetester.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_queueenginetester.cpp
@@ -502,6 +502,8 @@ void QueueEngineTester::init(const mqbconfm::Domain& domainConfig)
                                                 k_NULL_QUEUE_KEY,
                                                 k_PARTITION_ID,
                                                 d_mockDomain_mp.get(),
+                                                0,
+                                                // not counting Unconfirmed
                                                 d_allocator_p),
                          d_allocator_p);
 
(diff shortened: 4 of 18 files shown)


src/groups/mqb/mqbmock/mqbmock_queue.cpp
```diff
@@ -528,7 +528,8 @@ mqbi::QueueHandle* HandleFactory::makeHandle(
     const bsl::shared_ptr<mqbi::QueueHandleRequesterContext>& clientContext,
     mqbstat::QueueStatsDomain*                                stats,
     const bmqp_ctrlmsg::QueueHandleParameters&                handleParameters,
-    bslma::Allocator*                                         allocator)
+    BSLS_ANNOTATION_UNUSED mqbu::SingleCounter* parent,
```
678098, line 531: I am surprised to see the formatter misaligning the arguments list here.
Not needed to fix, but it might cause reformatting when we update clang-format

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -4485,6 +4486,7 @@ ClusterQueueHelper::ClusterQueueHelper(
 , d_numPendingReopenQueueRequests(0)
 , d_primaryNotLeaderAlarmRaised(false)
 , d_stopContexts(allocator)
+, d_counterOfUnconfirmed(0)
```
678098, line 4489: Would prefer `bsl::nullptr_t()`, `nullptr` or `NULL` here, because `0` is confusing. Looking at this line, It's possible to assume that this line sets the `0 initial value` of the counter, but here it sets the `0 pointer` to a missing parent.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -151,6 +151,8 @@ Test::Test()
                d_storageKey,
                d_partitionId,
                &d_domain,
+               0,
```
678098, line 154: Here, 0 is not misleading, but also might change to some `NULL`

src/groups/mqb/mqbblp/mqbblp_queueenginetester.cpp
```diff
@@ -502,6 +502,8 @@ void QueueEngineTester::init(const mqbconfm::Domain& domainConfig)
                                                 k_NULL_QUEUE_KEY,
                                                 k_PARTITION_ID,
                                                 d_mockDomain_mp.get(),
+                                                0,
```
678098, line 505: Here, 0 is not misleading, but also might change to some `NULL`

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
```diff
@@ -1252,16 +1262,7 @@ bsls::Types::Int64 QueueHandle::countUnconfirmed(unsigned int subQueueId) const
...
-        for (Downstreams::const_iterator itStream = d_downstreams.begin();
-             itStream != d_downstreams.end();
-             ++itStream) {
-            const bsl::shared_ptr<Downstream>& downstream = (*itStream);
-            if (downstream) {
-                if (downstream->d_data) {
-                    result += downstream->d_data->size();
-                }
-            }
-        }
```
678098, line 1264: Looking at the code, I am not sure now if this loop is actually slow.
It seems that we are not blocked here by any IO, and not blocked by the need to acquire mutexes or smth like this.
Can you estimate how many times we might call this `countUnconfirmed` and how many `d_downstream`s we might have on average?

Suppose we have 100k subqueues, and each queue has 10 downstreams. Then we'll have just 1kk additions here in total across all operation chain links, which doesn't look too bad.

My concern here, what if we made our normal workflow slower by a few ticks when we make these counter updates when it's not really necessary.

src/groups/mqb/mqbi/mqbi_queue.h
```diff
@@ -964,12 +965,13 @@ class QueueHandleFactory {
     /// Create a new handle, using the specified `allocator`, for the
     /// specified `queue` as requested by the specified `clientContext` with
     /// the specified `parameters`, and associated wit the specified
-    /// `stats`.
+    /// `stats`.  Use the specified 'counterOfUnconfirmed'.
```
678098, line 968: No argument `counterOfUnconfirmed` here now, only the `parent` which I would like to rename to avoid confusion

src/groups/mqb/mqbblp/mqbblp_remotequeue.t.cpp
```diff
@@ -289,6 +289,8 @@ TestBench::TestRemoteQueue::TestRemoteQueue(
                d_storageKey,
                1,  // partition
                &theBench.d_domain,
+               0,
```
678098, line 292: Here, 0 is not misleading, but also might change to some `NULL`

src/groups/mqb/mqbblp/mqbblp_queuestate.h
```diff
@@ -210,6 +212,7 @@ class QueueState {
                const mqbu::StorageKey& key,
                int                     partitionId,
                mqbi::Domain*           domain,
+               mqbu::SingleCounter*    parent,
```
678098, line 215: Not mentioned in the docs

src/groups/mqb/mqbblp/mqbblp_queueenginetester.cpp
```diff
@@ -502,6 +502,8 @@ void QueueEngineTester::init(const mqbconfm::Domain& domainConfig)
...
+                                                // not counting Unconfirmed
```
678098, line 506: This comment is misleading, since we actually count unconfirmed. We just don't pass a parent counter to `QueueState`

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -151,6 +151,8 @@ Test::Test()
...
+               // not counting unconfirmed
```
678098, line 155: This comment is misleading, since we actually count unconfirmed. We just don't pass a parent counter to `QueueState`

src/groups/mqb/mqbblp/mqbblp_remotequeue.t.cpp
```diff
@@ -289,6 +289,8 @@ TestBench::TestRemoteQueue::TestRemoteQueue(
...
+               // not counting unconfirmed
```
678098, line 293: This comment is misleading, since we actually count unconfirmed. We just don't pass a parent counter to `QueueState`

678098 (changes requested):
Have some comments, the most important one, are we sure that the loop over downstreams is actually slow?

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.h
```diff
@@ -486,6 +486,9 @@ class ClusterQueueHelper : public mqbc::ClusterStateObserver,
 
     StopContexts d_stopContexts;
 
+    mqbu::SingleCounter d_counterOfUnconfirmed;
```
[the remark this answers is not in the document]
dorjesinpo: I have actually renamed `d_unconfirmedCounter` thinking `counterOfUnconfirmed` is more descriptive (`<smth>Counter` can also mean that `<smth>` is a property of `Counter`).  But sure, let me rename it back

dorjesinpo:
> Have some comments, the most important one, are we sure that the loop over downstreams is actually slow?

For the ReliableBroadcast, yes.  Consider thread switching too.  AtomicInt seems to be a small price for replacing it.

src/groups/mqb/mqbu/mqbu_resourceusagemonitor.h
```diff
@@ -465,6 +466,20 @@ class ResourceUsageMonitor {
     print(bsl::ostream& stream, int level = 0, int spacesPerLevel = 4) const;
 };
 
+struct SingleCounter {
+    /// Thread-safe int value hierarchical tracker
+    /// Relies on an owner for creation/destruction
+
+    SingleCounter*  d_parent_p;
+    bsls::AtomicInt d_value;
```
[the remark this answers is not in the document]
dorjesinpo: I wanted to catch miscounting. `bsls::AtomicInt64` is a good idea

src/groups/mqb/mqbu/mqbu_resourceusagemonitor.h
```diff
@@ -658,6 +673,30 @@ inline ResourceUsageMonitorState::Enum ResourceUsageMonitor::state() const
...
+inline SingleCounter::SingleCounter(SingleCounter* parent)
+: d_parent_p(parent)
+, d_value(0)
+{
+    // NOTHING
+}
+
+inline void SingleCounter::update(int delta)
+{
+    d_value += delta;
```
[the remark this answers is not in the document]
dorjesinpo: I can't think of the negative case.  Still prefer to keep it signed.

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
dorjesinpo: I think Reliable Broadcast shutdown is going to suffer, up to the point of crushing, as we have it in some cases with overloaded proxies.  Consider thread switching as well.  We replace that at the cost of 2-3 atomic addition per PUSH/CONFIRM.  That cost is small.

src/groups/mqb/mqbblp/mqbblp_queueenginetester.cpp
dorjesinpo: That meant the caller is not counting
(`// no parent counter of unconfirmed`)

src/groups/mqb/mqbu/mqbu_resourceusagemonitor.h
```diff
@@ -658,6 +683,33 @@ inline ResourceUsageMonitorState::Enum ResourceUsageMonitor::state() const
...
+
+    BSLS_ASSERT_SAFE(d_value >= 0);
+
+    if (d_parent_p) {
```
678098, line 703: Do we necessary need to have a tree-like structure for counting unconfirmed with update on every node on the path to the root?
I think, we only need one top-level counter during shutdown.
I would like to avoid `if` here, so we don't have extra branching in the normal workflow.

We can still build `SingleCounter`s with a parent object, but we can go up the tree and set up the pointer to the parent counter in each node. So we just update the atomic here and don't call `d_parent_p->update(delta);`

src/groups/mqb/mqbu/mqbu_resourceusagemonitor.h
```diff
@@ -465,6 +466,30 @@ class ResourceUsageMonitor {
...
+
+struct SingleCounter {
+    /// Thread-safe int value hierarchical tracker
+    /// Relies on an owner for creation/destruction
+
+    SingleCounter* d_parent_p;
+    /// If not null, gets updated
+
+    bsls::AtomicInt64 d_value;
+    /// The value being tracked
```
678098, line 481: These comments should be on top of the field

dorjesinpo:
Reverting to the old way of counting.  A relevant PR is [https://github.com/bloomberg/blazingmq/pull/399](https://github.com/bloomberg/blazingmq/pull/399)

PR #367, opened 2024-07-22
mqbc::StorageMgr, RecoveryMgr: Thread safety improvements

kaikulimu:
This series of changes improved the thread safety of components integral to the FSM mode, including `mqbc::StorageManager`, `mqbc::RecoveryManager`, and other neighboring components.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_recoverymanager.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_recoverymanager.cpp
@@ -1722,14 +1722,15 @@ int RecoveryManager::replayPartition(
         bmqp::StorageMessageType::Enum storageMsgType =
             bmqp::StorageMessageType::e_UNDEFINED;
 
-        rc = mqbc::RecoveryUtil::incrementCurrentSeqNum(&currentSeqNum,
-                                                        &journalRecordBase,
-                                                        fti->journalFd(),
-                                                        toSequenceNum,
-                                                        pid,
-                                                        *destination,
-                                                        *d_clusterData_p,
-                                                        journalIt);
+        rc = mqbc::RecoveryUtil::incrementCurrentSeqNum(
+            &currentSeqNum,
+            &journalRecordBase,
+            fti->journalFd(),
+            toSequenceNum,
+            pid,
+            *destination,
+            d_clusterData_p->identity().description(),
+            journalIt);
         if (rc != 0) {
             break;  // BREAK
         }
--- a/src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
@@ -1510,6 +1510,17 @@ void StorageManager::stop()
                              bdlf::PlaceHolders::_2));  // latch
 }
 
+void StorageManager::initializeQueueKeyInfoMap(
+    BSLS_ANNOTATION_UNUSED const mqbc::ClusterState* clusterState)
+{
+    // executed by cluster *DISPATCHER* thread
+
+    // PRECONDITION
+    BSLS_ASSERT_SAFE(d_dispatcher_p->inDispatcherThread(d_cluster_p));
+
+    BSLS_ASSERT_OPT(false && "This method should only be invoked in FSM mode");
+}
+
 void StorageManager::setPrimaryForPartition(int                  partitionId,
                                             mqbnet::ClusterNode* primaryNode,
                                             unsigned int primaryLeaseId)
--- a/src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
@@ -360,7 +360,7 @@ void StorageManager::shutdownCb(int partitionId, bslmt::Latch* latch)
     mqbc::StorageUtil::shutdown(partitionId,
                                 latch,
                                 &d_fileStores,
-                                d_clusterData_p,
+                                d_clusterData_p->identity().description(),
                                 d_clusterConfig);
 }
 
@@ -483,6 +483,7 @@ void StorageManager::recoveredQueuesCb(int                    partitionId,
         &d_appKeysVec[partitionId],
         &d_appKeysLock,
         d_domainFactory_p,
+        &d_unrecognizedDomainsLock,
         &d_unrecognizedDomains[partitionId],
         d_clusterData_p->identity().description(),
         partitionId,
@@ -495,6 +496,7 @@ void StorageManager::recoveredQueuesCb(int                    partitionId,
 
     mqbc::StorageUtil::dumpUnknownRecoveredDomains(
         d_clusterData_p->identity().description(),
+        &d_unrecognizedDomainsLock,
         d_unrecognizedDomains);
 }
 
@@ -650,11 +652,12 @@ void StorageManager::clearPrimaryForPartitionDispatched(
     mqbs::FileStore* fs    = d_fileStores[partitionId].get();
     PartitionInfo&   pinfo = d_partitionInfoVec[partitionId];
 
-    mqbc::StorageUtil::clearPrimaryForPartition(fs,
-                                                &pinfo,
-                                                *d_clusterData_p,
-                                                partitionId,
-                                                primary);
+    mqbc::StorageUtil::clearPrimaryForPartition(
+        fs,
+        &pinfo,
+        d_clusterData_p->identity().description(),
+        partitionId,
+        primary);
 }
 
 void StorageManager::processStorageEventDispatched(
@@ -742,10 +745,15 @@ void StorageManager::processPartitionSyncEvent(
         return;  // RETURN
     }
 
+    PartitionInfo                    pinfo;
+    const ClusterStatePartitionInfo& cspinfo = d_clusterState.partition(pid);
+    pinfo.setPrimary(cspinfo.primaryNode());
+    pinfo.setPrimaryLeaseId(cspinfo.primaryLeaseId());
+    pinfo.setPrimaryStatus(cspinfo.primaryStatus());
     if (!mqbc::StorageUtil::validatePartitionSyncEvent(rawEvent,
                                                        pid,
                                                        source,
-                                                       d_clusterState,
+                                                       pinfo,
                                                        *d_clusterData_p,
                                                        false)  // isFSMWorkflow
     ) {
@@ -977,6 +985,7 @@ StorageManager::StorageManager(
 , d_allocators(d_allocator_p)
 , d_isStarted(false)
 , d_lowDiskspaceWarning(false)
+, d_unrecognizedDomainsLock()
 , d_unrecognizedDomains(allocator)
 , d_blobSpPool_p(clusterData->blobSpPool())
 , d_domainFactory_p(domainFactory)
@@ -1095,7 +1104,7 @@ void StorageManager::unregisterQueue(const bmqt::Uri& uri, int partitionId)
                                  &d_storagesLock,
                                  d_clusterData_p,
                                  partitionId,
-                                 d_partitionInfoVec[partitionId],
+                                 bsl::cref(d_partitionInfoVec[partitionId]),
                                  uri));
 
     d_fileStores[partitionId]->dispatchEvent(queueEvent);
@@ -1489,6 +1498,11 @@ int StorageManager::start(bsl::ostream& errorDescription)
 
 void StorageManager::stop()
 {
+    // executed by cluster *DISPATCHER* thread
+
+    // PRECONDITION
+    BSLS_ASSERT_SAFE(d_dispatcher_p->inDispatcherThread(d_cluster_p));
+
     if (!d_isStarted) {
         return;  // RETURN
     }
@@ -1502,8 +1516,8 @@ void StorageManager::stop()
     d_recoveryManager_mp->stop();
 
     mqbc::StorageUtil::stop(
-        d_clusterData_p,
         &d_fileStores,
+        d_clusterData_p->identity().description(),
         bdlf::BindUtil::bind(&StorageManager::shutdownCb,
                              this,
                              bdlf::PlaceHolders::_1,    // partitionId
@@ -1703,14 +1717,15 @@ void StorageManager::processStorageEvent(
             d_clusterData_p->membership().selfNodeStatus() ||
         isZero(d_clusterData_p->electorInfo().leaderMessageSequence());
     const ClusterStatePartitionInfo& pinfo = d_clusterState.partition(pid);
-    if (!mqbc::StorageUtil::validateStorageEvent(rawEvent,
-                                                 pid,
-                                                 source,
-                                                 pinfo.primaryNode(),
-                                                 pinfo.primaryStatus(),
-                                                 *d_clusterData_p,
-                                                 skipAlarm,
-                                                 false)) {  // isFSMWorkflow
+    if (!mqbc::StorageUtil::validateStorageEvent(
+            rawEvent,
+            pid,
+            source,
+            pinfo.primaryNode(),
+            pinfo.primaryStatus(),
+            d_clusterData_p->identity().description(),
+            skipAlarm,
+            false)) {                                       // isFSMWorkflow
         return;                                             // RETURN
     }
 
@@ -2059,7 +2074,7 @@ void StorageManager::processReplicaStatusAdvisory(
         d_clusterData_p,
         fs,
         partitionId,
-        d_partitionInfoVec[partitionId],
+        bsl::cref(d_partitionInfoVec[partitionId]),
         source,
         status));
 }
@@ -2119,6 +2134,7 @@ void StorageManager::gcUnrecognizedDomainQueues()
         d_dispatcher_p->inDispatcherThread(d_clusterData_p->cluster()));
 
     mqbc::StorageUtil::gcUnrecognizedDomainQueues(&d_fileStores,
+                                                  &d_unrecognizedDomainsLock,
                                                   d_unrecognizedDomains);
 }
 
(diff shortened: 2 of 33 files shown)


src/groups/mqb/mqbc/mqbc_clusterdata.h
```diff
@@ -79,13 +79,13 @@ class ClusterDataIdentity {
   private:
     // DATA
 
-    bsl::string d_name;
+    const bsl::string d_name;
```
hallfox, line 82: I would avoid adding these `const`s here, this can lead to unusually restrictive class design issues later. If the intent is to ensure these fields are never modified after class construction, then we should provide an interface that guarantees that.

src/groups/mqb/mqbc/mqbc_clusterdata.h
```diff
@@ -238,6 +238,12 @@ class ClusterData {
 
     // MANIPULATORS
 
+    /// Get a modifiable reference to this object's event scheduler.
+    bdlmt::EventScheduler* scheduler();
+
+    /// Get a modifiable reference to this object's buffer factory.
+    bdlbb::BlobBufferFactory* bufferFactory();
+
```
hallfox, line 246: Can the return value be `NULL` here? If not, should we even be returning a pointer type? We should specify.

src/groups/mqb/mqbc/mqbc_clusterdata.h
kaikulimu: Good point. Our interface already has such guarantees, so I will remove the `const`s.

src/groups/mqb/mqbc/mqbc_clusterdata.h
kaikulimu: Yeah they cannot be `NULL` here. I will add some asserts in the constructor, and change from pointer to reference here.

src/groups/mqb/mqbc/mqbc_storagemanager.h
```diff
@@ -768,6 +806,11 @@ class StorageManager
     /// THREAD: Executed by the cluster's dispatcher thread.
     virtual void stop() BSLS_KEYWORD_OVERRIDE;
 
+    /// Initialize the queue key info map based on information in the specified
+    /// `clusterState`.
+    virtual void initializeQueueKeyInfoMap(
+        const mqbc::ClusterState* clusterState) BSLS_KEYWORD_OVERRIDE;
```
678098, line 812: `clusterState` is a const parameter, it's not optional and required every time. Is it possible to convert it to const reference?

src/groups/mqb/mqbmock/mqbmock_storagemanager.cpp
```diff
@@ -0,0 +1,304 @@
...
+    return 0;
+}
+
+void StorageManager::stop()
+{
+    // NOTHING
+}
+
+void StorageManager::initializeQueueKeyInfoMap(
+    BSLS_ANNOTATION_UNUSED const mqbc::ClusterState* clusterState)
```
678098, line 55: Alternatively you might remove the annotation and arg variable name to silence the unused variable compiler warning

src/groups/mqb/mqbblp/mqbblp_recoverymanager.cpp
```diff
@@ -1722,14 +1722,15 @@ int RecoveryManager::replayPartition(
...
-                                                        *d_clusterData_p,
-                                                        journalIt);
+        rc = mqbc::RecoveryUtil::incrementCurrentSeqNum(
+            &currentSeqNum,
+            &journalRecordBase,
+            fti->journalFd(),
+            toSequenceNum,
+            pid,
+            *destination,
+            d_clusterData_p->identity().description(),
```
pniedzielski, line 1732: Just making sure I haven't missed anything in this commit---do any changes here add thread-safety, or are they only additional checks/documentation + changing the clusterData pointer to a const ref?

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
```diff
@@ -877,14 +883,16 @@ int RecoveryManager::createRecoveryFileSet(bsl::ostream&    errorDescription,
                                            mqbs::FileStore* fs,
                                            int              partitionId)
 {
+    // executed by the *QUEUE DISPATCHER* thread associated with 'partitionId'
+
     // PRECONDITIONS
-    BSLS_ASSERT_SAFE(fs);
+    BSLS_ASSERT_SAFE(fs && fs->inDispatcherThread());
```
pniedzielski, line 889: (in reference to the above comment, this looks like the exception)

src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
```diff
@@ -1510,6 +1510,17 @@ void StorageManager::stop()
...
 
+void StorageManager::initializeQueueKeyInfoMap(
+    BSLS_ANNOTATION_UNUSED const mqbc::ClusterState* clusterState)
+{
+    // executed by cluster *DISPATCHER* thread
+
+    // PRECONDITION
+    BSLS_ASSERT_SAFE(d_dispatcher_p->inDispatcherThread(d_cluster_p));
+
+    BSLS_ASSERT_OPT(false && "This method should only be invoked in FSM mode");
```
pniedzielski, line 1535: Is this really what we want?  This pair of asserts seems wrong...the second one can fire even if we are in the dispatcher thread?

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -3448,6 +3414,50 @@ void StorageManager::stop()
...
+void StorageManager::initializeQueueKeyInfoMap(const mqbc::ClusterState* clusterState)
+{
+    // executed by the *CLUSTER DISPATCHER* thread
+
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(d_dispatcher_p->inDispatcherThread(d_cluster_p));
+    BSLS_ASSERT_SAFE(clusterState);
+
+    if (d_isQueueKeyInfoMapVecInitialized) {
+        // The queue key info map vec should only be initialized once.
```
pniedzielski, line 3495: Does it make sense to log here?  Or do we want the semantics of this to be silently idempotent?  I see a comment later on that indicates this must be initialized only once, so it seems like it'd be good to know if we accidentally do it twice.

An `ASSERT` would feel heavy-handed, probably, but maybe some indication would be good.

src/groups/mqb/mqbi/mqbi_storagemanager.h
```diff
@@ -217,6 +221,10 @@ class StorageManager : public mqbi::AppKeyGenerator {
     /// Stop this storage manager.
     virtual void stop() = 0;
 
+    /// Initialize the queue key info map based on information in the specified
+    /// `clusterState`.
```
pniedzielski, line 224: With regard to the comment above, I think adding "The behavior of this function is undefined unless the queue key info map has not yet been initialized."  Doesn't mean we need to `ASSERT` inside, logging or even doing nothing is sufficient.

src/groups/mqb/mqbc/mqbc_storagemanager.t.cpp
```diff
@@ -201,6 +201,12 @@ struct TestHelper {
         BSLS_ASSERT_OPT(rc == 0);
     }
 
+    void setPartitionPrimary(mqbc::StorageManager* storageManager, int partitionId, unsigned int leaseId, mqbnet::ClusterNode* node)
+    {
+        d_cluster_mp->_state().setPartitionPrimary(partitionId, leaseId, node);
```
pniedzielski, line 209: 👍🏻 nice

src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
```diff
@@ -742,10 +745,15 @@ void StorageManager::processPartitionSyncEvent(
         return;  // RETURN
     }
 
+    PartitionInfo                    pinfo;
```
pniedzielski, line 748: Could you explain this change to me?  Is the worry that `clusterState` will be mutated?  And if so, could it be mutated while we're constructing this `PartitionInfo`?

src/groups/mqb/mqbc/mqbc_clusterdata.h
kaikulimu: d

src/groups/mqb/mqbc/mqbc_clusterdata.h
kaikulimu: The issue is that `mqbc::ClusterData` itself only has pointers to components such as `bdlmt::EventScheduler` and `bdlbb::BlobBufferFactory` (see [here](https://github.com/kaikulimu/blazingmq/blob/f3d03d74c4d6bb335080b1d4120d58bcaebe5fc3/src/groups/mqb/mqbc/mqbc_clusterdata.h#L155-L159)). Take the example of `bdlbb::BlobBufferFactory`, the actual owner is [`mqba::Application::d_bufferFactory`](https://github.com/bloomberg/blazingmq/blob/b042ca1fede898611beb53b6979b33c51a273644/src/groups/mqb/mqba/mqba_application.h#L135), and it's passed as pointer from `mqba::Application -> mqbblp::ClusterCatalog -> mqbblp::Cluster -> mqbc::ClusterData`. Given that, would it make sense to keep them as pointers?

src/groups/mqb/mqbc/mqbc_storagemanager.h
kaikulimu: Agreed

kaikulimu: d

src/groups/mqb/mqbmock/mqbmock_storagemanager.cpp
kaikulimu: Will keep it as `BSLS_ANNOTATION_UNUSED` to be consistent with other functions.

kaikulimu: d

src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
kaikulimu: The only caller of `initializeQueueKeyInfoMap()` is [`mqbc::ClusterStateManager`](https://github.com/kaikulimu/blazingmq/blob/f3d03d74c4d6bb335080b1d4120d58bcaebe5fc3/src/groups/mqb/mqbc/mqbc_clusterstatemanager.cpp#L271), and it should always use `mqbc` version of `StorageManager`. By design, this method should only be invoked in FSM mode. If you can think of a better way to express this, please let me know.

src/groups/mqb/mqbmock/mqbmock_storagemanager.cpp
```diff
@@ -0,0 +1,309 @@
...
+{
+}
+
+void StorageManager::registerQueue(
+    BSLS_ANNOTATION_UNUSED const bmqt::Uri& uri,
```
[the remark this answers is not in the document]
kaikulimu: I tried to change the alignment, then `git clang-format` changes it back to this. So this is what `clang format` wants :)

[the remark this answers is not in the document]
kaikulimu: d

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: By FSM machinery, `initializeQueueKeyInfoMap()` is called as the last step of transitioning from [`Healing Follower` to `Healed Follower`](https://github.com/kaikulimu/blazingmq/blob/f3d03d74c4d6bb335080b1d4120d58bcaebe5fc3/src/groups/mqb/mqbc/mqbc_clusterstatetable.h#L402-L405), or from [`Healing Primary` to `Healed Primary`](https://github.com/kaikulimu/blazingmq/blob/f3d03d74c4d6bb335080b1d4120d58bcaebe5fc3/src/groups/mqb/mqbc/mqbc_clusterstatetable.h#L519-L522). Normally, this method should only be called once. However, if there are fault scenarios where a node can go back-and-forth between healing and healed, then this method can be called more than once. I guess in that uncommon scenario, it's still a good thing to add a log. Will add.

kaikulimu: d

src/groups/mqb/mqbi/mqbi_storagemanager.h
kaikulimu: Per above, there are legal ways to call it more than once. Therefore, I plan to update the comment to say `Note that this method should only be called once; subsequent calls will be ignored.`

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -5224,7 +5224,7 @@ int FileStore::open(const QueueKeyInfoMap& queueKeyInfoMap)
         BALL_LOG_ERROR << partitionDesc() << "Failed to open in recovery mode,"
                        << " rc:" << rc << ", reason: ["
                        << errorDescription.str() << "].";
-        return rc_RECOVERY_MODE_FAILURE;  // RETURN
+        return rc * 10 + rc_RECOVERY_MODE_FAILURE;  // RETURN
```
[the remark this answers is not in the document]
kaikulimu: This `rc * 10` or `rc * 100` hack in used all over the place in the BlazingMQ codebase:
```
mqb (fsm-test__20240722__thread_safety)$ git grep -l "rc \* 10"
mqba/mqba_application.cpp
mqba/mqba_sessionnegotiator.cpp
mqbblp/mqbblp_cluster.cpp
mqbblp/mqbblp_clustercatalog.cpp
mqbblp/mqbblp_clusterorchestrator.cpp
mqbblp/mqbblp_clusterstatemanager.cpp
mqbblp/mqbblp_domain.cpp
mqbblp/mqbblp_storagemanager.cpp
mqbc/mqbc_clusterstateledgerutil.cpp
mqbc/mqbc_clusterutil.cpp
mqbc/mqbc_incoreclusterstateledger.cpp
mqbc/mqbc_incoreclusterstateledgeriterator.cpp
mqbc/mqbc_recoverymanager.cpp
mqbc/mqbc_recoveryutil.cpp
mqbc/mqbc_storagemanager.cpp
mqbc/mqbc_storageutil.cpp
mqbnet/mqbnet_transportmanager.cpp
mqbs/mqbs_filestore.cpp
mqbs/mqbs_filesystemutil.cpp
mqbsl/mqbsl_ledger.cpp
```

It is by convention that child calls always return non-positive values. But I do agree that there is a flaw, and the fix might be beyond the scope of this PR. We should create a JIRA card for this, and come up with a solution for the entire codebase. What do you think?

src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
kaikulimu: There is another call to `mqbc::StorageUtil::validatePartitionSyncEvent()` from [`mqbc::StorageManager`](https://github.com/kaikulimu/blazingmq/blob/f3d03d74c4d6bb335080b1d4120d58bcaebe5fc3/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L2755-L2763). That call is from the *QUEUE DISPATCHER* thread, and thus is not safe to access `clusterState`. Hence, I changed the signature of `validatePartitionSyncEvent()` to take in a `PartitionInfo` instead of a `ClusterState`.

The change you see here is simply a bi-product. There is no worry for mutation.

src/groups/mqb/mqbc/mqbc_clusterdata.h
kaikulimu: Will also make copy constructor and copy assignment operator private, since this object is not supposed to be copyable.

src/groups/mqb/mqbs/mqbs_filestore.cpp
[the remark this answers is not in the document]
pniedzielski: Yes, let's track this somewhere (internal + GitHub issue?), and come back to it later.  It's not worth blocking this PR.

src/groups/mqb/mqbi/mqbi_storagemanager.h
pniedzielski: ^ this makes sense to me.

src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
pniedzielski: lgtm then

src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
pniedzielski: I see.  In this case, the `false` makes sense.  Perhaps, "Only the FSM version of this method from `mqbc::StorageManager` should be invoked." as the string?

kaikulimu: Agreed. Will change

src/groups/mqb/mqbs/mqbs_filestore.cpp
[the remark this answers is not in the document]
kaikulimu: Will create a GitHub issue too.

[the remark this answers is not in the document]
kaikulimu: https://github.com/bloomberg/blazingmq/issues/403 is created to track this

src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
kaikulimu: d

kaikulimu:
@678098 @hallfox @pniedzielski Back for your review

pniedzielski (approved):
Changes look good to me!

PR #370, opened 2024-07-23
Refactor[BMQT]: turn copy operator for Subscription explicit

678098:


dorjesinpo (approved):
Thank you for fixing this.

src/groups/bmq/bmqt/bmqt_subscription.h
```diff
@@ -203,9 +203,19 @@ class Subscription {
 
     /// Create a new Subscription by copying values from the specified
     /// `other`.
-    Subscription(const Subscription& other);
+    explicit Subscription(const Subscription& other);
```
hallfox, line 206: We shouldn't need this `explicit here`, just the assignment operator

src/groups/bmq/bmqt/bmqt_subscription.h
```diff
@@ -203,9 +203,19 @@ class Subscription {
...
+    /// Assign to this object the value of the specified `rhs` object.
+    Subscription& operator=(const Subscription& rhs);
+
+#if defined(BSLS_COMPILERFEATURES_SUPPORT_RVALUE_REFERENCES) &&               \
+    defined(BSLS_COMPILERFEATURES_SUPPORT_NOEXCEPT)
+    /// Assign to this object the value of the specified `rhs` object.
+    /// After performing this action, the `rhs` object will be left in a
+    /// valid, but unspecified state.
+    Subscription& operator=(Subscription&& rhs);
+#endif
```
hallfox, line 218: We should probably use `bslmf::MovableRef` instead of preprocessor guards here

hallfox (changes requested):
See https://bde.bloomberg.com/bde-resources/doxygen/bde_api_prod/classbslmf_1_1MovableRef.html#a1dfc9a623accb2636b2079a4ff406c94 for how we can rewrite this to have rvalue references work on platforms with >=C++11 while also having fake moves on C++03

src/groups/bmq/bmqt/bmqt_subscription.h
678098: My bad

src/groups/bmq/bmqt/bmqt_subscription.h
678098: On the other hand, what do you think about removing explicit copy constructor and assignment at all? It should deal with the warnings, and also we re-implemented the default one anyway. We don't use allocators for these classes, so we don't need to implement allocator-aware assignment.

678098: The proposed change is in the latest commit

hallfox: This would also work, for some reason I thought there was an explicit destructor here that was forcing us into a rule-of-3/rule-of-5 situation

hallfox (approved):

PR #371, opened 2024-07-23
Refactor[BMQT]: make QueueOptions rule-of-three complaint

678098:
Remove 400 lines from build logs

### DIFF
--- a/src/groups/bmq/bmqt/bmqt_queueoptions.h
+++ b/src/groups/bmq/bmqt/bmqt_queueoptions.h
@@ -130,9 +130,6 @@ class QueueOptions {
     /// Assign to this object the value of the specified `rhs` object.
     QueueOptions& operator=(const QueueOptions& rhs);
 
-    /// Assign to this object the value of the specified `rhs` object.
-    QueueOptions& operator=(bslmf::MovableRef<QueueOptions> rhs);
-
     /// Set the maxUnconfirmedMessages to the specified `value`.  The
     /// behavior is undefined unless `value >= 0`. If the specified `value`
     /// is set to 0, it means that the consumer does not receive any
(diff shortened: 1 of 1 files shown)


src/groups/bmq/bmqt/bmqt_queueoptions.h
```diff
@@ -256,6 +262,42 @@ bsl::ostream& operator<<(bsl::ostream& stream, const QueueOptions& rhs);
...
+        d_info                    = rhs.d_info;
+        d_suspendsOnBadHostHealth = rhs.d_suspendsOnBadHostHealth;
+        d_hadSubscriptions        = rhs.d_hadSubscriptions;
+        if (d_allocator_p == rhs.d_allocator_p) {
+            d_subscriptions = rhs.d_subscriptions;
+            // No need to reassign `d_allocator_p`
+        }
+        else {
+            d_subscriptions = Subscriptions(rhs.d_subscriptions,
+                                            rhs.d_allocator_p);
```
678098, line 278: https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/group__bslstl__unorderedmap.html#ga010b0fbb7737d2e3a902daacda105193

From the `bsl::unordered_map` docs:
```
propagate to this object the allocator of rhs if the ALLOCATOR type has trait propagate_on_container_copy_assignment
```

The usual assignment operator might propagate the allocator and might not. Here, I make sure to construct the unordered map with the allocator from `rhs`.

hallfox: `bsl::allocator`, which the map uses, doesn't allow this propagation: https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/group__bslma__stdallocator.html

I also don't think we should be `rhs`'s allocator in this case, if allocator propagation were to happen on copy construction the class would handle it itself. Regardless I don't think we want to grab `rhs`'s allocator this way. I think we want to mimic the constructor's behavior here, which is to optionally supply the allocator we want or use the default allocator. From the properties of `[UsesBslmaAllocator` trait docs](https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/group__bslma__usesbslmaallocator.html):

> -  The allocator used by an object is not changed after construction (e.g., the assignment operator does not change the allocator used by an object). 

So what I think we really want here is to use `this`'s allocator to do each of these allocation operations.

hallfox (approved):
lgtm

PR #372, opened 2024-07-23
Refactor[MWC]: explicit operator= in mwcc containers

678098:
Removing the custom copy constructor `OrderedHashMap_SequentialIterator(const NcIter& other);` doesn't work since it produces compilation errors
```
In file included from /blazingmq/src/groups/bmq/bmqimp/bmqimp_brokersession.cpp:17:
In file included from /blazingmq/src/groups/bmq/bmqimp/bmqimp_brokersession.h:37:
In file included from /blazingmq/src/groups/bmq/bmqimp/bmqimp_messagecorrelationidcontainer.h:39:
/blazingmq/src/groups/bmq/bmqp/bmqp_requestmanager.h:1509:30: error: no viable conversion from 'OrderedHashMap_SequentialIterator<pair<...>>' to 'OrderedHashMap_SequentialIterator<const pair<...>>'
    for (RequestMapConstIter it = requestsCopy.begin();
                             ^    ~~~~~~~~~~~~~~~~~~~~
```

Implemented explicit `operator=` for `OrderedHashMap_SequentialIterator` and `OrderedHashMapWithHistory_Iterator`

### DIFF
--- a/src/groups/mwc/mwcc/mwcc_orderedhashmap.h
+++ b/src/groups/mwc/mwcc/mwcc_orderedhashmap.h
@@ -297,7 +297,7 @@ class OrderedHashMap_SequentialIterator {
     // MANIPULATORS
 
     /// Assign to this object the value of the specified `rhs` object.
-    OrderedHashMap_SequentialIterator& operator=(const NcIter&);
+    OrderedHashMap_SequentialIterator& operator=(const NcIter& rhs);
 
     /// Advance this iterator to the next element in the sequential list and
     /// return its new value.  The behavior is undefined unless this
--- a/src/groups/mwc/mwcc/mwcc_orderedhashmap.h
+++ b/src/groups/mwc/mwcc/mwcc_orderedhashmap.h
@@ -296,6 +296,9 @@ class OrderedHashMap_SequentialIterator {
 
     // MANIPULATORS
 
+    /// Assign to this object the value of the specified `rhs` object.
+    OrderedHashMap_SequentialIterator& operator=(const NcIter&);
+
     /// Advance this iterator to the next element in the sequential list and
     /// return its new value.  The behavior is undefined unless this
     /// iterator is in the range `[begin() .. end())` (i.e., the iterator is
@@ -885,6 +888,15 @@ inline OrderedHashMap_SequentialIterator<
 }
 
 // MANIPULATORS
+
+template <class VALUE>
+inline OrderedHashMap_SequentialIterator<VALUE>&
+OrderedHashMap_SequentialIterator<VALUE>::operator=(const NcIter& rhs)
+{
+    d_link_p = rhs.d_link_p;
+    return *this;
+}
+
 template <class VALUE>
 inline OrderedHashMap_SequentialIterator<VALUE>&
 OrderedHashMap_SequentialIterator<VALUE>::operator++()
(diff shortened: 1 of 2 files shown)


ddragan-bloomberg:
From what I gathered the `OrderedHashMap_SequentialIterator& operator=(const NcIter&)` definition exists so that a const iterator is constructible from a non-const itertator, but the same definition also might act as the copy constructor. Took me a while to figure this out. I think this is convoluted. A better way would be to:
- not define the copy constructor, not the copy operator;
- define a special from-const-iterator constructor.

678098:
> * not define the copy constructor, not the copy operator;
> * define a special from-const-iterator constructor.

If I remove the original `OrderedHashMapWithHistory_Iterator(const NcIter& other);` copy constructor, I will get compilation errors:

```
blazingmq/src/groups/mwc/mwcc/mwcc_orderedhashmapwithhistory.h:698:12: error: no matching conversion for functional-style cast from 'const iterator' (aka 'const OrderedHashMapWithHistory_Iterator<Value>') to 'const_iterator' (aka 'OrderedHashMapWithHistory_Iterator<const Value>')
    return const_iterator(d_first);
```

These errors will remain even if I define from-const-iterator constructor
```
typedef typename bsl::add_const<VALUE>::type      CType;
typedef OrderedHashMapWithHistory_Iterator<CType> CIter;
OrderedHashMapWithHistory_Iterator(const CIter& other);
```

ddragan-bloomberg:
I just realized that the assignment operator is also double-purpose. It allows to assign a non-const iterator to a const one, and services as the copy-assignment operator. So it should be explicitly declared and implemented.

-- this code is fine

src/groups/mwc/mwcc/mwcc_orderedhashmap.h
```diff
@@ -296,6 +296,9 @@ class OrderedHashMap_SequentialIterator {
 
     // MANIPULATORS
 
+    /// Assign to this object the value of the specified `rhs` object.
+    OrderedHashMap_SequentialIterator& operator=(const NcIter&);
```
pniedzielski: Nit: Because of the documentation comment, I think we want an `rhs` name here (like in the other commit in this PR):
`OrderedHashMap_SequentialIterator& operator=(const NcIter& rhs);`

src/groups/mwc/mwcc/mwcc_orderedhashmap.h
```diff
@@ -885,6 +888,15 @@ inline OrderedHashMap_SequentialIterator<
 }
 
 // MANIPULATORS
+
+template <class VALUE>
+inline OrderedHashMap_SequentialIterator<VALUE>&
+OrderedHashMap_SequentialIterator<VALUE>::operator=(const NcIter& rhs)
+{
+    d_link_p = rhs.d_link_p;
```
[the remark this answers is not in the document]
pniedzielski: This makes sense, this is such a cheap object that we shouldn't worry about self-assignment.

pniedzielski (approved):
Small nit, if you want to fix, go ahead and I'll reapprove; otherwise looks good.

src/groups/mwc/mwcc/mwcc_orderedhashmap.h
678098: Oh, I already force-pushed a fix for this. Might be GH lagging

pniedzielski: Huh, you're right!  Looks good, go ahead and merge.

PR #373, opened 2024-07-24
Fix sanitizer issues [BMQ,MWC]

alexander-e1off:
Revise cancelled/suppressed unit tests running under sanitizers. Fix some issues.

PR #374, opened 2024-07-24
Remove AIX support

hallfox:
We are dropping AIX support for BlazingMQ internally. As a result, it may be healthy to do a cleanup of the code added to specifically support this platform, as we will have no good way of testing these code paths.

### DIFF
--- a/src/groups/mqb/mqbu/mqbu_messageguidutil.t.cpp
+++ b/src/groups/mqb/mqbu/mqbu_messageguidutil.t.cpp
@@ -182,11 +182,7 @@ static void test2_multithread()
 
     const int k_NUM_THREADS = 10;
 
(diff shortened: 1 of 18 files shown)


src/groups/mqb/mqbu/mqbu_messageguidutil.t.cpp
```diff
@@ -318,9 +314,7 @@ static void test4_defaultHashUniqueness()
 
     mwctst::TestHelper::printTestName("DEFAULT HASH UNIQUENESS");
 
-#ifdef BSLS_PLATFORM_OS_AIX
```
678098, line 321: @alexander-e1off for awareness, might need to update your PR
https://github.com/bloomberg/blazingmq/pull/373/files

678098 (approved):

PR #378, opened 2024-07-26
UT[MWC]: C++20: Fix -Wvolatile warnings

hallfox:
C++20 has completely deprecated `volatile`, and GCC enables multiple warnings to catch invalid volatile uses.


pniedzielski (approved):
👏🏻 👏🏻 👏🏻 

(nit: only deprecated in the places it's broken, like thread-safety, not in places it has no replacement, like hardware register access: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1152r4.html and previous versions)

PR #379, opened 2024-07-26
Fix deprecated usage of C-style enums

hallfox:
Arithmetic expressions using C-style enums trigger a warning with -Wdeprecated-enum-float-conversion.


chrisbeard (approved):
👍

PR #383, opened 2024-07-31
[Do not merge]Feat[BMQTOOL]: auto adjust post rate in auto mode

678098:
We might want to automatically measure accurate throughput just before we hit the `LIMIT_MESSAGES` error. Previously, we had to manually run `bmqtool` with different settings to find the throughput, and this process was slow.

This branch is WIP to test this feature


678098:
Closing for now

PR #384, opened 2024-07-31
Feat[MQB, MWC]: report the number of tcp connections

waldgange:


678098 (changes requested):

src/groups/mqb/mqbstat/mqbstat_statcontroller.cpp
```diff
@@ -81,6 +82,20 @@ const bsls::Types::Int64 k_NS_PER_MESSAGE = 15 *
...
 const char k_PUBLISHINTERVAL_SUFFIX[] = ".PUBLISHINTERVAL";
 
+void portsDeleter(
+    bsl::unordered_map<bsl::string, bslma::ManagedPtr<mwcst::StatContext> >*
+                              map,
+    bslmt::Mutex*             mutex,
+    const mwcst::StatContext& context)
+{
+    // Lookup the port's StatContext and remove it from the 'map'
+    bslmt::LockGuard<bslmt::Mutex> guard(mutex);  // LOCK
```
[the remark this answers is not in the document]
waldgange: Unfortunately, the threadsafety of `StatContext` class is only partial.
https://github.com/bloomberg/blazingmq/blob/7d4b3d7caa075e5cca89995008e6fad30082b3f2/src/groups/mwc/mwcst/mwcst_statcontext.h#L96-L98
For example, I can't iterate over subcontexts that can be added/deleted without additional synchronisation anywhere except the `snapshot()` method or its synchronious pre- or post- steps, as handling the deleted subcontexts happens inside this method. Please find below some arguments in favor of the current imlementation:
Each channel stores its own `StatContext`, and this approach hasn't changed. What I have done is another 'layer' of `StatContext`s between channel contexts and their parent. This layer serves for breaking down the channels stats by ports. I really need to have only one `StatContext` for each port as I report aggregated stats. The inter-thread synchronisation is required here, because:
1. `portsDeleter()` is always called from `StatController`'s thread, but `TCPSessionFactory::channelStatContextCreator` and, hence,  constructors and destructors of channel-related `StatContext`s are called from `TCPSessionFactory`'s threads.
2. Although 'StatContext::addSubcontext()' is a thread-safe method, it will create a new `StatContext` instance even if there are sub-contexts with the same name.
So, in order to add a **channel** StatContext, I need to check if there is a **port** StatContext for that port and use it, otherways I have to create a new **port** `Statcontext`. After that I add a **channel** sub context to it. So I must have a **ManagedPtr**'s to the `StatContext`s of the **ports** and to lock a container (currently, the unordered_map), where they are stored. Otherways race conditions are possible, either with two insertions of the same name **port** `StatContext`s, or deletion during the adding a new **channel** subcontext.

If you have an idea how to reach it without the synchronization, please let me know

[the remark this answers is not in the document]
678098: > If you have an idea how to reach it without the synchronization, please let me know
@waldgange let me think about this for some time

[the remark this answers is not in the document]
678098: Hi @waldgange
I think the correct way to do it is to pass `ManagedPtr` to the `StatChannel` here, instead of `shared_ptr`:
https://github.com/bloomberg/blazingmq/blob/a6a6cefbb9b8cab1dc0b2bc0b1ee7c9bf3f8b05c/src/groups/mwc/mwcio/mwcio_statchannelfactory.cpp#L126-L133

So, `StatChannel` keeps ownership of this `StatContext`. When a `StatChannel` goes down, the `ManagedPtr` to its `StatContext` goes down too.

Note that it requires to change `MWC` interfaces. However, `StatChannel` and `StatChannelFactory` are used only within `BlazingMQ`, so I don't see any harm in this. Also, since we plan to own `ManagedPtr` in `StatChannel`, the `StatChannel` class is not copy constructible anymore, so we might explicitly delete copy constructor and assignment

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -1414,5 +1409,34 @@ bool TCPSessionFactory::isEndpointLoopback(const bslstl::StringRef& uri) const
...
+    const char                  pattern[] = ":(\\d{1,5})$";
+    bsl::string                 error(allocator);
+    size_t                      errorOffset;
+    BSLA_MAYBE_UNUSED const int rc = d_regex.prepare(
+        &error,
+        &errorOffset,
+        pattern,
+        bdlpcre::RegEx::k_FLAG_JIT);
+    BSLS_ASSERT_SAFE(rc == 0);
+    BSLS_ASSERT_SAFE(d_regex.isPrepared() == true);
```
678098, line 1428:
```suggestion
    BSLS_ASSERT_SAFE(d_regex.isPrepared());
```

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -1414,5 +1409,34 @@ bool TCPSessionFactory::isEndpointLoopback(const bslstl::StringRef& uri) const
...
+}
+
+bsl::string_view
+TCPSessionFactory::PortExtractor::extract(const bsl::string& endpoint) const
+{
+    bsl::string_view result;
+
+    BSLS_ASSERT_SAFE(0 == d_regex.match(&result, endpoint));
```
678098, line 1436: Is it possible to pass an incorrect endpoint here and cause the broker to crash on assert? I am thinking if we should crash here or handle incorrect `endpoint`

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -1414,5 +1409,34 @@ bool TCPSessionFactory::isEndpointLoopback(const bslstl::StringRef& uri) const
...
+
+    return result.substr(1);
```
678098, line 1438: Do we make `substr(1)` here because we capture the leading `:`?

src/groups/mwc/mwcio/mwcio_statchannel.cpp
```diff
@@ -81,11 +81,12 @@ StatChannel::StatChannel(const StatChannelConfig& config,
 {
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(config.d_statContext_sp);
+    d_config.d_statContext_sp->adjustValue(Stat::e_CONNECTIONS, 1);
 }
 
 StatChannel::~StatChannel()
 {
-    // NOTHING
+    d_config.d_statContext_sp->adjustValue(Stat::e_CONNECTIONS, -1);
```
678098, line 89: If we do per-channel stat contexts, then the number of connections will always be `0` or `1`, right?

src/groups/mwc/mwcst/mwcst_statcontext.h
```diff
@@ -799,6 +799,10 @@ class StatContext {
     /// all sibling subcontexts.
     int uniqueId() const;
 
+    /// Return the number of times this 'StatContext' had 'snapshot' called on
+    /// it.
+    bsls::Types::Int64 numSnapshots() const;
```
678098, line 804: This is useful, but also it's an interface change in `MWC`. Do we want to keep it?

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
waldgange: Yes, definetley. If we used the `match()` overload that puts results to a vector, we would just take the [1] element of the vector, that contains sub expresion of the regex, and the [0] element would contain the full match with `:` in the beginning. However current `match()` overload only returns full matched expression.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
waldgange: It never happens as an endpoint is always a `host:port` string. However, I can remove this assert and just return an empty string in case of not matching endpoint. Then broker will proceed working and all the channels with unknown ports will report their TCP connections numbers metric with an empty tag port

src/groups/mwc/mwcst/mwcst_statcontext.h
waldgange: I need to [check it](https://github.com/bloomberg/blazingmq/pull/384/files#diff-03491c2d54484fc4785eb1b838445ae48b73f16c1b0b491914c21c3e5b1322dfR603-R608) in `portsDeleter()` as before the first snapshot `numSubcontexts()` returns zero, and I don't want to delete new `StatContext`s in this case. It happens because `StatContext::addSubcontext()` puts the newly created contexts to `d_newSubcontexts` vector.
https://github.com/bloomberg/blazingmq/blob/7d4b3d7caa075e5cca89995008e6fad30082b3f2/src/groups/mwc/mwcst/mwcst_statcontext.cpp#L631
Then it moves them to `d_statContext` vector on next `StatContext::snapshot()`
https://github.com/bloomberg/blazingmq/blob/7d4b3d7caa075e5cca89995008e6fad30082b3f2/src/groups/mwc/mwcst/mwcst_statcontext.cpp#L339
However `StatContext::numSubcontexts()` doesn't take into account `d_newSubcontext` and will return 0 if there haven't been any stapshots yet.
https://github.com/bloomberg/blazingmq/blob/7d4b3d7caa075e5cca89995008e6fad30082b3f2/src/groups/mwc/mwcst/mwcst_statcontext.h#L1136-L1140
So if we don't check the number of snapshots in `portsDeleter()` we can unintentionally delete the `StatContext` that has just been added.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.h
```diff
@@ -97,6 +97,7 @@
 // BDE
 #include <bdlbb_blob.h>
 #include <bdlmt_eventscheduler.h>
+#include <bdlpcre_regex.h>
```
678098, line 100: Would ideally avoid exposing this include in the header, since it will go to all other source files.

678098 (commented):
Some thoughts

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -280,36 +282,35 @@ TCPSessionFactory::channelStatContextCreator(
...
-        parent = d_statController_p->channelsStatContext(
-            mqbstat::StatController::ChannelSelector::e_REMOTE);
-    }
+    mqbstat::StatController::ChannelSelector::Enum selector =
+        mwcio::ChannelUtil::isLocalHost(ipAddress)
+            ? mqbstat::StatController::ChannelSelector::e_REMOTE
+            : mqbstat::StatController::ChannelSelector::e_LOCAL;
 
+    mwcst::StatContext* parent = d_statController_p->channelsStatContext(
+        selector);
```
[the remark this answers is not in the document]
waldgange: You are right, it looks confusing, however I tried to keep the existing logic. Though your question motivated me to check the old repo's history. And I've figured out that it was the opposite before switching from BTE to NTC library. Seems like the expression in the braces has been broken unintentionally. So I will get it back as it was before.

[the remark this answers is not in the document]
678098: @waldgange thanks!

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -1414,5 +1422,73 @@ bool TCPSessionFactory::isEndpointLoopback(const bslstl::StringRef& uri) const
...
+    BSLS_ASSERT_SAFE(d_regex.isPrepared());
+}
+
+bslma::ManagedPtr<mwcst::StatContext>
+TCPSessionFactory::PortManager::addChannelContext(mwcst::StatContext* parent,
+                                                  const bsl::string&  endpoint,
+                                                  const bsl::string&  port)
+{
+    bdlma::LocalSequentialAllocator<2048> localAllocator(d_allocator_p);
+    mwcst::StatContextConfiguration statConfig(endpoint, &localAllocator);
```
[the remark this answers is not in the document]
waldgange: Unfortunately `statConfig` is not the same as it's initialized with different `enpoint` every time. This endpoint is also the name of the StatContext being created below. Hence, it can't be cached. I see two possible solutions in order to move preparation steps out from the lock:

1. To pass a pointer to the `TCPSessionFactory`'s mutex as a parameter to this method (or may be to the `PortManager`'s constructor, saving it as a class member) and lock it a bit later.
2. To add another mutex to this class and use it instead of the `TCPSessionFactory`'s mutex. In this case we can avoid locking the external mutex in both `addChannelContext()`  and `deleteChannelContext` methods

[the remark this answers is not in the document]
678098: Let's skip it for now

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -1414,5 +1422,73 @@ bool TCPSessionFactory::isEndpointLoopback(const bslstl::StringRef& uri) const
...
+    if (portIt != d_portMap.end()) {
+        channelStatContext = portIt->second.d_portContext->addSubcontext(
+            statConfig);
+        ++portIt->second.d_numChannels;
+    }
+    else {
+        mwcst::StatContextConfiguration     portConfig(port, &localAllocator);
+        bsl::shared_ptr<mwcst::StatContext> portStatContext =
+            parent->addSubcontext(
+                portConfig.storeExpiredSubcontextValues(true));
```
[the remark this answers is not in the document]
waldgange: Seems like this is a common practice:
https://github.com/bloomberg/blazingmq/blob/5aaf5966b9f192cb410223bda43092b5a69b7855/src/groups/mwc/mwcio/mwcio_statchannelfactory.cpp#L126-L127
https://github.com/bloomberg/blazingmq/blob/5aaf5966b9f192cb410223bda43092b5a69b7855/src/groups/mqb/mqbstat/mqbstat_clusterstats.cpp#L254-L258
https://github.com/bloomberg/blazingmq/blob/5aaf5966b9f192cb410223bda43092b5a69b7855/src/groups/mqb/mqbblp/mqbblp_cluster.cpp#L2582-L2584
and so on

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.h
waldgange: Ok, I can move this class (or its port extracting functionality as a separate class) to the source file. And create a static instance there.

waldgange: agreed to leave it as it is

678098 (approved):
LGTM!

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -1462,5 +1473,56 @@ bool TCPSessionFactory::isEndpointLoopback(const bslstl::StringRef& uri) const
...
+        channelStatContext = portStatContext->addSubcontext(statConfig);
+        d_portMap.emplace(port, PortContext({portStatContext, 1}));
+    }
+
+    return channelStatContext;
+}
+
+void TCPSessionFactory::PortManager::onDeleteChannelContext(
+    const bsl::uint16_t port)
```
678098, line 1518: For basic numeric types we don't typically specify const in arguments list

src/groups/mwc/mwcio/mwcio_ntcchannel.cpp
```diff
@@ -1392,6 +1392,16 @@ ntsa::Endpoint NtcChannel::peerEndpoint() const
     }
 }
 
+ntsa::Endpoint NtcChannel::sourceEndpoint() const
+{
+    if (d_streamSocket_sp) {
+        return d_streamSocket_sp->sourceEndpoint();
+    }
+    else {
+        return ntsa::Endpoint();
```
678098, line 1401: Might replace with ternary operator

678098 (dismissed):

678098 (approved):

PR #385, opened 2024-07-31
Config: add benchmark tools to broker run script

678098:


src/applications/bmqbrkr/etc/bmqbrkrcfg.json
```diff
@@ -1,6 +1,6 @@
 {
     "taskConfig": {
-        "allocatorType": "STACKTRACETEST",
```
678098, line 3: STACKTRACETEST allocator is for debug only, it is slow. For better user experience, we should provide COUNTING allocator as a default

PR #386, opened 2024-08-01
Feat[MQB]: count plugins allocations

678098:


678098:
Not ready yet, want to extend

PR #389, opened 2024-08-05
Performance[MQBSTAT]: lookup for per-appId metrics O(n) -> O(1)

678098:
Previously, we used a `bsl::list<ManagedPtr<mwcst::StatContext> >` for 2 roles:
- Store managed pointers to subcontexts associated with appIds
- Lookup to subcontexts for each specific appId

The problem is that the `list` assumes linear search, which might be too slow if we have a decent number of appIds.

However, due to the interfaces of `ManagedPtr` (on `Solaris`), we cannot just use a container which might reallocate its elements on resize. So instead we keep&update 2 collections together: `bsl::list` to hold `ManagedPointer`s and `bsl::unordered_map` for fast lookup to raw pointers to the needed `StatContext`s

Full list of changes:
- `mqbstat::QueueStatsDomain`: hold managed pointers in `bsl::list<StatSubContextMp> d_subContextsHolder`, lookup to subcontexts with `bsl::unordered_map<bsl::string, mwcst::StatContext*> d_subContextsLookup`
- `mqbstat::QueueStatsDomain`: pass allocator in constructor
- `mqbstat::QueueStatsDomain`: improve logging when no subcontext found
- `mqbc::ClusterDataIdentity`: add safety check to prevent possible `nullptr` dereference
- `mqbmock::Cluster`: add safety checks to find cluster misconfiguration early
- `mqbstat_queuestats.t.cpp`: add UT to check per-appId subcontexts and metrics

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_clusterdata.cpp
+++ b/src/groups/mqb/mqbc/mqbc_clusterdata.cpp
@@ -53,6 +53,8 @@ mqbc::ClusterDataIdentity clusterIdentity(const bslstl::StringRef& name,
     // Create client identity
     bmqp_ctrlmsg::ClientIdentity identity(allocator);
     if (!isRemote) {
+        BSLS_ASSERT_SAFE(netCluster->selfNode());
+
         identity.protocolVersion() = bmqp::Protocol::k_VERSION;
         identity.sdkVersion()      = bmqscm::Version::versionAsInt();
         identity.clientType()      = bmqp_ctrlmsg::ClientType::E_TCPBROKER;
--- a/src/groups/mqb/mqbmock/mqbmock_cluster.cpp
+++ b/src/groups/mqb/mqbmock/mqbmock_cluster.cpp
@@ -156,6 +156,10 @@ void Cluster::_initializeNetcluster()
                                       : k_LEADER_NODE_ID + 1;
     dynamic_cast<mqbnet::MockCluster*>(d_netCluster_mp.get())
         ->_setSelfNodeId(selfNodeId);
+
+    if (d_isClusterMember) {
+        BSLS_ASSERT_OPT(0 != d_netCluster_mp->selfNode());
+    }
 }
 
 void Cluster::_initializeNodeSessions()
@@ -231,7 +235,12 @@ Cluster::Cluster(bdlbb::BlobBufferFactory* bufferFactory,
 , d_processor()
 {
     // PRECONDITIONS
-    BSLS_ASSERT_OPT(isClusterMember || !isLeader);
+    if (isClusterMember) {
+        BSLS_ASSERT_OPT(!clusterNodeDefs.empty());
+    }
+    else {
+        BSLS_ASSERT_OPT(!isLeader);
+    }
 
     _initializeClusterDefinition(name,
                                  location,
--- a/src/groups/mqb/mqbstat/mqbstat_queuestats.cpp
+++ b/src/groups/mqb/mqbstat/mqbstat_queuestats.cpp
@@ -662,11 +662,9 @@ void QueueStatsDomain::updateDomainAppIds(
         return;  // RETURN
     }
 
-    bsl::unordered_set<bsl::string> remainingAppIds(
-        appIds.begin(),
-        appIds.end(),
-        appIds.size() * 3,  // initial bucket num
-        d_allocator_p);
+    bsl::unordered_set<bsl::string> remainingAppIds(appIds.begin(),
+                                                    appIds.end(),
+                                                    d_allocator_p);
 
     // 1. Remove subcontexts for unneeded appIds
     bsl::list<StatSubContextMp>::iterator it = d_subContextsHolder.begin();
--- a/src/groups/mqb/mqbstat/mqbstat_queuestats.cpp
+++ b/src/groups/mqb/mqbstat/mqbstat_queuestats.cpp
@@ -193,29 +193,6 @@ bool filterDirect(const mwcst::TableRecords::Record& record)
     return record.type() == mwcst::StatContext::e_TOTAL_VALUE;
 }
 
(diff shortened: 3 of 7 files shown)


src/groups/mqb/mqbc/mqbc_clusterdata.cpp
```diff
@@ -53,6 +53,8 @@ mqbc::ClusterDataIdentity clusterIdentity(const bslstl::StringRef& name,
     // Create client identity
     bmqp_ctrlmsg::ClientIdentity identity(allocator);
     if (!isRemote) {
+        BSLS_ASSERT_SAFE(netCluster->selfNode());
```
678098, line 56: On the line 65 below, `netCluster->selfNode()` is dereferenced without checks, which might lead to segfault. This might be achieved by using mock domain/cluster objects.

src/groups/mqb/mqbmock/mqbmock_cluster.cpp
```diff
@@ -156,6 +156,10 @@ void Cluster::_initializeNetcluster()
                                       : k_LEADER_NODE_ID + 1;
     dynamic_cast<mqbnet::MockCluster*>(d_netCluster_mp.get())
         ->_setSelfNodeId(selfNodeId);
+
+    if (d_isClusterMember) {
+        BSLS_ASSERT_OPT(0 != d_netCluster_mp->selfNode());
```
678098, line 161: Another early safety check for the same problem with dereferencing `selfNode` with mocks.

src/groups/mqb/mqbmock/mqbmock_cluster.cpp
```diff
@@ -231,7 +235,12 @@ Cluster::Cluster(bdlbb::BlobBufferFactory* bufferFactory,
 , d_processor()
 {
     // PRECONDITIONS
-    BSLS_ASSERT_OPT(isClusterMember || !isLeader);
+    if (isClusterMember) {
+        BSLS_ASSERT_OPT(!clusterNodeDefs.empty());
```
678098, line 239: Early safety check. Providing an empty `clusterNodeDefs` make it certain that `selfNode` will be nullptr.

src/groups/mqb/mqbstat/mqbstat_queuestats.cpp
```diff
@@ -609,25 +585,27 @@ void QueueStatsDomain::onEvent(EventType::Enum    type,
 
     BALL_LOG_SET_CATEGORY(k_LOG_CATEGORY);
 
-    if (!d_subContexts_mp) {
+    if (d_subContextsLookup.empty()) {
         BALL_LOGTHROTTLE_WARN(k_MAX_INSTANT_MESSAGES, k_NS_PER_MESSAGE)
-            << "[THROTTLED] No built sub contexts";
+            << "[THROTTLED] No built sub contexts for domain: "
+            << d_statContext_mp->name() << ", appId: " << appId;
```
678098, line 591: New log line looks like
`06AUG2024_02:21:59.442 10822:8281000960 WARN /Users/emalygin/work/blazingmq/src/groups/mqb/mqbstat/mqbstat_queuestats.cpp:602 MQBSTAT.QUEUESTATS [THROTTLED] No matching StatContext for domain: bmq://mock-domain/abc, appId: bar`

kaikulimu:
> However, due to the interfaces of `ManagedPtr` (on `Solaris`), we cannot just use a container which might reallocate its elements on resize.

Do you have source to this statement? It is quite inconvenient to have to use two data structures.

src/groups/mqb/mqbstat/mqbstat_queuestats.cpp
```diff
@@ -678,38 +656,56 @@ void QueueStatsDomain::setQueueContentRaw(bsls::Types::Int64 messages,
...
-    // Remove subcontexts if appIds are not present in updated AppIds
-    bsl::list<StatSubContextMp>::iterator it = d_subContexts_mp->begin();
-    while (it != d_subContexts_mp->end()) {
-        if (bsl::find(appIds.begin(), appIds.end(), it->get()->name()) ==
-            appIds.end()) {
-            it = d_subContexts_mp->erase(it);
+    bsl::unordered_set<bsl::string> remainingAppIds(
+        appIds.begin(),
+        appIds.end(),
+        appIds.size() * 3,  // initial bucket num
```
kaikulimu, line 668: Is hardcoding to `appIds.size() * 3` better than the default implementation-defined bucket num?

src/groups/mqb/mqbstat/mqbstat_queuestats.t.cpp
```diff
@@ -490,6 +495,171 @@ static void test4_queueStatsDomainContent()
...
+        stats.onEvent(mqbstat::QueueStatsDomain::EventType::e_CONFIRM_TIME,
+                      700,
+                      k_APPID_BAR);
+        stats.onEvent(mqbstat::QueueStatsDomain::EventType::e_CONFIRM_TIME,
+                      900,
+                      k_APPID_BAR);
+        stats.onEvent(mqbstat::QueueStatsDomain::EventType::e_CONFIRM_TIME,
+                      800,
+                      k_APPID_BAR);
+
```
kaikulimu, line 638:
```suggestion
```

kaikulimu (commented):
Minor comments

src/groups/mqb/mqbstat/mqbstat_queuestats.cpp
678098: Not a critical path, removed it.

src/groups/mqb/mqbstat/mqbstat_queuestats.t.cpp
678098: The empty line here is intended, it visually separates BAR and BAZ appIds

kaikulimu (approved):
lgtm

PR #400, opened 2024-08-15
Fix subscription md doc formatting

chrisbeard:
The wrapping here was causing an unintentional blockquote:
Screenshot 2024-08-15 at 10 38 37 AM

This is what it looks like after my changes:
Screenshot 2024-08-15 at 10 38 17 AM


pniedzielski (approved):
LGTM!

PR #408, opened 2024-08-26
mqbc::ClusterUtil: Revert to not populate appId infos in non-CSL mode

kaikulimu:
#318 introduces compatibility issues if we run a cluster with brokers of mixing versions 0.92.4 and 0.92.7, because bmqbrkr_0.92.4 asserts that appId infos are not populated in `QueueAssignmentAdvisory`. Therefore, we decide to split into two releases: where 0.92.8 will remove the assert, and 0.92.9 will populate the appId infos. Then, we can run a mix of bmqbrkr_0.92.8 and bmqbrkr_0.92.9 without error.

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_clusterutil.cpp
+++ b/src/groups/mqb/mqbc/mqbc_clusterutil.cpp
@@ -786,7 +786,8 @@ void ClusterUtil::populateQueueAssignmentAdvisory(
     ClusterState*                          clusterState,
     ClusterData*                           clusterData,
     const bmqt::Uri&                       uri,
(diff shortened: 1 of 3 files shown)


dorjesinpo (changes requested):
how about removing that assert?

kaikulimu:
Asserts were removed in https://github.com/bloomberg/blazingmq/pull/318/commits/78bce9b0a9e5de4486f0a953b70846d84efee940.

I reverted the appId population logic, but kept the assert removals

dorjesinpo (approved):

dorjesinpo (approved):

PR #410, opened 2024-08-28
small fixes to Python schema and tweak generators

jll63:
* Remove `xsdata[cli]` from Python pip requirements file, because it requires `ruff` and then in turn Rust.
* Changes to `schemagen`:
  * If `xsdata` is not available, require user to install it manually.
  * Add copyright notice to generated files.
  * `black`-format the generated files.
* `tweakgen`: add copyright notice to generated file, and `black`-format it.


kaikulimu (approved):

678098 (approved):

678098:
@jll63 we can merge this PR to fix python UTs https://github.com/bloomberg/blazingmq/pull/411
There are also issues with quotes in the schemagen script

678098:
@jll63 Need to also manually install `black` in CI

jll63:
> @jll63 Need to also manually install `black` in CI

I added `black` to `requirements.txt`. But I am not certain this is right. On one hand `black` is just like `xsdata[cli]` - it is needed by the two generators only. On the other hand, it does not pull heavy dependencies like `xsdata[cli]` does. 

If we decide to _not_ put `black` in the requirements, then I will have to make the two generators whine if it is not found, like `schemagen` also does with `xsdata`. Let's think about it over the weekend.

PR #416, opened 2024-09-04
Fix [MQB]: mqbc::StorageMgr: Transition to available only when all primary active

kaikulimu:
Fixes the flaky [`test_basic` of `test_restart.py`](https://github.com/bloomberg/blazingmq/blob/bd72cb45958abea96adb947c2f5e0c2bc95df94f/src/integration-tests/test_restart.py#L61). The issue was that a replica could advertise availability before all primaries are active; then, a proxy could repoen queue and post message with no avail. The fix is to transition to available only when all primaries are active.


src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -349,15 +360,24 @@ void StorageManager::setPrimaryStatusForPartitionDispatched(
...
+                  << "]: " << "Setting the status of primary: "
                   << pinfo.primary()->nodeDescription()
                   << ", primaryLeaseId: " << pinfo.primaryLeaseId()
-                  << ", from " << pinfo.primaryStatus() << " to " << value
-                  << ".";
+                  << ", from " << oldValue << " to " << value << ".";
 
     pinfo.setPrimaryStatus(value);
+    if (bmqp_ctrlmsg::PrimaryStatus::E_ACTIVE == value) {
+        d_fileStores[partitionId]->setPrimary(pinfo.primary(),
```
dorjesinpo, line 373: why is this in `setPrimaryStatusForPartition` and not in `setPrimaryForPartition`?

src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp
```diff
@@ -1655,17 +1655,32 @@ void ClusterOrchestrator::processPrimaryStatusAdvisory(
...
+                if (pinfo.primaryNode()) {
+                    BALL_LOG_OUTPUT_STREAM << " Ignoring advisory.";
+                }
+                else {
+                    BALL_LOG_OUTPUT_STREAM << " Since we have not received any"
+                                           << " information regarding the true"
+                                           << " primary, this advisory could "
+                                           << "be from the true one. Will"
+                                           << " buffer the advisory for now.";
+                    d_storageManager_p->bufferPrimaryStatusAdvisory(primaryAdv,
```
dorjesinpo, line 1680: Question, does `pinfo.primaryNode()` get assigned upon PrimaryStatusAdvisory or in the FSM flow there is another trigger?  We receive PrimaryStatusAdvisory and we do not have partition primary, why not assign the primary then?

dorjesinpo (commented):
Couple questions...

src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp
kaikulimu: In FSM mode, the source of truth for partition assignments are in the cluster state snapshot of the CSL file. As part of healing, a new leader [assigns partitions and then applies the assignments in its first CSL advisory](https://github.com/bloomberg/blazingmq/blob/5aaf5966b9f192cb410223bda43092b5a69b7855/src/groups/mqb/mqbc/mqbc_clusterstatemanager.cpp#L244-L258). Primary status adviosries can be stale; that's why we have a lot of checks in this function. My original idea was to simply ignore all primary status advisories and purely rely upon FSM for partition assignments. However, FSM can heal a replica but neglect to set a primary as active. Thus, I came up with the idea of buffering primary status advisories. If an advisory is not stale (i.e. matching primary node and leaseId), then we trust the availability advisory.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: FileStore should only set primary after the primary becomes active. The FileStore is unable to work with a passive primary. I will rename the function as `FileStore::setActivePrimary` to make the point clear.

kaikulimu:
@dorjesinpo Back to you

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -2379,6 +2418,68 @@ void StorageManager::do_processBufferedLiveData(const PartitionFSMArgsSp& args)
...
+                          << cit->first
+                          << " because primary node or leaseId is invalid. "
+                          << "Self-perceived [prmary, leaseId] is: ["
+                          << pinfo.primary()->nodeDescription() << ","
+                          << pinfo.primaryLeaseId() << "]";
+            continue;  // CONTINUE
+        }
+        pinfo.setPrimaryStatus(cit->first.status());
+        if (bmqp_ctrlmsg::PrimaryStatus::E_ACTIVE == cit->first.status()) {
+            d_fileStores[partitionId]->setActivePrimary(
```
dorjesinpo, line 2474: Do we need to 
```
  if(allParitionsAvailable()) {
            d_recoveryStatusCb(0);
```
or is it done implicitly and if so, maybe comment should explain that

dorjesinpo (commented):
one question

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -2379,6 +2418,68 @@ void StorageManager::do_processBufferedLiveData(const PartitionFSMArgsSp& args)
...
+                          << " Partition [" << partitionId
+                          << "]: " << "Ignoring primary status advisory "
+                          << cit->first
+                          << " because primary node or leaseId is invalid. "
+                          << "Self-perceived [prmary, leaseId] is: ["
+                          << pinfo.primary()->nodeDescription() << ","
+                          << pinfo.primaryLeaseId() << "]";
+            continue;  // CONTINUE
+        }
+        pinfo.setPrimaryStatus(cit->first.status());
```
kaikulimu, line 2472: What if you get a series of ACTIVE, then PASSIVE, then ACTIVE buffered advisories? How are you going to process them?

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: Will add a comment explaining how `d_recoveryStatusCb` is called a bit later.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: Currently, we do not do anything special when a primary becomes ACTIVE -> PASSIVE -> ACTIVE. Thus, we also do no-op here. Arguably, we can improve the logic, but it's beyond the scope of this PR.

dorjesinpo (approved):

PR #418, opened 2024-09-11
Add an integration test for broadcast mode

bbpetukhov:
Added an integration test for broadcast mode. It covers testing broadcast routing mode with consumers having different priorities and different subscriptions. The test makes sure that only high priority consumers receive messages and the messages are received in respect to customers' subscriptions.

**Describe your changes**
Added a new integration test `test_broadcast` to the `src/integration-tests/test_subscriptions.py`.

**Testing performed**
- The new test successfully passed on my machine locally
- GitHub Actions Workflow passed in the fork/branch

**Additional context**

### DIFF
--- a/src/integration-tests/test_subscriptions.py
+++ b/src/integration-tests/test_subscriptions.py
@@ -34,6 +34,7 @@ Existing tests:
 - test_non_overlapping
 - test_priorities
 - test_fanout
+- test_broadcast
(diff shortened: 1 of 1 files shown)


kaikulimu (approved):
Good job, Dmitrii! This is exactly what we expect for a broadcast test :)

PR #420, opened 2024-09-13
Feat[MQB]: Enhance queue consumption monitor alarm log with additional details

alexander-e1off:
When a queue starts to fill up, it is valuable to see information about which AppIds are impacted, and information about the messages in the queue.
Especially in the case of subscriptions (which we are enabling for everyone now), messages that match no subscription expression will build up in the put aside list.
To help make this situation clearer to operators and users (what apps are impacted, why are messages building up, how old is the head of the queue for each app, etc), we can log more information when the watermark alarm is triggered:
 - Sizes of put-aside list and redelivery list for each app for that queue;
 - Oldest message's timestamp in put aside list & its message properties;
 - Number of unconfirmed messages;
 - Total size of messages for each app (but seems it is already done by `storage()->capacityMeter()->printShortSummary()` );

This is to help debug why a message doesn't match a subscription.

Alarm log looks like this:
```
ERROR mqbblp_rootqueueengine.cpp:1765 ALARM [QUEUE_STUCK] Queue 'bmq://bmq.test.mem.fanout/my1?id=foo' Messages [current: 2 / 1,000], Bytes [current: 110  B / 1.00 MB], max idle time 20.00 s appears to be stuck. It currently has 1 consumers.
  1. bmqtool.tsk:199639@127.0.0.1~localhost:37430
    Handle Parameters .....: [ uri = "bmq://bmq.test.mem.fanout/my1" qId = 0 subIdInfo = NULL flags = 2 readCount = 1 writeCount = 0 adminCount = 0 ]
    Number of unconfirmed messages .....: 1
    UnconfirmedMonitors ....:
   STATE_NORMAL [Messages (STATE_NORMAL): 1 (820 - 820 - 1,024), Bytes (STATE_NORMAL): 55  B (25.60 MB - 25.60 MB - 32.00 MB)]
   STATE_NORMAL [Messages (STATE_NORMAL): 0 (820 - 820 - 1,024), Bytes (STATE_NORMAL): 0  B (25.60 MB - 25.60 MB - 32.00 MB)]

For appId: foo
Put aside list size: 1
Redelivery list size: 0
Number of messages: 2
Number of bytes: 110

Consumer subscription expressions:
x == 2
y == 4
y == 3
x == 1

Oldest message in a 'Put aside' list:
GUID                              Size        Timestamp (UTC)
4000000000033E07AD91F4D9C7E5D709       55  B  24SEP2024_09:09:06.777472+0000
Message Properties: [ sample_str (STRING) = "foo bar" x (INT32) = 10 ]

10 oldest messages in the queue:
Printing 2 message(s) [0-1 / 2] (total: 110  B)
       GUID                              Size        Timestamp (UTC)
    0: 4000000000033E07AD91F4D9C7E5D709       55  B  24SEP2024_09:09:06.777471+0000
    1: 4000010000344E57D5C6F4D9C7E5D709       55  B  24SEP2024_09:12:37.504283+0000

Current head of the queue:
GUID                              Size        Timestamp (UTC)
4000000000033E07AD91F4D9C7E5D709       55  B  24SEP2024_09:09:06.777471+0000

```

Implementation details:
- Log alarm logic is moved from `QueueConsumptionMonitor` into `RootQueueEngine` class, where more data is available;
- Callback is passed to `QueueConsumptionMonitor` and called in case of alarm to log alarm data;


dorjesinpo (changes requested):
Looks good.  Few questions.

- Maybe, we can print `Storage::numMessages` and `Storage::numBytes` as well?

- Can we get rid of `QueueEngineUtil_AppState::head()` and `QueueConsumptionMonitor::SubStreamInfo::d_headCb` now?

- Is `QueueConsumptionMonitor::onTransitionToIdle`  "level triggered" (vs "edge triggered")?  It can be noisy, how often we want that log?

@chrisbeard please take a look at the output

alexander-e1off:
Regarding `Maybe, we can print Storage::numMessages and Storage::numBytes as well?`. [CapacityMeter::printShortSummary()](https://github.com/bloomberg/blazingmq/blob/main/src/groups/mqb/mqbu/mqbu_capacitymeter.cpp#L446) already prints `numMessages` and `numBytes`, e.g. 
```
Messages [current: 2 / 1,000], Bytes [current: 66  B / 1.00 MB]
```
Aren't they the same as Storage::numMessages() and Storage::numBytes? In my test they printed the same values. Are there any possible scenarios when they will differ?

alexander-e1off:
Regarding `Can we get rid of QueueEngineUtil_AppState::head() and QueueConsumptionMonitor::SubStreamInfo::d_headCb now?` Now we print oldest message from `put aside` list, but what happens if we have empty `put aside` list and not empty `redelivery` list? Do we need to print oldest message from redelivery list too and then get rid of QueueEngineUtil_AppState::head()? Will double check if we can get rid of `d_headCb`, because it is used in logic to detect alarm condition.

Regarding `Is QueueConsumptionMonitor::onTransitionToIdle "level triggered" (vs "edge triggered")?` `  - `QueueConsumptionMonitor::onTransitionToIdle` is edge triggered, so `onTransitionToIdle` is called only once when state is transitioned from `Active` to `Idle`.

dorjesinpo:
> Aren't they the same as Storage::numMessages() and Storage::numBytes? In my test they

Looking at the `FileBackedStorage::numMessages` they are the same _if_ asking for the queue stats and not appId stats.

alexander-e1off:
> > Aren't they the same as Storage::numMessages() and Storage::numBytes? In my test they
> 
> Looking at the `FileBackedStorage::numMessages` they are the same _if_ asking for the queue stats and not appId stats.

Thanks, make sense, so I will add `Storage::numMessages` and `Storage::numBytes` per appId (appKey).

dorjesinpo:
> Regarding `Can we get rid of QueueEngineUtil_AppState::head() and QueueConsumptionMonitor::SubStreamInfo::d_headCb now?`

`d_headCb` is `QueueEngineUtil_AppState::head()` which is owned by the QueueEngine.  So, there seems to be no need to pass it to `RootQueueEngine::logAlarmCb`.
`QueueEngineUtil_AppState::head()`  should have answers about empty list(s) scenarios.  We may want to revisit the logic though.  Currently, we print put-aside if not empty, and the storage otherwise.

dorjesinpo:
> `QueueConsumptionMonitor::onTransitionToIdle` is edge triggered, so `onTransitionToIdle` is called only once when state is transitioned from `Active` to `Idle`.

Ok.  This can flap, we often see a lot of subsequent logs.  We are increasing the size of what's logged, so we may want to throttle the logging.

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1562,6 +1569,212 @@ void RootQueueEngine::onTimer(bsls::Types::Int64 currentTimer)
...
+bool RootQueueEngine::logAlarmCb(const mqbu::StorageKey& appKey,
+                                 const bool              enableLog) const
+{
+    // executed by the *QUEUE DISPATCHER* thread
+
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(d_queueState_p->queue()->dispatcher()->inDispatcherThread(
+        d_queueState_p->queue()));
+
+    // Construct AppId from appKey
```
dorjesinpo, line 1581: You can replace the App lookup code with
`Apps::const_iterator cItApp = d_apps.findByKey2(AppKeyCount(appKey, 0));`

`AppKeyCount` is an (hopefully, not for too long) artefact of our transitioning from non-CSL to CSL way of registering Apps.  In short, only registered Apps are supposed to have storage and consumption monitoring.  And registered Apps should have `0` as the "count" .

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1562,6 +1569,212 @@ void RootQueueEngine::onTimer(bsls::Types::Int64 currentTimer)
...
+    else if (!enableLog) {
+        // There are un-delivered messages, but log is disabled.
+        return true;  // RETURN
+    }
+
+    // Logging alarm info
+    bdlma::LocalSequentialAllocator<4096> localAllocator(d_allocator_p);
+
+    bmqt::Uri uri(&localAllocator);
+    uriBuilder.uri(&uri);
```
dorjesinpo, line 1615: `d_queueState_p->uri()`

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1562,6 +1569,212 @@ void RootQueueEngine::onTimer(bsls::Types::Int64 currentTimer)
...
+    d_queueState_p->storage()->capacityMeter()->printShortSummary(out);
+    out << ", max idle time "
+        << mwcu::PrintUtil::prettyTimeInterval(
+               d_queueState_p->queue()->domain()->config().maxIdleTime() *
+               bdlt::TimeUnitRatio::k_NANOSECONDS_PER_SECOND)
+        << " appears to be stuck. It currently has " << numConsumers
+        << " consumers." << ss.str() << '\n';
+
+    // Log un-delivered messages info
+    mqbi::Storage* const storage = d_queueState_p->storage();
```
dorjesinpo, line 1671: Use `storage` or `d_queueState_p->storage()` everywhere?

dorjesinpo (changes requested):
There may be a possibility to reduce `mqbblp_queueconsumptionmonitor.t` dependencies.  Let's explore it.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -338,58 +219,19 @@ TEST_F(Test, emptyQueue)
...
     ASSERT_EQ(logObserver.records().size(), expectedLogRecords);
 
     d_monitor.onTimer(k_MAX_IDLE_TIME + 1);
-    ASSERT_EQ(d_monitor.state(mqbu::StorageKey::k_NULL_KEY),
+    ASSERT_EQ(d_monitor.state(d_storageKey),
               QueueConsumptionMonitor::State::e_ALIVE);
     ASSERT_EQ(logObserver.records().size(), expectedLogRecords);
 }
 
-TEST_F(Test, logFormat)
```
alexander-e1off, line 356: This test originally tested the log format, but now logging is done by the callback, so this test makes no sense.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -727,122 +497,6 @@ TEST_F(Test, putAliveIdleSendAliveTwoSubstreams)
     ASSERT_EQ(d_monitor.state(key2), QueueConsumptionMonitor::State::e_ALIVE);
 }
 
-TEST_F(Test, putAliveIdleSendAliveTwoSubstreamsTwoConsumers)
```
alexander-e1off, line 730: Since consumers are removed, this test becomes the same as above, remove it.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -503,19 +498,7 @@ TEST_F(Test, putAliveIdleWithConsumer)
     ASSERT_EQ(logObserver.records().size(), ++expectedLogRecords);
     ASSERT(mwctst::ScopedLogObserverUtil::recordMessageMatch(
         logObserver.records().back(),
-        "ALARM \\[QUEUE_CONSUMER_MONITOR\\].*It currently has 2 consumers",
```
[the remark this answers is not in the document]
alexander-e1off: Refactored the test by simplifying logic and removing consumers, but still need `Test::d_queueState` and its dependencies, because it is required for `d_monitor`.

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
alexander-e1off: Replaced as proposed.

dorjesinpo (approved):
Thank you

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1613,6 +1617,204 @@ void RootQueueEngine::onTimer(bsls::Types::Int64 currentTimer)
...
+            const bmqp::MessagePropertiesInfo& logic =
+                storageIt_mp->attributes().messagePropertiesInfo();
+            bmqp::MessageProperties properties;
+            int ret = properties.streamIn(*appData, logic.isExtended());
+            if (!ret) {
+                out << "Message Properties: " << properties << '\n';
+            }
+            else {
+                BALL_LOG_WARN << "Failed to streamIn MessageProperties, rc = "
+                              << rc;
```
678098, line 1778:
```suggestion
                BALL_LOG_WARN << "Failed to streamIn MessageProperties, rc = "
                              << rc;
                out << "Message Properties: Failed to acquire [rc: " << rc << "]\n";
```
Do we want to print this in the alarm record itself?

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1613,6 +1617,204 @@ void RootQueueEngine::onTimer(bsls::Types::Int64 currentTimer)
...
+            }
+        }
+        else {
+            BALL_LOG_WARN << "Failed to get storage iterator for GUID: "
+                          << app->putAsideList().first() << ", rc = " << rc;
```
678098, line 1783:
```suggestion
            BALL_LOG_WARN << "Failed to get storage iterator for GUID: "
                          << app->putAsideList().first() << ", rc = " << rc;
            out << "'Put aside' list: Failed to acquire [rc: " << rc << "]\n";
```

678098 (changes requested):
Some comments

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1613,6 +1617,204 @@ void RootQueueEngine::onTimer(bsls::Types::Int64 currentTimer)
...
+         ++cIt) {
+        if (!cIt->expression().empty()) {
+            ss << cIt->expression() << '\n';
+            ++exprNum;
+        }
+    }
+    if (exprNum) {
+        if (exprNum == k_EXPR_NUM_LIMIT) {
+            out << k_EXPR_NUM_LIMIT << " of "
+                << " consumer subscription expressions: ";
```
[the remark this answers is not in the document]
alexander-e1off: Added `First` word in the beginning for clarity, e.g. `First 50 of consumer subscription expressions:`

alexander-e1off:
@678098 thank you, applied your suggestions.

678098 (approved):
LGTM!

Issue #426, opened 2024-09-27
session->LoadMessageEventBuilder Destructor SIGSEV different thread


kclonts:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

// thread 1
Session* sess = new Session();

// thread 2
{
MessageEventBuilder b;
session->LoadMessageEventBuilder(&b);
}

SIGSEV on destructor from g_newDeleteAllocatorSingleton

Both Async & Sync mode

Documentation states all Session methods are thread enabled

### Expected Behavior

no sigsev; MessageEventBuilder is valid

### Steps To Reproduce

see description

### BlazingMQ Version

0.92.10, 0.92.6, latest

pniedzielski:
Hi @kclonts,

Thanks for your report!  It seems like a bug, but just a little bit more information:

1.  What platform are you on?  OSX version, Linux, architecture, etc
2.  Can you give us a minimal code example (.cpp file) that shows this?

dorjesinpo:
You have to call session `stop` before the exit.  Without that, session (internal) threads may use allocators that have destructed.

dorjesinpo:
..also, `sess` leaks

pniedzielski:
I'm able to reproduce this issue with and without calling `stop` and deleting `sess`.  For this backtrace I also added pthread thread names to the `std::thread` we're creating to distinguish.  Error is indeed at end of scope on `t1`.

```
(lldb) thread backtrace all
thread backtrace all
  thread #1, queue = 'com.apple.main-thread'
    frame #0: 0x000000018817ab8c libsystem_kernel.dylib`__ulock_wait + 8
    frame #1: 0x00000001881bc48c libsystem_pthread.dylib`_pthread_join + 608
    frame #2: 0x00000001880f1aa8 libc++.1.dylib`std::__1::thread::join() + 36
    frame #3: 0x00000001006e0c08 a.out`main + 336
    frame #4: 0x0000000187e2f154 dyld`start + 2476
  thread #2, name = 'bmqFSMEvtQ'
    frame #0: 0x000000018817c5cc libsystem_kernel.dylib`__psynch_cvwait + 8
    frame #1: 0x00000001881ba55c libsystem_pthread.dylib`_pthread_cond_wait + 1228
[...]

PR #433, opened 2024-10-01
Fix[bmqa_message.h]: Enable message `d_groupId` field regardless of build

pniedzielski:
*Issue number of the reported bug or feature request: #426*

**Describe your changes**
We currently do some cleverness with where and when we enable the incomplete message group ID feature in `CMakeLists.txt`.  However, when using the `build-darwin.sh` and `build-ubuntu.sh` scripts, as we currently recommend, the libraries we install using `make install` have this feature enabled.

Unfortunately, enabling message group IDs changes the size of `bmqa_message`’s `MessageImpl`, so a library built with message group IDs enabled is ABI-incompatible with an application built with message group IDs disabled.  Because we don’t advertise this feature anywhere or export the definition from CMake, users are unlikely to know they need to compile with the flag `-DBMQ_ENABLE_MSG_GROUPID`, and will run into characteristic ABI-incompatibility issues: buffer overruns, stack clobbering, and segmentation faults.

This patch makes a minimal change until we decide what to do with message group IDs: the API is still sensitive to `BMQ_ENABLE_MSG_GROUPID`, but the message structure itself always carries a `d_groupId` field, regardless of its presence.  This field is default-initialized and is otherwise unused.

**Testing performed**
This change fixes the minimal example given in #426.  I will let the CI run all the tests to ensure no breakage.

### DIFF
--- a/src/groups/bmq/bmqa/bmqa_message.h
+++ b/src/groups/bmq/bmqa/bmqa_message.h
@@ -121,11 +121,11 @@ struct MessageImpl {
     // SubscriptionHandle this message is
     // associated with
 
-#ifdef BMQ_ENABLE_MSG_GROUPID
+    // #ifdef BMQ_ENABLE_MSG_GROUPID
     bsl::string d_groupId;
     // Optional Group Id this message is
     // associated with
-#endif
+    // #endif
 };
 
 // ===============================


pniedzielski:
We should probably also fix our build system.

pniedzielski:
Closed in favor of #437.

PR #437, opened 2024-10-02
Fix[CMakeLists.txt]: Remove `-DBMQ_ENABLE_MSG_GROUPID` from the build system

pniedzielski:
*Issue number of the reported bug or feature request: #426*

**Describe your changes**
We do not ever want to build with this flag when releasing, and users often manage to enable this flag accidentally.  Because message group IDs are not fully implemented, we remove this temporary definition.  It can be added in later if we ever come back to this feature.

**Testing performed**
This change fixes the minimal example given in https://github.com/bloomberg/blazingmq/issues/426. I will let the CI run all the tests to ensure no breakage.

**Additional context**
Supersedes #433.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,16 +197,6 @@ else()
   endif()
 endif()
 
-# TBD: TEMPORARY >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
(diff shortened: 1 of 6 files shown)


pniedzielski:
Looks like some of the unit tests fail without this flag.  I will guard them with `BMQ_ENABLE_MSG_GROUPID`.

hallfox (approved):

PR #441, opened 2024-10-04
Fix[bmqeval]: limit expression length to avoid stack overflow

678098:
If we pass a 1MB `NOT` expression, there will be a segfault, even if we just validate it:
`zsh: segmentation fault  ./src/groups/bmq/bmqeval_simpleevaluator.t.tsk 1`

This PR:
- Limits expression length to reasonable 128 bytes.
- Introduces a new error type for expression validation `e_TOO_LONG`.
- Improves error messages on errors with concrete info about limitations.
- Fixes a few warnings related to bmqeval UT.

### DIFF
--- a/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.t.cpp
+++ b/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.t.cpp
@@ -61,9 +61,9 @@ class MockPropertiesReader : public PropertiesReader {
     // MANIPULATORS
 
     /// Return a `bdld::Datum` object with value for the specified `name`.
-    /// Use the specified `allocator` for any memory allocation.
-    virtual bdld::Datum get(const bsl::string& name,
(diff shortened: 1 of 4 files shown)


src/groups/bmq/bmqeval/bmqeval_simpleevaluator.t.cpp
```diff
@@ -287,7 +314,7 @@ static void test3_evaluation()
     struct TestParameters {
         const char* expression;
         bool        expected;
-        bool        expectCompilationFailure;
+        bool        expectCompilationFailure = false;
```
hallfox, line 317: This is a modern C++ feature

678098: Moved compilation failures to its own test

hallfox (approved):

PR #442, opened 2024-10-07
Feat[MQB]: Do not record last confirm

bbpetukhov:
**Describe your changes**
The goal of this optimization is to avoid storing the last Confirm record in the journal file, because the Delete record follows right after this Confirm record, thus the last Confirm record is redundant. 

Also since we'll have to store one record less, I decreased from 3 to 2 the default static size of records array.

Previously there were no unit tests for FileBackedStorage. Added 22 unit tests for FileBackedStorage by analogy to InMemoryStorage, since both classes inherit from ReplicatedStorage and their usage scenarios are similar.

**Testing performed**
- Manual testing via bmqtool and bmq storage tool to make sure that the last Confirm record is not written. Tested both Priority and Fanout mode (w/ 3 appIds).
- CI Workflow passed successfully.

**Additional context**

Priority mode:

Screenshot 2024-09-16 124215
Screenshot 2024-09-16 124237

Fanout mode:

Screenshot 2024-10-08 212644
Screenshot 2024-10-08 212626


src/groups/mqb/mqbs/mqbs_filebackedstorage.cpp
```diff
@@ -389,6 +389,10 @@ FileBackedStorage::confirm(const bmqt::MessageGUID& msgGUID,
     RecordHandlesArray& handles = it->second.d_array;
     BSLS_ASSERT_SAFE(!handles.empty());
 
+    if (0 == --it->second.d_refCount) {
```
678098, line 392: Let's add a comment explicitly stating that we don't record the last confirm here for optimization

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+    return uniqueGUID;
+}
+
+// CLASSES
+
+// ===================
+// class MockDataStore
+// ===================
+
+/// Mimimal mock implementation of the `mqbs::DataStore` interface
```
678098, line 160:
```suggestion
/// Minimal mock implementation of the `mqbs::DataStore` interface
```

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+    ASSERT_EQ(d_tester.storage().get(&attributes, generateUniqueGUID(guids)),
+              mqbi::StorageResult::e_GUID_NOT_FOUND);
+    ASSERT_EQ(d_tester.storage().get(&appData,
+                                     &options,
+                                     &attributes,
+                                     generateUniqueGUID(guids)),
+              mqbi::StorageResult::e_GUID_NOT_FOUND);
+}
+
+TEST_F(Test, test8_remove_messageNotFound)
```
678098, line 1000: Here the test has a number, in other places test names don't contain a number

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+// ------------------------------------------------------------------------
+// Remove Messages Test
```
678098, line 1002: Here the name of the test differs from the title

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+// Iterator Test
+//
+// Testing:
+//   Verifies the iterator in absence of virtual storages
+//   in a 'mqbs::FileBackedStorage'.
+// ------------------------------------------------------------------------
+{
+    mwctst::TestHelper::printTestName("Iterator- No virtual storages Test");
+
+    mwcu::MemOutStream errDescription(s_allocator_p);
```
678098, line 1496: Not used
```suggestion
```

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+// Capacity Meter Test
+//
+// Testing:
+//   Verifies the capacity meter functionality with respect to
+//   limits on messages in a 'mqbs::FileBackedStorage'.
+// ------------------------------------------------------------------------
+{
+    mwctst::TestHelper::printTestName("Capacity Meter- Limit Messages");
+
+    mwcu::MemOutStream errDescription(s_allocator_p);
```
678098, line 1671: Not used
```suggestion
```

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+// Capacity Meter Test
+//
+// Testing:
+//   Verifies the capacity meter functionality with respect to
+//   limits on bytes in a 'mqbs::FileBackedStorage'.
+// ------------------------------------------------------------------------
+{
+    mwctst::TestHelper::printTestName("Capacity Meter - Limit Bytes");
+
+    mwcu::MemOutStream errDescription(s_allocator_p);
```
678098, line 1716: Not used
```suggestion
```

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+{
+    mwctst::TestHelper::printTestName("GARBAGE COLLECT");
+
+    // Set with TTL of 20: control GC test by manipulating secondsFromEpoch
+    // input
+    const int k_TTL = 20;
+
+    configure(k_TTL);
+
+    mwcu::MemOutStream             errDescription(s_allocator_p);
```
678098, line 1762: Not used
```suggestion
```

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+// Testing:
+//   Verifies the 'remove' in a 'mqbs::FileBackedStorage'. Check GUIDs that
+//   in storage as well as GUID not in storage
+// ------------------------------------------------------------------------
+{
+    mwctst::TestHelper::printTestName("REMOVE MESSAGE");
+
+    const int k_MSG_COUNT = 10;
+
+    mwcu::MemOutStream             errDescription(s_allocator_p);
```
678098, line 1033: Not used
```suggestion
```

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+// Get Test - with no virtual storages
+//
+// Testing:
+//   Verifies the 'get' operation in absence of virtual storages
+//   in a 'mqbs::FileBackedStorage'.
+// ------------------------------------------------------------------------
+{
+    mwctst::TestHelper::printTestName("GET - WITH NO VIRTUAL STORAGES");
+
+    mwcu::MemOutStream             errDescription(s_allocator_p);
```
678098, line 946: Not used
```suggestion
```

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+// ------------------------------------------------------------------------
+// GET MESSAGE SIZE
+//
+// Testing:
+//   - 'getMessageSize(...)'
+// ------------------------------------------------------------------------
+{
+    mwctst::TestHelper::printTestName("GET MESSAGE SIZE");
+
+    mwcu::MemOutStream             errDescription(s_allocator_p);
```
678098, line 909: Not used
```suggestion
```

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+// Put Test - with no virtual storages
+//
+// Testing:
+//   Verifies the 'put' operation in absence of virtual storages
+//   in a 'mqbs::FileBackedStorage'.
+// ------------------------------------------------------------------------
+{
+    mwctst::TestHelper::printTestName("PUT - WITH NO VIRTUAL STORAGES");
+
+    mwcu::MemOutStream             errDescription(s_allocator_p);
```
678098, line 872: Not used
```suggestion
```

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+    ASSERT_EQ(d_tester.storage().numVirtualStorages(), 1);
+
+    ASSERT(d_tester.storage().removeVirtualStorage(k_APP_KEY2));
+    ASSERT(!d_tester.storage().hasVirtualStorage(k_APP_KEY2, &dummyAppId));
+    ASSERT_EQ(d_tester.storage().numVirtualStorages(), 0);
+}
+
+TEST_F(BasicTest, put_withVirtualStorages)
+// ------------------------------------------------------------------------
+// Put Test  with virtual storages
```
678098, line 1163:
```suggestion
// Put Test with virtual storages
```

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+TEST_F(BasicTest, get_withVirtualStorages)
+// ------------------------------------------------------------------------
+// Get Test With virtual storages
+//
+// Testing:
+//   Verifies the 'get' operation in presence of virtual storages
+//   in a 'mqbs::FileBackedStorage'.
+// ------------------------------------------------------------------------
+{
+    mwctst::TestHelper::printTestName("Get- with Virtual Storage Test");
```
678098, line 1354:
```suggestion
    mwctst::TestHelper::printTestName("Get - with Virtual Storage Test");
```

678098 (changes requested):
Misc comments.
Let's also fix warnings in the new test file from CI check `UT [C++]`
https://productionresultssa13.blob.core.windows.net/actions-results/85f085aa-bee4-4859-9ef1-8c0188f9cca1/workflow-job-run-63f2f093-663a-5fa9-4a2e-8878fefbc7e7/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-10-09T14%3A33%3A14Z&sig=%2BhV9s4XueHcYfci70aJVwz63O7TxrThtpwct2kiMON4%3D&ske=2024-10-09T23%3A27%3A43Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-10-09T11%3A27%3A43Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-08-04&sp=r&spr=https&sr=b&st=2024-10-09T14%3A23%3A09Z&sv=2024-08-04
Search for `mqbs_filebackedstorage.t` in UT build log.

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+        for (int i = 0; i < msgCount; i++) {
+            bmqt::MessageGUID guid;
+
+            if (useSameGuids) {
+                guid = guidHolder->at(i);
+            }
+            else {
+                mqbu::MessageGUIDUtil::generateGUID(&guid);
+                guidHolder->push_back(guid);
+            }
```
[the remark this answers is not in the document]
bbpetukhov: > If we pass `useSameGuids = false` and also pass non-empty `*guidHolder`, we will be pushing back messages so it's more than `msgCount` messages in `*guidHolder`

In fact this is exactly the expected behavior. Function `addMessages` adds new messages and new number of new messages if `msgCount`. Thus the size of `guidHolder` vector can become bigger than `msgCount`. And many existing tests (also in InMemoryStorage.t.cpp) rely on this behavior.

However, indeed checking flag `useSameGuids` on every iteration of `for` loop can be optimized and be done before the loop to reduce code complexity.

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,1946 @@
...
+
+    return randomGUID;
+}
+
+static bmqt::MessageGUID
+generateUniqueGUID(const bsl::vector<bmqt::MessageGUID>& guids)
+{
+    bmqt::MessageGUID uniqueGUID;
+    do {
+        uniqueGUID = generateRandomGUID();
```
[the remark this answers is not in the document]
bbpetukhov: Indeed if we look from the current implementation perspective, the generated GUIDs are both **unique** and **not random**. And this gives us the opportunity to optimize test code.

However, if we look from the interface perspective: `generateUnique` and `generateRandom` functions have different design incentive as it's described in their names despite the fact that their implementation is the same now. Performance-wise it's not critical to have a few more function in the tests. 

Thus I'd suggest to keep the functions separate. But we can drop the for loop checking for uniqueness.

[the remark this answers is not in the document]
678098: @bbpetukhov good point, let's keep it as it is.
Maybe you can keep the `for-loop` and leave a small comment that actually we expect this loop to exit from the first iteration due to the current implementation of generator, but what the function will work even if generator changes.

src/groups/mqb/mqbs/mqbs_filebackedstorage.h
```diff
@@ -97,7 +97,7 @@ class FileBackedStorage BSLS_KEYWORD_FINAL : public ReplicatedStorage {
     // Most likely number of records for each guid (one each of message,
     // confirm & deletion record).  This number is correct for every queue
     // except for the fanout one, which has more than 1 confirm records.
```
678098, line 99:
```suggestion
    // The most probable number of records for each guid for priority queue.
    // Currently, the value is 2: one data record + one deletion record.
    // With last confirm optimization, we don't write a last confirm, and don't
    // count it here.
    // For fanout queues, the expected number of records is more than this:
    // one data record + (number of appIDs - 1) confirms + one deletion record,
    // where -1 due to last confirm optimization.
```
I would change the comment above, since it's outdated.

678098 (commented):
A few small comments and good to merge

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,2010 @@
...
+    // NOT IMPLEMENTED
+    BaseTester(const BaseTester&) BSLS_KEYWORD_DELETED;
+    BaseTester& operator=(const BaseTester&) BSLS_KEYWORD_DELETED;
+};
+
+// =============
+// struct Tester
+// =============
+struct Tester : public BaseTester {
+    MockDataStore d_dataStore;
```
678098, line 643: Actually, I think there is a bug here.

In `Tester`, there is a field `MockDataStore d_dataStore`.
We pass a pointer to this field to `d_replicatedStorage_mp.load(...)`. But `d_replicatedStorage_mp` is a field of the base class `BaseTester`.

So when we call a chain of destructors, we first destroy fields of the child `Tester` class, and only after this we destroy fields of the parent `BaseTester` class.

This means, that the object under the pointer to `d_dataStore` will be destroyed before the `d_replicatedStorage_mp` is destroyed.

And we also call `d_replicatedStorage_mp->removeAll(k_NULL_KEY);` in `~BaseTester()`, so we basically access the `d_dataStore` object which was already destroyed.

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
```diff
@@ -0,0 +1,2010 @@
...
+// struct GCTest
+// -------------
+// CREATORS
+GCTest::GCTest()
+: d_testerBuffer()
+{
+    // NOTHING
+}
+
+GCTest::~GCTest()
```
678098, line 777: Actually, I am thinking if we have too much test related classes here. I don't see the point of creating so many of them.
What we have:
`mwctst::Test`, `GCTest`, `Test`, `BasicTest`
Can we only have 2 of them? Just `mwctst::Test` and `Test`?

The same thing with testers classes.
We have: `Tester`, `BaseTester`.
Can we have only one `Tester`?

I have a feeling that we are overcomplicating things by this bloated class structure

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
678098: Also, if we pass a pointer to a field value somewhere else, we should probably protect ourselves from incorrect use of this class, by removing copy/assignment operators

678098 (changes requested):

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
bbpetukhov: Makes sense. I'll put all the code into the single Tester class

src/groups/mqb/mqbs/mqbs_filebackedstorage.t.cpp
bbpetukhov: Done. Although maybe there is now more test code duplication, but the code of the unit tests became more straightforward. 

Please take a look. And if everything is ok, I'll transfer the changes to InMemoryStorage unit tests too

bbpetukhov: Migrated the refactoring to the InMemoryStorage unit tests

678098 (approved):

PR #446, opened 2024-10-08
Feat[plugins]: report queue depth per appId to prometheus

678098:
Introduced try-catch block for plugins:
```
10OCT2024_14:04:34.987 (139982138242624) ERROR bmqprometheus_prometheusstatconsumer.cpp:263 #PLUGIN_ERROR Invalid metric name
10OCT2024_14:04:35.987 (139982138242624) ERROR bmqprometheus_prometheusstatconsumer.cpp:263 #PLUGIN_ERROR Invalid metric name
10OCT2024_14:04:36.987 (139982138242624) ERROR bmqprometheus_prometheusstatconsumer.cpp:263 #PLUGIN_ERROR Invalid metric name
```


src/plugins/bmqprometheus/bmqprometheus_prometheusstatconsumer.cpp
```diff
@@ -414,11 +418,18 @@ void PrometheusStatConsumer::captureQueueStats()
...
                 if (role == mqbstat::QueueStatsDomain::Role::e_PRIMARY) {
-                    value = mqbstat::QueueStatsDomain::getValue(
-                        *appIdIt,
-                        d_snapshotId,
-                        mqbstat::QueueStatsDomain::Stat::e_QUEUE_TIME_MAX);
-                    updateMetric(&queueTimeDataPoint, labels, value);
+                    for (DatapointDefCIter dpIt = bdlb::ArrayUtil::begin(defs);
+                         dpIt != bdlb::ArrayUtil::end(defs);
+                         ++dpIt) {
+                        const bsls::Types::Int64 value =
```
waldgange, line 424: Looks a bit confusing as we already have `value` in outer sope. Let's either reassign the existing valiable or change the name of this one

678098: I replaced it by introducing another for-loop for conflicting `value`.

678098:
@waldgange back to you, also note that I changed metric names:
1. Use a common dot notation everywhere, so `queue_gc_msgs` -> `queue.gc_msgs`
2. Renamed confusing metrics `queue_content_msgs` -> `queue.content_msgs_max` and the same with content_bytes (both of these metrics had implicit max value reported).

alexander-e1off:
> @waldgange back to you, also note that I changed metric names:
> 
> 1. Use a common dot notation everywhere, so `queue_gc_msgs` -> `queue.gc_msgs`
> 2. Renamed confusing metrics `queue_content_msgs` -> `queue.content_msgs_max` and the same with content_bytes (both of these metrics had implicit max value reported).

Prometheus does not support dot in [metric names](https://prometheus.io/docs/concepts/data_model/): 
```
Metric names may contain ASCII letters, digits, underscores, and colons. It must match the regex [a-zA-Z_:][a-zA-Z0-9_:]*.
```

waldgange:
> @waldgange back to you, also note that I changed metric names:
> 
> 1. Use a common dot notation everywhere, so `queue_gc_msgs` -> `queue.gc_msgs`
> 2. Renamed confusing metrics `queue_content_msgs` -> `queue.content_msgs_max` and the same with content_bytes (both of these metrics had implicit max value reported).

It's OK to change the suffixes, but metric names can't contain dots. This is a Prometheus-specific [restriction](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels), so we use underscores.
Moreover If you pass these names to `prometheus-cpp` it will raise an exception. And it will crash the broker.
Actually it does not look safe. I think it's worth wrapping the code in `onSnapshot()` method with a `try/catch` block to prevent the broker from crashing in future. Although metric names are constant, it's definitely possible to specify label names that also have the same restrictions.

678098:
@alexander-e1off @waldgange thanks, I will revert underscore change then.
I added a try-catch block around the code which operates metrics.
I am sure that it can throw `std::invalid_argument`, `std::runtime_error` or `std::length_error`, but to be sure, I catch a general exception type.

src/groups/mqb/mqbstat/mqbstat_statcontroller.cpp
```diff
@@ -576,7 +576,12 @@ void StatController::snapshotAndNotify()
     // StatConsumers will report all stats
     bsl::vector<StatConsumerMp>::iterator it = d_statConsumers.begin();
     for (; it != d_statConsumers.end(); ++it) {
-        (*it)->onSnapshot();
+        try {
+            (*it)->onSnapshot();
+        }
+        catch (const bsl::exception& e) {
+            BALL_LOG_ERROR << "#PLUGIN_ERROR " << e.what();
+        }
```
waldgange, line 584: I like this change! I was surprised we didn't have this wrapper before

waldgange (approved):
LGTM

PR #453, opened 2024-10-09
It: Include full path for admin stat it test failures

chrisbeard:
**Describe your changes**
This patch makes it a little easier to debug the metric & operation that causes an integration test for stats to fail.

Some examples:

Broker returning a stat that the test does not expect to be reported.
```
FAILED test_admin_client.py::test_queue_stats[single_node] - AssertionError: ('after-post.appIds.bar.values', 'Extra expected keys:', set(), 'Missing expected keys:', {'queue_history_abs'})
```


src/integration-tests/test_admin_client.py
```diff
@@ -90,6 +90,7 @@ def extract_stats(admin_response: str) -> dict:
 def expect_same_structure(
     entry: Union[dict, list, str, int],
     expected: Union[dict, list, str, int, dt.ValueConstraint],
+    path="",
```
678098, line 93:
```suggestion
    path: str = "",
```

678098 (approved):

678098 (approved):

PR #454, opened 2024-10-09
CI: Add CI checks with clang-tidy

hallfox:
This adds a new check to our workflows using clang-tidy. Right now we only enable analyzer checks, though we could look at expanding to more in the future. We are primarily interested in getting another pass over our code checking for C++03 syntax compatibility, since we've had a few instances of modern syntax constructs getting into the code base. `clang-tidy` will do a bit more than syntax checks, but I believe having will be a boon overall.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -17,8 +17,10 @@
         },
         {
             "name": "macos-arm64-vcpkg",
-            "description": "VCPKG based configuration for building on arm-based MacOS",
-            "toolchainFile": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+            "description":
+                "VCPKG based configuration for building on arm-based MacOS",
+            "toolchainFile":
+                "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
             "inherits": "base",
             "cacheVariables": {
                 "VCPKG_INSTALL_OPTIONS": "--allow-unsupported",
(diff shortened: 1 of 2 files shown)


CMakePresets.json
```diff
@@ -37,9 +41,42 @@
...
             "name": "linux-x64-vcpkg",
-            "description": "VCPKG based configuration for building on x86_64-based Linux",
-            "toolchainFile": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+            "description":
+                "VCPKG based configuration for building on x86_64-based Linux",
+            "toolchainFile":
+                "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
             "inherits": "base"
+        },
+        {
```
pniedzielski, line 50: This is a good change even beyond clang-tidy.  Simplifies CI a lot.

pniedzielski (commented):
This looks good to me; however I think the PR should also fix the issues `clang-tidy` finds, so we're not left with a failing CI.  Maybe it makes sense to start out with a very small set of checks and have more PRs to incrementally enable more.

pniedzielski:
@hallfox 

Work I am doing in the same vein as https://github.com/bloomberg/blazingmq/pull/706 is to make the `clang-format`, `black`, etc checks we do in CI first-class CMake targets.  It may make sense to add a `clang-tidy` CMake target first, fix the issues it finds, then update this PR to use that CMake target.

Thoughts?

CMakePresets.json
pniedzielski: Coming back to this and in reference to our (@hallfox, @emelialei88) offline discussion earlier today, it may be best to pull this out into its own PR (plus other useful presets).

I'll take a stab at this tomorrow and send it your way.

PR #461, opened 2024-10-14
Fix: m_bmqstoragetool::FileManagerImpl: Asserts not have side effects

kaikulimu:
If operations are contained within `BSLS_ASSERT`s, then when we run in optimized mode the asserts will be ignored and those operations will be skipped.

@waldgange @alexander-e1off @bbpetukhov  For awareness


678098 (approved):
This is a good change, let's merge it.
In the future, I think we should replace asserts in bmqstoragetool by exceptions. There is no point to continue trying to do anything if an error is not recoverable, and these checks should not be disabled by compilation options.

PR #470, opened 2024-10-18
Feat[MQB]:Enhance filebackedstorage/inmemorystorage alarm log with app subscription info

alexander-e1off:
When a queue starts to fill up, it is valuable to see information about which AppIds are impacted, and information about the messages in the queue.
Especially in the case of subscriptions (which we are enabling for everyone now), messages that match no subscription expression will build up in the put aside list.
To help make this situation clearer to operators and users (what apps are impacted, why are messages building up, how old is the head of the queue for each app, etc), we can log more information when the filebackedstorage alarm is triggered:
 - Sizes of put-aside list and redelivery list for each app for that queue;
 - Consumer subscriptions;
 - Oldest message's timestamp in put aside list & its message properties;

This is to help debug why a message doesn't match a subscription.

Enhanced log looks like this:
```
ALARM: ALARM [CAPACITY_STATE_FULL] for 'queue [bmq://bmq.test.persistent.fanout/my1]': [Messages (STATE_FULL): 3 (limit: 3), Bytes (STATE_NORMAL): 165  B (limit: 1.00 MB)]
For appId: foo

Put aside list size: 2
Redelivery list size: 0
Number of messages: 2
Number of bytes: 110  B

Consumer subscription expressions:
<Empty>
x == 1
y == 2

Oldest message in the 'Put aside' list:
GUID                              Size        Timestamp (UTC)
4000010000417F0461B0F00A8CE52964       55  B  18OCT2024_09:03:27.485301+0000
Message Properties: [ sample_str (STRING) = "foo bar" x (INT32) = 3 ]

For appId: bar

Put aside list size: 2
Redelivery list size: 0
Number of messages: 2
Number of bytes: 110  B

Consumer subscription expressions:
x == 1
y == 2

Oldest message in the 'Put aside' list:
GUID                              Size        Timestamp (UTC)
4000010000417F0461B0F00A8CE52964       55  B  18OCT2024_09:03:27.485301+0000
Message Properties: [ sample_str (STRING) = "foo bar" x (INT32) = 3 ]
```

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
@@ -1619,6 +1619,132 @@ void RootQueueEngine::onTimer(bsls::Types::Int64 currentTimer)
     d_consumptionMonitor.onTimer(currentTimer);
 }
 
+bsl::ostream&
+RootQueueEngine::logAppSubscriptionInfo(bsl::ostream&           stream,
+                                        const mqbu::StorageKey& appKey) const
+{
+    // executed by the *QUEUE DISPATCHER* thread
+
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(d_queueState_p->queue()->dispatcher()->inDispatcherThread(
+        d_queueState_p->queue()));
+
+    // Get AppState by appKey.
+    Apps::const_iterator cItApp = d_apps.findByKey2(AppKeyCount(appKey, 0));
+    if (cItApp == d_apps.end()) {
+        BALL_LOG_WARN << "No app found for appKey: " << appKey;
+        stream << "\nSubscription info: no app found for appKey: " << appKey;
+        return stream;  // RETURN
+    }
+
+    const AppStateSp& app = cItApp->value();
+    return logAppSubscriptionInfo(stream, app);
+}
+
+bsl::ostream&
+RootQueueEngine::logAppSubscriptionInfo(bsl::ostream&     stream,
+                                        const AppStateSp& appState) const
+{
+    mqbi::Storage* const storage = d_queueState_p->storage();
+
+    // Log un-delivered messages info
+    stream << "\nFor appId: " << appState->appId() << "\n\n";
+    stream << "Put aside list size: "
+           << bmqu::PrintUtil::prettyNumber(static_cast<bsls::Types::Int64>(
+                  appState->putAsideListSize()))
+           << '\n';
+    stream << "Redelivery list size: "
+           << bmqu::PrintUtil::prettyNumber(static_cast<bsls::Types::Int64>(
+                  appState->redeliveryListSize()))
+           << '\n';
+    stream << "Number of messages: "
+           << bmqu::PrintUtil::prettyNumber(
+                  storage->numMessages(appState->appKey()))
+           << '\n';
+    stream << "Number of bytes: "
+           << bmqu::PrintUtil::prettyBytes(
+                  storage->numBytes(appState->appKey()))
+           << "\n\n";
+
+    // Log consumer subscriptions
+    mqbblp::Routers::QueueRoutingContext& routingContext =
+        appState->routing()->d_queue;
+    mqbcmd::Routing routing;
+    routingContext.loadInternals(&routing);
+    const bsl::vector<mqbcmd::SubscriptionGroup>& subscrGroups =
+        routing.subscriptionGroups();
+    if (!subscrGroups.empty()) {
+        // Limit to log only k_EXPR_NUM_LIMIT expressions
+        static const size_t k_EXPR_NUM_LIMIT = 50;
+        bmqu::MemOutStream  ss(d_allocator_p);
+
+        size_t exprNum = 0;
+        for (bsl::vector<mqbcmd::SubscriptionGroup>::const_iterator cIt =
+                 subscrGroups.begin();
+             cIt != subscrGroups.end() && exprNum < k_EXPR_NUM_LIMIT;
+             ++cIt) {
+            if (!cIt->expression().empty()) {
+                ss << cIt->expression() << '\n';
+                ++exprNum;
+            }
+        }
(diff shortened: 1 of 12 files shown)


dorjesinpo (commented):
Looks good.  Couple small comments

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1620,6 +1620,131 @@ void RootQueueEngine::onTimer(bsls::Types::Int64 currentTimer)
...
+    mqbblp::Routers::QueueRoutingContext& routingContext =
+        appState->routing()->d_queue;
+    mqbcmd::Routing routing;
+    routingContext.loadInternals(&routing);
+    const bsl::vector<mqbcmd::SubscriptionGroup>& subscrGroups =
+        routing.subscriptionGroups();
+
+    // Limit to log only k_EXPR_NUM_LIMIT expressions
+    static const size_t k_EXPR_NUM_LIMIT = 50;
+    mwcu::MemOutStream  ss(d_allocator_p);
```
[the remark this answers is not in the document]
alexander-e1off: Added `if (!subscrGroups.empty())` upfront

src/groups/mqb/mqbu/mqbu_capacitymeter.h
```diff
@@ -218,6 +224,10 @@ class CapacityMeter {
     // SpinLock for synchronization of this
     // component
 
+    LogAppsSubscriptionInfoCb d_logAppsSubscriptionInfoCb;
```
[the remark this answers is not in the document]
alexander-e1off: Renamed as `LogEnhancedStorageInfoCb` to be more generic.

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.h
```diff
@@ -458,6 +463,15 @@ class RootQueueEngine BSLS_KEYWORD_FINAL : public mqbi::QueueEngine {
...
     loadInternals(mqbcmd::QueueEngine* out) const BSLS_KEYWORD_OVERRIDE;
+
+    /// Log appllication subscription info for the specified `appKey` into the
+    /// specified `stream`.
+    ///
+    /// THREAD: This method is called from the Queue's
+    /// dispatcher thread.
+    virtual bsl::ostream& logAppSubscriptionInfo(
+        bsl::ostream&           stream,
+        const mqbu::StorageKey& appKey) const BSLS_KEYWORD_OVERRIDE;
```
678098, line 474: There is another overload of `logAppSubscriptionInfo` at line 216. Since they look similar, I would propose to place their declarations nearby.

alexander-e1off: Put declarations nearby

src/groups/mqb/mqbs/mqbs_filebackedstorage.cpp
```diff
@@ -1029,12 +1029,11 @@ bsl::ostream&
 FileBackedStorage::logAppsSubscriptionInfoCb(bsl::ostream& stream) const
 {
     if (queue()) {
-        mqbi::Storage::AppIdKeyPairs appIdKeyPairs;
-        loadVirtualStorageDetails(&appIdKeyPairs);
+        mqbi::Storage::AppInfos appInfos;
```
dorjesinpo, line 1032: Please use allocator, when possible

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1619,6 +1619,132 @@ void RootQueueEngine::onTimer(bsls::Types::Int64 currentTimer)
...
+    mqbblp::Routers::QueueRoutingContext& routingContext =
+        appState->routing()->d_queue;
+    mqbcmd::Routing routing;
+    routingContext.loadInternals(&routing);
+    const bsl::vector<mqbcmd::SubscriptionGroup>& subscrGroups =
+        routing.subscriptionGroups();
+    if (!subscrGroups.empty()) {
+        // Limit to log only k_EXPR_NUM_LIMIT expressions
+        static const size_t k_EXPR_NUM_LIMIT = 50;
+        bmqu::MemOutStream  ss(d_allocator_p);
```
dorjesinpo, line 1679: Why not output to the `stream` directly and avoid using `ss`

dorjesinpo (changes requested):

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
alexander-e1off: In line 1692 there is checking of expressions number and there is different caption depending on check result to explain user that limited number of expressions is printed, before printing list of expressions. Is there better way to get number of non-empty subscriptions?

src/groups/mqb/mqbs/mqbs_filebackedstorage.cpp
alexander-e1off: Added allocator usage

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
dorjesinpo: Can we not use `subscrGroups.size()` upfront and log the header line before iterating?

alexander-e1off: I noticed that subscrGroups container is not empty even if there is no subscriptions, but it contains empty expressions (that's why I use `if (!cIt->expression().empty())` and `exprNum` counter to get real number of expressions.

alexander-e1off: @dorjesinpo got rid of using interim `ss`, added printing of empty subscription. Tested that there is no duplication in `subscrGroups`.

dorjesinpo (approved):
Thank you

PR #477, opened 2024-10-25
Perf[MQB]: do not build temporary functors for every routed message

678098:
Should save 1% out of 36.4% of the queue dispatcher thread and reduce the total number of allocations.

Before:
Screenshot 2024-10-25 at 19 08 45

After:
Screenshot 2024-10-25 at 23 27 28

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
@@ -637,6 +637,17 @@ QueueEngineUtil_AppsDeliveryContext::QueueEngineUtil_AppsDeliveryContext(
 , d_currentMessage(0)
 , d_queue_p(queue)
 , d_timeDelta()
+, d_currentAppView_p(0)
+, d_visitVisitor(
+      bdlf::BindUtil::bindS(allocator,
+                            &QueueEngineUtil_AppsDeliveryContext::visit,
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -637,6 +637,17 @@ QueueEngineUtil_AppsDeliveryContext::QueueEngineUtil_AppsDeliveryContext(
...
+, d_visitVisitor(
+      bdlf::BindUtil::bindS(allocator,
+                            &QueueEngineUtil_AppsDeliveryContext::visit,
+                            this,
+                            bdlf::PlaceHolders::_1))
+, d_broadcastVisitor(bdlf::BindUtil::bindS(
+      allocator,
+      &QueueEngineUtil_AppsDeliveryContext::visitBroadcast,
+      this,
+      bdlf::PlaceHolders::_1))
```
678098, line 650: The alternative to this is to made `Visitor` functor and inherit from it locally.
Like here:
https://github.com/bloomberg/blazingmq/pull/481

dorjesinpo: To implement new DeliveryContext interface...

dorjesinpo (approved):

PR #478, opened 2024-10-25
Perf[BMQ]: do not create tmp functors in PutEventBuilder

678098:
Constructing local `bdlb::ScopeExitAny` makes a copy of a functor and allocates memory.

Before:
Screenshot 2024-10-25 at 22 09 03

After:
Screenshot 2024-10-26 at 00 39 49

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,9 @@ settings.json
 .cache/
 .venv/
 
+# fuzz-test output folder
+**/boofuzz-results
+
 # Symlink from 'src/applications/bmqbrkr/run'
 src/applications/bmqbrkr/etc/etc
 
--- a/src/groups/bmq/bmqp/bmqp_puteventbuilder.cpp
+++ b/src/groups/bmq/bmqp/bmqp_puteventbuilder.cpp
@@ -49,16 +49,31 @@ bool isWordAligned(const bdlbb::Blob& blob)
 #endif
 }  // close unnamed namespace
 
+// ---------------------------------
+// class PutEventBuilder::ResetGuard
+// ---------------------------------
+
+inline PutEventBuilder::ResetGuard::ResetGuard(
+    PutEventBuilder& putEventBuilder)
+: d_putEventBuilder(putEventBuilder)
+{
+    // NOTHING
+}
+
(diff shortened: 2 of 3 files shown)


dorjesinpo (commented):
Strictly speaking, moving the `reset` to `packMessageInternal` is not fully equivalent to the existing functionality as the scope gets reduced.

678098:
> Strictly speaking, moving the `reset` to `packMessageInternal` is not fully equivalent to the existing functionality as the scope gets reduced.

I believe, all the previous code paths used `packMessageInternal` in the end.

We call it on any return in `PutEventBuilder::packMessageInOldStyle`, `PutEventBuilder::packMessage` and `PutEventBuilder::packMessageRaw`

dorjesinpo:
> > Strictly speaking, moving the `reset` to `packMessageInternal` is not fully equivalent to the existing functionality as the scope gets reduced.
> 
> I believe, all the previous code paths where I removed `bdlb::ScopeExitAny` used `packMessageInternal` in the end, on return.
> 
> We call it on any return in `PutEventBuilder::packMessageInOldStyle`, `PutEventBuilder::packMessage` and `PutEventBuilder::packMessageRaw`

That is true _until_ there are no _other_ `return` statements.
Explicit resets in every case seem safer despite looking redundant.  But that is minor.  We can keep it in `packMessageInternal` but then let's add a comment

678098:
> > > Strictly speaking, moving the `reset` to `packMessageInternal` is not fully equivalent to the existing functionality as the scope gets reduced.
> > 
> > 
> > I believe, all the previous code paths where I removed `bdlb::ScopeExitAny` used `packMessageInternal` in the end, on return.
> > We call it on any return in `PutEventBuilder::packMessageInOldStyle`, `PutEventBuilder::packMessage` and `PutEventBuilder::packMessageRaw`
> 
> That is true _until_ there are no _other_ `return` statements. Explicit resets in every case seem safer despite looking redundant. But that is minor. We can keep it in `packMessageInternal` but then let's add a comment

I understand this concern. Let me do the similar thing, but keep guards in the previous places

678098:
@dorjesinpo fixed the PR

dorjesinpo (approved):

PR #482, opened 2024-10-28
Perf[MQB]: throttle logs critical for broadcast

678098:
Logging every 500 cached PUTs expiration not only generates gigabytes of logs, but also potentially limits broadcast speed.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
@@ -453,6 +453,7 @@ RemoteQueue::RemoteQueue(QueueState*       state,
 , d_pendingMessagesTimerEventHandle()
 , d_ackWindowSize(ackWindowSize)
 , d_unackedPutCounter(0)
+, d_expiredPutNum(0)
 , d_subStreams(allocator)
 , d_statePool_p(statePool)
 , d_producerState()
@@ -473,6 +474,9 @@ RemoteQueue::RemoteQueue(QueueState*       state,
         1,
         5 * bdlt::TimeUnitRatio::k_NS_PER_S);
     d_throttledFailedConfirmMessages.initialize(
+        1,
+        5 * bdlt::TimeUnitRatio::k_NS_PER_S);
+    d_throttledCachedPutMessages.initialize(
         1,
         5 * bdlt::TimeUnitRatio::k_NS_PER_S);
     // 1 log per 5s interval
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
```diff
@@ -1340,9 +1349,9 @@ void RemoteQueue::expirePendingMessagesDispatched()
     }
     else {
         d_pendingMessagesTimerEventHandle.release();
-        BALL_LOG_INFO << d_state_p->uri() << ": "
-                      << "no more timer scheduled to check expiration of "
-                      << "pending PUSH messages";
+        BALL_LOG_DEBUG << d_state_p->uri() << ": "
+                       << "no more timer scheduled to check expiration of "
+                       << "pending PUSH messages";
```
678098, line 1354: The same severity as a few lines above

src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
```diff
@@ -1176,12 +1180,17 @@ void RemoteQueue::onAckMessageDispatched(const mqbi::DispatcherAckEvent& event)
...
+        d_expiredPutNum += numErased;
 
         erasePendingMessage(it);
 
-        BALL_LOG_INFO << d_state_p->uri() << ": erased window of " << numErased
-                      << " cached broadcasted PUTs upon "
-                      << bmqt::AckResult::toAscii(ackResult);
+        if (d_throttledCachedPutMessages.requestPermission()) {
+            BALL_LOG_INFO << d_state_p->uri() << ": erased window of "
+                          << numErased << " (" << d_expiredPutNum
```
dorjesinpo, line 1189: `d_expiredPutNum` is more like "total erased broadcast PUTs".  But what is the meaning of this counter; why do we need it?

src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
```diff
@@ -1176,12 +1180,17 @@ void RemoteQueue::onAckMessageDispatched(const mqbi::DispatcherAckEvent& event)
...
         size_t numErased = 1 + erasePendingMessages(it);
+        d_expiredPutNum += numErased;
 
         erasePendingMessage(it);
 
-        BALL_LOG_INFO << d_state_p->uri() << ": erased window of " << numErased
-                      << " cached broadcasted PUTs upon "
-                      << bmqt::AckResult::toAscii(ackResult);
+        if (d_throttledCachedPutMessages.requestPermission()) {
+            BALL_LOG_INFO << d_state_p->uri() << ": erased window of "
```
dorjesinpo, line 1188: This could very well be `BALL_LOG_DEBUG`.  This log does not have much importance.

dorjesinpo (commented):
Looks good ,but maybe can be simplified

src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
678098: If we are throttling logs, we still might want to know how many we of these we skipped. Might be useful for debug, but also I'm okay to remove it.

src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
678098: Will redo it to debug then and remove the counter. Let's keep it simple.

678098:
@dorjesinpo simplified

PR #483, opened 2024-10-28
Perf[BMQIO]: remove unnecessary function copy

678098:
IO thread should spend 7.5% (1.7 / 22.96) less cpu

Before:
Screenshot 2024-10-28 at 16 33 12

After:
Screenshot 2024-10-28 at 17 27 43

### DIFF
--- a/src/groups/bmq/bmqio/bmqio_ntcchannel.cpp
+++ b/src/groups/bmq/bmqio/bmqio_ntcchannel.cpp
@@ -406,7 +406,6 @@ NtcRead::~NtcRead()
     BSLS_ASSERT_OPT(d_numNeeded == 0);
     BSLS_ASSERT_OPT(d_complete);
     BSLS_ASSERT_OPT(!d_timer_sp);
-    BSLS_ASSERT_OPT(!d_callback);
 }
 
 // MANIPULATORS
@@ -447,8 +446,6 @@ void NtcRead::clear()
 
     d_numNeeded = 0;
     d_complete  = true;
-
-    d_callback = bmqio::Channel::ReadCallback();
 }
 
 // ACCESSORS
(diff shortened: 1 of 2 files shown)


src/groups/bmq/bmqio/bmqio_ntcchannel.cpp
```diff
@@ -758,11 +755,19 @@ void NtcChannel::processReadQueueLowWatermark(
 
             int numNeeded = 0;
             {
-                bmqio::Channel::ReadCallback readCallback = read->callback();
+                const bmqio::Channel::ReadCallback& readCallback =
+                    read->callback();
 
                 bslmt::UnLockGuard<bslmt::Mutex> unlock(&d_mutex);
                 readCallback(bmqio::Status(), &numNeeded, &d_readCache);
```
678098, line 762:
```
    bslmt::UnLockGuard<bslmt::Mutex> unlock(&d_mutex);
    // !!!!!!!!! timeout or canceled event might happen
    readCallback(bmqio::Status(), &numNeeded, &d_readCache);
```

https://github.com/bloomberg/blazingmq/blob/8b0c00f58a7336d83d5d083d23eb07bc7045a79c/src/groups/bmq/bmqio/bmqio_ntcchannel.cpp#L645-L655

Here, we have a non-trivial thread race. In the past, we could have emptied the `d_callback` of `NtcRead` between these 2 lines of code. We make a copy of `d_callback` before unlocking a mutex, so we don't access the empty callback. However, it means that we can send another callback from this copy **after** we sent TIMEOUT or EMPTY status.

My PR doesn't fix this race, so we still can send something over this callback. Instead, I make sure not to clear it from another thread, so this reference is always alive. And I also added a post-check for `isComplete()` to return early.

dorjesinpo (changes requested):
Maybe, we can eliminate `setComplete` now and use `clear` instead?

678098:
> Maybe, we can eliminate `setComplete` now and use `clear` instead?

Done. Made sure that we have `clear` down the code paths where I removed `setComplete`.

dorjesinpo (approved):

PR #492, opened 2024-10-30
CI: Add untracked generated documentation files when publishing

pniedzielski:
If generating the documentation creates files that do not already exist on the `gh-pages` branch, the current GitHub Action to generate documentation will not add them.  Instead, it will only update existing files.  This may happen when a new component is added, or when the Doxygen version differs from prior runs.  This bit us on our first run of the documentation-generation GitHub Action, because the version that GitHub Actions uses is different than the version that built our original documentation.

This patch explicitly adds all files generated under the `docs/docs/apidocs/cpp_apidocs` directory, including both untracked files and tracked, modified files.

### DIFF
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -49,5 +49,4 @@ jobs:
           cd docs/
           git config --global user.name "${{ github.actor }}"
           git config --global user.email "${{ github.actor }}@users.noreply.github.com"
-          git commit -s -am "Update C++ API docs from commit ${{ github.sha }} on main"
-          git push
+          git add docs/apidocs/cpp_apidocs/ && git commit -s -am "Update C++ API docs from commit ${{ github.sha }} on main" && git push || true


678098 (commented):
How does this PR perform if there are no changes to commit?

pniedzielski:
> How does this PR perform if there are no changes to commit?

I think the only way this will happen is when we manually run the workflow, because our doxygen invocation looks like `( cat Doxyfile ; echo "PROJECT_NUMBER=${{ github.sha }}" ) | doxygen -`.  Namely, we embed the commit SHA into the version, which is printed to the documentation header.  But, just in case, the behavior is to not commit any changes.

I ran the following commands to verify this:

```
mkdir workspace && cd workspace

# Build the docs, commit, and push to `gh-pages1` branch a first time, using constant version string `v1.0`.  This makes an update.
git clone https://github.com/pniedzielski/blazingmq.git
cd blazingmq/
rm -rf docs/
git clone --branch gh-pages1 https://github.com/pniedzielski/blazingmq.git docs
( cat Doxyfile ; echo "PROJECT_NUMBER=v1.0" ) | doxygen -
cd docs/
git add .
git commit -s -am "Manual update #1"
git push

cd ../..
rm -rf blazingmq/

# Perform same steps as above again.  No changes to `main`, `gh-pages1` in well-known state.
# `doxygen` generation should be idempotent, so should be no changes to commit.git clone https://github.com/pniedzielski/blazingmq.git
cd blazingmq/
rm -rf docs/
git clone --branch gh-pages1 https://github.com/pniedzielski/blazingmq.git docs
( cat Doxyfile ; echo "PROJECT_NUMBER=v1.0" ) | doxygen -
cd docs/
git add .
echo $?  # 0
git commit -s -am "Manual update #2"
echo $?  # 1
git push
echo $?  # 0
```

I believe the `git commit` failing with return code 1 will cause the action to exit (with failure).  I think we want the action to exit like this, but we don't want it to fail:

```
git add . && git commit -s -am "Manual update #2" && git push || true
```

I'll make this change tomorrow.

678098 (approved):

PR #494, opened 2024-10-31
Fix[mqbblp::RemoteQueue]: 0 deduplication timeout causes expired PUTs

678098:
`deduplicationTimeMs` parameter serves at least 2 roles:
- it defines if queue history is saved or not on PRIMARY
- it is used on REPLICA/PROXY to schedule PUT expiration if PRIMARY node is not available

With the second point, we got a bug. When we provide 0 timeout, we schedule PUT expiration almost immediately on REPLICA/PROXY, causing some of the PUTs to expire before reaching PRIMARY.

This PR handles 0 (and just for safety negative values, if someone wants to pass -1) for `deduplicationTimeMs`, and now we assume that "zero deduplication timeout means we don't want to expire messages for as long as we can". We use a default 5 minute timer in this case.

The sample WARN log from PROXY:
`
31OCT2024_17:33:11.257 (140251446843136) WARN mqbblp_remotequeue.cpp:488 Remote queue [@bmq://bmq.capmon.priority.sc/dummy]: cannot schedule PUT expiration timer with a non-positive timeout from config [0 ms], use a default PUT expiration timeout for scheduler instead [5.00 m]
`

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
@@ -366,7 +366,7 @@ void RemoteQueue::pushMessage(
 
             if (result != mqbi::StorageResult::e_SUCCESS) {
                 if (d_throttledFailedPushMessages.requestPermission()) {
-                    BALL_LOG_WARN << d_state_p->uri()
+                    BALL_LOG_WARN << "[THROTTLED] " << d_state_p->uri()
                                   << " failed to store broadcast PUSH ["
                                   << msgGUID << "], result = " << result;
                 }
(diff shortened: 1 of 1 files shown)


src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
```diff
@@ -483,6 +483,19 @@ RemoteQueue::RemoteQueue(QueueState*       state,
     os << '@' << d_state_p->uri().asString();
     d_state_p->setDescription(os.str());
 
+    if (deduplicationTimeMs <= 0) {
+        d_pendingPutsTimeoutNs = 5 *
```
dorjesinpo, line 487: let's have a `k_` constant instead of `5`
Or, the default dedup timeout

src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
```diff
@@ -483,6 +483,19 @@ RemoteQueue::RemoteQueue(QueueState*       state,
...
+                                 bdlt::TimeUnitRatio::k_NANOSECONDS_PER_MINUTE;
+        BALL_LOG_WARN << "Remote queue [" << d_state_p->description()
+                      << "]: cannot schedule PUT deduplication timer with a "
+                      << "non-positive timeout from config ["
+                      << deduplicationTimeMs << " ms], use a default PUT "
+                      << "deduplication timeout for scheduler instead ["
+                      << bmqu::PrintUtil::prettyTimeInterval(
```
dorjesinpo, line 494: let's print the minutes, easier to read?

dorjesinpo (commented):
looks good, minor comments

src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp
678098: Pretty time interval prints `5.00 m`, you want to print the full `5 minutes` instead, right?

dorjesinpo: If this is `5.00 m`, then it is ok.  Just did not want long string of zeroes

678098:
@dorjesinpo I moved 5 to unit's constants. Also, changed the log line so it mentions `expiration`, not `deduplication`.

dorjesinpo (approved):

PR #495, opened 2024-10-31
Fix[MQB]: Enable strong consistency CSL and fix leader activeness

kaikulimu:
1. Enable strong consistency by default everywhere we use CSL. Before, it was in eventual consistency everywhere. I updated the unit tests to also test in strong consistency. Most integration tests pass in strong consistency, except a few related to appIds. This proves the race condition around [`Domain::configure`](https://github.com/bloomberg/blazingmq/blob/527fa18619d64895ef9e2143ee5f399e67cf3430/src/groups/mqb/mqbblp/mqbblp_domain.cpp#L539-L573), which @dorjesinpo fixed in #493.

2. I enhanced the resiliency of the elector logic by prohibiting to set leader status directly from `e_UNDEFINED` to `e_ACTIVE`. It should always go from `e_UNDEFINED -> e_PASSIVE -> e_ACTIVE`. [Here](https://bbgithub.dev.bloomberg.com/yyan82/knowledge/blob/main/blazingmq/storage/ElectorInfo_LeaderStatus.png) is a small diagram of the current triggers to leader status transition. I am adding a new transition from `e_PASSIVE` to `e_ACTIVE` during `ElectorInfo::onHealedFollower()` in this PR. I also fixed the logic of some observers to the leader status via the `onClusterLeader` method.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_cluster.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_cluster.cpp
@@ -3549,10 +3549,13 @@ void Cluster::onClusterLeader(mqbnet::ClusterNode*                node,
     }
 
     d_clusterOrchestrator.updateDatumStats();
-    d_clusterData.stats().setIsLeader(
-        d_clusterData.membership().selfNode() == node
-            ? mqbstat::ClusterStats::LeaderStatus::e_LEADER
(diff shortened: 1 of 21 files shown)


kaikulimu:
@dorjesinpo Ready for review

dorjesinpo (approved):

PR #496, opened 2024-11-01
Fix[MQB]: remove maxDeliveryAttempts range check

emelialei88:
A user uncovered an unfortunate (and undocumented) limitation in the behavior of poison pill configurations, we do not allow users to set this to 1-4 (using the default broker configurations).

We should allow users to configure these as valid values for the number of delivery attempts.

### DIFF
--- a/src/integration-tests/test_poison_messages.py
+++ b/src/integration-tests/test_poison_messages.py
@@ -158,8 +158,8 @@ class TestPoisonMessages:
         # 2. send a message to the consumer
         # 3. open a consumer
         # 4. kill a consumer
-        # 5. force a leader change
-        # 6. open a consumer
+        # 5. open a consumer
+        # 6. force a leader change
         # 7. kill a consumer a again
         # 8. open a consumer
         # The message should still exist and be delivered to the consumer (the
@@ -443,7 +443,7 @@ class TestPoisonMessages:
             multi_node, proxy, tc.DOMAIN_FANOUT, ["?id=foo", "?id=bar", "?id=baz"]
         )
 
-    @max_delivery_attempts(2)
+    @max_delivery_attempts(3)
     def test_poison_rda_reset_priority_active(self, multi_node: Cluster):
         proxies = multi_node.proxy_cycle()
         # pick proxy in datacenter opposite to the primary's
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -35,3 +35,6 @@ src/applications/bmqbrkr/etc/etc
 
 # 'sim_cpp11_features.pl' backups
 *.bak
+
```
pniedzielski, line 38: This may not be the best to commit to the main repo, since it's for your personal usage.  [Git looks in a few place for files to ignore](https://git-scm.com/docs/gitignore): `.gitignore` is one, and is meant to be committed and distributed to everyone; `.git/info/exclude` is another, and only applies to your local repo.

src/integration-tests/test_poison_messages.py
```diff
@@ -443,7 +443,7 @@ def test_poison_proxy_and_replica_fanout(self, multi_node: Cluster):
             multi_node, proxy, tc.DOMAIN_FANOUT, ["?id=foo", "?id=bar", "?id=baz"]
         )
 
-    @max_delivery_attempts(2)
+    @max_delivery_attempts(3)
```
pniedzielski, line 446: It would be good to have a note in the commit message body about why we needed this change, since it was quite subtle.

src/integration-tests/test_poison_messages.py
```diff
@@ -158,8 +158,8 @@ def _crash_consumer_restart_leader(
         # 2. send a message to the consumer
         # 3. open a consumer
         # 4. kill a consumer
-        # 5. force a leader change
-        # 6. open a consumer
+        # 5. open a consumer
```
pniedzielski, line 161: And perhaps also mention this change (maybe this should be in its own commit).

pniedzielski (changes requested):
Good test and change.  I think we should try to squash some of these changes for the sake of the commit history, and maybe rip one change out into its own commit.

pniedzielski (approved):
Awesome, I'll wait for CI to fire one more time before merging, but LGTM

pniedzielski:
Test failure seems unrelated; will look into in a different changeset.

PR #497, opened 2024-11-02
Perf[BMQ,MQB]: do not build tmp functions on ACKs

678098:


dorjesinpo (commented):
Can we not avoid `BuildEventActionFunctor` (and the virtual `operator()()`) if we introduce template parameters to `ProtocolUtil::buildEvent` (`BuildConfirmFunctor` and `BuildAckFunctor`)?

dorjesinpo (approved):
Thank you!  Let's rebase

PR #498, opened 2024-11-03
Perf[BMQ,MQB]: stable GC history preventing allocations

678098:


678098:
>I hadn't considered moving gc into the critical path.

Actually, we do `gc` from the same dispatcher thread, but just more rarely.

>and since a PUT is only one message, any value > 1 will eventually reduce history to the desired size?

That is correct. However, there should also be some overhead of calling `gc`, making some initial checks and going into the loop etc. Need to check if it's inlined or not.

> I'm curious if you've got any measurements on the gc time, do we have a rough number for how long that takes? I suspect it's pretty cheap.

Yes, I think it is cheap, but I will also add benchmarks.

src/groups/mqb/mqbs/mqbs_inmemorystorage.cpp
```diff
@@ -506,15 +508,33 @@ int InMemoryStorage::gcExpiredMessages(
...
+            << "] seconds, garbage-collected [" << numMsgsDeleted
+            << "] messages due to TTL expiration. "
+            << "Timestamp (UTC) of the latest encountered message: "
+            << bdlt::EpochUtil::convertFromTimeT64(latestMsgTimestampEpoch)
+            << " (Epoch: " << latestMsgTimestampEpoch
+            << "). Current time (UTC): " << currentTimeUtc
+            << " (Epoch: " << secondsFromEpoch << ")."
+            << " Num messages remaining in the storage: "
+            << numMessages(mqbu::StorageKey::k_NULL_KEY) << ". Storage type: "
+            << (isPersistent() ? "persistent." : "in-memory.");
```
678098, line 514: I moved these logs from `FileStore` to `FileBackedStorage` and `InMemoryStorage`

dorjesinpo (approved):
let's rebase and merge

PR #504, opened 2024-11-07
Fix C++03 incompatibility from local struct in template parameter

hallfox:
It is apparently against C++03 rules to use locally declared structs in template parameters because their linkage is incomplete. This change moves the local predicate to the anonymous namespace to fix the linkage issue.


chrisbeard (approved):
lgtm

PR #514, opened 2024-11-13
Fix[gh-pages]: Improve website typography

pniedzielski:
This patch makes two changes to typography of our website:

  1. Decrease the weight of bold text from Rubik Bold to Rubik Semibold, which fits better in the body; and

  2. Increase the font size of `code` text to match optically the size of normal body text.

**Before:**
Screenshot 2024-11-13 at 12 45 15 PM

**After:**
Screenshot 2024-11-13 at 12 39 26 PM


678098 (approved):
Looking good

PR #516, opened 2024-11-13
Docs[website]: Remove website files from `main`

pniedzielski:
This patch removes the old website files from our `main` branch, except for those assets that are also used by our Doxygen documentation generation.  Because these files now live on the `gh-pages` branch, having an older, out-of-date copy of these files in `main` is confusing.

Testing
========================================================================

The documentation still builds correctly after applying this patch.


678098 (approved):

pniedzielski:
bmqio_ntcchannel.t failures in ubsan and msan are unrelated to this change.  Continuing with merge.

PR #524, opened 2024-11-18
Fix[schemagen]: Double quote to prevent globbing and work splitting

pniedzielski:
This patch fixes the shellcheck CI failures SC2086 introduced by b39062018c2ef3e6b793aef8fb0e217401bbc642.

### DIFF
--- a/src/python/bin/schemagen
+++ b/src/python/bin/schemagen
@@ -46,13 +46,13 @@ cd "${python_root}"
 function copyright_black {
     echo "$copyright
 
-$(cat $1)" > $1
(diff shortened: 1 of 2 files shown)


678098:
@pniedzielski let's also ad `black` to dependencies?
It was removed here as a transitive dependency.
https://github.com/bloomberg/blazingmq/commit/b39062018c2ef3e6b793aef8fb0e217401bbc642

pniedzielski:
Sounds good, I'll extend this PR to do that

PR #525, opened 2024-11-19
Perf[MQB]: fix linear iteration over all queues on any replication

678098:
- Notify only weak consistency queues on replicated batch, and only ones that were affected by storage event builder flush.
- Remove obsolete `d_hasNewMessages` flag from `mqbblp::LocalQueue`.
- Cancel notifications for queues that changed their consistency from weak to strong on domain reconfigure.
- Bugfix: set `FileBackedStorage::d_hasReceipts` flag on domain reconfigure when consistency changes.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_localqueue.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_localqueue.cpp
@@ -139,6 +139,7 @@ int LocalQueue::configure(bsl::ostream& errorDescription, bool isReconfigure)
         d_state_p->setStorage(storageMp);
     }
     else {
+        d_state_p->storage()->setConsistency(domainCfg.consistency());
         rc = d_state_p->storage()->configure(errorDescription,
                                              domainCfg.storage().config(),
                                              domainCfg.storage().queueLimits(),
--- a/src/groups/mqb/mqbblp/mqbblp_localqueue.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_localqueue.cpp
@@ -530,10 +530,6 @@ void LocalQueue::postMessage(const bmqp::PutHeader&              putHeader,
                 1);
         }
     }
-
-    // If 'FileStore::d_storageEventBuilder' is flushed, flush all relevant
-    // queues (call 'afterNewMessage' to deliver accumulated data)
-    d_state_p->storage()->flushQueues();
 }
 
 void LocalQueue::onPushMessage(
--- a/src/groups/mqb/mqbblp/mqbblp_localqueue.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_localqueue.cpp
@@ -185,6 +185,16 @@ int LocalQueue::configure(bsl::ostream& errorDescription, bool isReconfigure)
         domainCfg.storage().queueLimits().bytes());
 
     if (isReconfigure) {
+        if (domainCfg.consistency().isStrongValue()) {
+            /// We register notifications for weak consistency queues to
+            /// deliver messages after any replication.  For strong consistency
+            /// queues, we start to PUSH only on receipt.  We have to check and
+            /// possibly remove from notifications the queues that changed
+            /// their consistency level from weak to strong.
+            d_state_p->storageManager()->cancelReplicationNotification(
+                d_state_p->partitionId(),
+                d_state_p->key());
+        }
         if (domainCfg.mode().isFanoutValue()) {
             d_state_p->stats().updateDomainAppIds(
                 domainCfg.mode().fanout().appIDs());
--- a/src/groups/mqb/mqbblp/mqbblp_localqueue.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_localqueue.cpp
@@ -500,10 +500,8 @@ void LocalQueue::postMessage(const bmqp::PutHeader&              putHeader,
 
     if (BSLS_PERFORMANCEHINT_PREDICT_LIKELY(res ==
                                             mqbi::StorageResult::e_SUCCESS)) {
-        // Upd
         // Message has been saved in the storage, but we don't indicate the
-        // engine yet of the new message, instead we just update the
-        // 'd_hasNewMessages' flag.  This is because storage (replicated)
+        // engine yet of the new message.  This is because storage (replicated)
         // messages are nagled, and we don't want to indicate to a peer to
         // deliver a particular guid downstream, before actually replicating
         // that message.  So notification to deliver a particular guid
--- a/src/groups/mqb/mqbblp/mqbblp_localqueue.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_localqueue.cpp
@@ -294,7 +294,7 @@ void LocalQueue::configureHandle(
     // to it.  We need to make sure that storage/replication is in sync, and
     // thus, we force-flush the file store.
 
-    d_state_p->storage()->dispatcherFlush(true, false);
+    d_state_p->storage()->flushStorage();
 
     // Attempt to deliver all data in the storage.  Otherwise, broadcast
     // can get dropped if the incoming configure request removes consumers.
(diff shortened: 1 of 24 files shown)


src/groups/mqb/mqbblp/mqbblp_localqueue.cpp
```diff
@@ -515,7 +516,12 @@ void LocalQueue::postMessage(const bmqp::PutHeader&              putHeader,
                                    appData->length());
 
         if (haveReceipt && refCount) {
-            d_hasNewMessages = true;
+            if (!d_hasNewMessages) {
+                BSLS_ASSERT_SAFE(d_notifications_p);
+
+                d_hasNewMessages = true;
+                d_notifications_p->push_back(d_state_p->queue());
```
678098, line 523: We add each queue to the notification list only once, when we switch the flag from false to true

src/groups/mqb/mqbs/mqbs_filestore.h
```diff
@@ -220,6 +220,8 @@ class FileStore : public DataStore {
     typedef StorageCollectionUtil::StorageMapIter      StorageMapIter;
     typedef StorageCollectionUtil::StorageMapConstIter StorageMapConstIter;
 
+    typedef bsl::vector<mqbi::Queue*> ReplicationNotificationVec;
```
678098, line 223: Vector here is better than list because
1. we reserve and reuse its capacity
2. the memory block is continuous

src/groups/mqb/mqbs/mqbs_filebackedstorage.h
```diff
@@ -203,7 +203,7 @@ class FileBackedStorage BSLS_KEYWORD_FINAL : public ReplicatedStorage {
     bmqp::SchemaLearner::Context d_schemaLearnerContext;
     // Context for replicated data.
 
-    const bool d_hasReceipts;
+    bool d_hasReceipts;
```
678098, line 206: We set this flag once and didn't update it before. It is also used here for early return:

https://github.com/bloomberg/blazingmq/blob/c0f8832ce69bebefd5eaaf3637fe95fa8bb9f138/src/groups/mqb/mqbs/mqbs_filebackedstorage.cpp#L216-L221

src/groups/mqb/mqbi/mqbi_storage.h
```diff
@@ -408,6 +408,10 @@ class Storage {
                           const bsls::Types::Int64 messageTtl,
                           const int                maxDeliveryAttempts) = 0;
 
+    /// Set the consistency level associated to this storage to the specified
+    /// `value`.
+    virtual bool setConsistency(const mqbconfm::Consistency& value) = 0;
```
678098, line 413: I decided to introduce a separate setter instead of extending `configure`, because sometimes we call `configure` without setting new consistency.

dorjesinpo (commented):
Looks good, we just need to revisit `cancelReplicationNotification` logic

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -6917,29 +6924,50 @@ void FileStore::dispatcherFlush(bool storage, bool queues)
...
 {
     if (d_storageEventBuilder.messageCount() == 0) {
         // Empty 'd_storageEventBuilder' means it has been flushed and it is a
-        // good time to flush queues.
-        for (StorageMapIter it = d_storages.begin(); it != d_storages.end();
-             ++it) {
-            ReplicatedStorage* rs = it->second;
-            if (rs->queue()) {
-                rs->queue()->onReplicatedBatch();
+        // good time to flush weak consistency queues.
```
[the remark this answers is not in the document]
678098: I removed code duplication and also split `dispatcherFlush` -> `flushStorage`, `flushQueues`

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -6917,29 +6924,50 @@ void FileStore::dispatcherFlush(bool storage, bool queues)
...
+        for (bsl::unordered_set<mqbu::StorageKey>::iterator it =
+                 d_replicationNotifications.begin();
+             it != d_replicationNotifications.end();
+             it++) {
+            StoragesMap::iterator storageIt = d_storages.find(*it);
```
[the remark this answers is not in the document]
678098: Added a comment

src/groups/mqb/mqbs/mqbs_filestore.h
```diff
@@ -876,6 +882,12 @@ class FileStore : public DataStore {
     /// builder is empty (just flushed).
     void notifyQueuesOnReplicatedBatch();
 
+    /// When reconfigure from weak consistency to strong consistency:
+    /// Cancel replication notification (if any) for the specified `queueKey`.
+    /// No effect if `queueKey` is not found in this storage or if there are no
+    /// current registered notifications for this queue.
+    void cancelReplicationNotification(const mqbu::StorageKey& queueKey);
```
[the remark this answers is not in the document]
678098: This early return is helpful because it prevents us from doing this lookup a few lines below:

```
    RecordHandleMap::const_iterator it = d_handles.find(msgGUID);
```

If we can do early exit without accessing `OrderedHashMapWithHistory`, it will be nice.

src/groups/mqb/mqbi/mqbi_storage.h
```diff
@@ -503,11 +507,13 @@ class Storage {
...
-    /// If the specified `storage` is `true`, flush any buffered replication
-    /// messages to the peers.  If the specified `queues` is `true`, `flush`
-    /// all associated queues.  Behavior is undefined unless this node is
-    /// the primary for this partition.
-    virtual void dispatcherFlush(bool storage, bool queues) = 0;
+    /// Flush any buffered replication messages to the peers.  Behaviour is
+    /// undefined unless this cluster node is the primary for this partition.
+    virtual void flushStorage() = 0;
+
+    /// Flush all associated weak consistency queues.  Behaviour is
```
dorjesinpo, line 514: But we flush only if `d_storageEventBuilder` is empty.  See another comment about `flushQueues`

src/groups/mqb/mqbblp/mqbblp_localqueue.cpp
```diff
@@ -538,7 +533,7 @@ void LocalQueue::postMessage(const bmqp::PutHeader&              putHeader,
 
     // If 'FileStore::d_storageEventBuilder' is flushed, flush all relevant
     // queues (call 'afterNewMessage' to deliver accumulated data)
-    d_state_p->storage()->dispatcherFlush(false, true);
+    d_state_p->storage()->flushQueues();
```
dorjesinpo, line 536: This (already existing) logic is a bit unclear.  Please, check the following reasoning:
1.  `LocalQueue::postMessage` calls `put` which adds to the builder in the `FileStore`.
2.  `FileStore` tracks the builder size and  does `FileStore::flushIfNeeded` which used to call `notifyQueuesOnReplicatedBatch`
3.  Then, why do we need this call `dispatcherFlush(false, true)`/`flushQueues()` here?

If we don't, then there is no need to have `flushQueues` as an interface method.  It is entirely private business of `FileStore` (and something like `notifyQueuesOnReplicatedBatch` seems like a better name)

dorjesinpo (commented):
This is getting better and better.
One more concern about `flushQueue` API.

src/groups/mqb/mqbblp/mqbblp_localqueue.cpp
678098: That is correct, this call is unnecessary. Removed & simplified

src/groups/mqb/mqbi/mqbi_storage.h
678098: Updated the comment

src/groups/mqb/mqbs/mqbs_datastore.h
```diff
@@ -684,9 +684,11 @@ class DataStore : public mqbi::DispatcherClient {
 
     /// If the specified `storage` is `true`, flush any buffered replication
     /// messages to the peers.  If the specified `queues` is `true`, `flush`
-    /// all associated queues.  Behavior is undefined unless this node is
-    /// the primary for this partition.
-    virtual void dispatcherFlush(bool storage, bool queues) = 0;
+    /// all associated queues.
```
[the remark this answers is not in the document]
678098: Will clean this up

PR #527, opened 2024-11-21
IT[BMQ]: add tests to delete domain on disk

emelialei88:


emelialei88:
This issue is solved by adding a new admin command DOMAINS REMOVE

PR #528, opened 2024-11-24
fix PushStream & VirtualStorage ordinal

dorjesinpo:
Fixing bugs related to the Monolithic Storage  

1) VS ordinal recycling ([this](https://github.com/bloomberg/blazingmq/blob/e86bea84c894c70c19da5f12a69b28953ae20471/src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.cpp#L430) assert)
2) Empty `PushStream::Element` in `RelayQueueEngine::beforeMessageRemoved` ([this](https://github.com/bloomberg/blazingmq/blob/e86bea84c894c70c19da5f12a69b28953ae20471/src/groups/mqb/mqbblp/mqbblp_pushstream.cpp#L197) assert)

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_pushstream.h
+++ b/src/groups/mqb/mqbblp/mqbblp_pushstream.h
@@ -194,9 +194,11 @@ struct PushStream {
     void add(Element* element);
 
     /// Remove the specified `element` from both GUID and App corresponding to
-    /// the `element` (and specified when constructing the `element`).
-    /// Return the number of remaining Elements in the corresponding GUID.
-    unsigned int remove(Element* element);
+    /// the `element` (and specified when constructing the `element`).  If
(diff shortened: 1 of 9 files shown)


src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -673,7 +660,15 @@ inline unsigned int PushStream::remove(Element* element)
     // remove from the guid
     element->guid().remove(element, e_GUID);
 
-    return element->guid().numElements();
+    if (element->app().d_elements.numElements() == 0) {
```
pniedzielski, line 663: Centralizing this bit of logic as here seems right.  I've traced through the code paths leading up to this and I think this catches all places we need.

pniedzielski (approved):

PR #529, opened 2024-11-24
Feat[bmqstoragetool]: PrintManager

waldgange:
In order to make the output of the storagetool more grep-friendly or/and to be able to parse it the following changes have been done:
- a separate entity responsible for printing has been indroduced - the `PrintManager`
- the existing printing logic has been moved to one of the implementations of the `PrintManager`
- two new implementations of the `PrintManager` have been added: `JsonLine` (each Record is printed as a single-line json) and `JsonPretty` (each Record is printed as a multi-line json).
- a new parameter to specify the print mode has been added:
```
--print-mode             <print mode> 
          can be one of the following {human|json-prety|json-line}
          (default: human)
```


src/applications/bmqstoragetool/m_bmqstoragetool_recordprinter.h
```diff
@@ -0,0 +1,61 @@
...
+
+namespace BloombergLP {
+
+namespace m_bmqstoragetool {
+
+// ==================
+// class PrintManager
+// ==================
+
+class PrintManager {
```
syuzvinsky, line 38: I would never call this interface PrintManager because its name is too abstract even for an interface :).
Consider something like `MessagePrinter` or `MessagePrinterStrategy`.

src/applications/bmqstoragetool/bmqstoragetool.m.cpp
```diff
@@ -57,6 +57,12 @@ parseArgs(CommandLineArguments& arguments, int argc, const char* argv[])
          "path to a .bmq_csl file",
          balcl::TypeInfo(&arguments.d_cslFile),
          balcl::OccurrenceInfo::e_OPTIONAL},
+        {"print-mode",
+         "print mode",
+         "can be one of the following: [HUMAN, JSON_PRETTY, JSON_LINE]. "
+         "Defailt value is HUMAN",
```
syuzvinsky, line 63:
```suggestion
         "Default value is HUMAN",
```

src/applications/bmqstoragetool/bmqstoragetool.m.cpp
```diff
@@ -57,6 +57,12 @@ parseArgs(CommandLineArguments& arguments, int argc, const char* argv[])
          "path to a .bmq_csl file",
          balcl::TypeInfo(&arguments.d_cslFile),
          balcl::OccurrenceInfo::e_OPTIONAL},
+        {"print-mode",
+         "print mode",
+         "can be one of the following: [HUMAN, JSON_PRETTY, JSON_LINE]. "
```
syuzvinsky, line 62: Does the tool support lower-case values for this parameter? If not, it would be handy to have.

src/applications/bmqstoragetool/m_bmqstoragetool_filters.cpp
```diff
@@ -44,8 +44,10 @@ Filters::Filters(const bsl::vector<bsl::string>& queueKeys,
         mqbu::StorageKey                         key;
         bsl::vector<bsl::string>::const_iterator uriIt = queueUris.cbegin();
         for (; uriIt != queueUris.cend(); ++uriIt) {
-            if (queueMap.findKeyByUri(&key, *uriIt)) {
-                d_queueKeys.insert(key);
+            bsl::optional<mqbu::StorageKey> key = queueMap.findKeyByUri(
```
alexander-e1off, line 47: Do we need `optional` here? [Parameters::validateQueueNames](https://github.com/bloomberg/blazingmq/blob/main/src/applications/bmqstoragetool/m_bmqstoragetool_parameters.cpp#L449) guaranties that queue name is present in `queueMap`.

src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.cpp
```diff
@@ -76,6 +76,9 @@ int moveToLowerBound(mqbs::JournalFileIterator* jit,
             }
             left = jit->recordIndex();
         }
+        static int i = 0;
+        bsl::cout << (++i) << " advance(" << bsl::max((right - left) / 2, 1ULL)
```
alexander-e1off, line 80: Debug?

src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp
```diff
@@ -2068,6 +2356,12 @@ static void test24_summaryWithQueueDetailsTest()
     bmqtst::TestHelper::printTestName(
         "OUTPUT SUMMARY WITH QUEUE DETAILS TEST");
 
+    bmqtst::TestHelperUtil::ignoreCheckDefAlloc() = true;
```
alexander-e1off, line 2359: Debug?

src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp
```diff
@@ -2068,6 +2356,12 @@ static void test24_summaryWithQueueDetailsTest()
...
+    // Disable default allocator check for this test because
+    // EXPECT_CALL(PrinterMock::printRecordSummary(bsls::Types::Uint64,
```
alexander-e1off, line 2361: Debug?

src/applications/bmqstoragetool/m_bmqstoragetool_printer.cpp
```diff
@@ -0,0 +1,878 @@
...
+        return;  // RETURN
+    }
+
+    const bsl::vector<const char*> fields = {"BlazingMQ File Header",
+                                             "Journal File Header",
+                                             "Journal SyncPoint"};
+
+    PRINTER_TYPE1 printer(ostream, &fields);
+    {
+        bsl::ostringstream s;
```
alexander-e1off, line 115:
```suggestion
        bsl::ostringstream s(allocator);
```

src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.cpp
waldgange: Yes. I removed this

src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp
waldgange: No. It's a useful comment. It describes what causes the default allocator use.

src/applications/bmqstoragetool/m_bmqstoragetool_filters.cpp
waldgange: Are you suggesting to change the code to something like this:
`d_queueKeys.insert(queueMap.findKeyByUri(&key, *uriIt).value())` ?
Without checking if the optional really has the value? I agree that all the URIs are meant to be valid at this point. But are we winning anything with this potentially dangerous code? A temporary optional inside the function will be constructed anyway. But RVO will prevent it from extra copying to outside the function. The second copying will happen when we insert the value. The only way to avoid it is to return a const reference (which is not convinient, as we can't check if it's valid or dangling without an extra bool variable) or a pointer (which is not safe to return) to the internal value of the `queueMap`. Also `mqbu::StorageKey` is a very lightweight object. `sizeof(mqbu::StorageKey)` is only 5 bytes, `sizeof(bsl::optional<mqbu::StorageKey>)` is 6 bytes which is less than even a pointer. So it's not worth optimizing anythng here.

pniedzielski (approved):
I think this is as minimal a change as we can reasonably make, and it not just adds functionality but improves the code it touches.

I found just one minor copy-paste error.  I'm going to land this and make a quick PR, assign it to you, so we can turn this around faster.

pniedzielski:
msan CI failure unrelated to this PR.  We should fix it, but not letting this failure block this PR.

PR #530, opened 2024-11-25
CI: daily dependencies build in a separate workflow

678098:
Sometimes the packages used to build dependencies in the runner system change, and it causes the dependencies to rebuild when restored from the old cache. There is no way to rewrite the existing dependencies cache (they are immutable), so instead we generate dependency caches on a daily basis.

This PR:
- Adds the current date to a cache key, so cache rebuilds daily. It should allow to save time on other builds during that day.
- Moves dependencies build to a separate workflow file.

This dependency cache was used over the last 3 months and we spent 8 minutes to rebuild it on any PR push anyway:

Screenshot 2024-11-25 at 18 28 31

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,64 +13,22 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build_dependencies:
-    name: Build deps [ubuntu]
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - name: Get dependencies hash
-        id: get-hash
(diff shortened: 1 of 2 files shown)


678098:
> I'm surprised that moving the Ubuntu dependency installation into a different workflow works, but I guess it's that `get_dependencies` being its own job. I'm going to let the remaining tests pass before approving, but I think even now it's fairly clear it works.

This is called reusable (callable) workflow: https://docs.github.com/en/actions/sharing-automations/reusing-workflows

It runs in its own runner, so there are only a few ways to exchange binaries between the called one and the caller. Cache is okay for our use-case.

pniedzielski (approved):

PR #532, opened 2024-11-26
CI: rebuild deps cache from the main branch nightly

678098:
Caches generated from non-default branches are not cross-accessible, so it has to be done from the main branch.

We will do it at `0:10 UTC` at the beginning of a new day.


pniedzielski (approved):
👍🏻

PR #533, opened 2024-11-27
refactor: Move bmqtst globals into a helper struct

hallfox:
This is causing some conflicting symbol definitions with users of the old mwctst package. This should only be affecting test files and implementations of components in bmqtst.


pniedzielski (approved):
👍🏻

PR #535, opened 2024-12-03
refactor: Rename gtest conflicting test macros

hallfox:
bmqtst provides several gtest-like compatibility macros, but this prevents us from using gtest or gmock directly because it uses the exact same names as the test macros from that library. Since we want to support at least using gmock in our tests, we rename them to avoid conflict.


pniedzielski:
@hallfox DCO check failure and formatting check failure.  Could you do the clang formatting fixes in another commit, so I can still easily review the first commit?

pniedzielski (changes requested):
Looks good.  I see three remaining `ASSERT`s outside of `bmqst`; they don't prevent building but here they are:

```
src/groups/bmq/bmqp/bmqp_puteventbuilder.t.cpp
2281:    ASSERT(rawEvent.isValid());
2282:    ASSERT(rawEvent.isPutEvent());
2288:    ASSERT(putIter.isValid());
```

pniedzielski (approved):
👍🏻

PR #537, opened 2024-12-05
test_reconfigure_domains.py: Improve comments and readability

kaikulimu:
Subtle improvements, and use strong consistency domain instead of eventual consistency

### DIFF
--- a/src/integration-tests/test_reconfigure_domains.py
+++ b/src/integration-tests/test_reconfigure_domains.py
@@ -33,8 +33,8 @@ pytestmark = order(6)
 
 INITIAL_MSG_QUOTA = 10
 
-URI_PRIORITY_1 = f"bmq://{tc.DOMAIN_PRIORITY}/abcd-queue"
-URI_PRIORITY_2 = f"bmq://{tc.DOMAIN_PRIORITY}/qrst-queue"
+URI_PRIORITY_1 = f"bmq://{tc.DOMAIN_PRIORITY_SC}/abcd-queue"
+URI_PRIORITY_2 = f"bmq://{tc.DOMAIN_PRIORITY_SC}/qrst-queue"
 
 
 class TestReconfigureDomains:
@@ -49,8 +49,8 @@ class TestReconfigureDomains:
         self.reader.open(URI_PRIORITY_1, flags=["read"], succeed=True)
         self.reader.open(URI_PRIORITY_2, flags=["read"], succeed=True)
 
-    # Instruct 'writer to 'POST 'n' messages to the domain.
-    # Returns 'True' if all of them succeed, and a false-y value otherwise.
+    # Instruct writer to POST `n` messages to the `uri`.  Return true if all of
+    # them succeed, else a false-y value.
     def post_n_msgs(self, uri, n):
         results = (
             self.writer.post(uri, payload=[f"msg{i}"], wait_ack=True)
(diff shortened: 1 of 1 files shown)


src/integration-tests/test_reconfigure_domains.py
```diff
@@ -33,8 +33,8 @@
 
 INITIAL_MSG_QUOTA = 10
 
-URI_PRIORITY_1 = f"bmq://{tc.DOMAIN_PRIORITY}/abcd-queue"
-URI_PRIORITY_2 = f"bmq://{tc.DOMAIN_PRIORITY}/qrst-queue"
+URI_PRIORITY_1 = f"bmq://{tc.DOMAIN_PRIORITY_SC}/abcd-queue"
+URI_PRIORITY_2 = f"bmq://{tc.DOMAIN_PRIORITY_SC}/qrst-queue"
```
678098, line 37: Don't we want to test non-SC domain here anymore?
I am thinking if we can parametrize this with tweaks.
I am also ok if we just switch to SC here without extra work, or we can add an issue to support non-SC if it's worth it

src/integration-tests/test_reconfigure_domains.py
```diff
@@ -49,24 +49,28 @@ def setup_cluster(self, cluster: Cluster):
...
         )
         return all(res == Client.e_SUCCESS for res in results)
 
-    # Helper method which tells 'leader' to reload the domain config.
-    def reconfigure_to_n_msgs(self, cluster: Cluster, num_msgs, leader_only=True):
+    # Instruct the leader to reconfigure domain config in `cluster` to have a
+    # limit of `max_num_msgs`.  If 'leader_only' is true, the 'DOMAINS
+    # RECONFIGURE' command will only be issued to the leader.  Return true if
+    # reconfigure succeeds, else false.
+    def reconfigure_to_limit_n_msgs(self, cluster: Cluster, max_num_msgs):
```
678098, line 65:
```suggestion
    def reconfigure_to_limit_n_msgs(self, cluster: Cluster, max_num_msgs: int) -> bool:
```
Might add more type hints

src/integration-tests/test_reconfigure_domains.py
```diff
@@ -84,7 +88,7 @@ def test_reconfigure_domain_message_limits(self, multi_node: Cluster):
         assert not self.post_n_msgs(URI_PRIORITY_2, 1)
 
         # Modify the domain configuration to hold 2 more messages.
```
678098, line 90: Is this comment not correct?
```suggestion
        # Modify the domain configuration to hold 10 more messages.
```

src/integration-tests/test_reconfigure_domains.py
```diff
@@ -49,24 +49,28 @@ def setup_cluster(self, cluster: Cluster):
...
-            tc.DOMAIN_PRIORITY
-        ].definition.parameters.storage.domain_limits.messages = num_msgs
+            tc.DOMAIN_PRIORITY_SC
+        ].definition.parameters.storage.domain_limits.messages = max_num_msgs
         return cluster.reconfigure_domain(
-            tc.DOMAIN_PRIORITY, leader_only=leader_only, succeed=True
+            tc.DOMAIN_PRIORITY_SC, leader_only=True, succeed=True
         )
 
+    # Verify that reconfiguring domain message limits work as expected.
```
678098, line 73:
```suggestion
    # Verify that reconfiguring domain message limits works as expected.
```
Not sure about this

src/integration-tests/test_reconfigure_domains.py
```diff
@@ -119,6 +123,7 @@ def test_reconfigure_domain_message_limits(self, multi_node: Cluster):
         self.reader.confirm(URI_PRIORITY_1, "+1", succeed=True)
         assert self.post_n_msgs(URI_PRIORITY_2, 1)
 
+    # Verify that reconfiguring queue message limits work as expected.
```
678098, line 126:
```suggestion
    # Verify that reconfiguring queue message limits works as expected.
```
Also not sure here

678098 (commented):
A few comments

src/integration-tests/test_reconfigure_domains.py
kaikulimu: The problem of parameterize testing non-SC and SC domains exist for many integration test files. Will create an issue to support all of them

678098: Is there a significant difference in the broker logics so we are better to support both? If we don't expect differences in the logics, might just leave SC-one

kaikulimu: There is minimal difference in logics. I would advocate always testing SC mode, as non-SC is considered legacy.

kaikulimu: The difference is SC domains are more asynchronous as they need to wait for quorum of receipts from replicas. I will create an issue.

kaikulimu: https://github.com/bloomberg/blazingmq/issues/543

kaikulimu:
@678098 Back to you

kaikulimu:
@678098 Back to you

678098 (approved):
LGTM

PR #538, opened 2024-12-05
Refactor[BMQ,MQB]: remove redundant virtual, s_allocator_p

678098:
1. Remove `s_allocator_p` mentions from the code samples.
2. Put keyword `final` where it was missed.
3. Remove redundant `virtual` on methods that already marked as `override`.
See also: https://learn.microsoft.com/en-us/cpp/code-quality/c26435?view=msvc-170
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final


pniedzielski (approved):
For the first patch, I have no strong opinions either way (with or without `virtual`s in the derived classes).  But verified that these `virtual`s are redundant.

Second patch, 👍🏻 good changes

PR #541, opened 2024-12-09
Feat[BMQ]: Add an admin command to remove a domain

emelialei88:
The existing admin command "DOMAINS DOMAIN <name> PURGE" only add purge record to the journal files for the queues belonging to this domain. This may cause race condition where it's possible for a user to connect to a queue that's purged.

This PR adds a new admin command to:
- Reject all incoming messages related to this domain
- Purge the queues on this domain
- Clear DomainResolver cache
- Purge existing data on the queues
- Force expire the queues


src/groups/mqb/mqbcmd/mqbcmd_messages.h
```diff
@@ -53827,3 +53866,10 @@ inline bool InternalResult::isUndefinedValue() const
 // GENERATED BY @BLP_BAS_CODEGEN_VERSION@
 // USING bas_codegen.pl -m msg --noAggregateConversion --noExternalization
 // --noIdent --package mqbcmd --msgComponent messages mqbcmd.xsd
+// ----------------------------------------------------------------------------
+// NOTICE:
+//      Copyright 2024 Bloomberg Finance L.P. All rights reserved.
+//      Property of Bloomberg Finance L.P. (BFLP)
+//      This software is made available solely pursuant to the
+//      terms of a BFLP license agreement which governs its use.
+// ------------------------------- END-OF-FILE --------------------------------
```
678098, line 53875: And this autogenerated license needs to be removed

src/groups/mqb/mqbcmd/mqbcmd_parseutil.cpp
```diff
@@ -564,6 +570,34 @@ int parseDomainReconfigure(DomainReconfigure* reconfigure,
...
+
+    if (domain.empty()) {
+        *error = "DOMAINS REMOVE command must be followed by a "
+                 "domain name.";
+        return -1;  // RETURN
+    }
+
+    remove->domain() = domain;
+
+    if (equalCaseless(finalize, "finalize")) {
```
pniedzielski, line 589: Probably `"FINALIZE"` here too, even though functionally it's the same with `equalCaseless`

src/groups/mqb/mqbcmd/mqbcmd_parseutil.cpp
```diff
@@ -564,6 +570,34 @@ int parseDomainReconfigure(DomainReconfigure* reconfigure,
...
+        remove->finalize().makeValue(true);
+    }
+    else if (!finalize.empty()) {
+        *error = "Invalid optional key word '" + finalize +
+                 "' for DOMAINS REMOVE <domain> [finalize]";
```
pniedzielski, line 594: but definitely `[FINALIZE]` here.

src/groups/mqb/mqba/mqba_domainmanager.cpp
```diff
@@ -702,6 +702,124 @@ int DomainManager::processCommand(mqbcmd::DomainsResult*        result,
             return 0;  // RETURN
         }
     }
+    else if (command.isRemoveValue()) {
+        const bsl::string& name = command.remove().domain();
+
+        DomainSp domainSp;
+
+        if (0 != locateOrCreateDomain(&domainSp, name)) {
```
pniedzielski: We talked about this, but just to record it down: the reason we do this here is due to the lazy loading of domains that BlazingMQ does.  We don't want to fail if a user creates a domain configuration file, happens not to connect, and then issues this admin command, or creates a domain configuration file, uses the domain, purges the domain, and restarts the broker before issuing this admin command.

pniedzielski: But, this will load the domain into the resolver cache, so we do need to clear that on the second pass.

src/groups/mqb/mqba/mqba_domainmanager.cpp
```diff
@@ -702,6 +702,124 @@ int DomainManager::processCommand(mqbcmd::DomainsResult*        result,
...
+            bmqu::MemOutStream os;
+            os << "Trying to remove a nonexistent domain '" << name << "'";
+            result->makeError().message() = os.str();
+            return -1;  // RETURN
+        }
+
+        // First pass
+        if (command.remove().finalize().isNull()) {
+            BALL_LOG_INFO << "[First pass] DOMAINS REMOVE '" << name
+                          << "' called!!!";
```
pniedzielski: Not sure we need the triple exclamation mark exuberance in the logs

src/groups/mqb/mqba/mqba_domainmanager.cpp
```diff
@@ -702,6 +702,124 @@ int DomainManager::processCommand(mqbcmd::DomainsResult*        result,
...
+            // 1. Reject if there's any opened queue
+            if (domainSp->hasActiveQueue()) {
+                bmqu::MemOutStream os;
+                os << "Trying to remove the domain '" << name
+                   << "' while there are queues open";
+                result->makeError().message() = os.str();
+                return -1;  // RETURN
+            }
+
+            BALL_LOG_INFO << "BEFORE CHEKCING CLUSTER STATUS";
```
pniedzielski: Not sure we need these logs either, they look like they were just for debugging.

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -441,10 +441,12 @@ void Cluster::sendAck(bmqt::AckResult::Enum     status,
             d_throttledFailedAckMessages,
             BALL_LOG_INFO << description() << ": failed Ack "
                           << "[status: " << status << ", source: '" << source
-                          << "'" << ", correlationId: " << correlationId
+                          << "'"
```
pniedzielski: Are these clang-format related changes?

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
```diff
@@ -784,7 +784,7 @@ void ClusterProxy::processEvent(const bmqp::Event&   event,
                        << "Received unexpected event: " << event;
         BSLS_ASSERT_SAFE(false && "Unexpected event received");
         return;  // RETURN
-    }  // break;
+    }            // break;
```
pniedzielski: These are strange changes---again, clang-format ones?  (and do we need them to pass CI?)

src/groups/mqb/mqbmock/mqbmock_domain.cpp
```diff
@@ -149,6 +149,16 @@ int Domain::processCommand(mqbcmd::DomainResult*        result,
     return -1;
 }
 
+void Domain::removeDomainStart()
+{
+    BSLS_ASSERT_SAFE(false && "NOT IMPLEMENTED!");
```
pniedzielski, line 160: I think these may be better as `// NOTHING`.  This is a mock, so it probably shouldn't assert when two of its functions are called.

src/groups/mqb/mqbmock/mqbmock_queue.cpp
```diff
@@ -527,6 +527,13 @@ bmqp::SchemaLearner& Queue::schemaLearner() const
     return d_schemaLearner;
 }
 
+bool Queue::hasActiveHandle() const
+{
+    BSLS_ASSERT_SAFE(false && "NOT IMPLEMENTED!");
```
pniedzielski, line 532: Same here.

src/integration-tests/test_domain_remove.py
```diff
@@ -0,0 +1,277 @@
+# Copyright 2024 Bloomberg Finance L.P.
```
pniedzielski, line 1: New file, so 2025!

src/integration-tests/test_domain_remove.py
```diff
@@ -0,0 +1,277 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""
+This suite of test cases verifies the admin command
+"DOMAINS REMOVE <domain> [finalize]" work as expected
```
pniedzielski, line 19: `[FINALIZE]`

src/groups/mqb/mqba/mqba_commandrouter.cpp
```diff
@@ -315,6 +315,10 @@ void CommandRouter::setCommandRoutingMode()
             d_routingModeMp.load(new (*allocator) ClusterWideRoutingMode());
             // DOMAINS RECONFIGURE <domain>
         }
+        else if (domains.isRemoveValue()) {
+            d_routingModeMp.load(new (*allocator) ClusterWideRoutingMode());
+            // DOMAINS REMOVE <domain> [finalize]
```
pniedzielski, line 320: `[FINALIZE]`

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -3694,21 +3691,41 @@ void Cluster::gcQueueOnDomainDispatched(mqbcmd::ClusterResult* result,
     BSLS_ASSERT_SAFE(dispatcher()->inDispatcherThread(this));
 
     // 'true' implies immediate
-    if (const int rc = d_clusterOrchestrator.queueHelper().gcExpiredQueues(
-            true,
-            domainName)) {
+    const int rc =
+        d_clusterOrchestrator.queueHelper().gcExpiredQueues(true, domainName);
+    if (rc == -1 || rc == -3) {
+        // TBD: We allow the node to not be an active primary for *any*
```
pniedzielski, line 3697: Good note.  Leader != primary will break a lot of assumptions, so it's good to document them as we add them.

src/groups/mqb/mqba/mqba_domainmanager.cpp
```diff
@@ -782,26 +774,50 @@ int DomainManager::processCommand(mqbcmd::DomainsResult*        result,
...
+
+            int rc = locateDomain(&domainSp, name);
+            if (0 != rc) {
+                bmqu::MemOutStream os;
+                os << "Domain '" << name << "' doesn't exist";
+                result->makeError().message() = os.str();
+                return rc;  // RETURN
+            }
+
+            if (!domainSp->isRemoveComplete()) {
```
pniedzielski, line 807: I think we should make this a real failure, with `result->makeError().message()` being something descriptive.

pniedzielski (changes requested):
I think the only substantial change we need to make is to clear the domain resolver cache in the second phase as well (since loading the domain will repopulate the cache).

Great!

src/groups/mqb/mqba/mqba_domainmanager.cpp
emelialei88: Do you mean I should do more than returning `-1`?

src/groups/mqb/mqbcmd/mqbcmd_commandlist.cpp
```diff
@@ -73,6 +73,14 @@ struct CommandDefinition {
      "Clear the domain resolution cache entry of the optionally specified "
      "'domain', or clear all domain resolution cache entries if 'ALL' is "
      "specified."},
+    {"DOMAINS REMOVE <domain> [finalize]",
+     "Remove a domain with an optional keyword 'finalize'",
```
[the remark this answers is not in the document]
emelialei88: 1. We do check the state of the domain to make sure we only continue the second round when the first is completed.
2. Good catch!

pniedzielski:
I think we should add a test-case or two just to verify the change in that final commit.  Then should be good to go.

pniedzielski (approved):
🥳

PR #552, opened 2024-12-20
fix: RemoteQueue handles duplicates

dorjesinpo:
The following _broadcast_ sequence
1.  m1 arrives, gets inserted, does go out, d_appData is null
2.  m2 with the same GUID arrives, does not get inserted (duplicate), does go out
3.  upstream is lost
4.  soft NACK arrives for m1, data goes back to d_appData
5.  soft NACK arrives for m2, d_appData is not null

...results in the [assert](https://github.com/bloomberg/blazingmq/blob/6b34d5283821da0fd4b9264d45011f57eb0bf2a0/src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp#L1193)


pniedzielski (approved):
Change looks good.

PR #508, opened 2024-11-11
Feat[Storagetool]: Add searching by record sequence numbers and offsets

alexander-e1off:
Enhance storage tool to search by composite sequence numbers (primaryLeaseId, sequenceNumber) and offsets:
- Search up to a given sequence number or offset;
- Search starting from a given sequence number or offset;
- Search specific sequence number(s) or offsets;

Below are examples how to use new feature:

Filter messages within composite sequence numbers (primaryLeaseId, sequenceNumber) range:
```bash
bmqstoragetool --journal-file=<path> --seqnum-lt=<leaseId-sequenceNumber>
bmqstoragetool --journal-file=<path> --seqnum-gt=<leaseId-sequenceNumber>
bmqstoragetool --journal-file=<path> --seqnum-lt=<leaseId1-sequenceNumber1> --seqnum-gt=<leaseId2-sequenceNumber2>
```

Filter messages within record offsets range:
```bash
bmqstoragetool --journal-file=<path> --offset-lt=<offset>
bmqstoragetool --journal-file=<path> --offset-gt=<offset>
bmqstoragetool --journal-file=<path> --offset-lt=<offset1> --offset-gt=<offset2>
```
Filter messages with corresponding composite sequence numbers (defined in form `primaryLeaseId-sequenceNumber`):
```bash
bmqstoragetool --journal-file=<path> --seqnum=<leaseId-sequenceNumber_1> --seqnum=<leaseId-sequenceNumber_N>
```
Filter messages with corresponding record offsets:
```bash
bmqstoragetool --journal-file=<path> --offset=<offset_1> --offset=<offset_N>
```

PR #560, opened 2025-01-09
Fix[Storagetool]: use bsl::optional for `Parameters::Range` class members

alexander-e1off:
Addressing postponed review comment https://github.com/bloomberg/blazingmq/pull/508#discussion_r1869935596 - use `bsl::optional` for `Range` class members.


pniedzielski (approved):
Looks great!

PR #561, opened 2025-01-09
Fix[BMQ,MQB]: ensure Solaris build

678098:
1. Fix return value optimization problem with Solaris in `bmqp::BlobPoolUtil::createBlobPool`, by returning a shared pointer to the blob pool. The choice of shared pointer type is made to do it compatible with resource manager PR: https://github.com/bloomberg/blazingmq/pull/499
2. Update all places where blob shared pointer pool is used.
3. `bmqstoragetool`: remove incompatible initializer list usage.
4. `bmqio_ntcchannel.t`: disambiguate argument for `ntcf::System::createInterface` call.
5. Logger classes: explicitly implement `publish` override to get rid of a warning.
6. `bmqu::BlobIterator` remove unused `const char* operator->() const;` that produces a compilation warning `Warning: Questionable (non-class) return type for BloombergLP::bmqu::BlobIterator::operator->() const.`
7. Replace unrecognized `nullptr` usage by `NULL` or `bsl::nullptr_t()` where applicable.
8. `mqbstat`: pass enum template arguments in a way acceptable by Solaris compiler (`BrokerStats::EventType::Enum::e_CLIENT_DESTROYED` -> `BrokerStats::EventType::e_CLIENT_DESTROYED`).
9. storage tool: set allocator flag via `bmqtst::TestHelperUtil::ignoreCheckDefAlloc()`

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_blobpoolutil.cpp
+++ b/src/groups/bmq/bmqp/bmqp_blobpoolutil.cpp
@@ -51,15 +51,13 @@ BlobPoolUtil::createBlobPool(bdlbb::BlobBufferFactory* blobBufferFactory_p,
 
     bslma::Allocator* alloc = bslma::Default::allocator(allocator);
 
-    return bsl::shared_ptr<BlobSpPool>(
-        new BlobSpPool(
-            bdlf::BindUtil::bind(&createBlob,
-                                 blobBufferFactory_p,
-                                 bdlf::PlaceHolders::_1,   // arena
-                                 bdlf::PlaceHolders::_2),  // allocator
-            k_BLOB_POOL_GROWTH_STRATEGY,
-            alloc),
-        alloc);
+    return bsl::allocate_shared<BlobSpPool>(
+        alloc,
+        bdlf::BindUtil::bind(&createBlob,
+                             blobBufferFactory_p,
+                             bdlf::PlaceHolders::_1,   // arena
+                             bdlf::PlaceHolders::_2),  // allocator
+        k_BLOB_POOL_GROWTH_STRATEGY);
 }
 
 }  // close package namespace
--- a/src/groups/bmq/bmqp/bmqp_blobpoolutil.cpp
+++ b/src/groups/bmq/bmqp/bmqp_blobpoolutil.cpp
@@ -42,20 +42,24 @@ void createBlob(bdlbb::BlobBufferFactory* bufferFactory,
 // class BlobPoolUtil
 // ------------------
 
-BlobPoolUtil::BlobSpPool
+BlobPoolUtil::BlobSpPoolSp
 BlobPoolUtil::createBlobPool(bdlbb::BlobBufferFactory* blobBufferFactory_p,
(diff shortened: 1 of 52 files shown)


678098:
> * There are a couple integration test failures. Could they be related to your changes?
These failures are also present in the build from the latest main:
https://github.com/bloomberg/blazingmq/actions/runs/12712681267/job/35439252483

kaikulimu (dismissed):

src/groups/bmq/bmqp/bmqp_blobpoolutil.cpp
```diff
@@ -42,20 +42,24 @@ void createBlob(bdlbb::BlobBufferFactory* bufferFactory,
...
+
+    return bsl::shared_ptr<BlobSpPool>(
+        new (*alloc) BlobSpPool(
+            bdlf::BindUtil::bind(&createBlob,
+                                 blobBufferFactory_p,
+                                 bdlf::PlaceHolders::_1,   // arena
+                                 bdlf::PlaceHolders::_2),  // allocator
+            k_BLOB_POOL_GROWTH_STRATEGY,
+            alloc),
+        alloc);
```
hallfox, line 62:
```suggestion
    return bsl::allocate_shared<BlobSpPool>(
            alloc,
            bdlf::BindUtil::bind(&createBlob,
                                 blobBufferFactory_p,
                                 bdlf::PlaceHolders::_1,   // arena
                                 bdlf::PlaceHolders::_2),  // allocator
            k_BLOB_POOL_GROWTH_STRATEGY));
```

Slightly reduces the number of allocator shenanigans. `allocate_shared` should forward the allocator to the right `BlobSpPool` constructor and provide it to the resulting `shared_ptr`.

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -925,7 +925,7 @@ void Domain::removeDomainReset()
     bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);  // LOCK
 
     d_state            = e_PREREMOVE;
-    d_teardownRemoveCb = nullptr;
+    d_teardownRemoveCb = bsl::nullptr_t();
```
hallfox, line 928: I find the use of `nullptr_t`'s constructor a little controversial, I feel as if we should really only be using preexisting representations of a nullptr (e.g. NULL, nullptr) but I suppose this works?

678098: > I find the use of `nullptr_t`'s constructor a little controversial, I feel as if we should really only be using preexisting representations of a nullptr (e.g. NULL, nullptr) but I suppose this works?

Pure `nullptr` doesn't work on Solaris, `NULL` or `bsl::nullptr_t()` work, but `bsl::nullptr_t()` is more restrictive. We typically used `bsl::nullptr_t()` to reset callbacks (`bsl::function`)

kaikulimu (approved):

PR #562, opened 2025-01-10
Fix[Storagetool]: simplify CompositeSequenceNumber class

alexander-e1off:
Addressing postponed review comment https://github.com/bloomberg/blazingmq/pull/508#discussion_r1869819443 - simplify `CompositeSequenceNumber` class.

### DIFF
--- a/src/applications/bmqstoragetool/m_bmqstoragetool_compositesequencenumber.cpp
+++ b/src/applications/bmqstoragetool/m_bmqstoragetool_compositesequencenumber.cpp
@@ -67,42 +67,44 @@ CompositeSequenceNumber::fromString(bool*              success,
         const bsl::string secondPart = seqNumString.substr(separatorPos + 1);
 
         // Convert parts to numbers
+        size_t              posFirst, posSecond;
+        unsigned long       uLong;
+        bsls::Types::Uint64 seqNumber;
         try {
-            size_t posFirst, posSecond;
-
-            const unsigned long       uLong = bsl::stoul(firstPart, &posFirst);
-            const bsls::Types::Uint64 seqNumber = bsl::stoul(secondPart,
-                                                             &posSecond);
-            if (posFirst != firstPart.size() ||
-                posSecond != secondPart.size()) {
-                errorDescription
-                    << "Invalid input: non-numeric values encountered.";
-                break;  // BREAK
-            }
-
-            unsigned int leaseId = static_cast<unsigned int>(uLong);
-            if (uLong != leaseId) {
-                errorDescription << "Invalid input: number out of range.";
-                break;  // BREAK
-            }
-
-            if (leaseId == 0 || seqNumber == 0) {
-                errorDescription << "Invalid input: zero values encountered.";
-                break;  // BREAK
-            }
-
-            d_compositeSequenceNumber = bsl::make_pair(leaseId, seqNumber);
-
-            *success = true;
-            return *this;  // RETURN
+            uLong     = bsl::stoul(firstPart, &posFirst);
+            seqNumber = bsl::stoul(secondPart, &posSecond);
         }
         catch (const bsl::invalid_argument& e) {
             errorDescription
                 << "Invalid input: non-numeric values encountered.";
+            break;  // BREAK
         }
         catch (const bsl::out_of_range& e) {
             errorDescription << "Invalid input: number out of range.";
+            break;  // BREAK
         }
+
+        if (posFirst != firstPart.size() || posSecond != secondPart.size()) {
+            errorDescription
+                << "Invalid input: non-numeric values encountered.";
+            break;  // BREAK
+        }
+
+        unsigned int leaseId = static_cast<unsigned int>(uLong);
+        if (uLong != leaseId) {
+            errorDescription << "Invalid input: number out of range.";
+            break;  // BREAK
+        }
+
+        if (leaseId == 0 || seqNumber == 0) {
+            errorDescription << "Invalid input: zero values encountered.";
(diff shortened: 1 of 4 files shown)


src/applications/bmqstoragetool/m_bmqstoragetool_compositesequencenumber.h
```diff
@@ -139,19 +138,20 @@ bool operator<=(const CompositeSequenceNumber& lhs,
...
 
-inline unsigned int CompositeSequenceNumber::leaseId() const
+inline bsls::Types::Uint64 CompositeSequenceNumber::sequenceNumber() const
 {
-    return d_leaseId;
+    return d_compositeSequenceNumber.second;
 }
 
-inline bsls::Types::Uint64 CompositeSequenceNumber::sequenceNumber() const
+inline bsl::pair<unsigned int, bsls::Types::Uint64>
```
678098, line 151: Can return const ref to the field and avoid making a copy

src/applications/bmqstoragetool/m_bmqstoragetool_compositesequencenumber.cpp
```diff
@@ -71,24 +71,27 @@ CompositeSequenceNumber::fromString(bsl::ostream&      errorDescription,
...
-        if (d_leaseId == 0 || d_seqNumber == 0) {
+        if (leaseId == 0 || seqNumber == 0) {
             errorDescription << "Invalid input: zero values encountered.";
             return *this;  // RETURN
         }
 
-        d_isSet = true;
+        d_compositeSequenceNumber = bsl::make_pair(leaseId, seqNumber);
+
+        *success = true;
```
678098, line 94: I think it's possible to rewrite the code by using `do-while(false)` to make it more clear.

Smth like this:
```
do {
    // ...
    if (someError()) {
        errorDescription << "Bad thing happened!";
        break;
    }
    // ...

    *success = true;
    return *this;
} while (false);

*success = false;
return *this;
```

This way:
- No need to "touch" `success` variable twice (it's being reset in the beginning of the function)
- No need to create disposable exception objects

src/applications/bmqstoragetool/m_bmqstoragetool_parameters.cpp
```diff
@@ -413,13 +416,16 @@ Parameters::Parameters(const CommandLineArguments& arguments,
     }
     else if (!arguments.d_seqNumLt.empty() || !arguments.d_seqNumGt.empty()) {
         bmqu::MemOutStream errorDescr(allocator);
+        bool               success = false;
```
678098, line 419: `success` is not checked after, is it okay?

src/applications/bmqstoragetool/m_bmqstoragetool_parameters.cpp
```diff
@@ -429,11 +435,12 @@ Parameters::Parameters(const CommandLineArguments& arguments,
     if (!arguments.d_seqNum.empty()) {
         CompositeSequenceNumber seqNum;
         bmqu::MemOutStream      errorDescr(allocator);
+        bool                    success = false;
```
678098, line 438: Here `success` is also not checked

678098 (changes requested):
A few questions

src/applications/bmqstoragetool/m_bmqstoragetool_compositesequencenumber.cpp
```diff
@@ -28,28 +28,28 @@ namespace m_bmqstoragetool {
...
-, d_seqNumber(0)
-, d_isSet(false)
+: d_compositeSequenceNumber(0, 0)
 {
     // NOTHING
 }
 
 CompositeSequenceNumber::CompositeSequenceNumber(
     const unsigned int        leaseId,
     const bsls::Types::Uint64 sequenceNumber)
```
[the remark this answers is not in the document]
alexander-e1off: Removed const-qualify.

src/applications/bmqstoragetool/m_bmqstoragetool_compositesequencenumber.h
alexander-e1off: Fixed as proposed.

src/applications/bmqstoragetool/m_bmqstoragetool_parameters.cpp
alexander-e1off: Validation is  done in `CommandLineArguments::validate()`. Added assert.

src/applications/bmqstoragetool/m_bmqstoragetool_parameters.cpp
alexander-e1off: Validation is  done in `CommandLineArguments::validate()`. Added assert.

src/applications/bmqstoragetool/m_bmqstoragetool_compositesequencenumber.cpp
```diff
@@ -28,75 +28,84 @@ namespace m_bmqstoragetool {
...
+            *success = true;
             return *this;  // RETURN
         }
+        catch (const bsl::invalid_argument& e) {
+            errorDescription
+                << "Invalid input: non-numeric values encountered.";
+        }
+        catch (const bsl::out_of_range& e) {
+            errorDescription << "Invalid input: number out of range.";
+        }
```
678098, line 105: What about placing `try-catch` block around the minimal code that might produce exceptions? In this case, only `bsl::stoul` lines should be covered

678098 (approved):

PR #565, opened 2025-01-13
Feat: CQH calls Queue::getHandle

dorjesinpo:
For the upcoming Dynamic Apps feature, we need `ClusterQueueHelper` to be the caller of `Queue::getHandle` (and inspect the status to make Dynamic Apps updates if necessary).
So, instead of `Domain` calling `Queue::getHandle` after `Cluster::openQueue`, it is the `Cluster` (`ClusterQueueHelper`) calling `Queue::getHandle` as part of `Cluster::openQueue`.

Roughly, the sequence looks like this:
```
    //  CQH::processPeerOpenQueueRequest  QueueSessionManager::processOpenQueue
    //              \                       /
    //               V                     V
    //              DomainFactory::createDomain
    //              /                      \
    //             V                        V
    //  CQH::onGetDomain                  QueueSessionManager::onDomainOpenCb
    //          |                           /
    //          V                          /
    //      CQH::onGetDomainDispatched    /
    //                      \            /
    //                       V          V
    //                      Domain::openQueue
    //                              |
    //                              V
    //                          Cluster::openQueue
    //                              |
    //                              V
    //                              CQH::openQueue
    //                                      \
    //                                       V
    //                                  Queue::getHandle
    //                                      /
    //                                     V
    //                              CQH::onGetQueueHandle. (future Dynamic Apps work)
    //                                    /
    //                                   V
    //                      Domain::onOpenQueueResponse
    //                        /                 \
    //                       V                   V
    //  CQH::onGetQueueHandleDispatched     QueueSessionManager::onQueueOpenCb
    //
    ```

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
@@ -176,6 +176,7 @@ ClusterQueueHelper::OpenQueueContext::OpenQueueContext(
 , d_clientContext(clientContext)
 , d_callback(callback)
 {
+    BSLS_ASSERT_SAFE(domain);
     // NOTHING
 }
 
@@ -189,9 +190,20 @@ ClusterQueueHelper::OpenQueueContext::~OpenQueueContext()
 void ClusterQueueHelper::OpenQueueContext::setQueueContext(
     QueueContext* queueContext)
 {
+    BSLS_ASSERT_SAFE(!d_queueContext_p);
+    BSLS_ASSERT_SAFE(queueContext);
+
     d_queueContext_p = queueContext;
     // Bump 'd_inFlight' counter
     ++(d_queueContext_p->d_liveQInfo.d_inFlight);
+}
+
+ClusterQueueHelper::QueueContext*
+ClusterQueueHelper::OpenQueueContext::queueContext() const
+{
+    BSLS_ASSERT_SAFE(d_queueContext_p);
+
+    return d_queueContext_p;
 }
 
 // -----------------------------------------
@@ -302,7 +314,7 @@ ClusterQueueHelper::getNextSubQueueId(const OpenQueueContextSp& context)
         d_cluster_p->dispatcher()->inDispatcherThread(d_cluster_p));
     BSLS_ASSERT_SAFE(context);
 
-    QueueLiveState* queueInfo = &context->d_queueContext_p->d_liveQInfo;
+    QueueLiveState* queueInfo = &context->queueContext()->d_liveQInfo;
 
     // We use a unique subQueue id for every subStream of the queue opened with
     // upstream, and this is used to correlate messages with the subStream.
@@ -321,7 +333,7 @@ ClusterQueueHelper::getNextSubQueueId(const OpenQueueContextSp& context)
         bsl::numeric_limits<unsigned int>::max() / 2) {
         BMQTSK_ALARMLOG_ALARM("CLUSTER_STATE")
             << d_cluster_p->description() << " nextSubQueueId for queue "
-            << context->d_queueContext_p->uri()
+            << context->queueContext()->uri()
             << " in cluster is at 50% capacity, please schedule a bounce of"
             << " this broker." << BMQTSK_ALARMLOG_END;
     }
@@ -329,7 +341,7 @@ ClusterQueueHelper::getNextSubQueueId(const OpenQueueContextSp& context)
              bsl::numeric_limits<unsigned int>::max() / 10 * 9) {
         BMQTSK_ALARMLOG_PANIC("CLUSTER_STATE")
             << d_cluster_p->description() << " nextSubQueueId for queue "
-            << context->d_queueContext_p->uri()
+            << context->queueContext()->uri()
             << " in cluster is at 90% capacity, please urgently schedule a"
             << " bounce of this broker." << BMQTSK_ALARMLOG_END;
     }
@@ -338,7 +350,7 @@ ClusterQueueHelper::getNextSubQueueId(const OpenQueueContextSp& context)
               bmqp::QueueId::k_RESERVED_SUBQUEUE_ID)) {
         BALL_LOG_ERROR << d_cluster_p->description()
                        << " nextSubQueueId for queue "
-                       << context->d_queueContext_p->uri() << " in cluster "
+                       << context->queueContext()->uri() << " in cluster "
                        << d_cluster_p->name()
                        << " has reached capacity, aborting the broker to"
                        << " contain the damage !";
@@ -799,7 +811,7 @@ void ClusterQueueHelper::processPendingContexts(
 void ClusterQueueHelper::assignUpstreamSubqueueId(
     const OpenQueueContextSp& context)
 {
-    QueueLiveState&   info  = context->d_queueContext_p->d_liveQInfo;
+    QueueLiveState&   info  = context->queueContext()->d_liveQInfo;
     const bsl::string appId = bmqp::QueueUtil::extractAppId(
         context->d_handleParameters);
     StreamsMap::const_iterator it = info.d_subQueueIds.findByAppIdSafe(appId);
@@ -823,7 +835,7 @@ void ClusterQueueHelper::assignUpstreamSubqueueId(
             info.d_subQueueIds.insert(
                 appId,
                 upstreamSubId,
-                SubQueueContext(context->d_queueContext_p->uri(),
+                SubQueueContext(context->queueContext()->uri(),
                                 subQueueIdInfo,
                                 d_allocator_p));
         }
@@ -837,7 +849,7 @@ void ClusterQueueHelper::assignUpstreamSubqueueId(
         info.d_subQueueIds.insert(
             appId,
             context->d_upstreamSubQueueId,
-            SubQueueContext(context->d_queueContext_p->uri(),
+            SubQueueContext(context->queueContext()->uri(),
                             context->d_handleParameters.subIdInfo(),
                             d_allocator_p));
     }
@@ -851,7 +863,7 @@ void ClusterQueueHelper::processOpenQueueRequest(
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(
         d_cluster_p->dispatcher()->inDispatcherThread(d_cluster_p));
-    BSLS_ASSERT_SAFE(isQueueAssigned(*(context->d_queueContext_p)));
+    BSLS_ASSERT_SAFE(isQueueAssigned(*(context->queueContext())));
     // At this time, the Queue must have been assigned an id/partition.
 
     if (d_cluster_p->isRemote()) {
@@ -863,7 +875,7 @@ void ClusterQueueHelper::processOpenQueueRequest(
         return;  // RETURN
     }
 
-    const int pid = context->d_queueContext_p->partitionId();
+    const int pid = context->queueContext()->partitionId();
 
     if (hasActiveAvailablePrimary(pid)) {
         if (d_clusterState_p->isSelfPrimary(pid)) {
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp [hunk 1]
```diff
@@ -178,6 +160,40 @@ void countUnconfirmed(bsls::Types::Int64* result, mqbi::Queue* queue)
...
+, d_callback(callback)
+{
+    // NOTHING
+}
+
+ClusterQueueHelper::OpenQueueContext::~OpenQueueContext()
+{
+    BSLS_ASSERT_SAFE(d_queueContext_p);
+
+    --d_queueContext_p->d_liveQInfo.d_inFlight;
```
678098, line 186: This class API still allows to copy or assign the object to another one, and it might break refcounting done in destructor. I think copy/assignment constructors should be deleted to prevent this. It will also safeguard against non-obvious problems with temporal object copies

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -178,6 +160,40 @@ void countUnconfirmed(bsls::Types::Int64* result, mqbi::Queue* queue)
...
+// struct ClusterQueueHelper::OpenQueueContext
+// -------------------------------------------
+
+ClusterQueueHelper::OpenQueueContext::OpenQueueContext(
+    mqbi::Domain*                                             domain,
+    const bmqp_ctrlmsg::QueueHandleParameters&                handleParameters,
+    const bsl::shared_ptr<mqbi::QueueHandleRequesterContext>& clientContext,
+    const mqbi::Cluster::OpenQueueCallback&                   callback)
+: d_queueContext_p(0)
+, d_domain_p(domain)
```
678098, line 173: The `domain` might be NULL if we are in the proxy, right?

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -826,44 +844,44 @@ void ClusterQueueHelper::assignUpstreamSubqueueId(OpenQueueContext* context)
...
-    const OpenQueueContext& context)
+    const OpenQueueContextSp& context)
 {
     // executed by the cluster *DISPATCHER* thread
 
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(
         d_cluster_p->dispatcher()->inDispatcherThread(d_cluster_p));
-    BSLS_ASSERT_SAFE(isQueueAssigned(*(context.d_queueContext_p)));
+    BSLS_ASSERT_SAFE(isQueueAssigned(*(context->d_queueContext_p)));
```
678098, line 854: Here, and in other places, we dereference `d_queueContext_p` without checking that it's not null. We rely on the fact that it was correctly set before. We are on the control path, so a few extra checks will not harm the performance. What about adding something like
```suggestion
    BSLS_ASSERT_SAFE(context->d_queueContext_p);
    BSLS_ASSERT_SAFE(isQueueAssigned(*(context->d_queueContext_p)));
```

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -877,30 +895,31 @@ void ClusterQueueHelper::processOpenQueueRequest(
...
     // executed by the cluster *DISPATCHER* thread
 
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(
         d_cluster_p->dispatcher()->inDispatcherThread(d_cluster_p));
 
-    QueueLiveState& qinfo = context.d_queueContext_p->d_liveQInfo;
-    const int       pid   = context.d_queueContext_p->partitionId();
+    QueueLiveState& qinfo = context->d_queueContext_p->d_liveQInfo;
+    const int       pid   = context->d_queueContext_p->partitionId();
```
678098, line 920: Here we also dereference `context->d_queueContext_p` without proper checks

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -1141,12 +1161,12 @@ void ClusterQueueHelper::onOpenQueueResponse(
         failure = false;
     }
 
-    QueueContext&           qcontext = *context.d_queueContext_p;
+    QueueContext&           qcontext = *context->d_queueContext_p;
```
678098, line 1164: Dereferencing

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -1824,38 +1844,37 @@ bool ClusterQueueHelper::createQueue(
...
 
-    QueueContext*   queueContext = context.d_queueContext_p;
+    QueueContext*   queueContext = context->d_queueContext_p;
     QueueLiveState& qinfo        = queueContext->d_liveQInfo;
     const int       pid          = queueContext->partitionId();
     const bool      isPrimary    = !d_cluster_p->isRemote() &&
                            d_clusterState_p->isSelfPrimary(pid);
     bsl::shared_ptr<mqbi::Queue> queue = createQueueFactory(errorDescription,
-                                                            context,
+                                                            *context,
```
678098, line 1877: Missing check before that context is not a null shared ptr

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -193,149 +193,26 @@ int normalizeConfig(mqbconfm::Domain* defn,
...
-    if (info->uri().domain() != d_name) {
-        // Note: This method will fire on all domains which belong to the
-        //       cluster having the queue assignment, but we examine the domain
-        //       name from the 'uri' to guarantee that only one domain is
-        //       updated.
-
-        return;  // RETURN
+    if (status.category() == bmqp_ctrlmsg::StatusCategory::E_SUCCESS) {
+        // VALIDATION: The queue must exist at this point, i.e., have been
+        //             registered.FE(
```
678098, line 206:
```suggestion
        //             registered.
```

src/groups/mqb/mqbblp/mqbblp_domain.h
```diff
@@ -89,8 +89,7 @@ namespace mqbblp {
 // ============
 
 /// Domain implementation
-class Domain BSLS_KEYWORD_FINAL : public mqbi::Domain,
-                                  public mqbc::ClusterStateObserver {
+class Domain : public mqbi::Domain {
```
678098, line 92:
```suggestion
class Domain BSLS_KEYWORD_FINAL : public mqbi::Domain {
```

src/groups/mqb/mqbi/mqbi_cluster.h
```diff
@@ -224,7 +224,7 @@ class Cluster : public DispatcherClient {
     /// above).
     typedef bsl::function<void(
         const bmqp_ctrlmsg::Status&            status,
-        Queue*                                 queue,
+        QueueHandle*                           queueHandle,
```
678098, line 227: Arg name changed, might update docs for this function

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
dorjesinpo: No, I don't think Proxy can work without a domain

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -819,51 +849,51 @@ void ClusterQueueHelper::assignUpstreamSubqueueId(OpenQueueContext* context)
...
-    const OpenQueueContext& context)
+    const OpenQueueContextSp& context)
 {
     // executed by the cluster *DISPATCHER* thread
 
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(
         d_cluster_p->dispatcher()->inDispatcherThread(d_cluster_p));
-    BSLS_ASSERT_SAFE(isQueueAssigned(*(context.d_queueContext_p)));
+    BSLS_ASSERT_SAFE(isQueueAssigned(*(context->queueContext())));
```
678098, line 866:
```suggestion
    BSLS_ASSERT_SAFE(context);
    BSLS_ASSERT_SAFE(context->queueContext());
    BSLS_ASSERT_SAFE(isQueueAssigned(*(context->queueContext())));
```

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -769,19 +798,20 @@ void ClusterQueueHelper::processPendingContexts(
...
         processOpenQueueRequest(*it);
     }
 }
 
-void ClusterQueueHelper::assignUpstreamSubqueueId(OpenQueueContext* context)
+void ClusterQueueHelper::assignUpstreamSubqueueId(
+    const OpenQueueContextSp& context)
 {
-    QueueLiveState&   info  = context->d_queueContext_p->d_liveQInfo;
+    QueueLiveState&   info  = context->queueContext()->d_liveQInfo;
```
678098, line 814: Might rely on the checks done before calling this function, or might add them here as well
```suggestion
    // executed by the cluster *DISPATCHER* thread

    // PRECONDITIONS
    BSLS_ASSERT_SAFE(context);
    BSLS_ASSERT_SAFE(context->queueContext());

    QueueLiveState&   info  = context->queueContext()->d_liveQInfo;
```

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -819,51 +849,51 @@ void ClusterQueueHelper::assignUpstreamSubqueueId(OpenQueueContext* context)
...
     if (hasActiveAvailablePrimary(pid)) {
         if (d_clusterState_p->isSelfPrimary(pid)) {
-            // At primary.  Load the routing configuration and inject the
-            // downstream handle parameters into the "response"
-            BSLS_ASSERT_SAFE(context.d_domain_p);
+            // At primary.
+
+            // Load the routing configuration and inject the downstream
+            // handle parameters into the "response"
+            BSLS_ASSERT_SAFE(context->d_domain_p);
```
678098, line 886: Since you have added a check for `domain` arg in context constructor, it is not null by construction, and this check is not necessary

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -1933,40 +1977,41 @@ bsl::shared_ptr<mqbi::Queue> ClusterQueueHelper::createQueueFactory(
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(
         d_cluster_p->dispatcher()->inDispatcherThread(d_cluster_p));
-    BSLS_ASSERT_SAFE(isQueueAssigned(*(context.d_queueContext_p)));
 
-    const int pid = context.d_queueContext_p->partitionId();
+    QueueContext* queueContext = context.queueContext();
```
678098, line 1981:
```suggestion
    BSLS_ASSERT_SAFE(context);
    QueueContext* queueContext = context.queueContext();

    BSLS_ASSERT_SAFE(queueContext);
```

678098 (changes requested):

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
dorjesinpo: Yes, but `d_domain_p` can change so let's keep it.

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
dorjesinpo: (`OpenQueueContext::queueContext()` already contains assertion)

678098 (approved):
Let's run sanitizers and merge, the UT failure is unrelated

PR #573, opened 2025-01-16
CI: simplify sanitizers launch

678098:
- Shorter names for sanitizer jobs, allowing them to fully fit in GH interface
- Simplify .yaml calls for sanitizers, by moving everything in 1 file and using args matrix
- Simplify .yaml by reusing dependencies cache workflow
- Disable `-Wunsafe-buffer-usage` warning when building instrumented UTs, this warning generates too much noise in the logs (70k+ log lines) and it is basically impossible to get any useful information from it
- Ensure cache rebuild or access from one workflow at a time
- Update `bde` 4.8.0.0 -> 4.18.0.0, `bde-tools` 4.8.0.0 -> 4.13.0.0


docker/build_deps.sh
```diff
@@ -28,8 +28,8 @@ fetch_git() {
 }
 
 fetch_deps() {
-    fetch_git bloomberg bde-tools 4.8.0.0
-    fetch_git bloomberg bde 4.8.0.0
+    fetch_git bloomberg bde-tools 4.13.0.0
```
alexander-e1off, line 31: Should we also synchronize bde/ntf versions in [build-ubuntu.sh](https://github.com/bloomberg/blazingmq/blob/main/bin/build-ubuntu.sh#L84) and [build-darwin.sh](https://github.com/bloomberg/blazingmq/blob/main/bin/build-darwin.sh#L42)?

678098: Updated both scripts

alexander-e1off (approved):
LGTM!

PR #577, opened 2025-01-17
Fix[UT]: disable JIT for RegEx under memory sanitizer

678098:


src/groups/bmq/bmqt/bmqt_uri.cpp
```diff
@@ -222,12 +222,16 @@ void UriParser::initialize(bslma::Allocator* allocator)
...
+    // See the internal ticket 177953779.
     int regexOptions = bdlpcre::RegEx::k_FLAG_JIT;
-#if defined(__has_feature)
+#if defined(__has_feature)  // Clang-supported method for checking sanitizers.
 #if __has_feature(memory_sanitizer)
     regexOptions &= ~bdlpcre::RegEx::k_FLAG_JIT;
+#elif defined(__SANITIZE_MEMORY__)  // GCC-supported macros for checking MSAN.
+    regexOptions &= ~bdlpcre::RegEx::k_FLAG_JIT;
 #endif
 #endif
```
678098, line 233: So we have disabled this for bmqt::Uri before, should do the same for `PartitionIdExtractor`

pniedzielski:
@678098 looks like the merge will be manual but easy.  I'll reapprove when that's fixed up.

PR #469, opened 2024-10-18
fix: generate AppKey only once

dorjesinpo:
Upon queue registration in the storage,   use keys in the CSL instead of generating new ones to keep CSL and non-CSL consistent.

Secondary, switching to `unordered_set` from `vector` to optimize calculating Apps diff

PR #581, opened 2025-01-23
fix: always update CSL on QueueUpdateAdvisory

dorjesinpo:
1) After [#469](https://github.com/bloomberg/blazingmq/pull/469), Primaries use ClusterState (instead of Domain config) when opening queue.  
2) In legacy mode, Replicas do not update ClusterState upon QueueUpdateAdvisory from Primary (App Registration).  (They do upon QueueAssignmentAdvisory).
3) If an App is added (to Domain config) to an existing queue, and then a Replica becomes Primary and receives OpenQueue request, it will detect the difference between the exiting storage (with the App) and the current ClusterState (without the App).  The new Primary then will remove the storage for the App and abort upon the assert in `RootQueueEngine::initializeAppId` - `#QUEUE_STORAGE_NOTFOUND Virtual storage does not exist for AppId`

One way of fixing this is to allow CSL apply QueueUpdateAdvisory commits in legacy mode (this PR)
Another would be to add `processQueueUpdateAdvisory` symmetrically to `processQueueAssignmentAdvisory`

Best way is to apply CSL commits unconditionally.

### DIFF
--- a/src/integration-tests/test_appids.py
+++ b/src/integration-tests/test_appids.py
@@ -644,6 +644,12 @@ def test_open_authorize_restart_from_non_FSM_to_FSM(cluster: Cluster):
 
 
 def test_open_authorize_change_primary(multi_node: Cluster):
+    """Add an App to Domain config of an existing queue, and then force a
+    Replica to become new Primary.  Start new Consumer.  Make sure the Consumer
+    receives previously posted data.
+    This is to address the concern with Replica not processing QueueUpdates
+    before becoming Primary.
+    """
     leader = multi_node.last_known_leader
     proxies = multi_node.proxy_cycle()
 
@@ -660,7 +666,7 @@ def test_open_authorize_change_primary(multi_node: Cluster):
 
     # ---------------------------------------------------------------------
     # Authorize 'quux'.
-    set_app_ids(multi_node, default_app_ids + ["quux"])
+    set_app_ids(multi_node, all_app_ids)
 
     # ---------------------------------------------------------------------
     # Post a message.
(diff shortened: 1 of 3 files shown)


src/integration-tests/test_appids.py
```diff
@@ -641,3 +641,57 @@ def test_open_authorize_restart_from_non_FSM_to_FSM(cluster: Cluster):
...
+
+
+def test_open_authorize_change_primary(multi_node: Cluster):
+    leader = multi_node.last_known_leader
+    proxies = multi_node.proxy_cycle()
+
+    producer = next(proxies).create_client("producer")
+    producer.open(tc.URI_FANOUT, flags=["write,ack"], succeed=True)
+
+    all_app_ids = default_app_ids + ["new_app"]
```
kaikulimu, line 653: Why do you need to add `["new_app"]`?

src/integration-tests/test_appids.py
```diff
@@ -641,3 +641,57 @@ def test_open_authorize_restart_from_non_FSM_to_FSM(cluster: Cluster):
             consumers[app_id].close(f"{tc.URI_FANOUT}?id={app_id}", block=True)
             == Client.e_SUCCESS
         )
+
+
+def test_open_authorize_change_primary(multi_node: Cluster):
```
kaikulimu, line 646: Comment about what this test is trying to test.

kaikulimu (commented):
Minor comments

src/integration-tests/test_appids.py
dorjesinpo: It is a leftover.  Refactoring.

kaikulimu (approved):
LGTM

PR #583, opened 2025-01-23
Fix[bmq]: Flush output on test and contract failures

hallfox:
We're seeing failures from tests without any of the context we'd expect (e.g. [a bsls assertion failure without line number info](https://github.com/bloomberg/blazingmq/actions/runs/12932233787/job/36068012310?pr=565)). To fix, we should explicitly flush stdout/stderr depending on which mode we're failing in.

### DIFF
--- a/src/groups/bmq/bmqtst/bmqtst_testhelper.h
+++ b/src/groups/bmq/bmqtst/bmqtst_testhelper.h
@@ -791,7 +791,12 @@ static inline void
 _assert(bool result, const char* expression, const char* file, int line)
 {
     if (!result) {
-        printf("Error %s(%d): %s    (failed)\n", file, line, expression);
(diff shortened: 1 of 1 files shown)


dorjesinpo:
`test_shutdown_upstream_proxy[virtual_fsm]` failure is irrelevant and can be ignored

`mqbc_incoreclusterstateledger.t`, failure is irrelevant.  It is output proves the PR is working
`Error /home/runner/work/blazingmq/blazingmq/src/groups/mqb/mqbc/mqbc_incoreclusterstateledger.t.cpp(110): 'cslIter.header().recordType()' (UPDATE) == 'recordType' (SNAPSHOT)    (failed)`

We get `file`, `line`, `expression` now!

dorjesinpo (approved):
Thank you

PR #584, opened 2025-01-24
Fix[MQB]: use CSL to update state on QueueAssignmentAdvisory

emelialei88:
Similar to PR #581, this PR invokes CSL commit callback for QueueAssignmentAdvisories and the queue assignment part of LeaderAdvisory, effectively making CSL the source of truth.


src/groups/mqb/mqbblp/mqbblp_clusterstatemanager.cpp
```diff
@@ -1690,16 +1693,21 @@ void ClusterStateManager::processQueueAssignmentAdvisory(
...
-            d_state_p->assignQueue(uri,
-                                   queueKey,
-                                   queueInfo.partitionId(),
-                                   appIdInfos);
+                d_state_p->assignQueue(uri,
+                                       queueKey,
+                                       queueInfo.partitionId(),
+                                       appIdInfos);
+            }
+            // When this function is not buffered, called from
```
kaikulimu, line 1708: I don't think you should change this logic. It would break `mqbblp::ClusterStateManager::processLeaderAdvisory`. We invoke CSL commit callback for QueueAssignmentAdv but not for the queue assignment part of LeaderAdv yet. We can take care of LeaderAdv in a later PR.

Also, note that we should never be receiving a direct QueueAssignmentAdv (i.e. not part of LeaderAdv) anymore after the removal of https://github.com/bloomberg/blazingmq/blob/7b30704eba30fefa10fd5775517cec2485884a6a/src/groups/mqb/mqbc/mqbc_clusterutil.cpp#L1037

src/integration-tests/test_migrate_to_csl.py
```diff
@@ -0,0 +1,54 @@
...
+
+"""
+Testing migrating to CSL.
+"""
+import blazingmq.dev.it.testconstants as tc
+from blazingmq.dev.it.fixtures import (  # pylint: disable=unused-import
+    Cluster,
+    multi_node,
+)
+
```
kaikulimu, line 24: Please add `pytestmark = order(??)`

src/integration-tests/test_migrate_to_csl.py
```diff
@@ -0,0 +1,54 @@
...
+    for member in members:
+        assert member.outputs_regex(
+            "(Applying cluster message with type = UPDATE).*(queueAssignmentAdvisory)",
+            timeout,
+        )
+        assert member.outputs_regex(
+            "(Committed advisory).*queueAssignmentAdvisory", timeout
+        )
+        assert not member.outputs_regex(
+            "'QueueUnAssignmentAdvisory' will be applied to", timeout
```
kaikulimu, line 53: Looks correct for two reasons:
1. Should be `QueueAssignmentAdvisory` instead of `QueueUnAssignmentAdvisory`
2. This log line seems to be during https://github.com/bloomberg/blazingmq/blob/7b30704eba30fefa10fd5775517cec2485884a6a/src/groups/mqb/mqbc/mqbc_clusterutil.cpp#L1000-L1003, but should appear when we go through CSL path.

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -1009,34 +1009,6 @@ ClusterUtil::assignQueue(ClusterState*         clusterState,
...
-            key,
-            queueAdvisory.queues().back().partitionId(),
-            appInfos);
-        BSLS_ASSERT_SAFE(assignRc);
-
-        BALL_LOG_INFO << cluster->description()
-                      << ": Queue assigned: " << queueAdvisory;
-
-        // Broadcast 'queueAssignmentAdvisory' to all followers
-        clusterData->messageTransmitter().broadcastMessage(controlMsg);
```
kaikulimu, line 1037: Now, `QueueAssignmentAdvisory` should never be sent out. We should remove https://github.com/bloomberg/blazingmq/blob/7b30704eba30fefa10fd5775517cec2485884a6a/src/groups/mqb/mqbblp/mqbblp_cluster.cpp#L3079-L3087 and https://github.com/bloomberg/blazingmq/blob/7b30704eba30fefa10fd5775517cec2485884a6a/src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp#L1341 and https://github.com/bloomberg/blazingmq/blob/7b30704eba30fefa10fd5775517cec2485884a6a/src/groups/mqb/mqbi/mqbi_clusterstatemanager.h#L336-L339.

kaikulimu: Still, preserve enough code such that `mqbblp::ClusterStateManager::processBufferedQueueAdvisories()` and `mqbblp::ClusterStateManager::processLeaderAdvisory()` do not break.

kaikulimu (changes requested):
Some comments

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -1009,34 +1009,6 @@ ClusterUtil::assignQueue(ClusterState*         clusterState,
...
-            uri,
-            key,
-            queueAdvisory.queues().back().partitionId(),
-            appInfos);
-        BSLS_ASSERT_SAFE(assignRc);
-
-        BALL_LOG_INFO << cluster->description()
-                      << ": Queue assigned: " << queueAdvisory;
-
-        // Broadcast 'queueAssignmentAdvisory' to all followers
```
kaikulimu, line 1036: Thinking about compatibility between broker versions, we should still broadcast this advisory for now, such as old broker can still receive and process.

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -3076,15 +3076,6 @@ void Cluster::processClusterControlMessage(
                                  source),
             this);
     } break;  // BREAK
-    case MsgChoice::SELECTION_ID_QUEUE_ASSIGNMENT_ADVISORY: {
```
kaikulimu, line 3079: Thinking about compatibility between different broker versions, we still keep this case as an no-op, in case an old broker sends us this:
```
case MsgChoice::SELECTION_ID_QUEUE_ASSIGNMENT_ADVISORY: {
    // NO-OP
} break; // BREAK
```

emelialei88: I think maybe we should keep as what it is and not remove or change to NO-OP. When the old leader calls `assignQueue`, it will broadcast `SELECTION_ID_QUEUE_ASSIGNMENT_ADVISORY`. If we delete this case for new followers, they won't do anything upon receiving the advisory which is in correct.

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
emelialei88: Could you give an example?

kaikulimu: What I meant is: Old version followers return on commit callback of `e_CLUSTER_TYPE` events, so they rely on having this additionally broadcasted 'queueAssignmentAdvisory' to process correctly. Thus, new version leader should keep broadcasting this additional time for now,

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
kaikulimu: Old leader will also write QAA to CSL; old followers will return immediately, but new followers should process QAA as part of CSL commit callback. It should be safe to remove this branch.

kaikulimu (commented):
Be sure to test that these changes solve https://github.com/bloomberg/blazingmq/issues/683 before we merge this PR.

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -1062,14 +1062,6 @@ ClusterUtil::assignQueue(ClusterState*         clusterState,
...
         BSLS_ASSERT_SAFE(queueAdvisory.queues().size() == 1);
 
-        bmqp_ctrlmsg::QueueInfo& queueInfo = queueAdvisory.queues().back();
-
-        BSLA_MAYBE_UNUSED const bool assignRc = clusterState->assignQueue(
-            queueInfo);
-        BSLS_ASSERT_SAFE(assignRc);
-
-        BALL_LOG_INFO << cluster->description()
-                      << ": Queue assigned: " << queueAdvisory;
```
kaikulimu, line 1072: You should put the note next to the `broadcastMessage`, i.e.:

```
        // NOTE: We must broadcast this control message before applying to CSL,
        // because if CSL is running in eventual consistency it will
        // immediately apply a commit with a higher seqeuence number than the
        // QueueAssignmentAdvisory.  If we ever receive the commit before the
        // QAA, we will alarm due to out-of-sequence advisory.

        clusterData->messageTransmitter().broadcastMessage(controlMsg);
```

The `Do it before 'assignQueue'` comment no longer makes sense so we can remove.

src/integration-tests/test_migrate_to_csl.py
```diff
@@ -0,0 +1,245 @@
...
+    4. Reconfigure back to non-CSL and make sure queues are operational
+    """
+    uri_priority1 = f"bmq://{tc.DOMAIN_PRIORITY_SC}/q1"
+    uri_fanout1 = f"bmq://{tc.DOMAIN_FANOUT_SC}/f1"
+    uri_fanout1_foo = f"bmq://{tc.DOMAIN_FANOUT_SC}/f1?id=foo"
+    uri_fanout1_bar = f"bmq://{tc.DOMAIN_FANOUT_SC}/f1?id=bar"
+    uri_priority2 = f"bmq://{tc.DOMAIN_PRIORITY_SC}/q2"
+    uri_fanout2 = f"bmq://{tc.DOMAIN_FANOUT_SC}/f2"
+    uri_fanout2_foo = f"bmq://{tc.DOMAIN_FANOUT_SC}/f2?id=foo"
+    uri_fanout2_bar = f"bmq://{tc.DOMAIN_FANOUT_SC}/f2?id=bar"
```
kaikulimu: I would suggest using the [`test constants`](https://github.com/bloomberg/blazingmq/blob/30f913022f282ffe520477b70fe89397ebc7f58d/src/python/blazingmq/dev/it/testconstants.py#L29-L59) instead of defining your own.

src/integration-tests/test_migrate_to_csl.py
```diff
@@ -0,0 +1,245 @@
...
+            "(Committed advisory).*queueAssignmentAdvisory", timeout
+        )
+        assert not member.outputs_regex(
+            "Processing queueAssignmentAdvisory message", timeout
+        )
+
+
+@tweak.cluster.cluster_attributes.is_cslmode_enabled(False)
+@tweak.cluster.cluster_attributes.is_fsmworkflow(False)
+def test_reconfigure_from_non_CSL_to_CSL(cluster: Cluster):
```
kaikulimu: We already have the test [`test_restart_between_non_FSM_and_FSM`](https://github.com/bloomberg/blazingmq/blob/30f913022f282ffe520477b70fe89397ebc7f58d/src/integration-tests/test_restart.py#L200). Does your test here offer anything extra?

kaikulimu (approved):

PR #585, opened 2025-01-24
fix: Not delivering old data to new Apps

dorjesinpo:
Scenario I.  Not enough confirms.  "Stuck" message.

0.  Domain with at least 2 apps {`a1`, `a2`}
1.  Post `m1`.  `refCount == 2`
2.  Delete `a1` app
3.  Restart broker
4.  Open `a1` consumer, no data
5.  Open `a2` consumer, receive `m1`, confirm `m1`
6.  `m1` is still in the storage

This is because the restart does not notify storage about `a1` deletion.
Without restart, 5. will delete `m1` from the storage because 2. decrements `m1` refCount in memory (not in Journal, where it is always `2`) 

Scenario II.  Too many confirms
--------------------------------

0.  Domain with 1 apps `a1`
1.  Post `m1`.  `refCount == 1`
2.  Add an app `a2`
3.  Open `a1` consumer, receive `m1`
4.  Open `a2` consumer, receive `m1` (wrong `refCount`!)  
5.  `a1` confirms
6.  `a2` confirms.  Failure [reason: `GUID_NOT_FOUND`]

Solution.

---------

Not to deliver existing data to newly added App.  And do that upon restart as well!

The chronology of message/App events should be available in memory.  No need to store this information

This chronology can be restored in `recoveredQueuesCb` which iterates all existing data and app records in the Journal.

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_datastore.cpp
+++ b/src/groups/mqb/mqbs/mqbs_datastore.cpp
@@ -39,6 +39,11 @@ BSLMF_ASSERT(sizeof(DataStoreRecordHandle) ==
 void DataStoreConfigQueueInfo::addAppInfo(const bsl::string&      appId,
                                           const mqbu::StorageKey& appKey)
 {
+    // Comparing ages of Apps and messages relies on chronological order
+    // (hence 'OrderedHashMap'.
+    // The Legacy mode recreates Apps by reverse iterating Journal (w/ QList)
+    // The CSL node recreates Apps in chronological order.
+
     if (d_isCSL) {
         d_appIdKeyPairs.insert(bsl::make_pair(appKey, appId));
     }
@@ -51,6 +56,7 @@ void DataStoreConfigQueueInfo::addAppInfo(const bsl::string&      appId,
 
     if (ghost != d_ghosts.end()) {
         d_purgeOps.erase(ghost->second);
+        d_ghosts.erase(ghost);
     }
 }
 
@@ -64,8 +70,14 @@ void DataStoreConfigQueueInfo::addPurgeOp(const mqbu::StorageKey&   key,
         // This is not a ghost app
     }
     else {
-        PurgeOps::const_iterator cit = d_purgeOps.emplace(start, end);
-        d_ghosts.emplace(key, cit);
+        Ghosts::iterator ghost = d_ghosts.find(key);
+
+        // Keep only one PurgeOp per App - the last one (the first in reverse).
+        // This code relies on 'FileStore::recoverMessages' reverse iteration.
+        if (ghost == d_ghosts.end()) {
+            PurgeOps::const_iterator cit = d_purgeOps.emplace(start, end);
+            d_ghosts.emplace(key, cit);
+        }
     }
 }
 
--- a/src/groups/mqb/mqbs/mqbs_datastore.cpp
+++ b/src/groups/mqb/mqbs/mqbs_datastore.cpp
@@ -32,6 +32,66 @@ BSLMF_ASSERT(sizeof(DataStoreRecordHandle) ==
              sizeof(DataStoreConfig::RecordIterator));
 }  // close unnamed namespace
 
+// ------------------------------
+// class DataStoreConfigQueueInfo
+// ------------------------------
+
+void DataStoreConfigQueueInfo::addAppInfo(const bsl::string&      appId,
+                                          const mqbu::StorageKey& appKey)
+{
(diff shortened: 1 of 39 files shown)


src/integration-tests/test_appids.py
```diff
@@ -641,3 +642,150 @@ def test_open_authorize_restart_from_non_FSM_to_FSM(cluster: Cluster):
...
+
+
+def test_old_data_new_app(cluster: Cluster):
+    """Do this: m1, +new_app_1, m2, +new_app2, m3, +new_app3, m4, -new_app2
+    Old apps  receive  4
+    new_app_1 receives 3
+    new_app_2 receives 0 (after it gets deleted)
+    new_app_3 receives 1
+
+    Restart and receive the same (except for new_app_2)
```
kaikulimu, line 654: Why except for new_app_2? new_app_2 should still receive 0 after restart

src/integration-tests/test_appids.py
```diff
@@ -641,3 +642,150 @@ def test_open_authorize_restart_from_non_FSM_to_FSM(cluster: Cluster):
...
+
+        new_consumer_3 = next(proxies).create_client(new_app_3)
+        new_consumer_3.open(
+            f"{tc.URI_FANOUT}?id={new_app_3}", flags=["read"], succeed=True
+        )
+
+        leader.dump_queue_internals(tc.DOMAIN_FANOUT, tc.TEST_QUEUE)
+
+        for app_id in default_app_ids:
+            test_logger.info(f"Check if {app_id} has seen 2 messages")
```
kaikulimu, line 731:
```suggestion
            test_logger.info(f"Check if {app_id} has seen 4 messages")
```

src/integration-tests/test_appids.py
```diff
@@ -641,3 +642,150 @@ def test_open_authorize_restart_from_non_FSM_to_FSM(cluster: Cluster):
...
+            f"{tc.URI_FANOUT}?id={new_app_3}", block=True, succeed=True
+        )
+
+    _verify_clients()
+
+    cluster.stop_nodes()
+
+    cluster.start_nodes(wait_leader=True, wait_ready=True)
+
+    _verify_clients()
```
kaikulimu, line 791: Please verify that all messages are indeed removed from the storage after the consumers have confirmed all messages.

kaikulimu (changes requested):
Some comments

src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.cpp
```diff
@@ -448,42 +539,91 @@ int VirtualStorageCatalog::addVirtualStorage(bsl::ostream& errorDescription,
...
+        BSLS_ASSERT_SAFE(d_queue_p->queueEngine());
+        // QueueEngines use the key to look up the id
+
+        d_queue_p->queueEngine()->unregisterStorage(removing->appId(),
+                                                    appKey,
+                                                    removingOrdinal);
     }
 
-    return false;
+    if (d_areOrdinalsContinuous) {
```
pniedzielski, line 601: I'm fairly sure here that the code is more readable with `d_isProxy` or such.  In my mind, at least, when reading this component, that's what I've found myself reading it as, and the documentation comment says the same.  Willing to be overridden, though.

pniedzielski (changes requested):
Overall comments:

  - I noticed there are a few cases of `///` comments with `'this emphasis'`.  Note that the single tick format won't do anything in Doxygen anymore, so we need to do backticks in all these cases.  Minor changes, maybe don't worry about them because I'll catch them in a later PR for doxygenifying the codebase.
  
  - There are two test failures.  I don't *think* they're related, but can you make sure?

src/groups/mqb/mqbs/mqbs_datastore.cpp
```diff
@@ -32,6 +32,66 @@ BSLMF_ASSERT(sizeof(DataStoreRecordHandle) ==
...
+    const DataStoreRecordKey& current) const
+{
+    unsigned int count = 0;
+    for (PurgeOps::const_iterator cit = d_purgeOps.begin();
+         cit != d_purgeOps.end();) {
+        if (current < cit->first) {
+            // 'current' is not in any (more) range
+            return count;
+        }
+        else if (!(cit->second < current)) {
```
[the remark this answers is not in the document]
pniedzielski: Okay

kaikulimu (approved):
I approve.

pniedzielski (approved):

PR #586, opened 2025-01-26
mqba::Application: Temporarily disable admin command re-routing

kaikulimu:


kaikulimu:
@chrisbeard Looks like disabling admin command rerouting would break the `DOMAINS REMOVE` command that @emelialei88 added recently. Is that okay?

chrisbeard:
> @chrisbeard Looks like disabling admin command rerouting would break the `DOMAINS REMOVE` command that @emelialei88 added recently. Is that okay?

Yeah, that's fine to (temporarily) break.

chrisbeard (approved):
Nice, this is an easy patch for us to revert later

PR #589, opened 2025-01-28
Fix mqbblp::RecoveryManager: Clear sync peer promptly

kaikulimu:
See internal ticket 178057288

Clearing sync peer promptly prevents [this failure condition](https://bbgithub.dev.bloomberg.com/BMQ/blazingmq-mirror/blob/85fa14b429195474314a4fc1675b5bc923099a6e/src/groups/mqb/mqbblp/mqbblp_recoverymanager.cpp#L666-L667) to trigger when the sync peer node goes down after sync with that peer is already complete. Otherwise, we will call `onPartitionPrimarySyncStatus(partitionId, -1 /* status */);` and fail [this assert](https://bbgithub.dev.bloomberg.com/BMQ/blazingmq-mirror/blob/85fa14b429195474314a4fc1675b5bc923099a6e/src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp#L368-L369) later on because primary is already active.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp
@@ -1770,7 +1770,7 @@ void ClusterOrchestrator::processPrimaryStatusAdvisory(
     // TBD: may need to review the order of invoking these routines.
 
     BALL_LOG_INFO << d_clusterData_p->identity().description()
-                  << " PartitionId [" << primaryAdv.partitionId()
+                  << " Partition [" << primaryAdv.partitionId()
                   << "]: received primary status advisory: " << primaryAdv
                   << ", from: " << source->nodeDescription();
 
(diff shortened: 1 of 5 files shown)


kaikulimu:
Can ignore IncoreCSL unit test failure. Will be fixed in another branch

dorjesinpo (commented):
This logs `source` twice (instead of `primarySyncPeer`):[https://bbgithub.dev.bloomberg.com/BMQ/blazingmq-mirror/blob/85fa14b429195474314a4fc1675b5bc923099a6e/src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp#L813](https://bbgithub.dev.bloomberg.com/BMQ/blazingmq-mirror/blob/85fa14b429195474314a4fc1675b5bc923099a6e/src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp#L813)

Can we document/comment somewhere how do `RecoveryManager_PrimarySyncContext::primarySyncInProgress()` and `RecoveryManager_PrimarySyncContext::syncPeer()` relate?  What are the scenarios when one is `0` and the other is not (if any)?

dorjesinpo (approved):

PR #591, opened 2025-01-30
Feat[bmqtool]: collect ACK latencies

678098:
- Pass `m_bmqtool::Parameters` to `m_bmqtool::Application` as a const-ref, avoid passing and holding it as a pointer.
- Always put post timestamp to messages in bmqtool. Remove "smart" skipping of messages to add timestamp to.
- Use throttles to collect in average 1000 latencies per second.
- Remember post timestamps as correlation ids in producer. Use these ids to calculate ack latencies.

### DIFF
--- a/src/applications/bmqtool/m_bmqtool_application.cpp
+++ b/src/applications/bmqtool/m_bmqtool_application.cpp
@@ -71,6 +71,13 @@ namespace {
 
 BALL_LOG_SET_NAMESPACE_CATEGORY("BMQTOOL.APPLICATION");
 
+/// The maximum latencies logged per second
+const int k_MAX_LATENCIES_PER_SECOND = 1000;
+
+/// The average expected time between logged latencies
+const bsls::Types::Int64 k_NS_PER_LATENCY = bdlt::TimeUnitRatio::k_NS_PER_S /
+                                            k_MAX_LATENCIES_PER_SECOND;
+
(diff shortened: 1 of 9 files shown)


src/applications/bmqtool/m_bmqtool_application.cpp
```diff
@@ -516,15 +507,18 @@ void Application::generateLatencyReport()
     // could see the evolution over time and maybe some pattern (for example
     // initial latency being higher due to cache warmup, ...).
     int                                           idx = 0;
-    bsl::list<bsls::Types::Int64>::const_iterator it  = d_latencies.begin();
-    while (it != d_latencies.end()) {
-        if (idx++ % 10 == 0) {
-            output << "\n    ";
+    bsl::list<bsls::Types::Int64>::const_iterator it  = itStart;
+    while (it != latencies.cend()) {
+        if (idx++ > 0) {
```
waldgange, line 512: The statement has changed. The precious `if (idx++ % 10 == 0)` was true for idx ==0. The new one `if (idx++ > 0)` is false. Hence we we don't print '\n' on first iteration as we used to do.  Is it intended?

waldgange (approved):
I only have a concern about the condition in the while loop. Feel free to fix it if it's broken or to merge the PR

waldgange (approved):
LGTM

678098:
@waldgange rebased

waldgange (approved):

PR #592, opened 2025-01-31
Fix[IT]: CSL+FSM together or both disabled; rollback new python features

678098:
- Rollback new python features (`int | None` with `|`-notation, replaced by `Optional[int]`) blocking us from launching ITs on older python versions
- Redo ITs launch presets, by allowing CSL+FSM together only or both disabled


pniedzielski (approved):
LGTM!

Issue #219, opened 2024-03-18
Subscriptions expressions: Ability to check whether a property exists


mlongob:
### Is there an existing proposal for this?

- [X] I have searched the existing proposals

### Is your feature request related to a problem?

The expression language currently does not provide a clear way to check for the existence of a property.

Let's assume I have a queue with two types of messages.

Messages of type A have the following properties:
```
radius = 34
color = "blue"
```

Messages of type B have the following properties:
```
height: 53
width: 100
color = "red"
```

If I want to filter for messages that contain a `radius` field in a subscription, I can currently recur to the following workaround expressions:

```
radius == radius
radius >= 0 || radius < 0
```

however my intent to check for existence is not clear. 

### Describe the solution you'd like

I would love if the expression language allowed me to check for existence of fields explicitly. This can be done by either checking if a field:

```
!= NULL
```
for the existence check.

and 

``` 
== NULL
```
for the missing property check.

Alternatively an operator for exists() can be introduced. This alternative would require a unary negation operator.

PR #594, opened 2025-02-03
Feat[bmq]: support exists(x) function in subscription expressions

bbpetukhov:
[*Issue number of the reported bug or feature request: #<219>*](https://github.com/bloomberg/blazingmq/issues/219)

**Describe your changes**
- Added new function exists(x) to check if the variable x is defined in the current context

**Testing performed**
- Added various unit tests to cover exists(x)

**Additional context**
Created PR for the docs website update: https://github.com/bloomberg/blazingmq/pull/648

### DIFF
--- a/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.cpp
+++ b/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.cpp
@@ -386,11 +386,7 @@ bdld::Datum SimpleEvaluator::Exists::evaluate(EvaluationContext& context) const
     bdld::Datum value = context.d_propertiesReader->get(d_name,
                                                         context.d_allocator);
 
-    if (value.isError()) {
-        return bdld::Datum::createBoolean(false);
-    }
-
-    return bdld::Datum::createBoolean(true);
+    return bdld::Datum::createBoolean(!value.isError());
 }
 
 }  // close package namespace
--- a/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.cpp
+++ b/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.cpp
@@ -364,9 +364,9 @@ bdld::Datum SimpleEvaluator::Not::evaluate(EvaluationContext& context) const
     return bdld::Datum::createBoolean(!value.theBoolean());
 }
 
-// --------------------------
+// -----------------------------
 // class SimpleEvaluator::Exists
-// --------------------------
+// -----------------------------
 
 SimpleEvaluator::Exists::Exists(const bsl::string& name)
 : d_name(name)
--- a/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.cpp
+++ b/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.cpp
@@ -364,5 +364,34 @@ bdld::Datum SimpleEvaluator::Not::evaluate(EvaluationContext& context) const
     return bdld::Datum::createBoolean(!value.theBoolean());
 }
 
+// --------------------------
+// class SimpleEvaluator::Exists
+// --------------------------
+
+SimpleEvaluator::Exists::Exists(const bsl::string& name)
+: d_name(name)
+{
+}
+
+#if defined(BSLS_COMPILERFEATURES_SUPPORT_RVALUE_REFERENCES) &&               \
+    defined(BSLS_COMPILERFEATURES_SUPPORT_NOEXCEPT)
+SimpleEvaluator::Exists::Exists(bsl::string&& name) noexcept
+: d_name(bsl::move(name))
+{
+}
+#endif
+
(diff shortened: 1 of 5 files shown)


678098:
@bbpetukhov @chrisbeard some time ago, we had an idea of declaring internal properties or keywords with two underscores `__`, like `__timestamp`. Do you think we should do this? Should we also make checks when declaring properties for messages, to prevent setting properties with names `abs` or `exists`?

src/groups/bmq/bmqeval/bmqeval_simpleevaluator.cpp
```diff
@@ -364,5 +364,34 @@ bdld::Datum SimpleEvaluator::Not::evaluate(EvaluationContext& context) const
     return bdld::Datum::createBoolean(!value.theBoolean());
 }
 
+// --------------------------
+// class SimpleEvaluator::Exists
+// --------------------------
```
678098, line 369:
```suggestion
// -----------------------------
// class SimpleEvaluator::Exists
// -----------------------------
```

This is not a very important fix, just to be complete

src/groups/bmq/bmqeval/bmqeval_simpleevaluator.h
```diff
@@ -478,6 +478,41 @@ class SimpleEvaluator {
         evaluate(EvaluationContext& context) const BSLS_KEYWORD_OVERRIDE;
     };
 
+    // ---
+    // Exists
+    // ---
```
678098, line 483:
```suggestion
    // ------
    // Exists
    // ------
```

src/groups/bmq/bmqeval/bmqeval_simpleevaluator.h
```diff
@@ -478,6 +478,41 @@ class SimpleEvaluator {
...
+        /// evaluation context, as a boolean.
+        explicit Exists(bsl::string&& name) noexcept;
+#endif
+
+        // ACCESSORS
+
+        /// Evaluate `expression` passed to the constructor. If it is a
+        /// boolean, return the negated value as a boolean Datum.
+        /// Otherwise, set the error in the context to  e_TYPE, stop the
+        /// evaluation, and return a null datum.
```
678098, line 511:
```suggestion
        /// Evaluate `expression` passed to the constructor.  If it is a
        /// boolean, return the negated value as a boolean Datum.
        /// Otherwise, set the error in the context to e_TYPE, stop the
        /// evaluation, and return a null datum.
```

src/groups/bmq/bmqeval/bmqeval_simpleevaluator.t.cpp
```diff
@@ -520,6 +521,14 @@ static void test3_evaluation()
         {"i_0 != -9223372036854775807", true},  // -(2 ** 63) + 1
         {"i_0 != 9223372036854775807", true},   // 2 ** 63 - 1
         {"i_0 != -9223372036854775808", true},  // -(2 ** 63)
+
+        // exists
+        {"exists(i_42)", true},
+        {"exists(non_existing_property)", false},
```
678098, line 527: Also the interesting case when either a property doesn't exist or has a constrained value:
```suggestion
        {"exists(non_existing_property)", false},
        {"!exists(non_existing_property) || non_existing_property > 41", true},
```

src/groups/bmq/bmqeval/bmqeval_simpleevaluator.t.cpp
```diff
@@ -520,6 +521,14 @@ static void test3_evaluation()
         {"i_0 != -9223372036854775807", true},  // -(2 ** 63) + 1
         {"i_0 != 9223372036854775807", true},   // 2 ** 63 - 1
         {"i_0 != -9223372036854775808", true},  // -(2 ** 63)
+
+        // exists
```
678098, line 525: Let's add a comment that `exists` might mean a property or a function depending on its usage:
```suggestion
        // exists
        // Note that we allow to set a message property with name `exists`.
        // In this case the usage context defines whether it's a function call
        // or a property value.
```

src/groups/bmq/bmqeval/bmqeval_simpleevaluator.cpp
```diff
@@ -364,5 +364,34 @@ bdld::Datum SimpleEvaluator::Not::evaluate(EvaluationContext& context) const
...
+bdld::Datum SimpleEvaluator::Exists::evaluate(EvaluationContext& context) const
+{
+    bdld::Datum value = context.d_propertiesReader->get(d_name,
+                                                        context.d_allocator);
+
+    if (value.isError()) {
+        return bdld::Datum::createBoolean(false);
+    }
+
+    return bdld::Datum::createBoolean(true);
```
jll63, line 393: I would simply write `return bdld::Datum::createBoolean(!value.isError())` but it's subjective, you decide.

jll63 (approved):

678098 (approved):

PR #598, opened 2025-02-05
fix race between restoreState and onOpenQueueResponse [178106131]

dorjesinpo:
Race between processing reopen request cancellation (due to disconnect) and reacting to new primary.  In the former, we assume that since we got cancellation, there is no primary and we buffer the request waiting for primary (which is already there).


dorjesinpo:
`mqbc_incoreclusterstateledger.t ` failure is unrelated

678098 (approved):

PR #600, opened 2025-02-05
fix run --bmq-log-dir on darwin

dorjesinpo:
`run-tests --bmq-log-dir=/Users/vdzhitenov/tests` generates an unnecessary error on Darwin:
`"/proc/sys/kernel/core_pattern does not exist, cores will not be saved"`


jll63 (approved):

dorjesinpo:
already fixed by another PR

Issue #593, opened 2025-02-03
bmqtool: fix possibly unsafe format string passing


678098:
We allow to pass a sequential message pattern to bmqtool as argument, to be used here:
```
if (!d_parameters.sequentialMessagePattern().empty()) {
    char buffer[128];
    length = snprintf(
        buffer,
        sizeof(buffer),
        d_parameters.sequentialMessagePattern().c_str(),
        d_numMessagesPosted);
    msg.setDataRef(buffer, length);
}
```

This leads to compilation warning:

```
/blazingmq/src/applications/bmqtool/m_bmqtool_poster.cpp:143:21: warning: format string is not a string literal [-Wformat-nonliteral]
  143 |                     d_parameters.sequentialMessagePattern().c_str(),
```

In reality, we don't need such flexibility. Instead, we can pass `sequentialMessagePrefix` and append a mutable suffix with format string `%09d, d_numMessagesPosted`.

sameep-git:
I am new to open source, but I would like to contribute to this issue. 
I looked at m_bmqtool_poster.cpp to understand the warning. But, I could not find a usage for ```sequentialMessagePrefix``` anywhere else. Do you have any examples of using it?

678098:
> I am new to open source, but I would like to contribute to this issue. I looked at m_bmqtool_poster.cpp to understand the warning. But, I could not find a usage for `sequentialMessagePrefix` anywhere else. Do you have any examples of using it?

Hi @sameep-git, you are right that we don't have `sequentialMessagePrefix` parameter now. The idea of this PR is to replace `sequentialMessagePattern` usage with `sequentialMessagePrefix` and do not pass variable format string to `snprintf`.

For simplicity, you can reuse `sequentialMessagePattern` argument and make local changes here:
https://github.com/bloomberg/blazingmq/blob/0e8bbf26ffd8d5ef211c9347a61c774b7d196949/src/applications/bmqtool/m_bmqtool_poster.cpp#L132-L138

Codebase search shows the usage of this cmd argument in this integration test, in a few places:
https://github.com/bloomberg/blazingmq/blob/0e8bbf26ffd8d5ef211c9347a61c774b7d196949/src/integration-tests/test_puts_retransmission.py#L411-L412
https://github.com/bloomberg/blazingmq/blob/0e8bbf26ffd8d5ef211c9347a61c774b7d196949/src/integration-tests/test_puts_retransmission.py#L59
[...]

PR #603, opened 2025-02-07
Fix: m_bmqtool::Poster: remove unsafe format string passed

sameep-git:
Closes: #593

1. We remove `sequentialMessagePattern` from being passed as a format string by appending it with a mutable suffix for `d_numMessagesPosted`.
2. Remove the need for `%d` in the integration test file `test_puts_retransmission.py`.

### DIFF
--- a/src/integration-tests/test_puts_retransmission.py
+++ b/src/integration-tests/test_puts_retransmission.py
@@ -205,7 +205,6 @@ class TestPutsRetransmission:
                 + r"msg\s*"  # "msg"
                 + r"(\d+)"
             )  # %d
-            )
 
             re_confirm = re.compile(
                 "(?i)"  # case insensitive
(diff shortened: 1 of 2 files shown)


src/integration-tests/test_puts_retransmission.py
```diff
@@ -203,8 +203,9 @@ def error(test_logger, prefix, p1, p2=None):
                 + re.escape(uri[0])  # queue url
                 + r"\|(.+)\|"  # |GUID|
                 + r"msg\s*"  # "msg"
-                + r"(\d+)\|"
+                + r"(\d+)"
             )  # %d
+            )
```
678098, line 208: Extra bracket:
```suggestion
```

678098 (approved):
Nice work @sameep-git!

PR #607, opened 2025-02-12
Perf[MQB]: cache support for MESSAGE_PROPERTIES_EX on PUSH path

678098:
This PR gets rid of the repeated `bmqp::ProtocolUtil::hasFeature` calls per every PUSH message.

In this listing, ~2% of all CPU time is lost on these calls:

Screenshot 2025-02-12 at 19 04 39


dorjesinpo (approved):
thank you

PR #618, opened 2025-02-19
Fix[BMQ]: unsafe schemalearner use

dorjesinpo:
Cannot use `bmqp::SchemaLearner` in both FSM and event threads.
Instead, call the leaner in FSM and cache the result (schema) in the generated event.

### DIFF
--- a/src/groups/bmq/bmqa/bmqa_message.h
+++ b/src/groups/bmq/bmqa/bmqa_message.h
@@ -60,7 +60,6 @@
 // BMQ
 
 #include <bmqa_queueid.h>
-#include <bmqp_messageproperties.h>
 #include <bmqt_compressionalgorithmtype.h>
 #include <bmqt_correlationid.h>
 #include <bmqt_messageguid.h>
@@ -82,6 +81,10 @@ namespace bmqimp {
 class Event;
 }
 
+namespace bmqp {
+class MessageProperties_Schema;
+}
+
 namespace bmqa {
 
 // FORWARD DECLARATION
@@ -101,6 +104,8 @@ class MessageProperties;
 ///                      `correlationId`), then they should be reset in
 ///                      `bmqa::MessageIterator.nextMessage()`.
 struct MessageImpl {
+    // FORWARD DECLARATION
+
     // PUBLIC DATA
 
     /// Pointer to the Event this message is associated with
@@ -118,7 +123,7 @@ struct MessageImpl {
     /// SubscriptionHandle this message is associated with
     bmqt::SubscriptionHandle d_subscriptionHandle;
 
-    bmqp::MessageProperties::SchemaPtr d_schema_sp;
+    bsl::shared_ptr<const bmqp::MessageProperties_Schema> d_schema_sp;
 
 #ifdef BMQ_ENABLE_MSG_GROUPID
     /// Optional GroupId this message is associated with
--- a/src/groups/bmq/bmqa/bmqa_message.h
+++ b/src/groups/bmq/bmqa/bmqa_message.h
@@ -60,6 +60,7 @@
 // BMQ
 
 #include <bmqa_queueid.h>
+#include <bmqp_messageproperties.h>
 #include <bmqt_compressionalgorithmtype.h>
 #include <bmqt_correlationid.h>
 #include <bmqt_messageguid.h>
@@ -116,6 +117,8 @@ struct MessageImpl {
 
     /// SubscriptionHandle this message is associated with
     bmqt::SubscriptionHandle d_subscriptionHandle;
+
+    bmqp::MessageProperties::SchemaPtr d_schema_sp;
 
 #ifdef BMQ_ENABLE_MSG_GROUPID
     /// Optional GroupId this message is associated with
--- a/src/groups/bmq/bmqimp/bmqimp_event.h
+++ b/src/groups/bmq/bmqimp/bmqimp_event.h
@@ -163,6 +163,14 @@ class Event {
         enum Enum { e_UNINITIALIZED, e_READ, e_WRITE };
     };
 
+    struct MessageContext {
+        /// This event can contain multiple messages and needs to cache context
+        /// for each message.
+        bmqt::CorrelationId                d_correlationId;
+        unsigned int                       d_subscriptionHandleId;
+        bmqp::MessageProperties::SchemaPtr d_schema_sp;
+    };
+
   private:
     // PRIVATE TYPES
     typedef bsls::ObjectBuffer<bmqp::PutEventBuilder> PutEventBuilderBuffer;
@@ -251,8 +259,7 @@ class Event {
     // has been constructed in its buffer.  We
     // cannot rely on 'd_msgEventMode'
 
-    bsl::vector<bsl::pair<bmqt::CorrelationId, unsigned int> >
-        d_correlationIds;
+    bsl::vector<MessageContext> d_contexts;
     // Ordered list of correlationIds for all
     // the PUT/ACK/PUSH messages in this event.
     // For PUSH messages optional corresponding
@@ -466,13 +473,7 @@ class Event {
     /// undefined unless 0 <= `position` < numCorrrelationIds(), and event's
     /// type() is MESSAGEEVENT, `messageEventMode()` is READ and the
     /// underlying raw event is of type ACK, PUT or PUSH.
-    const bmqt::CorrelationId& correlationId(int position) const;
-
-    /// Return the subscriptionId at the specified 'position'. Behavior is
-    /// undefined unless 0 <= 'position' < numCorrrelationIds(), and event's
-    /// type() is MESSAGEEVENT, 'messageEventMode()' is READ and the
-    /// underlying raw event is of type PUSH.
-    unsigned int subscriptionId(int position) const;
+    const MessageContext& context(int position) const;
 
     // MANIPULATORS
 
@@ -500,9 +501,11 @@ class Event {
     /// maintained by this instance.  The behavior is undefined unless
     /// event's type() is MESSAGEEVENT, 'messageEventMode()' is READ and the
     /// underlying raw event is of type ACK, PUT or PUSH.
-    void addCorrelationId(const bmqt::CorrelationId& correlationId,
-                          unsigned int               subscriptionHandleId =
-                              bmqt::SubscriptionHandle::k_INVALID_HANDLE_ID);
+    void addContext(const bmqt::CorrelationId& correlationId,
+                    unsigned int               subscriptionHandleId =
+                        bmqt::SubscriptionHandle::k_INVALID_HANDLE_ID,
+                    bmqp::MessageProperties::SchemaPtr schema =
+                        bmqp::MessageProperties::SchemaPtr());
 
     /// Insert the specified `queue` to the queues and the specified
     /// `corrId` to the list of correlationIds associated with this event.
@@ -731,10 +734,10 @@ inline int Event::numCorrrelationIds() const
     BSLS_ASSERT_SAFE(d_rawEvent.isAckEvent() || d_rawEvent.isPutEvent() ||
                      d_rawEvent.isPushEvent());
 
-    return static_cast<int>(d_correlationIds.size());
+    return static_cast<int>(d_contexts.size());
 }
 
-inline const bmqt::CorrelationId& Event::correlationId(int position) const
+inline const Event::MessageContext& Event::context(int position) const
 {
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(type() == EventType::e_MESSAGE);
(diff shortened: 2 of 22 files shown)


src/groups/bmq/bmqp/bmqp_schemalearner.cpp
```diff
@@ -129,31 +138,52 @@ SchemaLearner::Context SchemaLearner::createContext(int foreignId)
...
+    BSLS_ASSERT_SAFE(contextHandle);
+
+    if (input.isRecycled()) {
+        contextHandle->d_schema_sp.reset();
+        // Must destroy previously learned Schema
+    }
+
+    const SchemaPtr& result = contextHandle->d_schema_sp;
+
+    if (!result) {
```
678098, line 176: Can we assume that it's an unlikely scenario? Or is it better to evaluate the expression without such assumptions?

src/groups/bmq/bmqp/bmqp_schemalearner.cpp
```diff
@@ -129,31 +138,52 @@ SchemaLearner::Context SchemaLearner::createContext(int foreignId)
...
+        // Learn new Schema by reading all MessageProperties.
+        MessageProperties mps(d_allocator_p);
+        int               rc = mps.streamIn(blob, input, result);
+
+        if (rc == 0) {
```
678098, line 181: If `rc != 0` we will silently return an empty `SchemaPtr`. Should we log something or skip the message?

src/groups/bmq/bmqp/bmqp_messageproperties.h
```diff
@@ -99,6 +99,8 @@ class MessageProperties_Schema {
                              bslma::Allocator*        basicAllocator);
     MessageProperties_Schema(const MessageProperties_Schema& other);
 
+    ~MessageProperties_Schema();
```
678098, line 102: Let's also add `BSLS_KEYWORD_FINAL` to the class head, since we declared a non-virtual destructor?

```
class MessageProperties_Schema BSLS_KEYWORD_FINAL {
```

src/groups/bmq/bmqp/bmqp_schemalearner.cpp
```diff
@@ -45,6 +45,8 @@ struct SchemaLearner::SchemaHandle {
     // If 'false', needs new translation.
 
     SchemaHandle(SchemaIdType id);
+
+    ~SchemaHandle();
```
678098, line 49: Let's also add `BSLS_KEYWORD_FINAL` to the class head, since we declared a non-virtual destructor?

```
struct SchemaLearner::SchemaHandle BSLS_KEYWORD_FINAL {
```

src/groups/bmq/bmqimp/bmqimp_queuemanager.cpp
```diff
@@ -288,26 +289,39 @@ void QueueManager::resetState()
...
+    // Use 'subscriptionHandle' instead of the internal
+    // 'info.d_subscriptionId' so that
+    // 'bmqimp::Event::subscriptionId()' returns 'subscriptionHandle'
+
+    queueEvent->insertQueue(info.d_subscriptionId, queue);
+
+    bmqp::MessageProperties::SchemaPtr schema = schemaLearner().learn(
+        schemaLearner().createContext(queueId),
+        bmqp::MessagePropertiesInfo(info.d_header),
+        bdlbb::Blob());
```
678098, line 322: Looking at `learn` implementation:

```
SchemaLearner::SchemaPtr
SchemaLearner::learn(Context&                     context,
                     const MessagePropertiesInfo& input,
                     const bdlbb::Blob&           blob)
{
    const SchemaIdType inputId = input.schemaId();

    if (!isPresentAndValid(inputId)) {
        // Nothing to do
        return SchemaPtr();  // RETURN
    }
```

We have an early return if `schemaId` is not present or valid. Does it happen for all messages that don't have message properties? If it can happen that frequently, we do a lot of unnecessary work here:
- Pass args and call `schemaLearner().createContext(queueId)`
- Within `schemaLearner().createContext(queueId)` make lookup in a map (we don't use its result)
- Construct `bdlbb::Blob()` that we don't use
- Pass args and call `SchemaLearner::learn` just to return early

Do you think it's worth to avoid these extra operations if we don't have schemaId?

src/groups/bmq/bmqimp/bmqimp_queuemanager.cpp
```diff
@@ -355,6 +362,33 @@ int QueueManager::onPushEvent(QueueManager::EventInfos* eventInfos,
...
+            schema = *schemaHolder;
+            if (!schema) {
+                // Learn new Schema by reading all MessageProperties.
+                bdlbb::Blob appData(d_allocator_p);
+                if (msgIterator.loadApplicationData(&appData)) {
+                    bmqp::MessageProperties mps(d_allocator_p);
+
+                    if (mps.streamIn(appData, input.isExtended()) == 0) {
+                        // Learn new schema.
+                        *schemaHolder = schema = mps.makeSchema(d_allocator_p);
```
678098, line 386: Same concern about the number of `SchemaPtr` copies.

src/groups/bmq/bmqimp/bmqimp_queuemanager.cpp
```diff
@@ -288,26 +289,32 @@ void QueueManager::resetState()
...
 }
 
-const QueueManager::QueueSp
-QueueManager::observePushEvent(bmqt::CorrelationId* correlationId,
-                               unsigned int*        subscriptionHandleId,
-                               const bmqp::EventUtilQueueInfo& info)
+void QueueManager::observePushEvent(Event*                          queueEvent,
+                                    const bmqp::EventUtilQueueInfo& info)
 {
+    bmqt::CorrelationId correlationId;
```
678098, line 295:
```suggestion
    // PRECONDITIONS
    BSLS_ASSERT_SAFE(queueEvent);

    bmqt::CorrelationId correlationId;
```

src/groups/bmq/bmqimp/bmqimp_event.h
```diff
@@ -496,13 +502,15 @@ class Event {
...
-    /// 'subscriptionId' to the list of correlationId-subscriptionId pairs
-    /// maintained by this instance.  The behavior is undefined unless
-    /// event's type() is MESSAGEEVENT, 'messageEventMode()' is READ and the
-    /// underlying raw event is of type ACK, PUT or PUSH.
-    void addCorrelationId(const bmqt::CorrelationId& correlationId,
-                          unsigned int               subscriptionHandleId =
-                              bmqt::SubscriptionHandle::k_INVALID_HANDLE_ID);
+    /// 'subscriptionId' and `schema` to the context maintained by this
+    /// instance.  The behavior is undefined unless event's type() is
+    /// MESSAGEEVENT, 'messageEventMode()' is READ and the underlying raw event
```
678098, line 507: There is no enum with the name `MESSAGEEVENT`, better to mention `e_MESSAGE`. This is a common problem for docs in this file. I can make this change later

src/groups/bmq/bmqa/bmqa_message.h
```diff
@@ -100,6 +104,8 @@ class MessageProperties;
 ///                      `correlationId`), then they should be reset in
 ///                      `bmqa::MessageIterator.nextMessage()`.
 struct MessageImpl {
+    // FORWARD DECLARATION
+
```
678098, line 108:
```suggestion
```

src/groups/bmq/bmqimp/bmqimp_event.h
dorjesinpo: Those checks are gone anyway, so the comment should be removed

src/groups/bmq/bmqp/bmqp_eventutil.cpp
```diff
@@ -372,13 +375,38 @@ Flattener::packMesage(const Protocol::SubQueueInfosArray& subQInfo)
         return result;  // RETURN
     }
 
+    int                               queueId = header.queueId();
+    const bmqp::MessagePropertiesInfo input(header);
+
+    bmqp::MessageProperties::SchemaPtr schema;
```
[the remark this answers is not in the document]
dorjesinpo: Agree, that is excessive.  Refactoring

src/groups/bmq/bmqimp/bmqimp_event.h
```diff
@@ -163,6 +163,19 @@ class Event {
         enum Enum { e_UNINITIALIZED, e_READ, e_WRITE };
     };
 
+    struct MessageContext {
+        /// This event can contain multiple messages and needs to cache context
+        /// for each message.
+        bmqt::CorrelationId                d_correlationId;
+        unsigned int                       d_subscriptionHandleId;
+        bmqp::MessageProperties::SchemaPtr d_schema_sp;
```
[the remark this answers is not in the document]
dorjesinpo: Can't unless explicitly define copy constructor

src/groups/bmq/bmqimp/bmqimp_queuemanager.h
```diff
@@ -342,6 +343,8 @@ class QueueManager {
     void updateSubscriptions(const bsl::shared_ptr<Queue>&         queue,
                              const bmqp_ctrlmsg::StreamParameters& config);
 
+    bmqp::SchemaLearner& schemaLearner();
```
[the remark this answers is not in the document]
dorjesinpo: We do need this (the call was missing, now it is engaged)

src/groups/bmq/bmqp/bmqp_schemalearner.cpp
```diff
@@ -129,31 +138,38 @@ SchemaLearner::Context SchemaLearner::createContext(int foreignId)
...
+    HandlesMap::iterator it = context->d_handles.find(inputId);
 
-    if (contextHandle) {
-        if (input.isRecycled()) {
-            contextHandle->d_schema_sp.reset();
-            // Must destroy previously learned Schema
-        }
+    if (it == context->d_handles.end()) {
+        HandlePtr newHandle(new (*d_allocator_p) SchemaHandle(inputId),
+                            d_allocator_p);
```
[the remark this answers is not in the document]
dorjesinpo: Should not be performance-critical.  We create new schema only when the message properties sequence changes

src/groups/bmq/bmqa/bmqa_message.h
```diff
@@ -117,6 +121,8 @@ struct MessageImpl {
     /// SubscriptionHandle this message is associated with
     bmqt::SubscriptionHandle d_subscriptionHandle;
 
+    bsl::shared_ptr<const bmqp::MessageProperties_Schema> d_schema_sp;
```
hallfox: This looks like a private component of `MessageProperties`, we should probably be referring to it through the typedef
```suggestion
    bmqp::MessageProperties::SchemaPtr d_schema_sp;
```

dorjesinpo:
> Hi @dorjesinpo, do you want to make any more changes to this PR?

Hi.  Other than the typedef change, no.  Was waiting for the next round of your review

src/groups/bmq/bmqimp/bmqimp_queuemanager.cpp
```diff
@@ -376,20 +415,22 @@ int QueueManager::onPushEvent(QueueManager::EventInfos* eventInfos,
...
             else {
-                eventInfos->back().d_ids.push_back(bmqp::EventUtilQueueInfo(
-                    subQueueInfos[0].id(),
-                    msgIterator.header(),
-                    msgIterator.applicationDataSize()));
+                eventInfos->back().d_ids.push_back(
+                    bmqp::EventUtilQueueInfo(msgIterator.header(),
+                                             subQueueInfos[0].id(),
+                                             msgIterator.applicationDataSize(),
+                                             schema));
```
678098: There are still extra `schema` copies here, can do the same change you did in another file:
```
    bmqp::MessageProperties::SchemaPtr* schema_p = 0;

    schema_p = d_schemaLearner.observe(d_schemaLearner.createContext(queueId),
                                       input);

    if (schema_p) {
        if (schema_p->get() == 0) {
            // Learn new Schema by reading all MessageProperties.
            bmqp::MessageProperties mps(d_allocator_p);

            if (mps.streamIn(d_appData, input.isExtended()) == 0) {
                // Learn new schema.
                *schema_p = mps.makeSchema(d_allocator_p);
            }
        }
    }
```

src/groups/bmq/bmqa/bmqa_message.h
dorjesinpo: See https://github.com/bloomberg/blazingmq/pull/618#discussion_r1967815484

We can return to that version, sure

src/groups/bmq/bmqimp/bmqimp_queuemanager.cpp
```diff
@@ -355,6 +367,38 @@ int QueueManager::onPushEvent(QueueManager::EventInfos* eventInfos,
...
+        const bmqp::MessagePropertiesInfo input(header);
+
+        bmqp::MessageProperties::SchemaPtr* schema_p = d_schemaLearner.observe(
+            d_schemaLearner.createContext(queueId),
+            input);
+
+        bmqp::MessageProperties::SchemaPtr schema;
+
+        if (schema_p) {
+            if (!schema_p->get() == 0) {
```
678098: Is this `!` a bug?
```suggestion
            if (schema_p->get() == 0) {
```

dorjesinpo: > Is this `!` a bug?

it is! Amended

src/groups/bmq/bmqimp/bmqimp_event.h
```diff
@@ -731,32 +752,20 @@ inline int Event::numCorrrelationIds() const
...
-
-inline unsigned int Event::subscriptionId(int position) const
-{
-    // PRECONDITIONS
-    BSLS_ASSERT_SAFE(type() == EventType::e_MESSAGE);
-    BSLS_ASSERT_SAFE(messageEventMode() == MessageEventMode::e_READ);
-    BSLS_ASSERT_SAFE(d_rawEvent.isPushEvent());
-    BSLS_ASSERT_SAFE(0 <= position);
-    BSLS_ASSERT_SAFE(static_cast<int>(d_correlationIds.size()) > position);
+    BSLS_ASSERT_SAFE(static_cast<int>(d_contexts.size()) > position);
```
678098, line 766: Would prefer to reorder so the smaller value is on the left
```suggestion
    BSLS_ASSERT_SAFE(position < static_cast<int>(d_contexts.size()));
```

678098 (approved):
Looks good

PR #620, opened 2025-02-19
fix Remove storage not found when removing queue

dorjesinpo:
```
// While transitioning to CSL, Replica can receive both
// QueueUnassignmentAdvisory and Queue DELETION.  The latter can delete
// the storage.
```
Removing the ERROR log in this case


kaikulimu (approved):
lgtm

PR #605, opened 2025-02-11
Refactor aligned printer

ariraein:
*Issue number of the reported bug or feature request: #<number>*

**Describe your changes**
A clear and concise description of the changes you have made.

**Testing performed**
Describe the testing you have performed to ensure that the bug has been addressed, or that the new feature works as planned.

**Additional context**
Add any other context about your contribution here.

PR #621, opened 2025-02-20
Feat[bmqstoragetool]: Add unit tests for m_bmqstoragetool::Printer implementations

waldgange:


src/applications/bmqstoragetool/m_bmqstoragetool_printer.t.cpp
```diff
@@ -0,0 +1,2577 @@
...
+/// Calculate record offset
+template <typename RECORD_TYPE>
+bsls::Types::Uint64 recordOffset(const RECORD_TYPE& record)
+{
+    BSLS_ASSERT_SAFE(record.header().sequenceNumber() > 0);
+    return k_HEADER_SIZE + (record.header().sequenceNumber() - 1) *
+                               mqbs::FileStoreProtocol::k_JOURNAL_RECORD_SIZE;
+}
+
+class my_CountingAllocator : public bslma::Allocator {
```
alexander-e1off, line 67: Where is this class used?

alexander-e1off (commented):
LGTM, one question.

src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp
```diff
@@ -1895,9 +1869,7 @@ static void test18_searchMessagesByOffsetsRange()
         if (rtype == RecordType::e_MESSAGE) {
             const MessageRecord& msg = *reinterpret_cast<const MessageRecord*>(
                 recordIter->second.buffer());
-            const bsls::Types::Uint64& offset =
-                msg.header().sequenceNumber() *
-                mqbs::FileStoreProtocol::k_JOURNAL_RECORD_SIZE;
+            const bsls::Types::Uint64& offset = recordOffset(msg);
```
678098, line 1872: Do we really need a const-ref to a basic type?

src/applications/bmqstoragetool/m_bmqstoragetool_commandprocessorfactory.t.cpp
```diff
@@ -48,11 +48,12 @@ static void test1_breathingTest()
         new (*bmqtst::TestHelperUtil::allocator()) FileManagerMock(),
         bmqtst::TestHelperUtil::allocator());
 
+    bmqu::MemOutStream resultStream(bmqtst::TestHelperUtil::allocator());
     bslma::ManagedPtr<CommandProcessor> cmdProcessor =
         CommandProcessorFactory::createCommandProcessor(
             &params,
             fileManager,
-            bsl::cout,
+            resultStream,
```
678098, line 56: Here in other places we pass a local `MemOutStream` and do nothing with it after call. Do we need to pass it then?

src/applications/bmqstoragetool/m_bmqstoragetool_printer.cpp
```diff
@@ -430,15 +440,16 @@ class HumanReadablePrinter : public Printer {
                       << queueOpRecordsCount << '\n';
 
             bsl::vector<const char*> fields(d_allocator_p);
+            fields.reserve(4);
```
678098, line 443: Will probably have conflict with https://github.com/bloomberg/blazingmq/pull/605

src/applications/bmqstoragetool/m_bmqstoragetool_printer.cpp
```diff
@@ -535,13 +546,31 @@ class JsonPrinter : public Printer {
     bsl::ostream&     d_ostream;
     bslma::Allocator* d_allocator_p;
     mutable bool      d_braceOpen;
+    mutable bool      d_firstRaw;
+
+    void openBraceIfNotOpen(const std::string& fieldName) const
+    {
+        if (!d_braceOpen) {
+            d_ostream << "  \"" << fieldName << "\": [\n";
+            d_braceOpen = true;
```
678098, line 555: Is it possible to have more than one layer of brackets? The alternative would be to use `int balance` and keep `balance >= 0`, `++balance` on open and `--balance` on close

src/applications/bmqstoragetool/m_bmqstoragetool_printer.t.cpp
waldgange: Oh, good notice! I used it for debugging the default allocator usage and forgot to delete. I'll delete it in the next commit

src/applications/bmqstoragetool/m_bmqstoragetool_printer.cpp
waldgange: Yes, it will. But it already has a conflict as both JsonPrinter and AllignedPrinter are used as template arguments in some places, e.g.
https://github.com/bloomberg/blazingmq/blob/d43b47354ad0551636e3484a61c35305fca73ab9/src/applications/bmqstoragetool/m_bmqstoragetool_recordprinter.h#L54-L55
So the interfaces of both printers must be changed in one commit. The author will notice compilation errors after rebase

src/applications/bmqstoragetool/m_bmqstoragetool_printer.cpp
waldgange: It's not supported at the moment. To suport this the class must become more complicated as we'll have to introduse and support such classes as Json objects and arrays. Now we deal only with primitives. Now this situation is handled manually outside the class

src/applications/bmqstoragetool/m_bmqstoragetool_journalfileprocessor.t.cpp
waldgange: No, I'll remove it

src/applications/bmqstoragetool/m_bmqstoragetool_commandprocessorfactory.t.cpp
waldgange: You a right, we don't need it. I'll bring bsl::cout back

678098 (approved):
Looks good

PR #626, opened 2025-02-21
CI: check C++03 standard build

678098:
The new c++03 build workflow can catch accidental usage of features not supported by solaris compiler, for example:
```
/home/runner/work/blazingmq/blazingmq/src/groups/bmq/bmqu/bmqu_blob.cpp:588:19: error: ‘ret’ does not name a type
  588 |     auto          ret = findOffset(&end, blob, start, length);
```

### DIFF
--- a/src/groups/bmq/bmqex/bmqex_executor.t.cpp
+++ b/src/groups/bmq/bmqex/bmqex_executor.t.cpp
@@ -112,7 +112,7 @@ class StatExecutor {
 
     /// Perform `++statistics()->d_postCallCount` followed by `f()`.
     template <class FUNCTION>
-    void post(bslmf::MovableRef<FUNCTION> f) const
+    void post(FUNCTION f) const
     {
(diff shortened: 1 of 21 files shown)


src/groups/bmq/bmqex/bmqex_executor.t.cpp
```diff
@@ -115,15 +115,15 @@ class StatExecutor {
     void post(FUNCTION f) const
     {
         ++d_statistics_p->d_postCallCount;
-        f();
+        bslmf::MovableRefUtil::access(f)();
```
hallfox, line 118: Why is this necessary? Is `FUNCTION` being used as a `MovableRef<T>`?

678098: Yes, it was one if the compilation errors

hallfox (dismissed):

PR #627, opened 2025-02-21
Feat[bmq]: Add interfaces in DTTracer to serialize and deserialize span

emelialei88:


pniedzielski:
Let's be careful about when we merge this, because it's a breaking change to the plugin interface.  We should definitely call attention to this.

src/groups/bmq/bmqpi/bmqpi_dttracer.h
```diff
@@ -56,6 +57,23 @@ class DTTracer {
         const bsl::shared_ptr<DTSpan>& parent,
         const bsl::string_view&        operation,
         const DTSpan::Baggage&         baggage = DTSpan::Baggage()) const = 0;
+
+    /// Serialize the underlying context for a specified `DTSpan` and return
```
678098, line 61:
```suggestion
    /// Serialize the underlying context for the specified `DTSpan` and return
```

src/groups/bmq/bmqpi/bmqpi_dttracer.t.cpp
```diff
@@ -35,6 +35,17 @@ struct DTTracerTestImp : public bsls::ProtocolTestImp<bmqpi::DTTracer> {
         const bsl::shared_ptr<bmqpi::DTSpan>& parent,
         const bsl::string_view&               operation,
         const bmqpi::DTSpan::Baggage& baggage) const BSLS_KEYWORD_OVERRIDE;
+
+    virtual int serializeSpan(bsl::vector<unsigned char>*           out,
```
678098, line 39:
```suggestion
    int serializeSpan(bsl::vector<unsigned char>*           out,
```

Can simplify, but also need to run formatter

src/groups/bmq/bmqpi/bmqpi_dttracer.t.cpp
```diff
@@ -35,6 +35,17 @@ struct DTTracerTestImp : public bsls::ProtocolTestImp<bmqpi::DTTracer> {
...
+                              const bsl::shared_ptr<bmqpi::DTSpan>& dtSpan)
+        const BSLS_KEYWORD_OVERRIDE;
+
+    virtual int deserializeAndCreateChildSpan(
```
678098, line 43:
```suggestion
   int deserializeAndCreateChildSpan(
```

Can simplify, but also need to run formatter

src/groups/bmq/bmqpi/bmqpi_dttracer.t.cpp
```diff
@@ -100,8 +126,13 @@ static void test1_breathingTest()
     BMQTST_ASSERT(tracer.testVirtualDestructor());
 
     PV("Verify that all methods are public and virtual");
-    bmqpi::DTSpan::Baggage empty;
+    bmqpi::DTSpan::Baggage     empty;
+    bsl::vector<unsigned char> emptyVec;
```
678098, line 130:
```suggestion
    bmqpi::DTSpan::Baggage     empty(bmqtst::TestHelperUtil::allocator());
    bsl::vector<unsigned char> emptyVec(bmqtst::TestHelperUtil::allocator());
```

src/groups/bmq/bmqpi/bmqpi_dttracer.h
```diff
@@ -56,6 +57,23 @@ class DTTracer {
...
+    /// the bit stream as `out`.  Return 0 on success, or a non-zero error code
+    /// on error.
```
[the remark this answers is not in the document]
pniedzielski: What is the state of `out` on error?  Probably we want it to be valid-but-unspecified.

src/groups/bmq/bmqpi/bmqpi_dttracer.h
```diff
@@ -56,6 +57,23 @@ class DTTracer {
...
+    virtual int serializeSpan(bsl::vector<unsigned char>*    out,
+                              const bsl::shared_ptr<DTSpan>& dtSpan) const = 0;
+
+    /// Deserialize the specified bit stream `in` into a span context.
+    /// Create a child span and pass into `out` based on the deserialized
+    /// parent span context, the given `operation`, and an optional key-value
+    /// tags `baggage`.  Return 0 on success, or a non-zero error code on
+    /// error.
```
pniedzielski, line 71:
```
/// Deserialize the specified `in` buffer as a span, and then construct a new `DTSpan` in `out`, representing `operation` as a child of the deserialized span and having the key-value tags defined by `baggage`.
```

pniedzielski: I think we can do better with the parameter names, too.  `out` and `in` don't tell me much.  `child` and `buffer` tell me more.

pniedzielski (changes requested):
We should be very careful with when we merge this change, as I noted above.  A few comments on the interface:

src/groups/bmq/bmqimp/bmqimp_brokersession.t.cpp
```diff
@@ -2834,6 +2834,49 @@ class DTTestTracer : public bmqpi::DTTracer {
...
         return result;
     }
+
+    int serializeSpan(bsl::vector<unsigned char>*           out,
+                      const bsl::shared_ptr<bmqpi::DTSpan>& dtSpan) const
+        BSLS_KEYWORD_OVERRIDE
+    {
+        bmqu::MemOutStream event(d_allocator_p);
+
+        out->resize(dtSpan->operation().size() + 1);
```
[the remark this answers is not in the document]
pniedzielski: I think for #1, it matters in C++03 but not in C++11.  `std::string` says:

> The elements of a `basic_string` are stored contiguously, that is, for a `basic_string` `s`, `&*(s.begin() + n) == &*s.begin() + n` for any `n` in `[​0​, s.size())`, and `*(s.begin() + s.size())` has value `CharT()` (a null terminator) (since C++11); or, equivalently, a pointer to `s[0]` can be passed to functions that expect a pointer to the first element of an array (until C++11), or a null-terminated array (since C++11) of `CharT`.

When deserializing below, we construct a string of length `std::distance(in.begin(), in.end())`.  In C++03, end of story, we don't null-terminate and read out from that buffer immediately after, and so this test has undefined behavior.  In C++11, `*(parentOperation.begin() + std::distance(in.begin(), in.end())) == '\0'` by the invariant of `basic_string`, so we're safe.

pniedzielski (dismissed):
Looks good.  We'll schedule the merge to minimize churn.

678098 (dismissed):

PR #632, opened 2025-02-26
Refactor[MQB]: simplify evaluator calls

678098:


src/groups/mqb/mqbblp/mqbblp_routers.cpp
```diff
@@ -657,7 +651,7 @@ Routers::RoundRobin::iterateGroups(const Visitor&               visitor,
             BSLS_ASSERT_SAFE(!group.d_highestSubscriptions.empty());
 
             if (group.d_canDeliver) {
-                if (group.evaluate(message->appData())) {
+                if (group.evaluate()) {
```
678098, line 654: We don't need to load appData here, instead:
1. We point properties reader to the current message iterator
2. Properties reader will read appData lazily, here:
https://github.com/bloomberg/blazingmq/blob/8be74d097e00500b917aca61728aac214847a1c1/src/groups/mqb/mqbblp/mqbblp_routers.cpp#L121

dorjesinpo (approved):
thank you

PR #640, opened 2025-02-28
Doc[bmq]: Fix description of PENDING queue state

chrisbeard:


pniedzielski (approved):
CI failure unrelated, caused by BDE.

PR #641, opened 2025-03-04
Fix: Remove expired elements from pushstream

dorjesinpo:
If `QueueEngineUtil_AppState::d_resumePoint` gets removed, an attempt to deliver it results in failure reading the data from the storage:
```
FATAL mqbblp_pushstream.cpp:86 Assertion failed: mqbi::StorageResult::e_SUCCESS == rc
```

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
@@ -738,7 +738,7 @@ bool QueueEngineUtil_AppsDeliveryContext::processApp(
 
         // Early return.
         // If all Apps return 'e_NO_CAPACITY_ALL', stop the iteration
-        // (d_numApps == 0).
+        // (d_numApps == d_numStops).
 
         ++d_numStops;
 
@@ -956,7 +956,8 @@ QueueEngineUtil_AppState::deliverMessages(bsls::TimeInterval*          delay,
                          result == Routers::e_NO_CAPACITY ||
                          result == Routers::e_NO_SUBSCRIPTION)) {
                 BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
-                d_putAsideList.add(start->guid());
+
+                putAside(start->guid());
                 // Do not block other Subscriptions. Continue.
             }
             else if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(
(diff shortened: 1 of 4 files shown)


src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1503,8 +1503,21 @@ void RelayQueueEngine::beforeMessageRemoved(const bmqt::MessageGUID& msgGUID)
...
     }
+    else {
+        PushStreamIterator del(storage(),
+                               &d_pushStream,
+                               d_pushStream.d_stream.find(msgGUID));
+
+        if (!del.atEnd()) {
+            del.removeAllElements();
+
+            del.advance();
```
pniedzielski, line 1518: The state munging of `d_iterator`, `d_currentElement`, `d_currentOrdinal` is tricky, and looking at the mqbblp_pushstream component, I often find I need to study the code to make sure they're all set as we need.

I think I've convinced myself that this is correct; it's just the same as above.  I've a nagging feeling about this component though.  Let's ship this, and I'm going to spend some time this weekend thinking about ways to simplify the usage of mqbblp_pushstream.

src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -956,7 +956,8 @@ QueueEngineUtil_AppState::deliverMessages(bsls::TimeInterval*          delay,
                          result == Routers::e_NO_CAPACITY ||
                          result == Routers::e_NO_SUBSCRIPTION)) {
                 BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
-                d_putAsideList.add(start->guid());
+
+                putAside(start->guid());
```
pniedzielski, line 960: 👍🏻

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
dorjesinpo: It is dimensional matrix, essentially, where one diminution is `d_iterator` (by GUIDs) and another is `d_currentElement`/`d_currentOrdinal `.

pniedzielski: I understand the data structure.  I'm saying the interface is poor.

PR #644, opened 2025-03-07
fix Continuous ordinals in Proxy

dorjesinpo:
Make ordinals continuous in Proxy as well to fix the bug with App removal resulting in out of bound ordinal upon PUSH in `RemoteQueue`

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.h
+++ b/src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.h
@@ -287,7 +287,7 @@ class VirtualStorageCatalog {
 
     /// Configure this object as Proxy to skip statistics and checking age of
     /// Apps vs age of messages.
-    void confgiureAsProxy();
+    void configureAsProxy();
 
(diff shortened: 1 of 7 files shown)


src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.h
```diff
@@ -368,9 +366,9 @@ inline void VirtualStorageCatalog::setDefaultRda(int maxDeliveryAttempts)
     }
 }
 
-inline void VirtualStorageCatalog::setDiscontinuousOrdinals()
+inline void VirtualStorageCatalog::confgiureAsProxy()
 {
-    d_isProxy = false;
+    d_isProxy = true;
```
pniedzielski, line 371: 👍🏻

pniedzielski (changes requested):
Typo but looks good.  Control flow becomes a little clearer.

pniedzielski (approved):
Changes look good, nice side effect of making the logic here just a little clearer.

PR #646, opened 2025-03-07
Fix call bmqt::UriParser::initialize

dorjesinpo:
Fixing `MemorySanitizer: use-of-uninitialized-value`


pniedzielski (approved):
Let's test it!

678098:
The minimal example of this bug is here: https://github.com/bloomberg/blazingmq/pull/577/files
See also the internal ticket 177953779

PR #647, opened 2025-03-07
mqbstat_queuestats.cpp: Publish per app id metrics from proxies too

kaikulimu:
Since we report confirmation time only between the last hop and the consumer, we need to report app id metric from proxies in case they are the last hop.


kaikulimu:
After discussion with team, we decided not to support publishing app id metrics from proxies.

chrisbeard:
Notably, proxies don't have the full domain configuration (i.e. list of appIds) so they don't create stat subcontexts for each appId. Reconfiguring on the cluster also don't propagate appId information to proxies.

PR #649, opened 2025-03-10
Fix[IT]: reduce storage file sizes to allow run in GH CI

678098:
## Before

1472 MiB per cluster node
5888 MiB per 4-node cluster

Screenshot 2025-03-10 at 15 39 50

## After

344 MiB per cluster node
1376 MiB per 4-node cluster

Screenshot 2025-03-10 at 15 30 02


kaikulimu (approved):
lgtm

PR #655, opened 2025-03-12
Fix gc'ing old message in the presence of new app

dorjesinpo:
Handling the scenario:
1) Post a message
2) Add new App
3) Expire the message

While updating stats, recognize the message age greater than the App,.

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.cpp
+++ b/src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.cpp
@@ -311,7 +311,13 @@ VirtualStorageCatalog::gc(const bmqt::MessageGUID& msgGUID)
     for (VirtualStoragesIter it = d_virtualStorages.begin();
          it != d_virtualStorages.end();
          ++it) {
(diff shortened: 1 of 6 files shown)


src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.cpp
```diff
@@ -372,6 +379,11 @@ bsls::Types::Int64 VirtualStorageCatalog::seek(DataStreamIterator*   it,
             break;  // BREAK
         }
         // else 'dataStreamMessage' is older than this VirtualStorage
+
+        if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(bytes)) {
+            BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
+            bytes += data.d_size;
```
pniedzielski, line 385: Is this meant to be `*bytes`?

dorjesinpo: Thank you! Amended

PR #657, opened 2025-03-14
fix write app deletion record

dorjesinpo:
Fixing [https://github.com/bloomberg/blazingmq/pull/585](https://github.com/bloomberg/blazingmq/pull/585) which did not write App deletion record when the App was empty

### DIFF
--- a/src/integration-tests/test_appids.py
+++ b/src/integration-tests/test_appids.py
@@ -758,6 +758,8 @@ def test_old_data_new_app(cluster: Cluster):
     new_app_1 = "new_app_1"
     _set_app_ids(cluster, default_app_ids + [new_app_1])
 
+    leader.capture(f"Registered appId '{new_app_1}'", timeout=5)
+
     # ---------------------------------------------------------------------
     # Post
     producer.post(tc.URI_FANOUT_SC, ["m2"], succeed=True, wait_ack=True)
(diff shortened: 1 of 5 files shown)


src/integration-tests/test_appids.py
```diff
@@ -752,6 +752,8 @@ def test_old_data_new_app(cluster: Cluster, domain_urls: tc.DomainUrls):
     new_app_1 = "new_app_1"
     set_app_ids(cluster, default_app_ids + [new_app_1], du)
 
+    leader.capture(f"Registered appId '{new_app_1}'", timeout=5)
```
kaikulimu, line 755: Does this log capture work to ensure synchronicity?

src/integration-tests/test_appids.py
```diff
@@ -871,3 +879,45 @@ def _verify_clients(andConfirm=False):
 
     leader.list_messages(du.domain_fanout, tc.TEST_QUEUE, 0, 100)
     assert leader.outputs_substr(f"Printing 0 message(s)", 5)
+
+
+def test_add_remove_add_app(cluster: Cluster, domain_urls: tc.DomainUrls):
+    """Trigger old message GC in the presence of new App.  Need to allocate
```
kaikulimu, line 885: Mention that the test is adding an app, then removing the app, then adding the same app again.

src/integration-tests/test_appids.py
dorjesinpo: The intent is to make sure the leader gets the reconfiguration event .

kaikulimu (approved):
lgtm

PR #660, opened 2025-03-14
Fix[MQB]: use CSL to update state on PartitionPrimaryAdvisory

emelialei88:
Similar to PR https://github.com/bloomberg/blazingmq/pull/581, this PR invokes CSL commit callback for PartitionPrimaryAdvisory and the partition primary assignment part of LeaderAdvisory, effectively making CSL the source of truth.


src/groups/mqb/mqbblp/mqbblp_clusterstatemanager.cpp
```diff
@@ -82,39 +82,16 @@ void ClusterStateManager::onCommit(
 
     // NOTE: Even when using old workflow, we still apply all advisories to the
     // CSL. We just don't invoke the commit callbacks.
-    // Make an exception for QueueUpdateAdvisory and QueueAssignmentAdvisory
+    // Make an exception for QueueUpdateAdvisory, QueueAssignmentAdvisory, and
+    // PartitionPrimaryAdvisory
```
kaikulimu, line 86:
```suggestion
    // Make an exception for QueueAssignmentAdvisory, QueueUpdateAdvisory, PartitionPrimaryAdvisory, and LeaderAdvisory
```

src/groups/mqb/mqbblp/mqbblp_clusterstatemanager.cpp
```diff
@@ -1909,14 +1567,6 @@ void ClusterStateManager::processLeaderAdvisory(
         return;  // RETURN
     }
 
-    // Process partition-primary mapping.  It's ok to process partition-primary
```
kaikulimu, line 1912: Why not just remove the entire `ClusterStateManager::processLeaderAdvisory`?

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -1465,29 +1441,21 @@ void ClusterUtil::sendClusterState(
         loadQueuesInfo(&advisory.queues(), clusterState);
     }
 
+    // Need to send the control message.
+    // Reason 1: The old version brokers should still receive the messages.
+    // Reason 2: For the new version brokers, `PartitionPrimary` and
```
kaikulimu, line 1446: Are you sure Reason 2 is valid? Wouldn't `ClusterStateManager::onCommit` for CSL Leader Advisory take care of this scenario?

emelialei88: `setLeaderStatus` is done in FSM mode not CSL. By removing `processLeaderAdvisory` will prevent the node to update the leader status to `ACTIVE`.

kaikulimu: Perhaps during `mqbblp::ClusterStateManager::onCommit`, you add a call to `setLeaderStatus`. Would that fix the issue?

src/groups/mqb/mqbblp/mqbblp_clusterstatemanager.cpp
emelialei88: I was hoping to deal with LeaderAdviory in a separate PR

PR #664, opened 2025-03-19
CI[Python]: Add pylint

waldgange:


.github/workflows/pylint.yaml
```diff
@@ -0,0 +1,41 @@
...
+          echo [pylint] > ${{ github.workspace }}/.lint-diffs
+          echo extensions=.py >> ${{ github.workspace }}/.lint-diffs
+      - name: Check the whole repo with pylint
+        run: |
+          pylint --output-format json -j 8 src > pylint.json || true
+          pylint-json2html < pylint.json > pylint.html
+      - name: Upload pylint results as artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: pylint_${{ env.RUN_UID }}
```
678098, line 36: `env.RUN_UID` doesn't exist in this context

waldgange: fixed

.github/workflows/pylint.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Check the whole repo with pylint
+        run: |
+          pylint --output-format json -j 8 src > pylint.json || true
+          pylint-json2html < pylint.json > pylint.html
+      - name: Upload pylint results as artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: pylint_result
+          path: ${{ github.workspace }}/pylint.html
+          retention-days: 5
```
678098, line 38: Artifact is uploaded as a compressed .zip file. Is it possible to save `.html` directly? What if compression level is set to zero?

678098: Maybe we can print linter results to GH Actions log then? The scheme with zip-compressed html requires too many manual steps when it can be avoided. Also the local file storage will be messy with these tmp files.
Another question, if we have a newly added linter error, will the GH Actions log print the exact place and error message, or will it just fail without telling where exactly?

678098 (approved):

PR #670, opened 2025-03-25
Feat[MQB]: refactor negotiator

emelialei88:


src/groups/mqb/mqbnet/mqbnet_initialconnectionhandler.h
```diff
@@ -0,0 +1,91 @@
...
+    /// initial connection.  The specified `context` is an in-out member
+    /// holding the initial connection context to use; and the
+    /// InitialConnectionHandler concrete implementation can modify some of the
+    /// members during the initial connection (i.e., between the
+    /// `initialConnect()` method and the invocation of the
+    /// `initialConnectionCb` method.  Note that if no initial connection is
+    /// needed, the `initialConnectionCb` may be invoked directly from inside
+    /// the call to `initialConnect()`.
+    virtual void
+    initialConnect(mqbnet::InitialConnectionHandlerContext* context,
```
dorjesinpo, line 83: See my other comment about `initialConnect` name

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.h
```diff
@@ -43,24 +44,27 @@ struct ConnectionType {
...
     bsl::shared_ptr<bmqio::Channel> d_channelSp;
 
     /// The callback to invoke to notify of the status of the negotiation.
-    mqbnet::Negotiator::NegotiationCb d_negotiationCb;
+    mqbnet::InitialConnectionHandler::InitialConnectionCb
+        d_initialConnectionCb;
 
     /// The negotiation message received from the remote peer.
-    bmqp_ctrlmsg::NegotiationMessage d_negotiationMessage;
+    bmqp_ctrlmsg::NegotiationMessage d_initialConnectionMessage_p;
```
dorjesinpo, line 67: The suffix `_p` denotes raw pointers,  This seems to be a value

dorjesinpo, line 67: This seems to be the last message received by the `InitialConnectionHandler` (before it is handed over to the negotiator)?

src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
```diff
@@ -0,0 +1,265 @@
...
+    if (rc != 0) {
+        bsl::string error(errStream.str().data(), errStream.str().length());
+        context->d_initialConnectionCb((rc * 10) +
+                                           rc_INVALID_NEGOTIATION_MESSAGE,
+                                       error,
+                                       session);
+        return;  // RETURN
+    }
+
+    switch (context->d_initialConnectionMessage_p.selectionId()) {
```
dorjesinpo, line 123: `decodeNegotiationMessage` does not have to cache the decoded message (`d_initialConnectionMessage_p`) if it is used in the same stack frame.

src/groups/mqb/mqba/mqba_initialconnectionhandler.h
```diff
@@ -0,0 +1,119 @@
...
+                      int*                              numNeeded,
+                      bdlbb::Blob*                      blob,
+                      const InitialConnectionContextSp& context);
+
+    /// Decode the initial connection messages received in the specified `blob`
+    /// and store it, on success, in the corresponding member of the specified
+    /// `context`, returning 0.  Return a non-zero code on error and
+    /// populate the specified `errorDescription` with a description of the
+    /// error.
+    int decodeNegotiationMessage(bsl::ostream& errorDescription,
```
dorjesinpo, line 92: When we add authentication, will this method read authentication messages?  If so, do we still want to call that `negotiation`?

dorjesinpo (changes requested):
Can we make the authenticator a separate PR if it is not used?

The prefix `initialConnect` appears many times, we may want to rethink the names.
The `connect` part does not match TCP listen context in which it is used.

We have `InitialConnectionHandlerContext` and `InitialConnectionContext`
Potentially, they can be even merged (the existing design has them split which is fine). We may need better names.

src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
```diff
@@ -0,0 +1,265 @@
...
+        mqbnet::ConnectionType::e_UNKNOWN;
+
+    // Reading for inbound request or continue to read
+    // after sending a request ourselves
+
+    if (context->isIncoming()) {
+        scheduleRead(initialConnectionContext);
+    }
+    else {
+        if (d_negotiator_mp->negotiateOutboundOrReverse(
```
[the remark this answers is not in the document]
emelialei88: No. It resides in `SessionNegotiator::negotiateOutboundOrReverse` now.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.h
```diff
@@ -401,10 +400,10 @@ class TCPSessionFactory {
         const bsl::shared_ptr<bmqio::Channel>&                  channel,
         const bsl::shared_ptr<bmqio::StatChannelFactoryHandle>& handle);
 
-    /// Asynchronously negotiate on the specified `channel` using the
+    /// Asynchronously initial connection on the specified `channel` using the
     /// specified `context`.
-    void negotiate(const bsl::shared_ptr<bmqio::Channel>&   channel,
-                   const bsl::shared_ptr<OperationContext>& context);
+    void initialConnect(const bsl::shared_ptr<bmqio::Channel>&   channel,
```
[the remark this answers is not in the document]
emelialei88: What about "handleInitialConnection"?

src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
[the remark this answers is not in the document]
emelialei88: I would imagine something similar. We read if it's an incoming request, otherwise maybe have a function like `authenticateOutboundOrReverse`.

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.h
emelialei88: Good catch!

emelialei88: We designed the protocol of `NegotiationMessage` to have both authentication and negotiation (maybe a debatable choice but since `NegotiationMessage` is a top-level structure, we either need to do this or add another top-level component for authentication. So this datatype should contain either authentication message or negotiation message.

src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
emelialei88: Do you mean in the `decodeNegotiationMessage` we shouldn't create an `Event`?

src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
```diff
@@ -262,72 +261,23 @@ void loadSessionDescription(bsl::string*                        out,
 // class SessionNegotiator
 // -----------------------
 
-void SessionNegotiator::readCallback(const bmqio::Status&        status,
-                                     int*                        numNeeded,
-                                     bdlbb::Blob*                blob,
-                                     const NegotiationContextSp& context)
+int SessionNegotiator::createSessionOnMsgType(
+    bsl::shared_ptr<mqbnet::Session>* session,
```
chrisbeard, line 265: We should check/assert that `session` is not null, e.g. `BSLS_ASSERT(session)`

src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
dorjesinpo: I meant that `decodeNegotiationMessage` does not need to assign to `d_negotiationMessage` (unless we need to keep the last received message in which case the name should be different).  The assignment can happen when we ready to pass the handling to the negotiator

dorjesinpo (changes requested):
A suggestion.  Why don't we keep the name `NegotiationContext`.  `InitialConnectionHandler` creates `NegotiationContext` when it is ready to pass the connection handling to the negotiator.   After all, this is the context for negotiator use.
That way, we clearly separate context for negotiation from the context for pre-negotiation phase (authentication).
And we have one less ``InitialConnection` prefix.

So, the chain looks like this `TCPSessionFactory_OperationContext` -> `InitialConnectionHandlerContext` (can be shortened to `InitialConnectionContext`) -> `NegotiationContext`

src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
```diff
@@ -0,0 +1,260 @@
...
+    if (negotiationMsg.has_value()) {
+        context->negotiationContext()->d_negotiationMessage =
+            negotiationMsg.value();
+        rc = d_negotiator_mp->createSessionOnMsgType(
+            &session,
+            context->negotiationContext());
+        if (rc == rc_CONTINUE_READ) {
+            scheduleRead(context);
+        }
+    }
```
dorjesinpo, line 136: we need to make sure `initialConnectionCompleteCb` always gets called.  Otherwise, the connection is stuck.

Regarding calling `initialConnectionCompleteCb`, there may be an option of calling it from `InitialConnectionHandler` only (never from the Negotiator) if the Negotiator is synchronous.  Could make the logic easier to follow.

dorjesinpo (changes requested):
Getting there.  Few more comments

dorjesinpo (changes requested):
some comments about error handling and we may shuffle the context data members

src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
```diff
@@ -262,71 +261,24 @@ void loadSessionDescription(bsl::string*                        out,
...
-    rc = decodeNegotiationMessage(errStream, context, inBlob);
-    if (rc != 0) {
-        bsl::string error(errStream.str().data(), errStream.str().length());
-        context->d_negotiationCb((rc * 10) + rc_INVALID_NEGOTIATION_MESSAGE,
-                                 error,
-                                 session);
-        return;  // RETURN
-    }
-
     switch (context->d_negotiationMessage.selectionId()) {
```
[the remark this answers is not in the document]
emelialei88: but `d_negotiationMessage` is not a pointer

src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
```diff
@@ -262,71 +261,24 @@ void loadSessionDescription(bsl::string*                        out,
...
 // -----------------------
 
-void SessionNegotiator::readCallback(const bmqio::Status&        status,
-                                     int*                        numNeeded,
-                                     bdlbb::Blob*                blob,
-                                     const NegotiationContextSp& context)
+int SessionNegotiator::createSessionOnMsgType(
+    bsl::ostream&                     errorDescription,
+    bsl::shared_ptr<mqbnet::Session>* session,
+    bool*                             isContinueRead,
```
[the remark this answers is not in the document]
emelialei88: @pniedzielski agrees with me on the use of bool instead of a return code. It seems to be more reasonable to have the return code indicating only either success or the type of error.

dorjesinpo (changes requested):
only one comment

dorjesinpo (approved):
Feel free to ignore my last comment.  It looks like `d_resultState_p` points to different types at different times.  Not entirely clean but we can leave this for now

src/groups/mqb/mqbnet/mqbnet_negotiationcontext.h
```diff
@@ -47,17 +48,12 @@ struct ConnectionType {
 // class NegotiationContext
 // ========================
 
-// VST for an implementation of NegotiatiorContext
+// VST for an implementation of NegotiationContext
```
pniedzielski, line 51: This ended up being a simple change. 👍🏻

pniedzielski (changes requested):
Overall approach is clean.  A few doc nits.  Otherwise we can ship in 94.x.

pniedzielski (approved):
Good to squash-merge

PR #673, opened 2025-03-27
Feat[MQB]: authn plugin interfaces

678098:


src/groups/mqb/mqbplug/package/mqbplug.mem
```diff
@@ -1,3 +1,4 @@
+mqbplug_authenticatorpluginfactory
```
emelialei88, line 1: It seems like our `mqbplug_authenticatorpluginfactory` has the same structure as `mqbplug_statconsumer`, but instead of naming it using the component name (i.e., `mqbplug_authenticator`), we call it `xxxpluginfactory`. Should we unify the way we name the plugins?

src/groups/mqb/mqbplug/mqbplug_authenticator.h
```diff
@@ -0,0 +1,227 @@
...
+
+  public:
+    // TRAITS
+    BSLMF_NESTED_TRAIT_DECLARATION(AuthenticationResult,
+                                   bslma::UsesBslmaAllocator)
+
+    // CREATORS
+
+    /// Construct this object using the optionally specified `allocator`.
+    explicit AuthenticationResult(bslma::Allocator* allocator = 0);
```
678098, line 97: We intend to pass an already constructed `AuthenticationResult` object to `Authenticator`, so it can set some fields. Due to this, we cannot use a constructor with all field values and cannot make fields `const`. We have to use setters to set field values and it's also useful to have `reset`

src/groups/mqb/mqbplug/mqbplug_authenticator.h
```diff
@@ -0,0 +1,227 @@
...
+    const bsl::string&                       principal() const;
+    const bsl::optional<bsls::Types::Int64>& lifetimeMs() const;
+
+    // MANIPULATORS
+    /// Reset the state of this structure.
+    void reset();
+
+    /// Set the corresponding field of this structure.
+    void setPrincipal(const bslstl::StringRef& principal);
+    void setLifetimeMs(bsls::Types::Int64 lifetimeMs);
```
678098, line 111: Note that with this interface we can only set empty `lifetimeMs` to a non-empty value.
We don't have a way to unset `lifetimeMs` other than calling `reset` for the whole structure.
I think we have a workflow where we need to set `lifetimeMs` and unset it right after.

src/groups/mqb/mqbplug/mqbplug_authenticator.h
```diff
@@ -0,0 +1,227 @@
...
+    /// Return the IP Address of the client.
+    virtual bslstl::StringRef clientIpAddress() const = 0;
+};
+
+// ==========================
+// class AuthenticationResult
+// ==========================
+
+/// The results of authentication.
+class AuthenticationResult BSLS_KEYWORD_FINAL {
```
678098, line 78: For this struct, I don't see why it should not be universal for all plugin types. If this is the case, we can keep one final implementation for it here.

src/groups/mqb/mqbplug/mqbplug_plugintype.h
```diff
@@ -42,7 +42,7 @@ namespace mqbplug {
 /// authentication, ...)
 struct PluginType {
     // TYPES
-    enum Enum { e_STATS_CONSUMER };
+    enum Enum { e_STATS_CONSUMER = 0, e_AUTHENTICATOR = 1 };
```
678098, line 45: I would like to fix the first enum `e_STATS_CONSUMER` by removing S from the name: `e_STAT_CONSUMER`. We have `mqbblp_statconsumer.*` sources without S in name

emelialei88 (approved):

emelialei88 (approved):
Might still need some tweaks but looks good for now.

PR #674, opened 2025-03-31
Fix: Using genCount for lastPush

dorjesinpo:
Using generation counter as the means to reset `PushStream::App::setLastPush` after each PUSH.


pniedzielski (approved):
The logic looks similar to as discussed.  I can't see any other issues.  Let's merge and test further?

PR #675, opened 2025-04-01
Fix; not switching thread context when closing on shutdown

dorjesinpo:
When close queue is the result of broker shutdown, it gets response immediately.  There is no need to switch thread context (to Cluster thread).

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.h
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.h
@@ -462,9 +462,9 @@ class ClusterQueueHelper BSLS_KEYWORD_FINAL
 
     StopContexts d_stopContexts;
 
-    /// When `true`, all cluster nodes support StopRequest V2 and this node
-    /// executes shutdown V2 logic.
-    bool d_supportShutdownV2;
+    /// When `true`, this node is shutting down using shutdown v2 logic.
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.h
```diff
@@ -464,7 +464,7 @@ class ClusterQueueHelper BSLS_KEYWORD_FINAL
 
     /// When `true`, all cluster nodes support StopRequest V2 and this node
     /// executes shutdown V2 logic.
-    bool d_supportShutdownV2;
+    bsls::AtomicBool d_supportShutdownV2;
```
pniedzielski, line 467: It seems like now this variable isn't whether this node should execute shutdown v2 logic, but rather whether the node is currently shutting down using shutdown v2 logic.

```
/// When `true`, this node is shutting down using shutdown v2 logic.  This can only be true when all cluster nodes support StopRequest V2.
bsls::AtomicBool d_v2ShuttingDown;
```
?

pniedzielski (approved):
Nit about comment.  Apply if you want, it's very minor.

Regardless this should help with the shutdown crash we saw.  Seems like shutdown crashes are a particular downside of our actor model...

pniedzielski (approved):

PR #678, opened 2025-04-02
Feat[MQBSTAT]: Add queue utilization metrics

alexander-e1off:
Add two new queue metrics representing queue utilization (in precents):
 - queue messages utilization: `queue.content_msgs / queue.cfg_msgs * 100%`;
 - queue bytes utilization: `queue.content_bytes/ queue.cfg_bytes * 100%`;


src/plugins/bmqprometheus/bmqprometheus_prometheusstatconsumer.cpp
```diff
@@ -360,7 +360,13 @@ void PrometheusStatConsumer::captureQueueStats()
                     {"queue_cfg_msgs", Stat::e_CFG_MSGS, false},
                     {"queue_cfg_bytes", Stat::e_CFG_BYTES, false},
                     {"queue_content_msgs_max", Stat::e_MESSAGES_MAX, false},
+                    {"queue_msgs_utilization",
+                     Stat::e_MESSAGES_UTILIZATION,
+                     false},
                     {"queue_content_bytes_max", Stat::e_BYTES_MAX, false},
+                    {"queue_butes_utilization",
```
chrisbeard, line 367:
```suggestion
                    {"queue_bytes_utilization",
```

src/groups/mqb/mqbstat/mqbstat_queuestats.cpp
```diff
@@ -100,6 +100,39 @@ bool filterDirect(const bmqst::TableRecords::Record& record)
...
+        avg = bmqst::StatUtil::average(
+            context.value(bmqst::StatContext::e_DIRECT_VALUE, currentValue),
+            latestSnapshot,
+            oldestSnapshot);
+    }
+    else {
+        avg = static_cast<double>(bmqst::StatUtil::value(
+            context.value(bmqst::StatContext::e_DIRECT_VALUE, currentValue),
+            latestSnapshot));
+    }
```
chrisbeard, line 122: What is the exact semantic of the calculated value? Is it "current utilization", "max utilization", etc?

alexander-e1off: I consider that utilization is calculated as average value of current messages/bytes within publish period divided by limit value. Snapshot interval is usually 1 sec, publish interval is usually 10 sec. So, the average value of messages/bytes for the last 10 snapshots is calculated. Added comments into the code describing this logic.

chrisbeard: I think we should keep the semantics aligned with the existing history metrics, which are either current or max. Given the choice between the two, I think max is the more useful option. Domains already provide the ability to alarm (internally in the broker) if capacity crosses a threshold (e.g. 80% full), so exposing that max would be the the closest stat to the stat triggering that alarm.

alexander-e1off: Agree, refactored to use max value for calculation of queue utilization.

src/groups/mqb/mqbstat/mqbstat_queuestats.h
```diff
@@ -110,8 +110,10 @@ class QueueStatsDomain {
             e_NB_CONSUMER,
             e_MESSAGES_CURRENT,
             e_MESSAGES_MAX,
+            e_MESSAGES_UTILIZATION,
```
chrisbeard, line 113: What do you think about ending with "_MAX" to make it clear it's peak utilization? Otherwise, lgtm.

alexander-e1off: Agree, added suffix `_MAX` to the new metrics.

chrisbeard (approved):
LGTM

PR #681, opened 2025-04-03
Fix: Remove unused parameter from `RelayQueueEngine`

pniedzielski:
During the fixes relating to monolithic storage, we left an unused parameter in `RelayQueueEngine::storePush` originally denoting whether the push was out-of-order or not.  This is no longer needed, so this patch removes the parameter.

*Issue number of the reported bug or feature request: #<number>*

**Describe your changes**
A clear and concise description of the changes you have made.

**Testing performed**

**Additional context**


pniedzielski:
@dorjesinpo this is low priority, but I don't want it to get lost

dorjesinpo (commented):
This looks good.  My only concern, there is some possibility of using `isOutOfOrder` in the last step(s) of Reliable Broadcast.  So, we have a choice of removing this and then we need to remove the corresponding `if` clause in `RelayQueueEngine`, or keeping this for a while

PR #682, opened 2025-04-04
Refactor: Move NegotiatorContext/NegotiationContext into their own components

emelialei88:
Take out NegotiationContext and NegotiatorContext into their own files

### DIFF
--- a/src/groups/mqb/mqba/mqba_negotiationcontext.h
+++ b/src/groups/mqb/mqba/mqba_negotiationcontext.h
@@ -1,0 +1,80 @@
+// Copyright 2025 Bloomberg Finance L.P.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


src/groups/mqb/mqba/mqba_negotiationcontext.h
```diff
@@ -0,0 +1,82 @@
...
+
+    /// True if this is a "reversed" connection (on either side of the
+    /// connection).
+    bool d_isReversed;
+
+    /// The type of the session being negotiated.
+    ConnectionType::Enum d_connectionType;
+};
+
+typedef bsl::shared_ptr<NegotiationContext> NegotiationContextSp;
```
pniedzielski, line 77: Maybe we don't want to pull this `typedef` out.  These are convenience `typedef`s, usually scoped to a `class`, and this looks like it's only used by `mqba_sessionnegotiator`.

pniedzielski (changes requested):
Small nit, sorry.  Otherwise looks good, we can rename and extend this type for authn use too.

pniedzielski (approved):

PR #684, opened 2025-04-07
fix: inlining closeQueue on shutdown

dorjesinpo:
[https://github.com/bloomberg/blazingmq/pull/675](https://github.com/bloomberg/blazingmq/pull/675) did not make closeQueue completely inline (called `RelayQueueEngine::onHandleReleased`)


pniedzielski (approved):
👍🏻

Issue #686, opened 2025-04-08
Automatically generate releases and release candidates when a tag is pushed


pniedzielski:
We're settling on `vXX.YY.ZZ` for the tag format of releases, and `vXX.YY.ZZ-rcWW` for the tag format of release candidates.  When a tag of one of these forms is pushed, we should automatically create either a draft GitHub prerelease or a release, with generated release notes.

PR #687, opened 2025-04-08
CI: Automatically create GitHub releases on tag push

pniedzielski:
*Issue number of the reported bug or feature request: #686*

**Describe your changes**
This PR extends our CI to automatically create a GitHub release or prerelease when tags of the correct form are generated.  Please see commit messages for details.

**Testing performed**
By its nature this is hard to test, but if it fails, we are only left with an incorrect release, which we can easily delete or manually create.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -6,8 +6,6 @@ on:
     tags:
       - v[0-9]+.[0-9]+.[0-9]+           # Releases
       - v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+  # Release Candidates
-    # Allow us to make manual images from `main`, tagged as `main`.
-    workflow_dispatch:
 
 env:
   REGISTRY: ghcr.io
@@ -52,7 +50,6 @@ jobs:
 
   build-and-publish-image:
     runs-on: ubuntu-latest
-    if: github.event_name == 'push'
     permissions:
       contents: read
       packages: write
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,10 +1,11 @@
-name: Publish
+name: Publish release
 
 on:
   push:
     # Publish `v*` tags as releases.
     tags:
-      - v*
+      - v[0-9]+.[0-9]+.[0-9]+           # Releases
+      - v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+  # Release Candidates
     # Allow us to make manual images from `main`, tagged as `main`.
     workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


pniedzielski:
@kaikulimu FYI, this PR works with tags of the form `vX.Y.Z` and `vX.Y.Z-rcW`.  So `BMQBRKR_X.Y.Z` tags won't trigger it.

I pushed tags for each release in the 0.93 series with the correct tag forms.  We can start using prereleases perhaps with the 0.94 series.

.github/workflows/publish.yaml
```diff
@@ -1,21 +1,54 @@
...
-    workflow_dispatch:
 
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
 
 jobs:
+  create-github-prerelease:
+    runs-on: ubuntu-latest
+    if: github.event_name == ’push’ && contains(github.ref_name, “rc”)
```
kaikulimu, line 16: Check `vXX.YY.ZZ-rcWW` is safer.

.github/workflows/publish.yaml
```diff
@@ -1,21 +1,54 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          release_name: Release Candidate ${{ github.ref_name }}
+          draft: true           # So we can write pre-release notes
+          prelease: true
+          generate_release_notes: true
+
+  create-github-release:
+    runs-on: ubuntu-latest
+    if: github.event_name == ’push’ && !contains(github.ref_name, “rc”)
```
kaikulimu, line 34: Check `vXX.YY.ZZ` is safer.

.github/workflows/publish.yaml
```diff
@@ -1,21 +1,54 @@
...
+  create-github-prerelease:
+    runs-on: ubuntu-latest
+    if: github.event_name == ’push’ && contains(github.ref_name, “rc”)
+    permissions:
+      contents: write
+
+    steps:
+      - name: Create release candidate
+        id: create_release_candidate
+        uses: actions/create-release@v1
```
kaikulimu, line 23: Use `softprops/action-gh-release@v2.2.1`

kaikulimu (changes requested):

.github/workflows/publish.yaml
pniedzielski: I've limited the workflow to run on tags of that shape in the most recent push.

```yaml
# Publish `v*` tags as releases.
tags:
  - v[0-9]+.[0-9]+.[0-9]+           # Releases
  - v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+  # Release Candidates
```

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet describes this syntax, but it should match the two checks you asked for.

.github/workflows/publish.yaml
pniedzielski: See above

kaikulimu (approved):

PR #691, opened 2025-04-08
Fix Broadcast advisory before apply to CSL; protect from older versions

kaikulimu:
Today, we witness alarm if we mix a 0.92.x leader with a 0.93.x follower. 0.92.x brokers has eventual consistency CSL and applies QueueAssignmentAdvisory/QueueUnAssignmentAdvisory along with its commit to CSL before broadcasting the corresponding legacy advisory. The commit will bump up the leader message sequence in a 0.93.x receiving broker and make it alarm when the corresponding legacy advisory arrives. The alarm will look something like:

```
08APR2025_20:18:46.333 (6109655040) ERROR mqbblp_clusterstatemanager.cpp:1575 ALARM [CLUSTER_STATE] Cluster (c2x2): got queueAssignmentAdvisory: [ sequenceNumber = [ electorTerm = 1 sequenceNumber = 11 ] queues = [ [ uri = "bmq://bmq.test.mmap.fanout/q1" key = [ D39618A6DD ] partitionId = 0 appIds = [ ] ] ] ] from current leader: [east2, 2], with smaller leader message sequence: [ electorTerm = 1 sequenceNumber = 11 ]. Current value: [ electorTerm = 1 sequenceNumber = 12 ]. Ignoring this advisory.
```

The fix in this PR has two parts:
1. Always broadcast legacy advisory before applying to CSL
2. Protect against out-of-sequence legacy advisory from 0.92.x broker, by allowing the advisory to possibly be exactly one sequence number behind.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
@@ -6348,13 +6348,19 @@ int ClusterQueueHelper::gcExpiredQueues(bool               immediate,
                                                            pid,
                                                            *d_clusterState_p);
 
-        // Apply 'queueUnassignedAdvisory' to CSL
-        d_clusterStateManager_p->unassignQueue(queueAdvisory);
-
         if (!d_cluster_p->isCSLModeEnabled()) {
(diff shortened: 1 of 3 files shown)


dorjesinpo (approved):
thank you

PR #692, opened 2025-04-08
IT: Connection loss test

waldgange:
Add integration test to detect connection loss between brokers or clients

### DIFF
--- a/src/integration-tests/test_connection_loss.py
+++ b/src/integration-tests/test_connection_loss.py
@@ -1,0 +1,202 @@
+# Copyright 2024 Bloomberg Finance L.P.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+This suite of test cases exercises connection losses.
+"""
+
+import json
+import re
+from typing import Dict
+
+from blazingmq.dev.it.fixtures import (  # pylint: disable=unused-import
+    Cluster,
+    order,
+    cluster,
+    multi_node,
+    tweak,
+    start_cluster,
+)
+from blazingmq.dev.it.process.admin import AdminClient
+from blazingmq.dev.it.process.client import Client
+from blazingmq.dev.it.process.proc import Process
+
+pytestmark = order(1)
+
+
+def test_broker_client(
+    cluster: Cluster,  # pylint: disable=redefined-outer-name
+) -> None:
+    """
+    Test: connection loss between a broker and a client.
+    - Start a broker and save the port it is listening.
+    - Start a tproxy redirecting to the broker started on the previous step.
+    - Start a client and connect it to the tproxy started on the previous step.
+    - Kill the tproxy to brake the connection between the client and the broker.
+
+    Concerns:
+    - The client is able to detect the connection loss.
+    - The connection is restored after tproxy restart.
+    """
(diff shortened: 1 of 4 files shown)


src/integration-tests/test_connection_loss.py
```diff
@@ -0,0 +1,202 @@
...
+    cluster,
+    multi_node,
+    tweak,
+    start_cluster,
+)
+from blazingmq.dev.it.process.admin import AdminClient
+from blazingmq.dev.it.process.client import Client
+from blazingmq.dev.it.process.proc import Process
+
+pytestmark = order(1)
```
kaikulimu, line 36:
```suggestion
pytestmark = order(2)
```

The `order` constitutes a logical dependency of tests, where only a higher level test can depend on the success of a lower level test. Now, `test_breathing.py` has order(1) and `test_restart.py` has order(2). I believe this test should be at least order(2).

src/integration-tests/test_connection_loss.py
```diff
@@ -0,0 +1,202 @@
...
+    try:
+        tproxy.start()
+        tproxy_port = tproxy.capture(r"Listening on (.+):(\d+)", timeout=2).group(2)
+
+        # Start a client
+        client: Client = proxy.create_client(f"client@{proxy.name}", port=tproxy_port)
+        client.start_session()
+        assert client.capture(r"CONNECTED", 5)
+
+        # Kill tproxy to brake the connection between broker and client
```
kaikulimu, line 70:
```suggestion
        # Kill tproxy to break the connection between broker and client
```

src/integration-tests/test_connection_loss.py
```diff
@@ -0,0 +1,202 @@
...
+                            tproxy.start()
+                            tproxy_port = tproxy.capture(
+                                r"Listening on (.+):(\d+)", timeout=2
+                            ).group(2)
+                            tproxies[node["name"]] = tproxy
+                            node["transport"]["tcp"]["endpoint"] = node["transport"][
+                                "tcp"
+                            ]["endpoint"].replace(str(broker_config.port), tproxy_port)
+                else:
+                    data["myClusters"][0]["elector"]["quorum"] = 4
```
kaikulimu, line 161: Could you have done something like
```
@tweak.cluster.elector.quorum(4)
```
at the top of the test?

src/integration-tests/test_connection_loss.py
```diff
@@ -0,0 +1,202 @@
...
+    - The connection loss leads to leader/primary divergence: Leader is "west2", but primary for all
+     partitions is "east1"
+    """
+
+    cluster = multi_node  # pylint: disable=redefined-outer-name
+    tproxies: Dict[str, Process] = {}
+
+    try:
+        for name in cluster.config.nodes:
+            with open(
```
kaikulimu, line 127: Can you refactor this tproxy interception logic into a reusable function? We can put this function in `cluster.py` or some other file that fits. We might add more scenarios in the future which uses tproxy.

kaikulimu (changes requested):
Looks nice. Some comments

src/integration-tests/test_connection_loss.py
waldgange: This will set `quorum = 4` in all 4 `clusters.json` files for each node. I need to set it only in 3 of them.
I intended to use `tweak` for that but it doesn't allow different setups for different config files. Also it doesn't support access by index to elements of a list. It makes it impossible to set different values for different nodes in a `clusters.json` file. 
Although it requires intervention to the `tweak` module, I agree that it would be very useful to have such option not only for this test but also for other scenarios. Maybe it's worth adding new `tweak` features in a separate PR? What do you think?

waldgange: Though I can get rid of the for loop with `@tweak.cluster.elector.quorum(4)`. So I only need to edit one file and set quorum value back to 0

waldgange:
@kaikulimu I've addressed the review notes. Please take a look

src/integration-tests/test_connection_loss.py
kaikulimu: @jll63 Can you advise on whether a new `tweak` feature would make sense?

kaikulimu (approved):

PR #693, opened 2025-04-09
CI: Can preset leader via env var BLAZINGMQ_LEADER_NAME for integration tests

bbpetukhov:
**Describe your changes**
Added a new env var BLAZINGMQ_LEADER_NAME which is respected in Integration tests runner. It allows to preset leader node in the cluster fixture.

The preset of leader is done via altering quorum for nodes. If the evn var BLAZINGMQ_LEADER_NAME is specified, for the selected leader quorum is set to 1, for all other nodes -- 100. This approach allows the specified node to win in leader election.

**Testing performed**
- Created a new test which simply starts the cluster
- Tested setting BLAZINGMQ_LEADER_NAME to all 4 nodes (i.e. ['east1', 'east2', 'west1', 'west2']) and making sure that the specified node is selected as leader.

### DIFF
--- a/src/python/blazingmq/dev/it/cluster.py
+++ b/src/python/blazingmq/dev/it/cluster.py
@@ -38,6 +38,10 @@ logger = logging.getLogger(__name__)
 
 CORE_PATTERN_PATH = "/proc/sys/kernel/core_pattern"
 
+QUORUM_DEFAULT = 0
+QUORUM_TO_ENSURE_LEADER = 1
+QUORUM_TO_ENSURE_NOT_LEADER = 100
+
 
 def _match_broker(broker, **kw):
     datacenter = kw.get("datacenter", None)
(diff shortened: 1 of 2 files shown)


src/python/blazingmq/dev/it/cluster.py
```diff
@@ -731,9 +751,13 @@ def _find_cores_dir(self) -> Optional[Path]:
 
         return None
 
-    def _start_broker(self, broker: cfg.Broker, array, cluster_name):
+    def _start_broker(
+        self, broker: cfg.Broker, nodes: List[Broker], cluster_name: Union[str, None]
```
jll63, line 755: I called it `array, not `nodes`, because it can be `self._nodes`, `self._virtual_nodes` or `self._proxies`. `brokers` would be a better name.

jll63 (changes requested):

src/python/blazingmq/dev/it/cluster.py
bbpetukhov: Renamed to "brokers"

jll63 (approved):

PR #696, opened 2025-04-09
WIP Feat[MQB]: add authentication with basic logic

emelialei88:
Broker side:
- Added support for the authentication protocol.
- Introduced a new `AuthenticationEvent` event type.
- Implemented the authenticator component to handle authentication logic, using `AuthenticationContext` to encapsulate relevant context data.
- Updated `InitialConnectionHandler` to handle the authentication flow.

Client side:
- Renamed NegotiationChannelFactory to InitialConnectionChannelFactory, which now handles both authentication and negotiation.
- Updated `bmqimp_application` to pass in an `AuthenticationMessage`.


src/groups/bmq/bmqp/bmqp_ctrlmsg.xsd
```diff
@@ -1609,12 +1609,50 @@
       </documentation>
     </annotation>
     <choice>
+      <element name='authenticateRequest'      type='tns:AuthenticateRequest'/>
+      <element name='authenticateResponse'     type='tns:AuthenticateResponse'/>
```
chrisbeard, line 1613: From our discussion today, we're leaning toward moving these types to a new, top-level complexType that just contains the authentication protocols.

src/groups/bmq/bmqimp/bmqimp_application.cpp
```diff
@@ -597,12 +599,15 @@ Application::Application(
...
-      NegotiatedChannelFactoryConfig(&d_statChannelFactory,
-                                     negotiationMessage,
-                                     sessionOptions.connectTimeout(),
-                                     d_blobSpPool_sp.get(),
-                                     allocator),
+, d_initialConnectionChannelFactory(
+      InitialConnectionChannelFactoryConfig(
+          &d_statChannelFactory,
+          authenticationMessage,
+          sessionOptions.connectTimeout(),  // TODO: different for authn?
```
emelialei88, line 606: Should we have a different timeout for authentication?

src/groups/bmq/bmqimp/README.md
```diff
@@ -13,13 +13,13 @@ its components directly, but rather use the accessor public APIs from 'bmqa'.
...
+Component                                | Provides ...
+-----------------------------------------|-----------------------------------------------------------
+`bmqimp_application`                     | the top level object to manipulate a session with bmqbrkr.
+`bmqimp_brokersession`                   | the implementation of a session with the bmqbrkr.
+`bmqimp_event`                           | a value-semantic type representing an event.
+`bmqimp_eventqueue`                      | a thread safe queue of pooled bmqimp::Event items.
+`bmqpimp_eventsstats`                    | a mechanism to keep track of Events statistics.
+`bmqpimp_queue`                          | a type object to represent information about a queue.
+`bmqimp_stat`                            | utilities for stat manipulation.
+`bmqimp_initialconnectionchannelfactory` | a channel factory that authenticates a peer and and negotiates with it.
```
678098, line 25:
```suggestion
`bmqimp_initialconnectionchannelfactory` | a channel factory that authenticates a peer and negotiates with it.
```

src/groups/bmq/bmqa/bmqa_session.cpp
```diff
@@ -246,6 +246,14 @@ int SessionUtil::createApplication(SessionImpl* sessionImpl)
 
     static bsls::AtomicInt s_sessionInstanceCount(0);
 
+    // Authentication Message
+    bmqp_ctrlmsg::AuthenticationMessage authenticationMessage;
+    bmqp_ctrlmsg::AuthenticateRequest&  ar =
+        authenticationMessage.makeAuthenticateRequest();
+    bsl::string str = "username:password";
+    ar.mechanism()  = "basic";
+    ar.data()       = bsl::vector<char>(str.begin(), str.end());  // hexBinary
```
678098, line 255: A few notes:

1. Binary hex introduces `+100%` protocol overhead to `data.size()` after encoding.
If we use `base64`, we can have `+33%` protocol overhead.
Probably it's not a problem, because we don't intend to send authn requests too often and don't want to send too much data here.

2. Should binary hex be enforced? If so, it's worth to add checks in both client library (before sending a request) and on the broker side (when request is received).
This also means that `"username:password"` data provided here will be rejected.

3. Another way to enforce binary hex is to hide it from the library user.
Let the user to just assign `bsl::vector<char> data` to anything, and convert it to binary hex when message is packed.

4. What variations of capital letters does this hex binary support?
`abcdef`, `ABCDEF`, or both `abcdefABCDEF`?

src/groups/bmq/bmqimp/bmqimp_application.h
```diff
@@ -226,7 +226,8 @@ class Application {
         const bsl::shared_ptr<bmqp::HeartbeatMonitor>& monitor);
     bsl::shared_ptr<bmqp::HeartbeatMonitor>
     createMonitor(const bsl::shared_ptr<bmqio::Channel>& channel);
-    void startHeartbeat(const bsl::shared_ptr<bmqio::Channel>&         channel,
+    void
+         startHeartbeat(const bsl::shared_ptr<bmqio::Channel>&         channel,
```
678098, line 230: This is a formatter bug @dorjesinpo encountered

src/groups/bmq/bmqimp/bmqimp_initialconnectionchannelfactory.cpp
```diff
@@ -0,0 +1,591 @@
...
+#include <bslma_default.h>
+#include <bsls_annotation.h>
+#include <bsls_assert.h>
+
+namespace BloombergLP {
+namespace bmqimp {
+
+namespace {
+
+BALL_LOG_SET_NAMESPACE_CATEGORY("BMQIMP.NEGOTIATEDCHANNELFACTORY");
```
678098, line 46:
```suggestion
BALL_LOG_SET_NAMESPACE_CATEGORY("BMQIMP.INITIALCONNECTIONCHANNELFACTORY");
```

src/groups/mqb/mqba/mqba_initialconnectionhandler.h
```diff
@@ -39,6 +40,7 @@ namespace mqba {
 
 // FORWARD DECLARATION
 class SessionNegotiator;
+class Authenticator;
```
678098, line 43:
```suggestion
```

This forward declaration is not correct (`mqbnet` namespace needed) and not used (`#include <mqbnet_authenticator.h>` present)

678098 (commented):
Continue looking, but here are some comments

PR #702, opened 2025-04-16
Fix: Support setting thread names on more platforms than Linux

pniedzielski:
This series of patches allows us to set thread names on all platforms that BDE's `bslmt::ThreadUtil` supports, which currently is Linux, Solaris, Darwin, and Windows.


chrisbeard (approved):
Looks good, nice cleanup :+1:

PR #704, opened 2025-04-17
Fix: remove assertion for proxy receive an unexpected event

emelialei88:


chrisbeard:
Let's remove the assert in the case above, too?

If we really wanted to, we could refactor this switch a bit to simplify all of the cases for unexpected events.

dorjesinpo (changes requested):
I don't think we need to change any `onDispatcherEvent` as it is our internal logic.
We do need to change `processEvent` everywhere to make sure we do not assert on unknown events (seems like only `ClusterProxy` does it)

dorjesinpo (approved):
thank you

PR #705, opened 2025-04-17
Doc[MQB]: fix doxygen comments for clusterstats enums

chrisbeard:
**Describe your changes**
While reviewing https://github.com/bloomberg/blazingmq/pull/678 and looking at a similar approach for partitions, I noticed some docs needed fixing in this package.

### DIFF
--- a/src/groups/mqb/mqbstat/mqbstat_clusterstats.h
+++ b/src/groups/mqb/mqbstat/mqbstat_clusterstats.h
@@ -64,8 +64,8 @@ class ClusterStats {
     struct PartitionEventType {
         // TYPES
         enum Enum {
+            /// Time in nanoseconds it took for the rollover operation.
             e_PARTITION_ROLLOVER
-            // Time in nanoseconds it took for the rollover operation.
         };
     };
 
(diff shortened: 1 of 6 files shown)


src/groups/mqb/mqbstat/mqbstat_clusterstats.h
```diff
@@ -64,8 +64,8 @@ class ClusterStats {
     struct PartitionEventType {
         // TYPES
         enum Enum {
+            /// Time in nanoseconds it took for the rollover operation.
```
pniedzielski, line 67: I find this, in combination with the `struct`'s comment above, baffling.  

src/groups/mqb/mqbs/mqbs_filestore.cpp:2914 looks like it's the only use, which makes it a little more clear what's going on, but yikes

src/groups/mqb/mqbstat/mqbstat_clusterstats.h
```diff
@@ -306,23 +304,17 @@ class ClusterNodeStats {
     struct ClusterNodeStatsIndex {
         enum Enum {
             /// Value:      Number of ack messages delivered to the client
-            e_STAT_ACK
-
-            ,
-            e_STAT_CONFIRM
+            e_STAT_ACK,
             // Value:      Number of confirm messages delivered to the client
```
pniedzielski, line 308: Should these be `///` too?

src/groups/mqb/mqbstat/mqbstat_printer.h
```diff
@@ -67,14 +67,14 @@ class Printer {
...
     struct Context {
+        /// Stat Context pointer
         bmqst::StatContext* d_statContext_p;
-        // Stat Context pointer
 
+        /// Table
         bmqst::Table d_table;
-        // Table
 
+        /// tip
```
pniedzielski, line 76: 😮‍💨

pniedzielski (changes requested):
I think you missed `///` in one place.  Other comments don't need to be addressed, but at least want to keep a record of things we need to revisit.

pniedzielski (approved):

PR #708, opened 2025-04-22
fixing CS shutdown V2

dorjesinpo:
Bugfix: `ClientSession` calls method on destructed `Latch` in ShutdownV2

### DIFF
--- a/src/groups/mqb/mqba/mqba_clientsession.cpp
+++ b/src/groups/mqb/mqba/mqba_clientsession.cpp
@@ -911,44 +911,38 @@ void ClientSession::initiateShutdownDispatched(
     BALL_LOG_INFO << description()
                   << ": initiateShutdownDispatched. Timeout: " << timeout;
 
+    bdlb::ScopeExitAny scopeExit(callback);
+
     if (d_operationState == e_DEAD) {
         // The client is disconnected.  No need to wait for tearDown.
-        callback();
         return;  // RETURN
     }
 
     if (d_operationState == e_SHUTTING_DOWN) {
         // More than one cluster calls 'initiateShutdown'?
-        callback();
-        return;  // RETURN
-    }
-
-    if (d_operationState == e_SHUTTING_DOWN_V2) {
-        // More than one cluster calls 'initiateShutdown'?
-        callback();
         return;  // RETURN
     }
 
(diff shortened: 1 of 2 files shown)


pniedzielski (approved):
Can you confirm my understanding below?

Relevant latch lives on the stack:
https://github.com/bloomberg/blazingmq/blob/f8e22d6a8e7e7552e37dd554e828dfd7d6a342ad/src/groups/mqb/mqba/mqba_application.cpp#L603

Our callback calls a method on the latch: https://github.com/bloomberg/blazingmq/blob/f8e22d6a8e7e7552e37dd554e828dfd7d6a342ad/src/groups/mqb/mqba/mqba_application.cpp#L620

Before this change, we always set `d_shutdownCallback` to this `callback`, with both shutdown v1 and shutdown v2.  Then shutdown v2 explicitly calls the callback.  Then, in shutdown v2 `tearDownAllQueuesDone` calls `callback` again, long after the latch has been destructed.

After this change, shutdown v2 does not set `d_shutdownCallback`.  It cals the `callback` once by virtue of the new scoped exit object.

If my understanding is correct, let's ship it.

dorjesinpo:
> Can you confirm my understanding below?
> 
> Relevant latch lives on the stack:
> 
> https://github.com/bloomberg/blazingmq/blob/f8e22d6a8e7e7552e37dd554e828dfd7d6a342ad/src/groups/mqb/mqba/mqba_application.cpp#L603
> 
> Our callback calls a method on the latch:
> 
> https://github.com/bloomberg/blazingmq/blob/f8e22d6a8e7e7552e37dd554e828dfd7d6a342ad/src/groups/mqb/mqba/mqba_application.cpp#L620
> 
> Before this change, we always set `d_shutdownCallback` to this `callback`, with both shutdown v1 and shutdown v2. Then shutdown v2 explicitly calls the callback. Then, in shutdown v2 `tearDownAllQueuesDone` calls `callback` again, long after the latch has been destructed.
> 
> After this change, shutdown v2 does not set `d_shutdownCallback`. It cals the `callback` once by virtue of the new scoped exit object.
> 
> If my understanding is correct, let's ship it.

Your understanding is correct.  In V2, there is no need to cache the callback, since the flow is simplified.  We got the error when blending V1 and V2.

PR #713, opened 2025-04-23
CI requirements.txt: Black to Ruff

kaikulimu:


pniedzielski (changes requested):
In this same commit, we need to change:

  1. `python_formatting_check` in `.github/workflows/formatting-check`;
  2. calls to `black` in `src/python/bin/schemagen`; and
  3. calls to `black` in `src/python/bin/tweakgen`.

kaikulimu:
@pniedzielski Addressed feedback, and fixed the new Ruff formatting errors. Back to you

pniedzielski (approved):
🐶

.github/workflows/formatting-check.yaml
```diff
@@ -36,14 +36,9 @@ jobs:
...
-          # TODO black no longer runs on 3.12.5. 
-          # Remove once the python version is >= 3.12.6
-          # See: https://github.com/psf/black/pull/4447
-          python-version: '3.12.4'
-      - name: black style check
+      - name: ruff style check
         run: |
-          pip3 install git+https://github.com/psf/black
-          black .
+          pip3 install ruff
```
678098, line 41: > The Ruff formatter is an extremely fast Python formatter, written in Rust. It’s over 30x faster than [Black](https://github.com/psf/black) and 100x faster than [YAPF](https://github.com/google/yapf), formatting large-scale Python projects in milliseconds — all while achieving >99.9% Black compatibility.

Nice!

src/python/requirements.txt
```diff
@@ -9,5 +8,6 @@ pytest-repeat
 pytest-rerunfailures
 pytest-xdist
 pyyaml
+ruff
```
hallfox, line 11: This should be in a requirements-dev.txt

kaikulimu: done

pniedzielski: 👍🏻 I may want to do this in the integration-test PR as well.

pniedzielski (approved):

PR #715, opened 2025-04-23
Fix: update BALL enums globally

pniedzielski:
The changes in commit bb9886c1aef0994f692836814ccb1e19b6be42b5 are meant to fix deprecation warnings from later versions of BDE than our CI builds with.  However, this commit does not apply the change globally, so we still get these warnings; furthermore it causes test failures when building against BDE versions as well.

This patch applies the change globally across the entire project.


pniedzielski:
Ideally #716 should be merged first, so we can verify the deprecation warnings this fixes in CI.

pniedzielski:
As expected, we see deprecation warnings in https://github.com/bloomberg/blazingmq/actions/runs/14630619119/job/41052385182.  This PR should fix many of those.  I'll rebase and re-run CI to verify.

chrisbeard (approved):

PR #717, opened 2025-04-23
Fix[mqbc]: Disable failing `mqbc_incoreclusterstateledger` test

pniedzielski:
This patch temporarily disables the failing test case 11 of the `mqbc_incoreclusterstateledger` component.  This has been consistently failing, and needs to be fixed, but until then let’s avoid the constant CI failure.


pniedzielski:
Issue #718 actually fixing this failure.

kaikulimu:
https://github.com/bloomberg/blazingmq/pull/688 intends to be the long-term fix, but I agree we should disable for now.

kaikulimu (approved):
LGTM

PR #719, opened 2025-04-24
fix: enable early exit in RelayQE

dorjesinpo:
Consider 10 consumers with 10 as maxUnconfirmed threshold; combined 100.  
    The delivery to individual consumer starts when number of unconfirmed 
    messages is less or equal to 8 (80% of 10); 
    The delivery to Proxy starts when the number is less or equal to 80 (80% of
    100).
    Consider uneven data consumption by consumers. 8 consumers have 9 pending 
    messages each, another one has 8 and the last one has 0.  The total is 80
    and that is enough for the Primary to send 20 messages.  Upon receipt, the
    first 8 consumers are already full, the other two get 2 and 10 messages
    respectively. And Proxy is left with 8 undelivered messages.  
    Those messages should NOT go into the linear put-aside list.

### DIFF
--- a/src/integration-tests/test_maxunconfirmed.py
+++ b/src/integration-tests/test_maxunconfirmed.py
@@ -22,6 +22,8 @@ from blazingmq.dev.it.fixtures import (  # pylint: disable=unused-import
 )
 from blazingmq.dev.it.process.client import Client
 
+from blazingmq.dev.it.util import wait_until
+
 pytestmark = order(4)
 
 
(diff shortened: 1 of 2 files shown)


src/integration-tests/test_subscriptions.py
```diff
@@ -1934,3 +1934,100 @@ def test_poison(cluster: Cluster, domain_urls: tc.DomainUrls):
     delayed = [producer.post(x=2)]  # pylint: disable=unused-variable
 
     consumer2.expect_messages(expected2, confirm=True)
+
+
+def test_stuck_downstream(multi_node: Cluster, domain_urls: tc.DomainUrls):
```
678098, line 1939: I think this test should not belong to this file, because:
1. It doesn't touch subscription feature as we configure it during negotiation and doesn't test it. The only relation to subscriptions feature is the internal representation in the broker, so we threat all the consumers with the same appId and empty subscription as the same subscription group.
2. It doesn't depend or use high-level `Consumer`/`Producer` classes introduced for this series of tests, which suggests that it's not related.

I would suggest to move it to another file, somewhere where we test routing/delivery.

src/integration-tests/test_maxunconfirmed.py
```diff
@@ -95,3 +95,100 @@ def test_maxunconfirmed(self, multi_node: Cluster, domain_urls: tc.DomainUrls):
...
+        assert (
+            producer.post(
+                uri, payload=["set#1"], wait_ack=True, block=True, succeed=True
+            )
+            == Client.e_SUCCESS
+        )
+
+    # all 10 consumers are full
+    for i in range(0, 10):
+        assert wait_until(lambda: len(consumers[i].list(uri, block=True)) == 10, 3)
```
678098, line 141: Missing import in this file:

`from blazingmq.dev.it.util import wait_until`

678098 (changes requested):

PR #726, opened 2025-04-25
CI[build-darwin.sh]: do not brew install every time

678098:
We do not call `apt-get install` in `build-ubuntu.sh` every time, and it's not necessary to call `brew install` in `build-darwin.sh`

### DIFF
--- a/bin/build-darwin.sh
+++ b/bin/build-darwin.sh
@@ -2,7 +2,7 @@
 
 # This script builds BlazingMQ and all of its dependencies.
 #
-# Required prerequisites:
+# Required tools:
 # - Clang
 # - CMake
 # - git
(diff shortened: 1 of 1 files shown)


bin/build-ubuntu.sh
```diff
@@ -1,28 +1,12 @@
...
-    pkg-config \
-    bison \
-    libfl-dev \
-    libbenchmark-dev \
-    libgmock-dev \
-    libtool \
-    libz-dev
-PREREQUISITES
+
+echo -e "Before running this script, install the following prerequisites, if not present yet," \
```
pniedzielski, line 5: I like this change!

pniedzielski (approved):
👍🏻

PR #729, opened 2025-04-30
CI: do not cache UT binaries in GH Actions

678098:
Follow up after this PR https://github.com/bloomberg/blazingmq/pull/707
Adding `all.t` to `all` makes us build UTs in workflow steps where we didn't intend to build UTs. We also cache these built UTs and it takes a lot of cache.
This PR removes usage of `all` target in workflows and changes it to specific targets.


pniedzielski (approved):
Yes, this is dramatic.  LGTM

PR #730, opened 2025-04-30
Fix: remove trailing NOTICE footers from src files

chrisbeard:
**Describe your changes**
These look like legacy copyright footers or something, they can be removed.


678098 (approved):
The IT failure is unrelated

PR #732, opened 2025-05-01
Fix[MQB]: do not call release callback on close if queue handle is unknown

678098:


src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -2127,6 +2127,10 @@ void ClusterQueueHelper::onHandleReleased(
 {
     // executed by the *QUEUE* dispatcher thread
 
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(handle);
+    BSLS_ASSERT_SAFE(requester);
```
678098, line 2132: These precondition checks are useful because they can catch an empty `handle` or `requester` in the original thread, before passing it to another thread's stack

dorjesinpo (approved):
thank you

PR #734, opened 2025-05-02
Fix[MQB]: fix queue assignment capacity late update in CSL mode

678098:
# PR Changes

- `ClusterUtil::assignQueue`: reordered preconditions checks. First, we check if the domain is built and good. After this, we check if we have a queue state built.
- Made more clear early returns if queue state exists already, and queue is in the process of assigning, assigned or being unassigned.
- Do not use `num_assigned` counter to decide if we can assign a new queue. Instead, look at the current number of allocated queue infos.

# IT Failure

We first observed a failure in integration test, we allow to assign 6 queues when we have a limit of 5 queues for this domain:

```
=========================== short test summary info ============================
FAILED test_maxqueues.py::TestMaxQueues::test_max_queue_restore[multi_node_fsm-eventual_consistency] - assert [-6, 0, 0, 0, 0, 0, ...] == [-6, -6, 0, 0, 0, 0, ...]
```

# Bug description

1. Assignment starts in `ClusterUtil::processQueueAssignmentRequest` when we have a **CLUSTER** dispatcher event. It is executed on the sole cluster thread on the current primary (if it's not a primary, the assignment request is refused with an error code).

2. There, we call `ClusterUtil::assignQueue`:

```
    assignQueue(clusterState,
                clusterData,
                ledger,
                cluster,
                uri,
                allocator,
                &status);

    clusterData->messageTransmitter().sendMessage(response, requester);
    // end
}
```

3. Within `ClusterUtil::assignQueue` we check if we have capacity to assign +1 queue to this domain:

```
    const int registeredQueues = domIt->second->numAssignedQueues();
    const int maxQueues        = domIt->second->domain()->config().maxQueues();
    if (maxQueues != 0) {
        const int requestedQueues = registeredQueues + 1;
        if (requestedQueues > maxQueues) {
            local::panic(domIt->second->domain());
        }
    // ...
```

4. Note that `ClusterState::DomainState::numAssignedQueues()` uses an internal variable `d_numAssignedQueues` that is updated only with `ClusterState::DomainState::adjustQueueCount(int by)`.

5. After verifying that we have enough capacity to open a new queue (on the previous step 4), we set the state for this queue to `k_ASSIGNING`.

And now we have difference between non-CSL and CSL:

- For non-CSL mode, we update `d_numAssignedQueues` right away in `clusterState->assignQueue`. We don't have problems in this case. If we have another assign queue request right after this one, we will have a correct counter `domIt->second->numAssignedQueues()`. 

- For CSL mode we only call `ledger->apply(queueAdvisory)` and return OK for the assignment request. After this, we finish processing this **CLUSTER** dispatcher event without updating `d_numAssignedQueues` counter. The counter for this queue will be updated in another **CLUSTER** dispatcher event when we commit queue assignment advisory. **BUG**: after the initial **CLUSTER** queue assignment event and the event that updates the counter we can have intermediate queue assignment events that will check `d_numAssignedQueues` counter, and this counter was not updated yet! It allows to assign more queues than the configured maximum for this domain.

Example timeline (**CLUSTER** thread):
```
[Configure domain max queues == 5]                ->

[Assign Queue "foo" Request]                      ->
[d_numAssignedQueues == 4 is less than maximum 5] ->
[Set state "foo": k_ASSIGNING]                    ->

[Assign Queue "bar" Request]                      ->
[d_numAssignedQueues == 4 is less than maximum 5] ->
[Set state "bar": k_ASSIGNING]                    ->

[Commit queue assignment advisory for "foo"]      ->
[Set state "foo": k_ASSIGNED]                     ->
["foo": d_numAssignedQueues++]                    ->

[Commit queue assignment advisory for "bar"]      ->
[Set state "bar": k_ASSIGNED]                     ->
["bar": d_numAssignedQueues++]                    ->

[d_numAssignedQueues == 6 now, we allowed to assign more queues than configured maximum 5]
```

# Solution

We have a container associated with a domain that holds allocated queue infos: `domIt->second->queuesInfo()`.
In fact this container holds queues that are in `k_ASSIGNED` and `k_ASSIGNING` states, so we can use this container's size as a measure to limit the number of assigning queues.

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_clusterstate.h
+++ b/src/groups/mqb/mqbc/mqbc_clusterstate.h
@@ -440,7 +440,7 @@ class ClusterState {
     typedef UriToQueueInfoMap::iterator                UriToQueueInfoMapIter;
     typedef UriToQueueInfoMap::const_iterator          UriToQueueInfoMapCIter;
 
-    struct DomainState {
+    struct DomainState BSLS_KEYWORD_FINAL {
       private:
         // DATA
         int               d_numAssignedQueues;
@@ -455,7 +455,7 @@ class ClusterState {
 
         /// Create a new DomainState using the specified `allocator` for
         /// memory allocations.
-        DomainState(bslma::Allocator* allocator);
+        explicit DomainState(bslma::Allocator* allocator);
 
         // MANIPULATORS
 
--- a/src/groups/mqb/mqbc/mqbc_clusterutil.cpp
+++ b/src/groups/mqb/mqbc/mqbc_clusterutil.cpp
@@ -883,14 +883,10 @@ ClusterUtil::assignQueue(ClusterState*         clusterState,
             k_ASSIGNMENT_WHILE_UNAVAILABLE;  // RETURN
     }
 
-    ClusterStateQueueInfo::State::Enum previousState =
-        ClusterStateQueueInfo::State::k_NONE;
-
     ClusterState::DomainStates& domainStates = clusterState->domainStates();
     DomainStatesIter domIt = domainStates.find(uri.qualifiedDomain());
 
     UriToQueueInfoMapIter queueIt;
-
     if (domIt == domainStates.end()) {
         ClusterState::DomainStateSp domainState;
         domainState.createInplace(allocator, allocator);
@@ -902,29 +898,7 @@ ClusterUtil::assignQueue(ClusterState*         clusterState,
         queueIt = domIt->second->queuesInfo().find(uri);
     }
 
-    if (queueIt == domIt->second->queuesInfo().end()) {
-        QueueInfoSp queueInfo;
-
-        queueInfo.createInplace(allocator, uri, allocator);
-
-        queueIt = domIt->second->queuesInfo().emplace(uri, queueInfo).first;
-    }
-    else {
-        previousState = queueIt->second->state();
-    }
-
-    if (previousState == ClusterStateQueueInfo::State::k_ASSIGNING) {
-        BALL_LOG_INFO << cluster->description() << "queueAssignment of '"
-                      << uri << "' is already pending.";
-        return QueueAssignmentResult::k_ASSIGNMENT_OK;
-    }
-
-    if (previousState == ClusterStateQueueInfo::State::k_ASSIGNED) {
-        BALL_LOG_INFO << cluster->description() << "queueAssignment of '"
-                      << uri << "' is already done.";
-        return QueueAssignmentResult::k_ASSIGNMENT_OK;
-    }
-
+    // There is nothing we can do if we don't have a built logical domain.
     if (domIt->second->domain() == 0) {
         BSLS_ASSERT_SAFE(clusterData->domainFactory());
         clusterData->domainFactory()->createDomain(
@@ -949,6 +923,28 @@ ClusterUtil::assignQueue(ClusterState*         clusterState,
         }
     }
 
+    ClusterStateQueueInfo::State::Enum previousState =
+        ClusterStateQueueInfo::State::k_NONE;
+    if (queueIt != domIt->second->queuesInfo().end()) {
+        // If we have a queue state in the map, we can extract this state.
+        // For k_ASSIGNED or k_ASSIGNING states we don't need to do anything
+        // here and can return early.
+        // If the state is k_UNASSIGNING, we proceed with assigning.
+        previousState = queueIt->second->state();
+
+        if (previousState == ClusterStateQueueInfo::State::k_ASSIGNING) {
+            BALL_LOG_INFO << cluster->description() << "queueAssignment of '"
+                          << uri << "' is already pending.";
+            return QueueAssignmentResult::k_ASSIGNMENT_OK;  // RETURN
+        }
+
+        if (previousState == ClusterStateQueueInfo::State::k_ASSIGNED) {
+            BALL_LOG_INFO << cluster->description() << "queueAssignment of '"
+                          << uri << "' is already done.";
+            return QueueAssignmentResult::k_ASSIGNMENT_OK;  // RETURN
+        }
+    }
+
     struct local {
         static void panic(mqbi::Domain* domain)
         {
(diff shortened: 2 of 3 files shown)


src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -996,13 +1004,21 @@ ClusterUtil::assignQueue(ClusterState*         clusterState,
...
     // Set the queue as assigning (no longer pending unassignment)
     queueIt->second->setState(ClusterStateQueueInfo::State::k_ASSIGNING);
 
     BALL_LOG_INFO << "Cluster [" << cluster->description()
-                  << "]: Transition: " << previousState << " -> "
-                  << ClusterStateQueueInfo::State::k_ASSIGNING << " for ["
-                  << uri << "].";
+                  << "]: Transition: " << ClusterStateQueueInfo::State::k_NONE
+                  << " -> " << ClusterStateQueueInfo::State::k_ASSIGNING
+                  << " for [" << uri << "].";
```
678098, line 1021: I can only see log lines with `Transition: NONE -> ASSIGNING` transitions. Our previous code did not allow to get to this log command with state other than `k_NONE`.

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -949,6 +923,36 @@ ClusterUtil::assignQueue(ClusterState*         clusterState,
...
+
+        if (state == ClusterStateQueueInfo::State::k_UNASSIGNING) {
+            BALL_LOG_WARN << cluster->description() << "queueAssignment of '"
+                          << uri
+                          << "' is failed: unassignment is in progress.";
+            return QueueAssignmentResult::k_ASSIGNMENT_REJECTED;  // RETURN
+        }
+    }
+
+    BSLS_ASSERT_SAFE(queueIt == domIt->second->queuesInfo().end());
```
678098, line 954: This precondition means that in the second half of this function we want to enqueue a new queue state in this container

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
kaikulimu: If you look at the [queue state transition table](https://github.com/bloomberg/blazingmq/blob/1d4a3d9c67dfd57cee16f8f800cec8e27a5ec873/src/groups/mqb/mqbc/mqbc_clusterstate.h#L880-L892), it is possible to go from `k_UNASSIGNING` to `k_ASSIGNING` if a queue gets assigned again while in the middle of pending unassignment.

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -949,6 +923,36 @@ ClusterUtil::assignQueue(ClusterState*         clusterState,
...
+                          << uri << "' is already done.";
+            return QueueAssignmentResult::k_ASSIGNMENT_OK;  // RETURN
+        }
+
+        if (state == ClusterStateQueueInfo::State::k_UNASSIGNING) {
+            BALL_LOG_WARN << cluster->description() << "queueAssignment of '"
+                          << uri
+                          << "' is failed: unassignment is in progress.";
+            return QueueAssignmentResult::k_ASSIGNMENT_REJECTED;  // RETURN
+        }
```
kaikulimu, line 951: Actually, if we get a new assignment while the queue is pending unassignment, we set the queue back to assigning. That is the `k_UNASSIGNING` to `k_ASSIGNING` transition in the [queue state transition table](https://github.com/bloomberg/blazingmq/blob/1d4a3d9c67dfd57cee16f8f800cec8e27a5ec873/src/groups/mqb/mqbc/mqbc_clusterstate.h#L880-L892). Therefore, no need to return here.

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
kaikulimu: Since `k_UNASSIGNING` to `k_ASSIGNING` is possible, this precondition no longer makes sense.

src/groups/mqb/mqbc/mqbc_clusterstate.h
```diff
@@ -440,7 +440,7 @@ class ClusterState {
     typedef UriToQueueInfoMap::iterator                UriToQueueInfoMapIter;
     typedef UriToQueueInfoMap::const_iterator          UriToQueueInfoMapCIter;
 
-    struct DomainState {
+    struct DomainState BSLS_KEYWORD_FINAL {
       private:
         // DATA
         int               d_numAssignedQueues;
```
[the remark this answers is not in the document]
678098: `d_numAssignedQueues` covers queues that are strictly `k_ASSIGNED`.
`queuesInfo()` contains queues in all states: `k_ASSIGNED`, `k_ASSIGNING`, `k_UNASSIGNING`.
It is indeed possible to calculate the number of `k_ASSIGNED` on demand, but it requires linear iteration over `queuesInfo()` every time.
We need to report this value in 2 cases: on admin command (rare event) and on metrics update (on every queue assign/unassign). Consider N queues being opened on a domain.
On every queue we'll need to iterate over all queues in the container and report stats. It has `O(N^2)` time complexity and it's not scalable.

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -970,30 +966,56 @@ ClusterUtil::assignQueue(ClusterState*         clusterState,
...
+
+        queueInfo.createInplace(allocator, uri, allocator);
+
+        queueIt = domIt->second->queuesInfo().emplace(uri, queueInfo).first;
+    }
+    else {
+        // Note that we already have `queueIt` and allocated resources for this
+        // queue.  No need to allocate new QueueInfo and check capacity.
+        BSLS_ASSERT_SAFE(previousState ==
+                         ClusterStateQueueInfo::State::k_UNASSIGNING);
```
678098, line 1018: Note that we will cannot have an alarm if we transit `k_UNASSIGNING` -> `k_ASSIGNING`, because we already have allocated resources and valid iterator `queueIt`

src/groups/mqb/mqbc/mqbc_clusterstate.h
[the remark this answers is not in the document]
kaikulimu: Makes sense. I agree

kaikulimu (approved):

PR #735, opened 2025-05-05
Feat[MQB]: Add partition utilization metrics

waldgange:


src/python/blazingmq/dev/it/data/data_metrics.py
```diff
@@ -380,122 +216,38 @@ def check(self, value: Any) -> bool:
     "appIds": {
         "bar": {
             "values": {
-                "queue_ack_msgs": 0,
-                "queue_ack_msgs_abs": 0,
-                "queue_ack_time_avg": 0,
-                "queue_ack_time_max": 0,
```
678098, line 386: I think we should keep these 0 values to check observed metrics against them

src/groups/mqb/mqbstat/mqbstat_clusterstats.cpp
```diff
@@ -175,6 +175,26 @@ bsls::Types::Int64 ClusterStats::getValue(const bmqst::StatContext& context,
...
     }
+    case Stat::e_PARTITION_DATA_UTILIZATION_MAX: {
+        const bsls::Types::Int64 value = STAT_RANGE(rangeMax,
+                                                    e_PARTITION_DATA_BYTES);
+        const bsls::Types::Int64 limit =
+            STAT_SINGLE(value, e_PARTITION_CFG_DATA_BYTES);
+        return (value == bsl::numeric_limits<bsls::Types::Int64>::min() ||
+                limit == 0)
+                   ? 0
+                   : (100 * value / limit);
```
678098, line 186: Can we really have `val[e_PARTITION_CFG_DATA_BYTES] == 0`?
If so, I would reorder the operations and allow early return, so we don't compute `STAT_RANGE` for `O(n)`:

```
limit = ...
if (UNLIKELY(limit == 0)) {
   UNLIKELY_HINT;
   return 0;  // RETURN
}
value = ...
return (value == ::min()) ? 0 : (100 * value / limit);
```

678098: Also, we can define a constant in this unit, the same can be done

```
namespace {
/// The default utilization value reported when we cannot
/// compute utilization.
const bsls::Types::Int64 k_UNDEFINED_UTILIZATION_VALUE = 0; 
}  // close unnamed namespace
```

src/groups/mqb/mqbstat/mqbstat_clusterstats.cpp
```diff
@@ -316,6 +336,14 @@ ClusterStats::setPartitionCfgBytes(bsls::Types::Int64 dataBytes,
     d_statContext_mp->setValue(
         ClusterStatsIndex::e_PARTITION_CFG_JOURNAL_BYTES,
         journalBytes);
+    bsl::vector<bsl::shared_ptr<bmqst::StatContext> >::const_iterator it =
+        d_partitionsStatContexts.cbegin();
```
678098, line 340: We can also report `e_CFG_CSL_BYTES` and `e_CSL_UTILIZATION_MAX` (we have one CSL file serving all partitions, so these metrics are per cluster, not per separate partitions).

src/python/blazingmq/dev/it/data/data_metrics.py
waldgange: Why? We neither report nor save the removed metrics in AppId StatContexts. We save them in the parent (Queue) context. We only report those metrics that I left intact: https://github.com/bloomberg/blazingmq/blob/a9e2eb5a383eb05f9bda112a6ba77cdd244bcbe4/src/groups/mqb/mqbstat/mqbstat_queuestats.cpp#L458-L479

src/groups/mqb/mqbstat/mqbstat_clusterstats.cpp
waldgange: 1) We can set the limit to 0. There is no config validation preventing that. Though the broker will crash before reaching the point of calculating the metric. It will try to write a record to the file and crash. So it's almost certain that `val[e_PARTITION_CFG_DATA_BYTES]` can't be 0 at this point. Otherwise it's a logic error. Do you think it's worth getting rid of this check at all? Probably replace it with BSLS_ASSERT_SAFE(limit != 0)
2) More often we will get `STAT_RANGE(rangeMax, e_PARTITION_DATA_BYTES) == -INT_MAX` if there were no values reported during the las t publish interval. So probably we can reorder the operations vice versa. calculate the value first and if it's not -INT_MAX then calculate the limit.
What do you think?

src/groups/mqb/mqbstat/mqbstat_clusterstats.cpp
waldgange: Good suggestion. But we also need to report outstanding bytes in the CSL file as well. Currently we don't report anything CSL-file related. Though it's not as straightforward as journal and data files utilization. Let's do it in a separate PR?

678098: @waldgange yes, let's do it in a separate PR

src/groups/mqb/mqbstat/mqbstat_clusterstats.cpp
678098: 1. `Do you think it's worth getting rid of this check at all? Probably replace it with BSLS_ASSERT_SAFE(limit != 0)`

Yes, we can do it. But also we need to ensure that partition file sizes metrics are recorded strictly before we report metrics. We don't want to introduce a crash here if for some reason the update hasn't reached stat context.

2. Let's reorder.

src/python/blazingmq/dev/it/data/data_metrics.py
678098: We report these metrics in JSON admin interface and validate them, even 0 values.

src/groups/mqb/mqbstat/mqbstat_queuestats.cpp
```diff
@@ -222,8 +227,13 @@ QueueStatsDomain::getValue(const bmqst::StatContext& context,
     }
     case QueueStatsDomain::Stat::e_BYTES_UTILIZATION_MAX: {
         // Calculate max queue bytes utilization, in percents.
-        return 100 * STAT_RANGE(rangeMax, DomainQueueStats::e_STAT_BYTES) /
-               STAT_SINGLE(value, DomainQueueStats::e_CFG_BYTES);
+        const bsls::Types::Int64 limit =
+            STAT_SINGLE(value, DomainQueueStats::e_CFG_BYTES);
+        return limit == 0
+                   ? 0
```
678098, line 233: There is the same magic constant `0` and `limit == 0` in this file.
Do you think it's worth to apply the same changes?

678098 (commented):
One question

src/groups/mqb/mqbstat/mqbstat_queuestats.cpp
waldgange: I've replaced the magic number by a constant. The check of `limit == 0` must stay here because for queues it's a real situation. We can reconfigure a domain setting the limit to 0 and the broker wil crash exactly in this place

678098 (approved):
Looks good

PR #736, opened 2025-05-06
Fix[MQBBLP] Refactor logic of consumption monitor

alexander-e1off:
Refactored logic of consumption monitor as follows:
 - Get rid of timer ticks (`onTimer()` methods and logic that calls them);

Question: Since there is no timer ticks anymore, should we also remove `consumptionMonitorPeriodMs` parameter from `clusters.json`?


src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -403,14 +490,228 @@ int RootQueueEngine::configure(bsl::ostream& errorDescription,
...
+
+        if (searchResult == mqbi::StorageResult::e_SUCCESS) {
+            // The message is still in the queue, trigger alarm and mark this
+            // appId as a stale.
+            logAlarm(cItApp);
+            d_consumptionMonitor.staleAppIds().insert(oldestMsgAppId);
+        }
+    }
+
+    // Try to find any other oldest message in queue (for all apps, except
```
dorjesinpo, line 641: What if `oldestMsgGuid` is gone but there is another message that is old and needs to be reported?

dorjesinpo (changes requested):
Still reviewing, some interim results.
Can we keep `QueueConsumptionMonitor` component and its UT(s)?  Having UTs is always better.
The logic, we may need to run in a discussion, let's get in touch.

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1732,33 +1732,47 @@ RootQueueEngine::logAppSubscriptionInfo(bsl::ostream&     stream,
...
     if (!headIt) {
         // No un-delivered messages, do nothing.
-        return false;  // RETURN
+        return alarmTime;  // RETURN
     }
+
+    // Get the arrival time delta of the oldest un-delivered message.
+    bsls::Types::Int64 arrivaTimelDelta = 0;
+    mqbs::StorageUtil::loadArrivalTimeDelta(&arrivaTimelDelta,
+                                            headIt->attributes());
```
dorjesinpo, line 1765: minor, this can be optimized.  If `enableLog == false`, we do not need `alarmTime` and do not need to call `attributes`.  We only need to check if there are any messages left or not.

dorjesinpo (changes requested):
looks good. there may be a possibility to simplify code a little, see inline.  Thank you

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -138,16 +106,23 @@ QueueConsumptionMonitor::QueueConsumptionMonitor(QueueState*       queueState,
...
 {
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(d_queueState_p);
+    BSLS_ASSERT_SAFE(d_scheduler_p);
     BSLS_ASSERT_SAFE(d_loggingCb);
 }
 
+QueueConsumptionMonitor::~QueueConsumptionMonitor()
+{
+    reset();
```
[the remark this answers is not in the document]
alexander-e1off: Agree, removed `reset()` and added asserts `dtor()`

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -167,6 +149,28 @@ QueueConsumptionMonitor::setMaxIdleTime(bsls::Types::Int64 value)
...
+    // If alarm event was already scheduled
+    if (d_alarmEventHandle) {
+        // If monitor is disabled (e.g. in case of reconfigure),
+        // cancel the event.
+        if (value == 0) {
+            d_scheduler_p->cancelEventAndWait(&d_alarmEventHandle);
+            d_alarmEventHandle.release();
+        }
+        else {
+            // TODO: need to reschedule the event at the new maxIdleTime from
```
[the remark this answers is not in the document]
alexander-e1off: Reworked the logic as follows:
1. Don't clean substream's info if new maxIdleTime !=0. If substream was in idle state, just continue to track it;
2. If alarm event was already scheduled, cancel it and schedule execution of `alarmEventDispatched()` which will find the earliest alarm time for the new maxIdleTime and schedule the event. As a consequence, if new maxIdleTime is less then previous one, ALARM could be logged immediately, e.g. we put message at 0 sec, and maxIdleTime = 10 sec. Then at 6th sec we set maxIdleTime = 5 sec, so, calculated alarm time = now - 6 + 5 => in the past => log ALARM.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -199,99 +203,306 @@ void QueueConsumptionMonitor::reset()
...
-            // Queue is 'alive' because at least one message was sent
-            // since the last 'timer'.
+        SubStreamInfo&     info = iter->second;
+        const bsl::string& id   = iter->first;
 
-            info.d_messageSent        = false;
-            info.d_lastKnownGoodTimer = d_currentTimer;
+        // Call alarm callback to log alarm if condition is met.
+        const bsls::TimeInterval alarmTime =
+            d_loggingCb(id, now, info.d_state == State::e_ALIVE);
```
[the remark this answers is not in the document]
alexander-e1off: Agree, reworked logic as follows:
1. Dont call `d_loggingCb` if `info.d_state == State::e_IDLE`;
2. Added assert  `info.d_idleEventHandle`;
3. Added assert `info.d_state == State::e_ALIVE` before calling `d_loggingCb`;

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
alexander-e1off: Optimized as proposed:
1. Updated `logAlarmCb()` signature to return bool `haveUndelivered` and receive pointer to store `alarmTime`;
2. `logAlarmCb` doesn't calculate `alarmTime` if `enableLog == false`;

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -63,6 +63,35 @@ static mqbconfm::Domain getDomainConfig()
...
+    ~QueueConsumptionMonitorTest() { reset(); }
+
+    // MODIFIERS
+
+    void alarmEventDispatched()
+    {
+        QueueConsumptionMonitor::alarmEventDispatched();
+    }
+
+    void idleEventDispatched(const bsl::string id)
```
678098, line 89:
```suggestion
    void idleEventDispatched(const bsl::string &appId)
```

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.h
```diff
@@ -274,18 +233,48 @@ class QueueConsumptionMonitor {
...
+    /// Alarm event dispatcher, executed in queue dispatcher thread.
+    /// It checks if there are substreams that meet alarm condition and trigger
+    /// the alarm for them.  If there are undelivered messages (among
+    /// substreams) it reschedules alarm event.
+    void alarmEventDispatched();
+
+    /// Idle event dispatcher, executed in queue dispatcher thread.
+    /// If there are no un-delivered messages the specified `id`, it calls
+    /// onTransitionToAlive(). Otherwise, it reschedules idle event.
+    void idleEventDispatched(const bsl::string id);
```
678098, line 276:
```suggestion
    void idleEventDispatched(const bsl::string &id);
```

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -63,6 +63,35 @@ static mqbconfm::Domain getDomainConfig()
...
+    : QueueConsumptionMonitor(queueState, loggingCb, allocator)
+    {
+        // NOTHING
+    }
+
+    ~QueueConsumptionMonitorTest() { reset(); }
+
+    // MODIFIERS
+
+    void alarmEventDispatched()
```
678098, line 84: All overridden methods of this class should be marked with `BSLS_KEYWORD_OVERRIDE`

678098 (commented):
Haven't checked absolutely everything, but here are some comments

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -138,16 +111,23 @@ QueueConsumptionMonitor::QueueConsumptionMonitor(QueueState*       queueState,
                                                  const LoggingCb&  loggingCb,
                                                  bslma::Allocator* allocator)
 : d_queueState_p(queueState)
-, d_maxIdleTime(0)
-, d_currentTimer(0)
+, d_scheduler_p(queueState->scheduler())
```
dorjesinpo, line 114: instead of caching `d_scheduler_p`, we can use `d_queueState_p->scheduler()` everywhere

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -158,13 +138,40 @@ QueueConsumptionMonitor::setMaxIdleTime(bsls::Types::Int64 value)
...
-         ++iter) {
-        iter->second = SubStreamInfo();
+    d_maxIdleTimeSec = value;
+
+    if (d_maxIdleTimeSec == 0) {
+        // Monitoring is disabled, cancel the idle events and reset all
+        // substreams.
+        cancelIdleEvents();
+
+        for (SubStreamInfoMapIter iter = d_subStreamInfos.begin(),
```
dorjesinpo, line 153: two iterations of `d_subStreamInfos` (the first is `cancelIdleEvents`)

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -158,13 +138,40 @@ QueueConsumptionMonitor::setMaxIdleTime(bsls::Types::Int64 value)
...
+
+    // If alarm event was already scheduled
+    if (d_alarmEventHandle) {
+        // Cancel the event and execute alarmEventDispatched() to reschedule
+        // alarm event for the new maxIdleTime.
+        d_scheduler_p->cancelEventAndWait(&d_alarmEventHandle);
+        d_alarmEventHandle.release();
+
+        if (d_maxIdleTimeSec > 0) {
+            d_queueState_p->queue()->dispatcher()->execute(
```
dorjesinpo, line 169: This is already the queue dispatcher thread.  Do we just need to call `alarmEventDispatched` directly?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -199,75 +206,57 @@ void QueueConsumptionMonitor::reset()
...
-                }
-            }
+    SubStreamInfo& info = subStreamInfo(id);
+    if (info.d_state == State::e_IDLE) {
+        // substream was in idle state
+        onTransitionToAlive(&info, id);
+        if (!d_alarmEventHandle) {
+            // Schedule alarm event to check if there are other un-delivered
+            // messages.
+            scheduleAlarmEvent(calculateAlarmTime(0, d_scheduler_p->now()));
```
dorjesinpo, line 259: does this ignore the (next) oldest message arrival time? 
also, let's say a stream is `IDLE` and there are two messages with the same arrival time. The first gets sent.  Do we transition to `ALIVE`?  We still have am old enough message 

Does this look like we need to call `d_loggingCb` or did we intend some optimization?

dorjesinpo (commented):
Question about `QueueConsumptionMonitor::onMessageSent` logic and few minor ones.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
alexander-e1off: Changed `cancelIdleEvents()` to `cancelIdleEventsAndResetStates(bool resetStates = true)`

alexander-e1off:
@dorjesinpo, @678098 thanks for your review comments. I've fixed them, please look at changes. Thank you!

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -199,100 +193,340 @@ void QueueConsumptionMonitor::reset()
...
+        return;  // RETURN
+    }
+
+    SubStreamInfo& info = subStreamInfo(appId);
+
+    if (info.d_idleEventHandle) {
+        info.d_idleEventHandle.release();
+    }
+
+    BSLS_ASSERT_SAFE(info.d_state == State::e_IDLE);
```
dorjesinpo, line 532: Is this guaranteed? 
Suppose `cancelEventAndWait(&subStreamInfo->d_idleEventHandle)` could not cancel because "the event has already been dispatched".  Then `idleEventDispatched` can get executed (late) after the state transition, for example in `onTransitionToAlive`

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
dorjesinpo: I guess, this reasoning is no longer valid. We always need `alarmTime`

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1756,8 +1756,24 @@ bool RootQueueEngine::logAlarmCb(const bsl::string& appId,
...
+    // Calculate alarm time
+    const bsls::TimeInterval alarmTime =
+        d_consumptionMonitor.calculateAlarmTime(arrivaTimelDelta, now);
+    if (alarmTime_p) {
+        *alarmTime_p = alarmTime;
+    }
+    // If the alarm time is in the future, don't log the alarm.
+    if (alarmTime > now) {
```
dorjesinpo: Minor.  This logic is a bit hard to follow.  The indirect indicator if that is comparing `alarmTime` twice - second time by the caller.
 Splitting this into two functions, one for retrieving `alarmTime`/ `haveUndelivered` and another for logging alarm (in `alarmEventDispatched`) could make it easier to follow.
 The name `d_loggingCb` does not suit the former.

dorjesinpo (changes requested):
I think, we cannot rely on state being the expected one in timers events due to possible race.

dorjesinpo:
@alexander-e1off please assign back to me when ready

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
alexander-e1off: Agree, added into `if` condition checking the state.

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
alexander-e1off: Agree, refactored `logAlarmCb` into two callbacks: `haveUndeliveredCb` and `logAlarmCb`.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
678098: Since you've added it to `if` condition, this assert is always true and can be removed.

pniedzielski:
Question about the following situation:

Let's say I have a priority queue with several consumers, equal priority.  If the head of the queue is a poison pill message, we'll keep attempting delivery of the poison pill, and we'll keep delivering later messages in the queue to other consumers as well.  Do I understand this change correctly that the stale queue alarm will fire in this situation, even though the broker is making progress delivering?

src/integration-tests/test_alarms.py
```diff
@@ -119,6 +258,174 @@ def test_alarms_subscription_mismatch(cluster: Cluster, domain_urls: tc.DomainUr
...
+        consumers[app_id] = consumer
+        consumer.open(f"{uri_fanout}?id={app_id}", flags=["read"], succeed=True)
+
+    producer.post(uri_fanout, ["msg1"], succeed=True, wait_ack=True)
+
+    for app_id in app_ids:
+        consumers[app_id].confirm(f"{uri_fanout}?id={app_id}", "*", succeed=True)
+
+    producer.post(uri_fanout, ["msg1"], succeed=True, wait_ack=True)
+    producer.post(uri_fanout, ["msg1"], succeed=True, wait_ack=True)
```
dorjesinpo: "msg3"

src/integration-tests/test_alarms.py
```diff
@@ -119,6 +258,174 @@ def test_alarms_subscription_mismatch(cluster: Cluster, domain_urls: tc.DomainUr
...
+    assert leader.alarms("QUEUE_STUCK", 1)
+    assert leader.capture(r"For appId: bar")
+    leader.drain()
+
+    # Then test no alarm while consumer 'foo' slowly confirms
+    time.sleep(1)
+    consumers["foo"].confirm(f"{uri_fanout}?id=foo", "*", succeed=True)
+
+    time.sleep(1)
+    consumers["foo"].confirm(f"{uri_fanout}?id=foo", "*", succeed=True)
```
dorjesinpo: what is the reason for this `confirm`; there seems to be no `max_unconfirmed_messages ` settings, so everything should be confirmed by the previous `confirm`?

src/integration-tests/test_alarms.py
```diff
@@ -119,6 +258,174 @@ def test_alarms_subscription_mismatch(cluster: Cluster, domain_urls: tc.DomainUr
...
+        consumer = next(proxies).create_client(app_id)
+        consumers[app_id] = consumer
+        consumer.open(f"{uri_fanout}?id={app_id}", flags=["read"], succeed=True)
+
+    producer.post(uri_fanout, ["msg1"], succeed=True, wait_ack=True)
+
+    for app_id in app_ids:
+        consumers[app_id].confirm(f"{uri_fanout}?id={app_id}", "*", succeed=True)
+
+    producer.post(uri_fanout, ["msg1"], succeed=True, wait_ack=True)
```
dorjesinpo, line 362: "msg2"

src/integration-tests/test_alarms.py
```diff
@@ -119,6 +258,174 @@ def test_alarms_subscription_mismatch(cluster: Cluster, domain_urls: tc.DomainUr
...
+    leader.drain()
+
+    # Then test no alarm while consumer 'foo' slowly confirms
+    time.sleep(1)
+    consumers["foo"].confirm(f"{uri_fanout}?id=foo", "*", succeed=True)
+
+    time.sleep(1)
+    consumers["foo"].confirm(f"{uri_fanout}?id=foo", "*", succeed=True)
+
+    producer.post(uri_fanout, ["msg1"], succeed=True, wait_ack=True)
```
dorjesinpo: "msg4"

src/integration-tests/test_alarms.py
```diff
@@ -119,6 +258,174 @@ def test_alarms_subscription_mismatch(cluster: Cluster, domain_urls: tc.DomainUr
...
+    time.sleep(1)
+
+    # Consumer 'bar' picks the last message
```
dorjesinpo: "the last message"?  or, the first message?

src/integration-tests/test_alarms.py
```diff
@@ -119,6 +258,174 @@ def test_alarms_subscription_mismatch(cluster: Cluster, domain_urls: tc.DomainUr
...
+# Consumption monitor fanout mode tests
+# -------------------------------------
+
+
+@tweak.domain.max_idle_time(3)
+def test_fanout_no_alarms_for_a_slow_queue(
+    cluster: Cluster, domain_urls: tc.DomainUrls
+):
+    """
+    Test no broker ALARMS in fanout mode for a slowly moving queue.
```
dorjesinpo: "no broker ALARMS" seems misleading since there are alarms in the test

src/integration-tests/test_alarms.py
```diff
@@ -28,11 +28,62 @@
...
+
+
 @tweak.domain.max_idle_time(3)
-def test_no_alarms_for_a_slow_queue(cluster: Cluster, domain_urls: tc.DomainUrls):
+def test_priority_no_alarms_for_a_slow_queue(
+    cluster: Cluster, domain_urls: tc.DomainUrls
+):
     """
-    Test no broker ALARMS for a slowly moving queue.
+    Test no broker ALARMS in priority mode for a slowly moving queue.
```
dorjesinpo, line 86: "no broker ALARMS" seems misleading since there are alarms in the test

The logic seems confusing

1. max_unconfirmed_messages=1
2. we post and confirm 1 message
3. we post 2 more messages
4. wait 4 secs for alarm
5. wait 1 sec and confirm the second message
6. wait 1 sec and confirm the third (last) message
7. why open one more consumer? is it supposed to receive anything?

It seems that at some point, we should see ""no longer appears to be stuck.".  After 6?

What is the point of 2 messages at step 3. and two separate confirm steps: 5. and 6.?

src/integration-tests/test_alarms.py
```diff
@@ -74,13 +126,99 @@ def test_no_alarms_for_a_slow_queue(cluster: Cluster, domain_urls: tc.DomainUrls
...
+    consumer = proxy.create_client("consumer")
+    consumer.open(uri_priority, flags=["read"], succeed=True)
+
+    # Post a message
+    producer.post(uri_priority, ["msg1"], succeed=True, wait_ack=True)
+
+    # Wait more than max idle time and check no alarm is triggered
+    assert not leader.alarms("QUEUE_STUCK", 2)
+
+    # Close consumer queue and post one more message
```
dorjesinpo: what is the state of the stream now, is it `ALIVE`? 
What is the reason for closing consumer?  It should populate redelivery list.  Do we ignore redelivery list in our consumption monitor logic? So, the state of the stream will be `ALIVE` after 1 sec?

src/integration-tests/test_alarms.py
```diff
@@ -74,13 +126,99 @@ def test_no_alarms_for_a_slow_queue(cluster: Cluster, domain_urls: tc.DomainUrls
...
+    consumer.close(uri_priority, succeed=True)
+
+    producer.post(uri_priority, ["msg2"], succeed=True, wait_ack=True)
+
+    # Test that alarm is triggered
+    assert leader.alarms("QUEUE_STUCK", 2)
```
dorjesinpo, line 148: this is because the argument `2` is less than idle time `1`

src/integration-tests/test_alarms.py
```diff
@@ -74,13 +126,99 @@ def test_no_alarms_for_a_slow_queue(cluster: Cluster, domain_urls: tc.DomainUrls
...
+    # Post a message
+    producer.post(uri_priority, ["msg1"], succeed=True, wait_ack=True)
+
+    # Wait more than max idle time and check no alarm is triggered
+    assert not leader.alarms("QUEUE_STUCK", 2)
+
+    # Close consumer queue and post one more message
+    consumer.close(uri_priority, succeed=True)
+
+    producer.post(uri_priority, ["msg2"], succeed=True, wait_ack=True)
```
dorjesinpo, line 145: if we (our contract) ignore redelivery list, this is necessary to make `haveUndeliveredCb` return something

src/integration-tests/test_alarms.py
```diff
@@ -74,13 +126,99 @@ def test_no_alarms_for_a_slow_queue(cluster: Cluster, domain_urls: tc.DomainUrls
...
+
+    # Test that alarm is triggered
+    assert leader.alarms("QUEUE_STUCK", 2)
+    assert leader.capture(r"For appId: __default")
+    leader.drain()
+
+    # Open consumer queue and post one more message
+    consumer.open(uri_priority, flags=["read"], succeed=True)
+    producer.post(uri_priority, ["msg3"], succeed=True, wait_ack=True)
```
dorjesinpo: what is the reason for this post?

src/integration-tests/test_alarms.py
```diff
@@ -119,6 +258,174 @@ def test_alarms_subscription_mismatch(cluster: Cluster, domain_urls: tc.DomainUr
...
+
+    producer.post(uri_fanout, ["msg1"], succeed=True, wait_ack=True)
+
+    for app_id in app_ids:
+        consumers[app_id].confirm(f"{uri_fanout}?id={app_id}", "*", succeed=True)
+
+    producer.post(uri_fanout, ["msg1"], succeed=True, wait_ack=True)
+    producer.post(uri_fanout, ["msg1"], succeed=True, wait_ack=True)
+
+    time.sleep(4)
```
dorjesinpo: could this be `3` instead of `4`?

src/integration-tests/test_alarms.py
```diff
@@ -119,6 +258,174 @@ def test_alarms_subscription_mismatch(cluster: Cluster, domain_urls: tc.DomainUr
...
+    producer.post(uri_fanout, ["msg1"], succeed=True, wait_ack=True)
+    time.sleep(1)
+
+    # Consumer 'bar' picks the last message
+    consumer_bar = next(proxies).create_client("bar")
+    consumer_bar.open(f"{uri_fanout}?id=bar", flags=["read"], succeed=True)
+
+    # Test that no alarm is triggered
+    time.sleep(1)
+    assert not leader.alarms("QUEUE_STUCK", 1)
```
dorjesinpo: should we see "no longer appears to be stuck." now?

src/integration-tests/test_alarms.py
```diff
@@ -119,6 +258,174 @@ def test_alarms_subscription_mismatch(cluster: Cluster, domain_urls: tc.DomainUr
...
+    consumers = {}
+    for app_id in app_ids:
+        consumer = next(proxies).create_client(app_id)
+        consumers[app_id] = consumer
+        consumer.open(f"{uri_fanout}?id={app_id}", flags=["read"], succeed=True)
+
+    producer.post(uri_fanout, ["msg1"], succeed=True, wait_ack=True)
+
+    # Wait more than max idle time and check no alarm is triggered
+    time.sleep(2)
```
dorjesinpo: could this be `1` instead of `2`?

src/integration-tests/test_alarms.py
```diff
@@ -119,6 +258,174 @@ def test_alarms_subscription_mismatch(cluster: Cluster, domain_urls: tc.DomainUr
...
+
+    # Test that alarm is triggered
+    assert leader.alarms("QUEUE_STUCK", 1)
+    assert leader.capture(r"For appId: foo")
+    leader.drain()
+
+    # Open consumer queue and post one more message
+    consumers["foo"].open(f"{uri_fanout}?id=foo", flags=["read"], succeed=True)
+    producer.post(uri_fanout, ["msg3"], succeed=True, wait_ack=True)
```
dorjesinpo: is this post important? would we see "no longer appears to be stuck." without it?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -199,100 +197,345 @@ void QueueConsumptionMonitor::reset()
...
-                    info.d_lastKnownGoodTimer = d_currentTimer;
-                    onTransitionToAlive(&info, id);
-                }
-            }
+        // There are no un-delivered messages or alarm time of the oldest
+        // message is in the future, so we can transition the substream to
+        // alive state.
+        onTransitionToAlive(&info, appId);
+
+        if (!d_alarmEventHandle && oldestMsgIt) {
```
dorjesinpo: I wonder if `d_alarmEventHandle` can be `true` if `info.d_state == State::e_IDLE`.  Probably, we can assert it cannot?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -199,100 +197,345 @@ void QueueConsumptionMonitor::reset()
...
+        // reschedule the idle event.
+        scheduleIdleEvent(&info, appId);
+        return;  // RETURN
+    }
+
+    // There are no un-delivered messages or alarm time of the oldest message
+    // is in the future, so we can transition the substream to alive state.
+    onTransitionToAlive(&info, appId);
+
+    if (!d_alarmEventHandle && oldestMsgIt) {
```
dorjesinpo: very minor, all (two) calls to `onTransitionToAlive` are followed by the same 
```
        if (!d_alarmEventHandle && oldestMsgIt) {
            // Since there is the oldest un-delivered message with alarm time
            // in the future, schedule the alarm event.
            scheduleAlarmEvent(alarmTime);
        }
```

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -81,8 +190,18 @@ struct Test : bmqtst::Test {
     ~Test() BSLS_KEYWORD_OVERRIDE;
 
     // MANIPULATORS
-    void putMessage(const bsl::string& id = bsl::string());
-    bool loggingCb(const bsl::string& id, bool enableLog);
+    void putMessage(const bsl::string& appId         = bsl::string(),
+                    bool               isUndelivered = true);
```
dorjesinpo: the `isUndelivered` seems strange.  Why not call `d_monitor.onMessagePosted()` in that one case when `isUndelivered == false`?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -311,74 +455,97 @@ BMQTST_TEST_F(Test, putAliveIdleEmptyAlive)
...
+    d_monitor.setMaxIdleTime(0);
+
     BMQTST_ASSERT_EQ(d_monitor.state(d_id),
                      QueueConsumptionMonitor::State::e_ALIVE);
     BMQTST_ASSERT_EQ(logObserver.records().size(), 0u);
 
-    d_monitor.onTimer(k_MAX_IDLE_TIME * 2);
-    BMQTST_ASSERT_EQ(d_monitor.state(d_id),
-                     QueueConsumptionMonitor::State::e_ALIVE);
+    d_monitor.alarmEventDispatched();
```
dorjesinpo: why do we call `alarmEventDispatched` after `setMaxIdleTime(0)`?

should we check for `!isAlarmScheduled()`?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -311,74 +455,97 @@ BMQTST_TEST_F(Test, putAliveIdleEmptyAlive)
...
 
-    d_monitor.onTimer(k_MAX_IDLE_TIME * 2 + k_MAX_IDLE_TIME * 2);
-    BMQTST_ASSERT_EQ(d_monitor.state(d_id),
-                     QueueConsumptionMonitor::State::e_ALIVE);
+    // Enable monitoring again
+    d_monitor.setMaxIdleTime(2);
+
+    d_monitor.alarmEventDispatched();
```
dorjesinpo, line 537: should we check for `isAlarmScheduled()`?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -311,74 +455,97 @@ BMQTST_TEST_F(Test, putAliveIdleEmptyAlive)
...
+    BMQTST_ASSERT(d_monitor.isAlarmScheduled());
+
+    d_monitor.alarmEventDispatched();
+
+    BMQTST_ASSERT_EQ(d_monitor.state(d_id),
+                     QueueConsumptionMonitor::State::e_IDLE);
+
+    d_monitor.setMaxIdleTime(2);
+
+    // No change in state if new value is set
```
dorjesinpo, line 513: this is interesting.  What do we want in our contract?  Stay in the `IDLE` state even if by the new config the state would be `ALIVE` or adapt to the new config and transition to `ALIVE`?
Note that the latter is very easy to do if we call `idleEventDispatched` from `setMaxIdleTime`

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -390,23 +557,27 @@ BMQTST_TEST_F(Test, reset)
...
 {
-    const bsls::Types::Int64 k_MAX_IDLE_TIME = 10;
-
-    d_monitor.setMaxIdleTime(k_MAX_IDLE_TIME);
+    d_monitor.setMaxIdleTime(1);
 
     d_monitor.registerSubStream(d_id);
 
-    putMessage();
+    putMessage(d_id, false);
```
dorjesinpo: why `false`? should the message not be considered as undelivered?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -444,59 +612,39 @@ BMQTST_TEST_F(Test, putAliveIdleSendAliveTwoSubstreams)
...
+    putMessage(id2);
 
-    d_monitor.onTimer(2 * k_MAX_IDLE_TIME);
-    BMQTST_ASSERT_EQ(d_monitor.state(id1),
-                     QueueConsumptionMonitor::State::e_ALIVE);
-    BMQTST_ASSERT_EQ(d_monitor.state(id2),
-                     QueueConsumptionMonitor::State::e_ALIVE);
-    BMQTST_ASSERT_EQ(logObserver.records().size(), expectedLogRecords);
+    BMQTST_ASSERT(d_monitor.isAlarmScheduled());
+
```
dorjesinpo, line 616: BMQTST_ASSERT(d_monitor.isAlarmScheduled());

dorjesinpo (changes requested):
The components look good.  There are comment about UTs and ITs.

Two main questions,

1. Should the state adapt to reconfiguration?  We currently seem to do that when deconfiguring (`setMaxIdleTime(0)`).  What about increasing/decreasing time interval?
2. We ignore redelivery list in the monitor logic.  Just making sure.  If a message ever got delivered, it is never "stale" even if the consumer drops.

src/integration-tests/test_alarms.py
alexander-e1off: Refactored the test logic to be better readable, removed unneeded alarm.

src/integration-tests/test_alarms.py
alexander-e1off: Now our contract includes checking of redelivery list.

src/integration-tests/test_alarms.py
alexander-e1off: Refactored the test logic to be better readable.

src/integration-tests/test_alarms.py
alexander-e1off: We have `max_idle_time = 1`, and for better readability I wanted to state that we wait ` more than max idle time`. leader.alarms("QUEUE_STUCK", 1) will try to wait one more second, but I assume alarm will be already logged, so no wait in `leader.alarms`.

src/integration-tests/test_alarms.py
alexander-e1off: Outdated, test logic is refactored.

src/integration-tests/test_alarms.py
alexander-e1off: Outdated, test logic is refactored.

src/integration-tests/test_alarms.py
alexander-e1off: Outdated, refactored the test logic.

src/integration-tests/test_alarms.py
alexander-e1off: Outdated, refactored the test logic.

src/integration-tests/test_alarms.py
alexander-e1off: Outdated, refactored the test logic.

src/integration-tests/test_alarms.py
alexander-e1off: Outdated, refactored the test logic.

src/integration-tests/test_alarms.py
alexander-e1off: Outdated, refactored the test logic.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
alexander-e1off: `d_alarmEventHandle` can be `true` if `info.d_state == State::e_IDLE`, because this alarm event is common for all apps (with contrast with `d_idleEventHandle`, which is per each App), and can be scheduled by another App. 

Below is scenario with two Apps:
1. `MaxIdleTime` = 5 sec;
2. At timestamp = 0 sec `Msg1_AppId_1` arrived, alarm event at 5 sec is scheduled;
3. At timestamp = 2 sec `Msg2_AppId_1` arrived;
4. At timestamp = 3 sec `Msg1_AppId_2` arrived;
5. At timestamp = 5 sec `alarmEventDispatched()` is called:
5.1 It sets `AppId_1 state = IDLE`; 
5.2 It schedules next alarm event at 3+5=8 sec (for `Msg1_AppId_2`)
6. At timestamp = 6 sec `Msg1_AppId_1` was delivered, in `onMessageSent()`:
6.1 It checks that oldest message for `AppId_1` is `Msg2_AppId_1` now, and it's alarm time is in the future (2+5=7 sec);
6.2 Since alarm time is in the future (7 sec), call `onTransitionToAlive()`;
6.3 Check if need to schedule event: if alarm event is already scheduled (`d_alarmEventHandle` != 0), in our case it was scheduled by `Msg1_AppId_2`,  we need to remember the `alarmTime` it was scheduled (8 sec) and compare it with current (7sec).
If current `alarmTime` is less, we need to reschedule the event.

Now this logic is enhanced to handle case above and moved into `QueueConsumptionMonitor::scheduleOrRescheduleAlarmEventIfNeeded()` method.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
alexander-e1off: Moved this logic into `QueueConsumptionMonitor::scheduleOrRescheduleAlarmEventIfNeeded()` method

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
alexander-e1off: Removed `isUndelivered`

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
alexander-e1off: Removed call off `alarmEventDispatched` after setMaxIdleTime(0), added `!d_monitor.isAlarmScheduled()` check

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
alexander-e1off: After transition disabling->enabling monitor, alarm event is not scheduled, all apps in ALIVE state, so need some trigger (message posting) to schedule it and then check `isAlarmScheduled()`. Added this logic.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
alexander-e1off: Added comment in `setMaxIdleTime()` that `No change in app states when new max idle time value is set.` if new value is not zero.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
alexander-e1off: Removed second argument, updated test

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
alexander-e1off: What is needed to do here? We already have `isAlarmScheduled()` check.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -199,99 +200,372 @@ void QueueConsumptionMonitor::reset()
...
+    if (subStreamInfo->d_idleEventHandle) {
+        // Cancel the idle event if it was scheduled.
+        d_queueState_p->scheduler()->cancelEventAndWait(
+            &subStreamInfo->d_idleEventHandle);
+        subStreamInfo->d_idleEventHandle.release();
+    }
+
+    subStreamInfo->d_state = State::e_ALIVE;
+
+    bdlma::LocalSequentialAllocator<2048> localAllocator(0);
```
678098: 1. Let's pass the allocator passed on `QueueConsumptionMonitor` construction?
2. I think it's enough to have a block of <1024> bytes for both `UriBuilder` and `Uri`. Maybe even 512 is enough, but I am less sure.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -199,99 +200,372 @@ void QueueConsumptionMonitor::reset()
...
+
+    bmqt::UriBuilder uriBuilder(d_queueState_p->uri(), &localAllocator);
+    uriBuilder.setId(appId);
+
+    bmqt::Uri uri(&localAllocator);
+    uriBuilder.uri(&uri);
+
+    BALL_LOG_INFO << "Queue '" << uri << "' no longer appears to be stuck.";
```
678098: Usually, INFO log severity is always enabled. However, in case it doesn't, it is good not to construct these tmp `uriBuilder`, `uri` objects. You can achieve it by constructing them within INFO BLOCK

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -199,99 +200,372 @@ void QueueConsumptionMonitor::reset()
...
+        idleTime,
+        bdlf::BindUtil::bind(
+            &QueueConsumptionMonitor::executeIdleInQueueDispatcher,
+            this,
+            appId));
+}
+
+void QueueConsumptionMonitor::executeAlarmInQueueDispatcher()
+{
+    // executed by the *SCHEDULER DISPATCHER* thread
```
678098:
```suggestion
    // executed by the *SCHEDULER* thread
```

Scheduler thread is not a dispatcher thread.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -199,99 +200,372 @@ void QueueConsumptionMonitor::reset()
...
+    d_queueState_p->queue()->dispatcher()->execute(
+        bdlf::BindUtil::bind(&QueueConsumptionMonitor::alarmEventDispatched,
+                             this),
+        d_queueState_p->queue());
+}
+
+void QueueConsumptionMonitor::executeIdleInQueueDispatcher(
+    const bsl::string& appId)
+{
+    // executed by the *SCHEDULER DISPATCHER* thread
```
678098:
```suggestion
    // executed by the *SCHEDULER* thread
```

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -199,99 +200,372 @@ void QueueConsumptionMonitor::reset()
...
+    // Forward event to the queue dispatcher thread
+    d_queueState_p->queue()->dispatcher()->execute(
+        bdlf::BindUtil::bind(&QueueConsumptionMonitor::idleEventDispatched,
+                             this,
+                             appId),
+        d_queueState_p->queue());
+}
 
-    d_currentTimer = currentTimer;
+void QueueConsumptionMonitor::cancelIdleEventsAndResetStates(bool resetStates)
```
678098:
```suggestion
void QueueConsumptionMonitor::cancelIdleEvents(bool resetStates)
```
The name of the function suggests that it always resets states. However, this behaviour is controlled by the bool arg passed to the function. It's better to remove `reset states` from the function name and mention reset states in function docs.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -199,99 +200,372 @@ void QueueConsumptionMonitor::reset()
...
-            BSLS_ASSERT_SAFE(info.d_state == State::e_ALIVE);
+    if (d_alarmEventHandle) {
+        d_alarmEventHandle.release();
+    }
+
+    if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(d_maxIdleTimeSec == 0)) {
+        // monitoring is disabled
+        BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
+        return;  // RETURN
+    }
```
678098:
```suggestion
    if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(d_maxIdleTimeSec == 0)) {
        // monitoring is disabled
        BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
        return;  // RETURN
    }

    if (d_alarmEventHandle) {
        d_alarmEventHandle.release();
    }
```

Possible to rearrange these instructions. If monitoring is enabled, there will be no change. If monitoring is disabled, the check for `d_alarmEventHandle` will be skipped by early return.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -199,99 +200,372 @@ void QueueConsumptionMonitor::reset()
...
 {
     // executed by the *QUEUE DISPATCHER* thread
 
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(d_queueState_p->queue()->dispatcher()->inDispatcherThread(
         d_queueState_p->queue()));
 
-    if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(d_maxIdleTime == 0)) {
+    if (d_alarmEventHandle || d_maxIdleTimeSec == 0) {
+        // Event is already scheduled or monitoring is disabled.
```
678098:
```suggestion
    if (d_alarmEventHandle || isMonitoringDisabled()) {
```

Can make a small inline const function and use it here:

```
inline bool isMonitoringDisabled() const
{
    return d_maxIdleTimeSec == 0;
}
```

The function name will serve a role of a self-commenting code, and it will be possible to remove the actual comment

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -199,99 +200,372 @@ void QueueConsumptionMonitor::reset()
...
+    bsls::TimeInterval alarmTime = now;
+    alarmTime.addNanoseconds(-arrivalTimeDeltaNs);
+    alarmTime.addSeconds(d_maxIdleTimeSec);
+
+    return alarmTime;
+}
+
+bool QueueConsumptionMonitor::isAlarmScheduled() const
+{
+    return d_alarmEventHandle ? true : false;
```
678098:
```suggestion
    return d_alarmEventHandle != 0;
```

678098 (commented):
Have more comments

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
alexander-e1off: Should I also change it [here](https://github.com/alexander-e1off/blazingmq/blob/ea8f576799e085f3e85e8fa7721e7255429a46e7/src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp#L1264) ?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
dorjesinpo: Objection.
If alarm event is schedules and then we reconfigure and `cancelEventAndWait` does not succeed, we are going to have `alarmEventDispatched` which then leave `d_alarmEventHandle` unreleased. Later we can reconfigure again

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -495,22 +523,23 @@ void QueueConsumptionMonitor::idleEventDispatched(const bsl::string& appId)
...
+        // schedule the idle event.
+        if (alarmTime <= now) {
+            scheduleIdleEvent(&info, appId);
+
+            return;  // RETURN
+        }
+
+        // Since there is the oldest un-delivered message with alarm time in
+        // the future, schedule or reschedule the alarm event if needed.
+        scheduleOrRescheduleAlarmEventIfNeeded(alarmTime);
```
dorjesinpo, line 552: can it be possible to have `info.d_state == State::e_IDLE` and `d_alarmEventHandle` ?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -249,24 +250,25 @@ void QueueConsumptionMonitor::onMessageSent(const bsl::string& appId)
...
+                // Since the idle event was scheduled, it will continue to
+                // track state.
+                BSLS_ASSERT_SAFE(info.d_idleEventHandle);
+
+                return;  // RETURN
+            }
+
+            // Since there is the oldest un-delivered message with alarm time
+            // in the future, schedule or reschedule the alarm event if needed.
+            scheduleOrRescheduleAlarmEventIfNeeded(alarmTime);
```
dorjesinpo, line 274: can it be possible to have info.d_state == State::e_IDLE and d_alarmEventHandle ?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -310,14 +312,40 @@ void QueueConsumptionMonitor::scheduleAlarmEvent(
...
     d_queueState_p->scheduler()->scheduleEvent(
         &d_alarmEventHandle,
-        alarmTime,
+        d_scheduledAlarmTime,
         bdlf::BindUtil::bind(
             &QueueConsumptionMonitor::executeAlarmInQueueDispatcher,
             this));
 }
 
+void QueueConsumptionMonitor::scheduleOrRescheduleAlarmEventIfNeeded(
```
dorjesinpo, line 340: the rational for `scheduleOrRescheduleAlarmEventIfNeeded` is not clear. 
In which scenarios we are not sure whether it is scheduled or not?

src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -1474,5 +1474,55 @@ void QueueEngineUtil_AppState::reportStats(
...
+    // PRECONDITIONS
+
+    BSLS_ASSERT_SAFE(d_queue_p->storage());
+
+    bool isFound = false;
+
+    // Check in the following order: redelivery list, put aside list and resume
+    // point.
+    if (!d_redeliveryList.empty()) {
+        isFound = getOldestMessageIteratorFromList(out, d_redeliveryList);
```
dorjesinpo: very minor, feel free to ignore, the check for `empty()` can be done in one place (in `getOldestMessageIteratorFromList`)

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
```diff
@@ -158,38 +142,56 @@ QueueConsumptionMonitor::setMaxIdleTime(bsls::Types::Int64 value)
...
+        cancelIdleEvents(true);
+    }
+
+    // No change in app states when new max idle time value is set.
+
+    // If alarm event was already scheduled
+    if (d_alarmEventHandle) {
+        // Cancel the event and execute alarmEventDispatched() to reschedule
+        // alarm event for the new maxIdleTime.
+        d_queueState_p->scheduler()->cancelEventAndWait(&d_alarmEventHandle);
```
dorjesinpo: this `cancelEventAndWait` may not succeed (see the contract saying "the event has already been dispatched").  So, `alarmEventDispatched ` may be called second time.
That may be ok, we will call `scheduleAlarmEvent`, then `d_alarmEventHandle.release()`, then `scheduleAlarmEvent` again.  Please check this reasoning
We can avoid that by checking `cancelEventAndWait` return code?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -155,53 +278,86 @@ Test::~Test()
...
+Test::haveUndeliveredCb(bsls::TimeInterval*       alarmTime_p,
+                        const bsl::string&        appId,
+                        const bsls::TimeInterval& now) const
 {
-    BALL_LOG_SET_CATEGORY("MQBBLP.QUEUECONSUMPTIONMONITORTEST");
+    bslma::ManagedPtr<mqbi::StorageIterator> oldestMsgIt;
+
+    const bool haveUndelivered = d_haveUndelivered.contains(appId);
+    if (haveUndelivered) {
+        bslma::Allocator*    alloc = bslma::Default::defaultAllocator();
```
dorjesinpo: maybe we can use `bmqtst::TestHelperUtil::allocator()` instead (cache it in `Test`)?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -218,18 +374,18 @@ BMQTST_TEST_F(Test, emptyQueue)
...
     d_monitor.registerSubStream(d_id);
 
-    d_monitor.onTimer(k_MAX_IDLE_TIME);
     BMQTST_ASSERT_EQ(d_monitor.state(d_id),
                      QueueConsumptionMonitor::State::e_ALIVE);
-    BMQTST_ASSERT_EQ(logObserver.records().size(), expectedLogRecords);
 
-    d_monitor.onTimer(k_MAX_IDLE_TIME + 1);
+    // Simulate alarm event dispatching
+    d_monitor.alarmEventDispatched();
```
dorjesinpo: simulating an `alarmEventDispatched` which would not happen / must not happen.
note that `setMaxIdleTime` calls `alarmEventDispatched` internally

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -311,74 +452,102 @@ BMQTST_TEST_F(Test, putAliveIdleEmptyAlive)
...
 
     bmqtst::ScopedLogObserver logObserver(ball::Severity::e_INFO,
                                           bmqtst::TestHelperUtil::allocator());
 
-    d_monitor.setMaxIdleTime(k_MAX_IDLE_TIME * 2);
+    // Disable monitoring
+    d_monitor.setMaxIdleTime(0);
+
+    BMQTST_ASSERT(!d_monitor.isAlarmScheduled());
+
```
dorjesinpo, line 524: we transition from `IDLE` to `ALIVE` _without_ logging anything if monitoring gets disabled.
either seems ok, but needs a mention in the contract.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -311,74 +452,102 @@ BMQTST_TEST_F(Test, putAliveIdleEmptyAlive)
...
-    BMQTST_ASSERT_EQ(d_monitor.state(d_id),
-                     QueueConsumptionMonitor::State::e_ALIVE);
+    // Enable monitoring again
+    d_monitor.setMaxIdleTime(2);
 
-    d_monitor.onTimer(k_MAX_IDLE_TIME * 2 + k_MAX_IDLE_TIME * 2);
-    BMQTST_ASSERT_EQ(d_monitor.state(d_id),
-                     QueueConsumptionMonitor::State::e_ALIVE);
+    // Simulate message posting to schedule alarm event
+    d_monitor.onMessagePosted();
```
dorjesinpo, line 533: what happens without posting new message?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -311,74 +452,102 @@ BMQTST_TEST_F(Test, putAliveIdleEmptyAlive)
...
+    BMQTST_ASSERT(d_monitor.isAlarmScheduled());
+
+    d_monitor.alarmEventDispatched();
 
-    d_monitor.onTimer(k_MAX_IDLE_TIME * 2 + k_MAX_IDLE_TIME * 2 + 1);
     BMQTST_ASSERT_EQ(d_monitor.state(d_id),
-                     QueueConsumptionMonitor::State::e_IDLE);
+                     QueueConsumptionMonitor::State::e_IDLE)
+
+    BMQTST_ASSERT_EQ(logObserver.records().size(), 1u);
```
dorjesinpo, line 542: what is tin hat one log record?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -390,23 +559,25 @@ BMQTST_TEST_F(Test, reset)
...
-                     QueueConsumptionMonitor::State::e_ALIVE);
+                     QueueConsumptionMonitor::State::e_IDLE);
 
     bmqtst::ScopedLogObserver logObserver(ball::Severity::e_INFO,
                                           bmqtst::TestHelperUtil::allocator());
-
     d_monitor.reset();
-    d_monitor.onTimer(k_MAX_IDLE_TIME + 1);
+
+    BMQTST_ASSERT(!d_monitor.isAlarmScheduled());
```
dorjesinpo, line 571: should we check the state?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
```diff
@@ -535,44 +685,52 @@ BMQTST_TEST_F(Test, usage)
...
-    monitor.onTimer(T += 15);  // nothing is logged
-    BMQTST_ASSERT_EQ(logObserver.records().size(), 2u);
-    // 15 seconds later - T + 105s
-    monitor.onTimer(T += 15);  // log ALARM
-    BMQTST_ASSERT_EQ(logObserver.records().size(), 3u);
-    // 15 seconds later - T + 120s
+    // Simulate idle event dispatching
+    monitor.idleEventDispatched(d_id);
+
+    // remain ALARM
```
dorjesinpo: but the state is not `ALARM`

src/integration-tests/test_alarms.py
```diff
@@ -41,46 +92,179 @@ def test_no_alarms_for_a_slow_queue(cluster: Cluster, domain_urls: tc.DomainUrls
...
+    consumer.confirm(uri_priority, "+1", succeed=True)
+    consumer.wait_push_event()
 
-    consumer1.confirm(uri_priority, "*", succeed=True)
+    time.sleep(1)
+    consumer.confirm(uri_priority, "+1", succeed=True)
+    consumer.wait_push_event()
 
+    # Test that no alarm is triggered
+    time.sleep(2)
```
dorjesinpo: we've already slept 1+1, and `alarms` is another 1.  These 2 secs seem redundant.

src/integration-tests/test_alarms.py
```diff
@@ -41,46 +92,179 @@ def test_no_alarms_for_a_slow_queue(cluster: Cluster, domain_urls: tc.DomainUrls
...
+    )
+
+    # Post "msg1", it is delivered to consumer
     producer.post(uri_priority, ["msg1"], succeed=True, wait_ack=True)
-    producer.post(uri_priority, ["msg1"], succeed=True, wait_ack=True)
 
-    time.sleep(4)
+    # Wait more than max idle time and check no alarm is triggered
+    time.sleep(2)
+    assert not leader.alarms("QUEUE_STUCK", 1)
```
dorjesinpo: can shave a sec by removing `time.sleep(2)` and  changing `assert not leader.alarms("QUEUE_STUCK", 2)`

src/integration-tests/test_alarms.py
```diff
@@ -41,46 +92,179 @@ def test_no_alarms_for_a_slow_queue(cluster: Cluster, domain_urls: tc.DomainUrls
...
+
+    # Post "msg2", it is not delivered to consumer due to max_unconfirmed_messages=1
+    producer.post(uri_priority, ["msg2"], succeed=True, wait_ack=True)
 
-    # First, test the alarm
+    # Wait more than max idle and check that alarm is triggered
+    time.sleep(2)
```
dorjesinpo: can shave a sec

src/integration-tests/test_alarms.py
```diff
@@ -41,46 +92,179 @@ def test_no_alarms_for_a_slow_queue(cluster: Cluster, domain_urls: tc.DomainUrls
...
+    # Disable consumption monitor (maxIdleTime = 0)
+    cluster.config.domains[du.domain_priority].definition.parameters.max_idle_time = 0
+    cluster.reconfigure_domain(du.domain_priority, succeed=True)
+
+    # Close consumer queue and post one more message
+    consumer.close(du.uri_priority, succeed=True)
+    producer.post(du.uri_priority, ["msg2"], succeed=True, wait_ack=True)
+
+    # Test that alarm is not triggered
+    time.sleep(2)
```
dorjesinpo: can shave a sec

src/integration-tests/test_alarms.py
```diff
@@ -41,46 +92,179 @@ def test_no_alarms_for_a_slow_queue(cluster: Cluster, domain_urls: tc.DomainUrls
...
+    # Decrease max idle time from 3 to 1 second
+    cluster.config.domains[du.domain_priority].definition.parameters.max_idle_time = 1
+    cluster.reconfigure_domain(du.domain_priority, succeed=True)
+
+    # Within 2 sec (more than new max idle time 1 sec but less than old one 3 sec) check that alarm is triggered
+    assert leader.alarms("QUEUE_STUCK", 2)
+    assert leader.capture(r"For appId: __default")
+    leader.drain()
+
+    # Confirm messages, queue is empty
```
dorjesinpo: the queue is not empty, see `wait_push_event` below

src/integration-tests/test_alarms.py
```diff
@@ -41,46 +92,179 @@ def test_no_alarms_for_a_slow_queue(cluster: Cluster, domain_urls: tc.DomainUrls
...
+    consumer.confirm(du.uri_priority, "*", succeed=True)
+    consumer.wait_push_event()
+
+    # Check that queue is no longer in alarm state
+    assert leader.outputs_regex(r"no longer appears to be stuck.", 1)
+
+    # Post 2 more messages, msg4 is not delivered to consumer due to max_unconfirmed_messages=1
```
dorjesinpo: neither msg3 nor msg4 get delivered because the consumer has confirmed only one message out of 4
we may not need msg4

src/integration-tests/test_alarms.py
```diff
@@ -41,46 +92,179 @@ def test_no_alarms_for_a_slow_queue(cluster: Cluster, domain_urls: tc.DomainUrls
...
+
+    # Increase max idle time from 1 to 3 seconds
+    cluster.config.domains[du.domain_priority].definition.parameters.max_idle_time = 3
+    cluster.reconfigure_domain(du.domain_priority, succeed=True)
+
+    # Within 2 secs (more than old max idle time 1 sec) check no alarm is triggered
+    assert not leader.alarms("QUEUE_STUCK", 2)
+
+    # Wait 2 seconds more (more than new max idle time 3 sec) and check that alarm is triggered
+    time.sleep(2)
```
dorjesinpo: can shave a sec

src/integration-tests/test_alarms.py
```diff
@@ -111,6 +295,7 @@ def test_alarms_subscription_mismatch(cluster: Cluster, domain_urls: tc.DomainUr
     time.sleep(2)
```
dorjesinpo: can shave a sec

src/integration-tests/test_alarms.py
dorjesinpo: > Refactored the test logic to be better readable.

still, the same question.  does `bar` get an alarm before its consumer opens? and if it does, then does it get "no longer appears to be stuck." after consumer opens?

src/integration-tests/test_alarms.py
dorjesinpo: > We have `max_idle_time = 1`, and for better readability I wanted to state that we wait ` more than max idle time`. leader.alarms("QUEUE_STUCK", 1) will try to wait one more second, but I assume alarm will be already logged, so no wait in `leader.alarms`.

there is no alarm.  so, we wait `2` + `1` (in `alarms`).  Extra sec.  Overall, these tests seem to sleep a lot, not ideal

src/integration-tests/test_alarms.py
```diff
@@ -119,6 +304,182 @@ def test_alarms_subscription_mismatch(cluster: Cluster, domain_urls: tc.DomainUr
...
+    producer.post(uri_fanout, ["msg2"], succeed=True, wait_ack=True)
+
+    # Apps except 'foo' confirm all messages, msg2 for 'foo' remains in queue
+    for app_id in app_ids:
+        if app_id != "foo":
+            consumers[app_id].confirm(f"{uri_fanout}?id={app_id}", "*", succeed=True)
+            consumers[app_id].wait_push_event()
+
+    # Wait more than max idle time
+    time.sleep(2)
```
dorjesinpo: potentially unnecessary wait (depending on timing)
better not have the `sleep` and use `2` in `alarms`.  Then, the alarm gets caught as soon as possible

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
alexander-e1off: Originally I also thought this way. After proposed performance improvement, I've checked that the case when`alarmEventDispatched() ` is called with `d_maxIdleTimeSec` == 0 is possible only when we call [setMaxIdleTime(0)](https://github.com/alexander-e1off/blazingmq/blob/stale-alarm/src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp#L161) or [reset()](https://github.com/alexander-e1off/blazingmq/blob/stale-alarm/src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp#L206) and `cancelEventAndWait` did not succeed. But in both methods we explicitly call `d_alarmEventHandle.release()`, so no need to release it again in `alarmEventDispatched()` and can early return if `d_maxIdleTimeSec` == 0.  I agree that potentially it's error prone, because in future we can add some methods that will not release `d_alarmEventHandle`. Should I rollback this change?

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
alexander-e1off: As I stated in my previous [comment](https://github.com/bloomberg/blazingmq/pull/736#discussion_r2151750010) `d_alarmEventHandle` can be true if info.d_state == State::e_IDLE, because this alarm event is common for all apps (with contrast with d_idleEventHandle, which is per each App), and can be scheduled by another App. We know that old message caused `IDLE` state is gone (delivered or GC-ed), but we have `next oldest message` for this app with alarm time in the future. So we call `scheduleOrRescheduleAlarmEventIfNeeded()` to check that alarm event is already scheduled by another app, and check if scheduled alarm time is earlier than `next oldest message` one. If yes, we do nothing. Otherwise, reschedule alarm event at `next oldest message` time, or schedule event if it was not scheduled (by another app).

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
alexander-e1off: The same as previous comment

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
alexander-e1off: We call  `scheduleOrRescheduleAlarmEventIfNeeded()` in case when app is in IDLE state, message caused it is gone (delivered or GCed) and there is `next old message` for this app that have alarm time in the future. 
`scheduleOrRescheduleAlarmEventIfNeeded()` logic is the following:
- If alarm event is already scheduled by another app at time earlier than `next old message` time, just do nothing;
- If alarm event is already scheduled by another app at time later than `next old message` time, reschedule alarm event at `next old message` time;
- If alarm event is not scheduled, schedule it at `next old message` time;

src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
alexander-e1off: Agree, moved into getOldestMessageIteratorFromList()

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
alexander-e1off: Added handling of `cancelEventAndWait` return code.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
alexander-e1off: Agree, added member `d_allocator_p` in Test and use it everywhere.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
alexander-e1off: Agree, removed it. `setMaxIdleTime` calls `alarmEventDispatched` internally only if alarm was scheduled. In this test no messages are posted.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
alexander-e1off: After disabling monitor, substreams states are reset to 'alive' and scheduled alarms are cancelled. So, when we enable monitor again, the state will be ALIVE without any activity (posting the message). To check that monitor works after enabling, we post a message.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
alexander-e1off: ALARM log - we posted message after enabling the monitor.

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
alexander-e1off: After reset, there is no any applications (we do `d_subStreamInfos.clear()`).

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
alexander-e1off: changed ALARM to IDLE

src/integration-tests/test_alarms.py
alexander-e1off: Agree, removed time.sleep(2)

src/integration-tests/test_alarms.py
alexander-e1off: Agree, changed comments to be clear and removed msg4.

src/integration-tests/test_alarms.py
alexander-e1off: To follow the test name/logic, removed special handle for `bar` and handle it as others.

src/integration-tests/test_alarms.py
alexander-e1off: replaced `time.sleep(2)` with `leader.alarms("QUEUE_STUCK", 2)`

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.t.cpp
alexander-e1off: Updated contract as following
```
/// Monitoring can be
/// disabled by setting the maximum idle time to zero. In this case substreams
/// states are reset to 'alive' and scheduled alarm is cancelled. No any
/// logs are generated when monitoring is disabled.
```

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
dorjesinpo: Considering that we want to call `release()` only when `cancelEventAndWait` succeeds, yes, we should `release()` in `alarmEventDispatched`

src/groups/mqb/mqbblp/mqbblp_queueconsumptionmonitor.cpp
dorjesinpo: thank you! may bad, overlooked the shared nature of `d_alarmEventHandle`

alexander-e1off:
@dorjesinpo thanks for your comments, I've addressed them, please have a look at changes and comments. Thanks.

src/integration-tests/test_alarms.py
```diff
@@ -41,46 +92,201 @@ def test_no_alarms_for_a_slow_queue(cluster: Cluster, domain_urls: tc.DomainUrls
...
-    # Consumer2 picks the last message
-    consumer2.open(
-        uri_priority, flags=["read"], max_unconfirmed_messages=1, succeed=True
+    # Consumer dropping the connection, msg1 is in redelivery list
+    consumer.stop_session(block=True)
+
+    # Wait more than max idle and check that alarm is triggered
+    assert leader.alarms("QUEUE_STUCK", 2)
+    assert leader.capture(r"For appId: __default")
+
```
dorjesinpo, line 188: can we `assert leader.capture(r"Redelivery list size: 1")`?

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1743,24 +1748,45 @@ bool RootQueueEngine::logAlarmCb(const bsl::string& appId,
...
-        return true;  // RETURN
+    if (headIt) {
+        // Get the arrival time delta of the oldest un-delivered message.
+        bsls::Types::Int64 arrivaTimelDelta = 0;
+        mqbs::StorageUtil::loadArrivalTimeDelta(&arrivaTimelDelta,
+                                                headIt->attributes());
+        // Calculate alarm time
+        const bsls::TimeInterval alarmTime =
+            d_consumptionMonitor.calculateAlarmTime(arrivaTimelDelta, now);
+        if (alarmTime_p) {
```
dorjesinpo: seems like the logic of the monitor always needs `alarmTime_p`

dorjesinpo (approved):
thank you

dorjesinpo:
@alexander-e1off let's resolve conflicts, reapprove and merge

dorjesinpo (approved):
thank you

PR #738, opened 2025-05-07
Fix[MQB]: thread race on operation logging [179164551]

678098:
- Introduce new component `bmqsys::OperationLogger` used to log operation times.
- Remove the internal state of `mqba::ClientSession` used to measure operation times.
- Instead, use instances of `bmqsys::OperationLogger` in `mqba::ClientSession`, to remove unsafe multi-thread access to `mqba::ClientSession` field.

```
bmqsys_operationlogger.h:202 bmqtool.tsk:25638: open queue [queue: 'bmq://bmq.test.persistent.priority/qqq'] took: 2.31 ms (2313125 nanoseconds)

bmqsys_operationlogger.h:202 bmqtool.tsk:60635: configure queue [queueId: 0, queue: 'bmq://bmq.test.persistent.priority/qqq'] took: 1.43 ms (1429375 nanoseconds)

bmqsys_operationlogger.h:202 bmqtool.tsk:49305: close queue [queue: 'bmq://bmq.test.persistent.priority/qqq'] took: 1.22 ms (1223709 nanoseconds)
```

### DIFF
--- a/src/groups/mqb/mqba/mqba_clientsession.cpp
+++ b/src/groups/mqb/mqba/mqba_clientsession.cpp
@@ -771,9 +771,10 @@ void ClientSession::tearDownAllQueuesDone(const bsl::shared_ptr<void>& session)
 }
 
 void ClientSession::onHandleConfigured(
-    const bmqp_ctrlmsg::Status&           status,
-    const bmqp_ctrlmsg::StreamParameters& streamParameters,
-    const bmqp_ctrlmsg::ControlMessage&   streamParamsCtrlMsg)
+    const bmqp_ctrlmsg::Status&                     status,
+    const bmqp_ctrlmsg::StreamParameters&           streamParameters,
+    const bmqp_ctrlmsg::ControlMessage&             streamParamsCtrlMsg,
+    const bsl::shared_ptr<bmqsys::OperationLogger>& opLogger)
 {
     // executed by the *ANY* thread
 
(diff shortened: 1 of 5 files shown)


src/groups/mqb/mqba/mqba_clientsession.cpp
```diff
@@ -2770,38 +2757,62 @@ void ClientSession::processEvent(
...
-                controlMessage);
+                controlMessage,
+                opLogger);
         } break;
         case MsgChoice::SELECTION_ID_CONFIGURE_QUEUE_STREAM:
         case MsgChoice::SELECTION_ID_CONFIGURE_STREAM: {
-            d_beginTimestamp = bmqsys::Time::highResolutionTimer();
+            bsl::shared_ptr<bmqsys::OperationLogger> opLogger =
+                bsl::allocate_shared<bmqsys::OperationLogger>(
+                    d_state.d_allocator_p);
```
678098, line 2808: Note: I wanted to use object pool to reuse operation loggers, however, this is not a good approach here.
1. `MemOutStream` used in `bmqsys::OperationLogger` doesn't reuse the internal buffer, so we don't benefit from reusing the same object. `MemOutStream::reset()` frees the internal buffer completely, and we don't have other way to save it. `bsl::stringstream` doesn't have this either.
2. `mqba::ClientSession` has a non-trivial lifetime. There can be dispatcher event associated with client sessions that were already destructed. If client session owns an object pool, we cannot return items to this object pool after client session destruction.

PR #740, opened 2025-05-08
Feat[MQB]: Add authn plugin config and controller

emelialei88:
This PR:
1. Add config schema for Authentication plugins
2. Add an `AuthenticationController` that's responsible for:
   - Loading Authentication factories and creating Authentication plugins upon start
   - Calling the proper `Authenticator` plugin to authenticate
3. Provide two Authentication plugins `bmqauthnfail` and `bmqauthnpass`. Both have `Basic` as `mechanism`, but one fail and another pass no matter what.

### DIFF
--- a/src/plugins/bmqauthnfail/bmqauthnfail_failauthenticator.cpp
+++ b/src/plugins/bmqauthnfail/bmqauthnfail_failauthenticator.cpp
@@ -28,6 +28,7 @@
 #include <bsl_memory.h>
 #include <bsl_optional.h>
 #include <bsl_string.h>
+#include <bsla_unused.h>
 #include <bslma_allocator.h>
 #include <bslma_managedptr.h>
 #include <bsls_assert.h>
@@ -93,9 +94,9 @@ bslstl::StringRef FailAuthenticator::mechanism() const
 }
 
 int FailAuthenticator::authenticate(
-    BSLS_ANNOTATION_UNUSED bsl::ostream&            errorDescription,
+    BSLA_UNUSED bsl::ostream&                       errorDescription,
     bsl::shared_ptr<mqbplug::AuthenticationResult>* result,
-    BSLS_ANNOTATION_UNUSED const mqbplug::AuthenticationData& input) const
+    BSLA_UNUSED const mqbplug::AuthenticationData& input) const
 {
     *result = bsl::allocate_shared<FailAuthenticationResult>(d_allocator_p,
                                                              "",
--- a/src/plugins/bmqauthnpass/bmqauthnpass_passauthenticator.cpp
+++ b/src/plugins/bmqauthnpass/bmqauthnpass_passauthenticator.cpp
@@ -28,6 +28,7 @@
 #include <bsl_memory.h>
 #include <bsl_optional.h>
 #include <bsl_string.h>
+#include <bsla_unused.h>
 #include <bslma_allocator.h>
 #include <bslma_managedptr.h>
 #include <bsls_assert.h>
@@ -93,9 +94,9 @@ bslstl::StringRef PassAuthenticator::mechanism() const
 }
 
 int PassAuthenticator::authenticate(
-    BSLS_ANNOTATION_UNUSED bsl::ostream&            errorDescription,
+    BSLA_UNUSED bsl::ostream&                       errorDescription,
     bsl::shared_ptr<mqbplug::AuthenticationResult>* result,
-    BSLS_ANNOTATION_UNUSED const mqbplug::AuthenticationData& input) const
+    BSLA_UNUSED const mqbplug::AuthenticationData& input) const
 {
     *result = bsl::allocate_shared<PassAuthenticationResult>(d_allocator_p,
                                                              "",


src/plugins/bmqauthnbasic/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# Basic Authenticator plugin
+# --------------------------
+
+if (NOT BMQ_TARGET_AUTHNBASIC_NEEDED )
+  return()
+endif()
+
+find_package( PkgConfig REQUIRED )
+
+bmq_add_plugin( bmqauthnbasic )
```
pniedzielski, line 10: The plugin should have `version` and `versiontag` components, to make sure that all static linking happens with the correct build.  See the prometheus plugin for an example.

src/plugins/bmqauthnbasic/bmqauthnbasic_basicauthenticator.h
```diff
@@ -0,0 +1,186 @@
...
+
+// ============================================================================
+//                             INLINE DEFINITIONS
+// ============================================================================
+
+// -----------------------------
+// class BasicAuthenticationData
+// -----------------------------
+
+inline const bsl::vector<char>& BasicAuthenticationData::authPayload() const
```
pniedzielski, line 159: Virtual methods can't be declared inline; this will result in the vtable being emitted in each translation unit, which in this case will likely cause an issue when the plugin is loaded.

src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -527,6 +541,7 @@ void Application::stop()
     STOP_OBJ(d_dispatcher_mp, "Dispatcher");
     STOP_OBJ(d_configProvider_mp, "ConfigProvider");
     STOP_OBJ(d_statController_mp, "StatController");
+    STOP_OBJ(d_authenticationController_mp, "AuthenticationController");
```
chrisbeard, line 544: The ordering here looks suspicious, I'd expect this to come before stat controller.

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include <bslstl_vector.h>
+#include <mqbauthn_authenticationcontroller.h>
+
+#include <mqbscm_version.h>
+// BMQ
+#include <bmqtsk_alarmlog.h>
+#include <bmqu_memoutstream.h>
+
+// MQB
+#include <bdlb_stringrefutil.h>
```
chrisbeard, line 26: nit: this is a bde lib

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.h
```diff
@@ -0,0 +1,105 @@
...
+    /// Used to instantiate 'Authenticator'
+    /// plugins at start-time.
+    mqbplug::PluginManager* d_pluginManager_p;
+
+    /// Registered authenticators
+    /// Mapping an authentication mechanism to a mqbplug::Authenticator
+    AuthenticatorMap d_authenticators;
+
+    /// Fallback principal
+    bsl::string d_principal;
```
chrisbeard, line 62: It looks like we are going with the concept that this is optional. If so, it's probably clearer for optionality to be reflected in the type here?

src/plugins/CMakeLists.txt
```diff
@@ -1 +1,4 @@
 add_subdirectory( "bmqprometheus" )
+add_subdirectory( "bmqauthnbasic" )
+add_subdirectory( "bmqauthnpass" )
+add_subdirectory( "bmqauthnfail" )
```
chrisbeard, line 4: As the pool of plugins expands, we might want a doc (maybe just a readme in `src/plugins`?) that gives an overview of the various plugins that come out of the box and whether they're intended to be used in production vs serve as examples.

src/groups/mqb/mqbplug/mqbplug_authenticator.h
```diff
@@ -138,9 +120,10 @@ class Authenticator {
     /// - Return a non-zero plugin-specific return code upon failure, and
     ///   populate the specified `errorDescription` with a brief reason for
     ///   logging purposes.
-    virtual int authenticate(bsl::ostream&             errorDescription,
-                             AuthenticationResult*     result,
-                             const AuthenticationData& input) const = 0;
+    virtual int
```
chrisbeard, line 123: Recently we discussed the potential pros of returning an enum here. Is that a TODO or did we change direction?

emelialei88: @678098 Evgeny made this comment: https://docs.google.com/document/d/1ggKVwu2bz7F79bNqMC2Z52Cv4m9oP1zEMHtTcU4RlUk/edit?disco=AAABiVZ0j-U

src/groups/mqb/mqbplug/mqbplug_authenticator.h
```diff
@@ -172,53 +155,42 @@ class AuthenticatorPluginFactory : public PluginFactory {
...
+    // STATIC CLASS METHODS
 
-inline const bsl::string& AuthenticationResult::principal() const
-{
-    return d_principal;
-}
+    /// Find authenticator config with the specified `name`
+    static const mqbcfg::AuthenticatorPluginConfig*
+    findAuthenticatorConfig(bslstl::StringRef name);
+};
```
chrisbeard, line 168: I know we used this pattern for [StatConsumer](https://github.com/bloomberg/blazingmq/blob/763b0330fae6c9d97cd49a4d522d85dc22af7699/src/groups/mqb/mqbplug/mqbplug_statconsumer.h#L138-L144), but maybe we can find a better approach.

src/groups/mqb/mqba/mqba_application.cpp
emelialei88: What's the reasoning? Is it because we expect `AuthenticationController` to be dependent on `StatController`?

emelialei88: What's the reasoning? Is it because we expect `AuthenticationController` to be dependent on `StatController`?

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.h
emelialei88: This might have not been finalized yet. Quote a comment from Evgeny @678098:
> We return int return code in many places, so it's fine to return it here. However, I think enums for RCs will be plugin-specific, so we can only define 0 value as "no error"

src/groups/mqb/mqbplug/mqbplug_authenticator.h
emelialei88: This might have not been finalized yet. Quote a comment from Evgeny @678098:
> We return int return code in many places, so it's fine to return it here. However, I think enums for RCs will be plugin-specific, so we can only define 0 value as "no error"

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+            // Add the authenticator into the collection
+            AuthenticatorMap::const_iterator cit = d_authenticators.find(
+                authenticator->mechanism());
+            if (cit != d_authenticators.cend()) {
+                errorDescription << "Attempting to create duplicate "
+                                    "authenticator with mechanism '"
+                                 << authenticator->mechanism();
+                return rc_DUPLICATE_MECHANISM;
+            }
```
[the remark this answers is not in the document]
emelialei88: I'm not fully sure I understand. We check `mechanism()` only after `start()` for the authenticator is called, and it should guarantee a valid result. Why do we want to call `mechanism()` before `start()`?

src/groups/mqb/mqbplug/mqbplug_authenticator.h
```diff
@@ -53,62 +53,44 @@ namespace mqbplug {
...
 
-    /// Return the corresponding data field.
-    const bsl::string&                       principal() const;
-    const bsl::optional<bsls::Types::Int64>& lifetimeMs() const;
-
-    // MANIPULATORS
-    /// Reset the state of this structure.
-    void reset();
+    /// Return the principal.
+    virtual const bsl::string& principal() const = 0;
```
emelialei88, line 90: Here, why don't we also return `bslstl::StringRef`?

src/groups/mqb/mqbplug/mqbplug_authenticator.h
emelialei88: Any alternative you have in mind?

pniedzielski (changes requested):
Here and below

pniedzielski (approved):

PR #741, opened 2025-05-08
Fix[bmqeval]: Fix or silence warnings in bmqeval

pniedzielski:
This PR contains a series of patches to fix or silence warnings in the bmqeval package.  Warnings for generated code cannot be fixed, so we silence them instead.

### DIFF
--- a/src/groups/bmq/bmqeval/bmqeval_simpleevaluatorparser.y
+++ b/src/groups/bmq/bmqeval/bmqeval_simpleevaluatorparser.y
@@ -13,6 +13,10 @@
 %define api.namespace { BloombergLP::bmqeval }
 %code requires
 {
+    #pragma clang diagnostic push
+    #pragma clang diagnostic ignored "-Wsuggest-destructor-override"
+    #pragma clang diagnostic ignored "-Wswitch-enum"
+
     #include <bsl_functional.h>
     #include <string>
     #include <bmqeval_simpleevaluator.h>
@@ -177,3 +181,5 @@ expression
 void SimpleEvaluatorParser::error(const std::string& message) {
     ctx.d_os << message << " at offset " << scanner.lastTokenLocation();
 }
+
+#pragma clang diagnostic pop
--- a/src/groups/bmq/bmqeval/bmqeval_simpleevaluatorparser.y
+++ b/src/groups/bmq/bmqeval/bmqeval_simpleevaluatorparser.y
@@ -75,6 +75,7 @@
 %token PLUS "+" MINUS "-";
 %token TIMES "*" DIVIDES "/" MODULUS "%";
 %token EQ "=" NE "<>" LT "<" LE "<=" GE ">=" GT ">";
+%token NOT "!";
 %token <bsl::string> EXISTS "exists";
 
 %left OR;
@@ -83,7 +84,7 @@
 %left LT LE GE GT;
 %left PLUS MINUS;
 %left TIMES DIVIDES MODULUS;
-%right NOT "!";
+%precedence NOT;
 
 %type<bsl::string> variable;
 %type<SimpleEvaluator::ExpressionPtr> expression;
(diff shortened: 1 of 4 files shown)


src/groups/bmq/bmqeval/bmqeval_simpleevaluatorparser.y
```diff
@@ -13,6 +13,10 @@
 %define api.namespace { BloombergLP::bmqeval }
 %code requires
 {
+    #pragma clang diagnostic push
```
kaikulimu, line 16: Is it fine that `#pragma clang diagnostic push` is inside the `%code requires {}` block but the `#pragma clang diagnostic pop` at the end is not?

src/groups/bmq/bmqeval/bmqeval_simpleevaluatorparser.y
```diff
@@ -75,6 +75,7 @@
 %token PLUS "+" MINUS "-";
 %token TIMES "*" DIVIDES "/" MODULUS "%";
 %token EQ "=" NE "<>" LT "<" LE "<=" GE ">=" GT ">";
+%token NOT "!";
```
kaikulimu, line 78: Can @jll63 and @678098 take a look and make sure the changes at Line 78, 87, and 172 are not breaking anything?

src/groups/bmq/bmqeval/bmqeval_simpleevaluatorparser.y
pniedzielski: I think so.  When I look at the generated parser code, these two pragmas end up in correct locations, in the correct order.

src/groups/bmq/bmqeval/bmqeval_simpleevaluatorparser.y
```diff
@@ -83,7 +84,7 @@
 %left LT LE GE GT;
 %left PLUS MINUS;
 %left TIMES DIVIDES MODULUS;
-%right NOT "!";
+%precedence NOT;
```
jll63, line 87: Not sure about this. Why not leave associativity to `%right`? I think that the change makes consecutive NOTs an error, but there is no reason for that. `!!(a>10)` is legal. Do we have unit tests with `!!`? If not we should add them, because we use the same precedences and associativities as [in C](https://en.cppreference.com/w/c/language/operator_precedence).

Since we're at it, precedence of unary minus should be higher than logical not, and we should add unary plus (Ankur was always pumping the brakes on adding operators - e.g. it was a struggle to get modulus past him).

678098: https://github.com/bloomberg/blazingmq/blob/b47f2d7f94bbb0f81988b34113dc9b48c173e246/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.t.cpp#L148
We only have this one, but this test case is an expected error due to the expression length. We don't have other tests with multiple `!`

pniedzielski: I'll add a unit test for `!!(a>10)` and the spellings with `~` too, that's a good catch.  It looks like we're also never testing the precedence of NOT vs. AND and OR, so I'll add those tests.

I'll shift the precedence of unary arithmetic operators in another PR?

This line was flagged by turning warnings on in Bison, which says it does not make sense to define an associativity for this operator given our grammar, since there's no shift-reduce conflict after shifting a NOT.  No shift-reduce conflict, so no place for associativity to enter into the picture.  Even cases like `!x!=y` will have no ambiguity, as this is already tokenized to `NOT x NE y` by the lexer.

Instead of associativity, the reason `!!(a>10)` will parse to `(NOT (NOT (NE a 10)))` is by virtue of our [unary negation production](https://github.com/bloomberg/blazingmq/blob/142da16decb19f02137b5c07c0e31ca90b3ddfa6/src/groups/bmq/bmqeval/bmqeval_simpleevaluatorparser.y#L159).  We shift a `NOT`, and the only thing we can do next is shift another another `NOT`, then shift and reduce through `(a > 10)` to get an `expression`.  Using this production, we reduce `NOT expression` to `expression`, and then reduce one more time `NOT expression` to `expression`, parsing the whole expression.  No shift-reduce conflict involved, so no need for associativity.

pniedzielski: @jll63 and @678098, commit 9ca9bbc contains the tests mentioned above, and they pass.

kaikulimu (approved):
LGTM. Thanks for adding more tests.

jll63 (approved):

PR #744, opened 2025-05-12
CI: 

bbpetukhov:
**Describe your changes**
Added a new workflow to be run manually. It builds 2 versions of BlazingMQ broker binaries and runs a new batch of integration tests which covers broker binary upgrade via stop/restart.

The build job is reused between existing Tests workflow and a new manually run one.

image

**Testing performed**
- Existing Tests workflow passed, i.e. build, UTs and ITs passed
- New manual workflow passed

**Additional context**
n/a


kaikulimu:
#693 has been approved. Feel free to rebase on top of that.

kaikulimu:
Note: https://github.com/bloomberg/blazingmq/issues/761 is a related issue.

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,136 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref }}
+          fetch-depth: 0
+      
+      - name: Get commit SHA
+        id: get_sha
+        run: echo "sha=$(git rev-parse HEAD)" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
bbpetukhov: inputs.ref can be either commit hash, branch name or tag.
This step resolves inputs.refs to commit hash

.github/workflows/manual.yaml
```diff
@@ -0,0 +1,234 @@
...
+          BLAZINGMQ_LEADER_NAME: "${{ github.event.inputs.leader_name }}"
+
+          # Broker default path for proxy nodes
+          BLAZINGMQ_BROKER: "${{ github.workspace }}/repo1/build/blazingmq/src/applications/bmqbrkr/bmqbrkr.tsk"
+          BLAZINGMQ_TOOL: "${{ github.workspace }}/repo1/build/blazingmq/src/applications/bmqtool/bmqtool.tsk"
+
+          BLAZINGMQ_BROKER_EAST1: "${{ github.workspace }}/repo1/build/blazingmq/src/applications/bmqbrkr/bmqbrkr.tsk"
+          BLAZINGMQ_BROKER_EAST2: "${{ github.workspace }}/repo2/build/blazingmq/src/applications/bmqbrkr/bmqbrkr.tsk"
+          BLAZINGMQ_BROKER_WEST1: "${{ github.workspace }}/repo2/build/blazingmq/src/applications/bmqbrkr/bmqbrkr.tsk"
+          BLAZINGMQ_BROKER_WEST2: "${{ github.workspace }}/repo1/build/blazingmq/src/applications/bmqbrkr/bmqbrkr.tsk"
```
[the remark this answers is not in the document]
bbpetukhov: Good idea. Thank you

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,136 @@
+name: Build Ubuntu
+
+description: |
+  This is a reusable workflow to build BlazingMQ on Ubuntu. 
+  Optionally run unit tests if the target is "all.t".
```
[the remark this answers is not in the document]
bbpetukhov: @kaikulimu do you think it's still worth it to extract running UTs from the build workflow?

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,136 @@
...
+            bison \
+            libfl-dev \
+            libbenchmark-dev \
+            libgmock-dev \
+            libz-dev
+
+      - name: Install cached non packaged dependencies
+        if: steps.build-cache-restore-step.outputs.cache-hit != 'true'
+        working-directory: deps
+        run: ../docker/build_deps.sh
```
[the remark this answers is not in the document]
kaikulimu: `dependencies.yaml` calls:
```
sudo apt-get update
sudo apt-get install -qy build-essential \
    gdb \
    ...
...
../docker/build_deps.sh
```

It seems you are doing the exact same thing here. The code here confuses me. Perhaps we could use a Zoom call.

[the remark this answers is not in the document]
kaikulimu: We should document `dependencies.yaml` properly too.

.github/workflows/manual.yaml
```diff
@@ -0,0 +1,237 @@
...
+    uses: ./.github/workflows/build-ubuntu.yaml
+    with:
+      ref: ${{ github.event.inputs.ref2 }}
+      target: "bmqbrkr bmqtool"
+
+  integration_tests_ubuntu_upgrade_version:
+    name: IT [${{ matrix.cluster }}/${{ matrix.mode }}/${{ matrix.consistency }}]
+    strategy:
+      matrix:
+        mode: ["legacy_mode", "fsm_mode"]
```
kaikulimu:
```suggestion
        mode: ["legacy_mode"]
```
Turn off FSM mode for now, as it is not ready to run with upgrading brokers yet

.github/workflows/manual.yaml
```diff
@@ -0,0 +1,237 @@
...
+          name: ${{ github.run_id }}_failure_logs_${{ env.RUN_UID }}
+          path: ${{ github.workspace }}/repoit/src/integration-tests/failure-logs
+          retention-days: 5
+
+
+  integration_tests_ubuntu_two_versions_in_cluster:
+    name: IT [${{ matrix.cluster }}/${{ matrix.mode }}/${{ matrix.consistency }}]
+    strategy:
+      matrix:
+        mode: ["legacy_mode", "fsm_mode"]
```
kaikulimu:
```suggestion
        mode: ["legacy_mode"]
```
Turn off FSM mode for now, as it is not ready to run with upgrading brokers yet

kaikulimu:
https://github.com/bbpetukhov/blazingmq/actions/runs/15193733870/job/42733259905
This failure from running Dmitrii's broker upgrade IT: The assert is from FSM-only code, but somehow it shows up in `IT [multi/legacy_mode/strong_consistency]` step

kaikulimu:
When I test ran with [BMQBRKR_0.92.8](https://github.com/bloomberg/blazingmq/releases/tag/BMQBRKR_0.92.8) and [BMQBRKR_0.93.18](https://github.com/bloomberg/blazingmq/releases/tag/BMQBRKR_0.93.18) with their SHAs, I got https://github.com/bbpetukhov/blazingmq/actions/runs/15285095339/job/42993360371, where only `[IT [multi3/legacy_mode/eventual_consistency]](https://github.com/bbpetukhov/blazingmq/actions/runs/15285095339/job/42993360373#logs)` and `[IT [multi3/legacy_mode/strong_consistency]](https://github.com/bbpetukhov/blazingmq/actions/runs/15285095339/job/42993360386#logs)` pass, and everything else failed.

That is okay!

The two tests that actually pass are what we actually want to achieve in this first step of manually testing 2 builds. Here are my two cents on the failed tests:
- [IT [multi3/fsm_mode/eventual_consistency]](https://github.com/bbpetukhov/blazingmq/actions/runs/15285095339/job/42993360390#logs): FSM mode is not ready for testing 2 builds yet. This is okay. 
- [IT [multi3/fsm_mode/strong_consistency]](https://github.com/bbpetukhov/blazingmq/actions/runs/15285095339/job/42993360380#logs): FSM mode is not ready yet.
- [IT [multi/legacy_mode/eventual_consistency]](https://github.com/bbpetukhov/blazingmq/actions/runs/15285095339/job/42993360371#logs)
- [IT [multi/legacy_mode/strong_consistency]](https://github.com/bbpetukhov/blazingmq/actions/runs/15285095339/job/42993360374#logs): There is exactly one test with assert: `test_restart_from_non_FSM_to_FSM[multi_node-strong_consistency].log`. The assert is `test_restart_from_non_FSM_to_FSM[multi_node-strong_consistency].log`, which will be fixed by https://github.com/bloomberg/blazingmq/pull/661 at Commit https://github.com/bloomberg/blazingmq/pull/661/commits/ee5de9812df62b67a3bab97db23d77d2fe287c45
- [IT [multi/fsm_mode/eventual_consistency]](https://github.com/bbpetukhov/blazingmq/actions/runs/15285095339/job/42993360377#logs): FSM mode is not ready yet.
- [IT [multi/fsm_mode/strong_consistency]](https://github.com/bbpetukhov/blazingmq/actions/runs/15285095339/job/42993360391#logs): FSM mode is not ready yet.

bbpetukhov:
> @bbpetukhov Why when you run manual tests such as https://github.com/bbpetukhov/blazingmq/actions/runs/14976029383, it only triggered two ITs? But when I run manual tests such as https://github.com/bbpetukhov/blazingmq/actions/runs/15285095339, it triggered eight ITs?

My run happen 2 weeks ago and then only fsm mode and only strong consistency were enabled. Since then we enabled legacy mode and eventual consistency thus multiplying number of test runs by factor of 2 twice.

.github/workflows/build-ubuntu.yaml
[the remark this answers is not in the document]
bbpetukhov: Added params save_build_as_artifacts and run_unit_tests:
https://github.com/bbpetukhov/blazingmq/commit/baf55c222c268edc8bd80bb2a790439a0c393ac5

kaikulimu:
Looking 92.8 vs 93.18 test has most test failures in common with 92.6 vs 92.8 -- mostly strong consistency tests in `test_graceful_shutdown.py` and `test_puts_retransmission.py`. We can deduce ITs have been failing quite frequently on 92.x brokers, and upgrading to 93.x would fix most of them.

On 93.15 vs 93.18 test, there was a broker assert in `test_appids.py::test_csl_repair_after_stop[multi_node-strong_consistency]` and a test failure in `test_maxunconfirmed.py::test_stuck_downstream[multi_node-strong_consistency]`. Those we need to be more concerned about.

kaikulimu:
One weakness I find is testing multiple broker versions together is that already-fixed failures creep in as false alarms in the IT. See my comment above to see how many of those false alarms pollute our tests..

kaikulimu (commented):
Some comments regarding turning off FSM flags.

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,157 @@
...
+
+on:
+  workflow_call:
+    inputs:
+      ref:
+        description: "Ref to build"
+        required: true
+        type: string
+      target:
+        description: "Build target"
```
kaikulimu:
```suggestion
        description: "Build targets, space-delimited"
```

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,157 @@
...
+  This is a reusable workflow to build BlazingMQ on Ubuntu. 
+  Optionally run unit tests if the target is "all.t".
+  Given a ref and a target, this workflow will build the project and will save the build results both as artifact and in cache.
+  The cache key is generated based on the ref and the target, and is returned via workflow output.
+
+on:
+  workflow_call:
+    inputs:
+      ref:
+        description: "Ref to build"
```
kaikulimu:
```suggestion
        description: "Branch name, tag, or commit SHA to build"
```

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,157 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref }}
+          fetch-depth: 0
+      
+      # Since input.ref can be a branch, tag or commit SHA,
+      # we need to get the commit SHA to use it in the cache key.
+      - name: Get commit SHA
+        id: get_sha
```
kaikulimu:
```suggestion
        id: get_sha
        if: inputs.save_build_as_artifacts == true
```
Do we need this step if we are not saving build?

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,157 @@
...
+        run: echo "sha=$(git rev-parse HEAD)" >> $GITHUB_ENV
+
+      - name: Generate cache key
+        id: get_cache_key
```
kaikulimu, line 61:
```suggestion
        id: get_cache_key
        if: inputs.save_build_as_artifacts == true
```
Similarly, do we need this step if we are not saving build?

.github/workflows/build-ubuntu.yaml
kaikulimu: Also, might be good idea to combine `id: get_sha` and `id: get_cache_key` into one step, if `id: get_cache_key` is the only user of `env.sha`

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,157 @@
...
+        description: "Build target"
+        required: true
+        type: string
+      save_build_as_artifacts:
+        description: "Whether to save the build as artifacts"
+        required: true
+        type: boolean
+        default: true
+      run_unit_tests:
+        description: "Whether to run unit tests"
```
kaikulimu:
```suggestion
        description: "Whether to run unit tests.  Can only be set to true if `all.t` is one of the build targets"
```

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,157 @@
...
+          cmake --build build/blazingmq --parallel 8 --target ${{ inputs.target }}
+
+      - name: Clean-up build directories before caching
+        if: steps.build-cache-restore-step.outputs.cache-hit != 'true' # Variable type is string, thus using quotes
+        run: |
+          find . -name "*.o" -type f -delete
+          find . -name "*.a" -type f -delete
+
+      - name: Run UT [c++]
+        if: inputs.target == 'all.t' && inputs.run_unit_tests == true # Variable type is boolean, thus no quotes
```
kaikulimu: `inputs.target` containing `all.t` should be sufficient, such as `inputs.target = "bmqbrkr bmqtool all.t"`

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,157 @@
...
+        run: |
+          cd ${{ github.workspace }}/build/blazingmq
+          ctest --output-on-failure -L \^unit\$
+
+      - name: Fail if requrested to run UTs, they were not built
+        if: inputs.target != 'all.t' && inputs.run_unit_tests == true # Variable type is boolean, thus no quotes
```
kaikulimu: `inputs.target` not containing `all.t` should be the condition

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,157 @@
...
+        run: |
+          echo "Unit tests were requested to run, but the target is not 'all.t'."
+          echo "Please set the target to 'all.t' to build unit tests."
```
kaikulimu:
```suggestion
          echo "Please set the target to include 'all.t' to build unit tests."
```

src/python/blazingmq/dev/it/fixtures.py
```diff
@@ -636,6 +630,65 @@ def multi_node(request):
     yield from cluster_fixture(request, request.param)
 
 
+###############################################################################
+# multi3_node cluster
```
kaikulimu: Let's change from `multi3` cluster to `multi6` cluster. In production, we do rolling upgrades to 6-node clusters, so let's modify the IT to use 6-node cluster to better reflect reality. Can be [`east1`, `east2`, `east3`, `west1`, `west2`, `west3`]

kaikulimu (changes requested):
We are close! Should be able to merge, after addressing this feedback (and turning off FSM flags).

src/python/blazingmq/dev/it/fixtures.py
bbpetukhov: The thing is we do already have cluster multi having 4 nodes. I thought in production we should be using odd number of nodes in a cluster to uniquely determine majority while voting during leader election. That's why I created a new cluster fixture with 3 nodes to better represent real-life use-cases.

If having even number of nodes (i.e. 4) is fine, we may just keep using "multi" fixture.

.github/workflows/build-ubuntu.yaml
bbpetukhov: Done

.github/workflows/build-ubuntu.yaml
bbpetukhov: Done

.github/workflows/build-ubuntu.yaml
bbpetukhov: This condition
> if: inputs.save_build_as_artifacts == true
may not be correct, since the cache is save always. For example if running UTs and not uploading build results as artifacts, the cache is still created to be used if the job is run again. However this may be redundant.

src/python/blazingmq/dev/it/fixtures.py
kaikulimu: The truth is that most of our production clusters today are 6-nodes. In the future, we will move towards a 7-node cluster across four data centers architecture -- where there will be [2, 2, 2, 1] nodes in each data center respectively. If you want to better prepare for the future, we can do:

```
east1, east2, west1, west2, north1, north2, south1
```

kaikulimu: Also see my private message on examples of present 6-node clusters and futuristic 7-node clusters. We cannot share the names of the cluster nodes here due to business security reasons.

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,157 @@
...
+      - name: Run UT [c++]
+        if: contains(inputs.target, 'all.t') && inputs.run_unit_tests == true # Variable type is boolean, thus no quotes
+        run: |
+          cd ${{ github.workspace }}/build/blazingmq
+          ctest --output-on-failure -L \^unit\$
+
+      - name: Fail if requrested to run UTs, they were not built
+        if: contains(inputs.target, 'all.t') == false && inputs.run_unit_tests == true # Variable type is boolean, thus no quotes
+        run: |
+          echo "Unit tests were requested to run, but the target is not 'all.t'."
```
kaikulimu:
```suggestion
          echo "Unit tests were requested to run, but the target does not include 'all.t'."
```

.github/workflows/build-ubuntu.yaml
kaikulimu: I thought the reason to skip uploading artifacts is because UTs generate ~2GB of build artifacts which is too large. By the same logic, shouldn't we skip saving cache too? It seems weird that we are not okay uploading artifacts but okay with saving cache. 

Can you explain the reasoning for the difference?

src/python/blazingmq/dev/it/fixtures.py
bbpetukhov: Ok, so shall we update straight to 7-node cluster?

kaikulimu: Okay, let's update to 7-node cluster

bbpetukhov: Done

.github/workflows/build-ubuntu.yaml
bbpetukhov: Generally artifacts and cache are used here for different purposes. 
Build results saved to artifacts to be later used in other jobs (e.g. used in ITs).
Storing build results to cache helps to speed-up rerun of the workflow for the same commit sha.

However you are right, indeed we moved running UTs into this workflow to avoid storing huge artifacts. Thus we may also skip storing huge build results in cache to optimize storage usage (even though total cache storage is limited to 10gb)

bbpetukhov: Also merged 2 steps into single "Generate cache key from commit SHA"

src/integration-tests/test_redeploy.py
```diff
@@ -0,0 +1,213 @@
...
+    # Post and receive final message after all nodes have been redeployed
+    messagesCounter.post_message(producer, uri_priority)
+    messagesCounter.assert_posted(consumer, uri_priority)
+
+
+@start_cluster(start=True, wait_leader=True, wait_ready=True)
+def test_redeploy_one_by_one(multi7_node: Cluster, domain_urls: tc.DomainUrls):
+    """
+    Test to upgrade binaries of cluster nodes one by one.
+    Every time a node is upgraded, only this node is restarted.
```
kaikulimu, line 205: Can I suggest a revision before I review this fully?
```suggestion
    Every time a node is upgraded, this node is restarted, then every other node is restarted (but not at the same time).
```
This is closer to what we do in production. We upgrade one node; this node turns over the weekend, and other nodes also turn over the weekend. Next week, we upgrade one more node, and nodes slowly turn over the weekend.

So, with a N-node cluster, we will restart N^2 times.

bbpetukhov: I think this scenario is addressed in another test `test_redeploy_whole_cluster_restart` where broker version upgrade is applied to the nodes one by one and after each node upgrade the whole cluster (all nodes) is restarted.

src/integration-tests/test_redeploy.py
```diff
@@ -0,0 +1,213 @@
...
+    Test to upgrade binaries of cluster nodes one by one.
+    Every time a node is upgraded, all the nodes are restarted.
+    """
+
+    # Ensure the leader is not changing during cluster restart.
+    # Otherwise cluster gets stuck on leader reelection (in legacy mode)
+    QUORUM_TO_ENSURE_NOT_LEADER = 100
+    for node in multi7_node.nodes():
+        if node is not multi7_node.last_known_leader:
+            node.set_quorum(QUORUM_TO_ENSURE_NOT_LEADER)
```
kaikulimu: Is this really necessary? Also, note that setting quorum only persists until the node restarts. In this case, we are restarting the cluster 7 times, but your `QUORUM_TO_ENSURE_NOT_LEADER` should only affect the first restart.

bbpetukhov: While I was investigating test failures, I discovered that the reason of test timeouts is that cluster can not select a leader: it first selects one, than another and then Unknown leader, so finally test fails by timeout.

Helping cluster to select a leader (predefining it beforehand) helped to fix the test.

bbpetukhov: I thought the quorum value is preserved after a node is started. If it's not the case, maybe it's better to preset it on every restart.

kaikulimu: If you change the quorum value in the config file, it is persisted after restart. If you send an admin command to change quorum, such as [`Broker::set_quorum`](https://github.com/bloomberg/blazingmq/blob/30f913022f282ffe520477b70fe89397ebc7f58d/src/python/blazingmq/dev/it/process/broker.py#L238, it is *not* persisted after restart.

bbpetukhov: I implemented presetting the leader to be 'east1' always during whole test run. Defining the leader is done via changing its config file (setting quorum to 1) and setting quorum for all other nodes to 7 (i.e. cluster size).

kaikulimu: Your workflow has a field `Leader name for ITs (auto selected if empty)`. Should it be the selected value if chosen instead of `east1`?

bbpetukhov: Agreed to remove the field `Leader name for ITs (auto selected if empty)`. 
Done.

src/integration-tests/test_redeploy.py
```diff
@@ -0,0 +1,235 @@
...
+
+    for broker in multi7_node.nodes():
+        # Stop all nodes
+        multi7_node.stop_nodes(prevent_leader_bounce=True)
+
+        # Update binary for the given broker
+        multi7_node.update_broker_binary(broker.config, NEW_VERSION_SUFFIX)
+
+        # Restart all nodes to apply binary update
+        multi7_node.start_nodes(wait_leader=True, wait_ready=False)
```
kaikulimu, line 190:
```suggestion
        # Update binary for the given broker
        multi7_node.update_broker_binary(broker.config, NEW_VERSION_SUFFIX)
        
        # Restart the given broker to apply binary update
        broker.stop()
        broker.start()
        
        # Restart every other node one by one
        for node in nodes:
            node.stop()
            node.start()
```

kaikulimu: Try if this modification fixes the current failures. The biggest difference is that the whole cluster is never entirely stopped at any time.

bbpetukhov: Today I tried this approach. First of all to make it work, I had to add extra waiting commands:

```
        broker.stop()
        multi7_node.make_sure_node_stopped(broker)
        broker.start()
        broker.wait_until_started()
```

However the test still does not pass because of many "STORAGE_OUT_OF_SYNC" errors which lead to broker abort, so the rest of nodes can not get quorum to elect a new leader and get the cluster to the healthy state.

I plan to try to implement a restart mechanism. So if a node is aborted due to partition inconsistency, it may restart and try to sync again.

bbpetukhov: After performing exhausting testing I am assured that the integration test "test_redeploy_whole_cluster_restart" is finally stable.
The most important change is done in the config file:
https://github.com/bbpetukhov/blazingmq/blob/refs/heads/manual-ci/src/integration-tests/test_redeploy.py#L60
setting startupWaitDurationMs to a small value for one of the nodes, so it becomes the single source of truth and prevent races during partitions recovery.

After branch rebase the new workflows passes:
https://github.com/bbpetukhov/blazingmq/actions/runs/16493242052

bbpetukhov: Also another integration test "test_redeploy_one_by_one" where the nodes are restarted one by one does not need any tuning. I've just removed redundant waiting for events (i.e. leader election) which are not expected to happen.

kaikulimu (approved):
lgtm

PR #745, opened 2025-05-12
Docs[mqbc_clusterutil.cpp]: better logs and docs

678098:


src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -1390,7 +1390,7 @@ void ClusterUtil::updateAppIds(ClusterData*                    clusterData,
                        << queueAdvisory << ", rc: " << rc;
     }
     else {
-        BALL_LOG_INFO << "Unregister appIds " << printRemoved
+        BALL_LOG_INFO << "Unregistered appIds " << printRemoved
                       << " and registered appIds " << printAdded << " for '"
```
chrisbeard, line 1394: `ledger->apply()` is an async update, so unsure what is best to print here. Also, the `uri` logged at the end of this log might be empty, which is not intuitive. Maybe that should print domain or uri, whichever was used?

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019-2023 Bloomberg Finance L.P.
+// Copyright 2025 Bloomberg Finance L.P.
```
kaikulimu, line 1: 2019-2025 to be consistent

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -1390,9 +1390,18 @@ void ClusterUtil::updateAppIds(ClusterData*                    clusterData,
...
-                      << uri << "'.";
+        BALL_LOG_INFO_BLOCK
+        {
+            BALL_LOG_OUTPUT_STREAM << "Advisory applied: unregister appIds "
+                                   << printRemoved << " and register appIds "
+                                   << printAdded << " for ";
+            if (!uri.empty()) {
+                BALL_LOG_OUTPUT_STREAM << " uri = [" << uri << "]";
+            }
+            if (!domainName.empty()) {
```
chrisbeard, line 1401: This should always be available/non-empty. Looks like `uri` is the only optional one (not a big fan of this API, but that's a separate discussion).

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
kaikulimu: 2014-03-07 is the date of our first commit. So sure, 2014-[currentYear] would work.

kaikulimu (approved):

678098:
The IT failure is unrelated

PR #746, opened 2025-05-13
Feat[MQB]: add authentication with basic logic

emelialei88:
Handing authentication logic on the broker side:
- Added authentication protocol.
- Introduced a new `AuthenticationEvent` event type.
- Implemented the authenticator component to handle authentication logic, using `AuthenticationContext` to encapsulate relevant context data.
- Updated `InitialConnectionHandler` to handle the authentication flow.

At this stage, we only handle incoming authentication requests and always allow them to succeed. Outbound, reverse authentication requests initiated by a broker, and re-authentication are not supported.

We also updated the unit test `m_bmqstoragetool_cslprinter.t` to align with changes made by `bascodegen`.


src/groups/bmq/bmqp/bmqp_ctrlmsg.xsd
```diff
@@ -1755,4 +1755,68 @@
...
+        is attempting to authenticate with the supplied authentication material
+        to initiate a session.
+
+        mechanism.: The authentication mechanism the client intends to use
+        data......: The client's authentication material
+      </documentation>
+    </annotation>
+    <sequence>
+      <element name='mechanism' type='string'/>
+      <element name='data'      type='hexBinary'  minOccurs='0'/>
```
chrisbeard, line 1799: @678098 brought up using `base64Binary` instead, we should consider whether that's the better option.

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,199 @@
...
+{
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(
+        context->d_authenticationMessage.isAuthenticateRequestValue());
+    BSLS_ASSERT_SAFE(context->d_initialConnectionContext_p->isIncoming() ||
+                     context->d_isReversed);
+
+    bmqp_ctrlmsg::AuthenticateRequest& authenticateRequest =
+        context->d_authenticationMessage.authenticateRequest();
+
```
emelialei88, line 59: As @678098 suggested, we might want to enforce the encoding mechanism. Should we add a function for format checking?

src/groups/bmq/bmqp/bmqp_ctrlmsg.xsd
```diff
@@ -1755,4 +1755,68 @@
...
+
+        status.....: Status of the request
+        lifetimeMs.: The duration (in milliseconds) the client's session is
+                     valid for. The client must re-authenticate before the
+                     lifetime expires or it will be disconnected.
+      </documentation>
+    </annotation>
+    <sequence>
+      <element name='status'      type='tns:Status' />
+      <element name='lifetimeMs'  type='long'   minOccurs='0'/>
```
pniedzielski, line 1818: What is the expected behavior when this value is not in the `AuthenticateResponse`?

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,199 @@
...
+// limitations under the License.
+
+// mqba_authenticator.h                                           -*-C++-*-
+#include <mqba_authenticator.h>
+
+#include <mqbscm_version.h>
+
+/// Implementation Notes
+///====================
+/// TODO
```
pniedzielski, line 23: `TODO`, but let's make sure to document this as it gets more complicated.  The authentication workflow is pretty simple, but let's do our best to describe how it goes.

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,199 @@
...
+// -------------------
+// class Authenticator
+// -------------------
+
+int Authenticator::onAuthenticationRequest(
+    bsl::ostream&                  errorDescription,
+    const AuthenticationContextSp& context)
+{
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(
```
pniedzielski, line 52: Fine to have preconditions, but they aren't preconditions if they're not documented in the function contract: `The behavior of this function is undefined unless ...`.  Otherwise, how would a caller of this function know it will crash if `!context->d_initialConnectionContext_p->isIncoming() && !context->d_isReversed`?  We'd be lying to them.

Second concern: are these preconditions checked or enforced by the caller?  I think they are, but the second one below has me worried.  Is there any way we could get a bad message over the wire from a client and crash?

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,199 @@
...
+        context->d_authenticationMessage.isAuthenticateRequestValue());
+    BSLS_ASSERT_SAFE(context->d_initialConnectionContext_p->isIncoming() ||
+                     context->d_isReversed);
+
+    bmqp_ctrlmsg::AuthenticateRequest& authenticateRequest =
+        context->d_authenticationMessage.authenticateRequest();
+
+    BALL_LOG_DEBUG
```
pniedzielski, line 60: I am a little concerned about logging arbitrary data that we get from the network, but I think the `operator<<` for `authenticateRequest` as generated is fine with this.  This is another place where `base64` may help compress the logs.

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    if (rc == rc_SUCCESS) {
+        *isContinueRead = true;
+    }
+
+    return rc;
+}
+
+int Authenticator::authenticationOutboundOrReverse(
```
pniedzielski, line 190: The reverse will be tricky, I think...

src/groups/mqb/mqba/mqba_authenticator.h
```diff
@@ -0,0 +1,168 @@
...
+    // NOT IMPLEMENTED
+
+    /// Copy constructor and assignment operator not implemented.
+    Authenticator(const Authenticator&);             // = delete
+    Authenticator& operator=(const Authenticator&);  // = delete
+
+  private:
+    // PRIVATE MANIPULATORS
+
+    /// Invoked when received a `ClientIdentity` authentication message with
```
pniedzielski, line 99: "Invoked when a `ClientIdentity` authentication message is received with the specified `context`"

Same below

src/groups/mqb/mqba/mqba_authenticator.h
```diff
@@ -0,0 +1,168 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// mqba_authenticator.h                                           -*-C++-*-
+#ifndef INCLUDED_MQBA_AUTHENTICATOR
+#define INCLUDED_MQBA_AUTHENTICATOR
+
+/// @file mqba_authenticator.h
+///
+/// @brief
```
pniedzielski, line 22: I think this documentation comment will be the most helpful of all of them.  Fine to have it blank as you're working but I think this is going to be a treasure-trove of useful info.  At least, `@brief` will be good, so the libmqb documentation has a description for this component.

src/groups/mqb/mqbnet/mqbnet_authenticator.h
```diff
@@ -0,0 +1,64 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// mqbnet_authenticator.h -*-C++-*-
+#ifndef INCLUDED_MQBNET_AUTHENTICATOR
+#define INCLUDED_MQBNET_AUTHENTICATOR
+
+//@PURPOSE:
```
pniedzielski, line 20: Component docs are important here in this file too.  Let's use the Doxygen format, like the public libbmq libraries we've already converted over.

src/groups/mqb/mqbnet/mqbnet_authenticator.h
```diff
@@ -0,0 +1,64 @@
...
+//
+//@CLASSES:
+//
+//@DESCRIPTION:
+
+// MQB
+#include <mqbnet_authenticationcontext.h>
+
+// BDE
+#include <bsl_memory.h>
```
pniedzielski, line 30: `#include <bsl_iostream.h>` as well because of `bsl::ostream` below.

dorjesinpo (commented):
We need to at least document/comment how are we going to enforce authentication.  Is `AuthenticationContext` going to have a state/status that someone (`InitialConnectionHandler`? `ClientSession`?) is going to check?

src/groups/mqb/mqba/mqba_authenticator.cpp
emelialei88: I'm not sure I understand the second concern
`isIncoming` is set by us when `TcpSessionFactory` receives an incoming call. `d_isReversed` is default to be true unless we receive an reversed connection.

src/groups/mqb/mqba/mqba_authenticator.cpp
emelialei88: I could get rid of printing the data, but sometime maybe we just want to debug and see what it is. Is setting it to debug mode still not okay?

src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
```diff
@@ -231,25 +246,40 @@ int InitialConnectionHandler::decodeInitialConnectionMessage(
...
-        errorDescription << "Invalid negotiation message received (failed "
-                         << "decoding ControlEvent): [rc: " << rc << "]:\n"
-                         << bmqu::BlobStartHexDumper(&blob);
-        return rc_INVALID_CONTROL_EVENT;  // RETURN
+        *authenticationMsg = authenticationMessage;
     }
+    else if (event.isControlEvent()) {
+        const int rc = event.loadControlEvent(&negotiationMessage);
+        if (rc != 0) {
+            BALL_LOG_ERROR << "Invalid response from broker [reason: 'control "
```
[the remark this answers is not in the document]
emelialei88: > And we do need to log/dump the blob
Is this a question? Do you mean we need to keep the error log of event or I should log even there's not an error?

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+    // Always succeeds for now
+    // TODO: For later implementation, plugins will perform authentication,
+    // taking the `AuthenticationContext` and updates it with the
+    // authentication result.
+    response.status().category() = bmqp_ctrlmsg::StatusCategory::E_SUCCESS;
+    response.status().code()     = 0;
+    response.lifetimeMs()        = 10 * 60 * 1000;
+
+    authenticationContext->state().testAndSwap(
```
dorjesinpo, line 97: checking return value would look good

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,228 @@
...
+    int                rc = rc_SUCCESS;
+
+    switch (authenticationMsg.selectionId()) {
+    case bmqp_ctrlmsg::AuthenticationMessage::
+        SELECTION_ID_AUTHENTICATE_REQUEST: {
+        rc = onAuthenticationRequest(errStream, authenticationMsg, context);
+    } break;  // BREAK
+    case bmqp_ctrlmsg::AuthenticationMessage::
+        SELECTION_ID_AUTHENTICATE_RESPONSE: {
+        rc = onAuthenticationResponse(errStream, authenticationMsg, context);
```
dorjesinpo, line 202: Will there be a case when brokers receive AuthenticationResponse?  If so, need a comment.

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+    authenticationContext->state().testAndSwap(
+        mqbnet::AuthenticationContext::State::e_AUTHENTICATING,
+        mqbnet::AuthenticationContext::State::e_AUTHENTICATED);
+
+    int rc = sendAuthenticationMessage(errorDescription,
+                                       authenticationResponse,
+                                       authenticationContext);
+
+    return rc;
```
dorjesinpo, line 105: making a note.  when it becomes asynchronous, we need to 
1) handle authentication failure in the auth thread (how? by disconnecting the channel?)
2) to make sure there is an authentication timeout and somebody enforces it

src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
```diff
@@ -175,17 +178,27 @@ int InitialConnectionHandler::processBlob(
...
+                                            &authenticationMsg,
                                             &negotiationMsg);
 
-    if (rc != 0) {
+    if (rc != rc_SUCCESS) {
         return (rc * 10) + rc_INVALID_NEGOTIATION_MESSAGE;  // RETURN
     }
 
-    if (negotiationMsg.has_value()) {
+    // Authentication or Negotiation based on the type of message received.
```
dorjesinpo, line 193: We need a check to reject negotiation message in between authentication request and authentication response.
Also a check to reject (another) authentication request in between authentication request and authentication response.
This could be checking `InitialConnectionContext::d_authenticationCtxSp` although explicit state would be better (we may need more checks or a mini FSM even)

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,228 @@
...
+    bsl::shared_ptr<mqbnet::AuthenticationContext> authenticationContext =
+        bsl::allocate_shared<mqbnet::AuthenticationContext>(
+            d_allocator_p,
+            context.get(),           // initialConnectionContext
+            authenticationMsg,       // authenticationMessage
+            false,                   // isReversed
+            State::e_AUTHENTICATING  // state
+        );
+
+    context->setAuthenticationContext(authenticationContext);
```
dorjesinpo, line 87: What would we give to Client/Admin/ClusterNode session for future authorization(s) if `Authenticator::handleAuthentication` was never called?  What's the plan?  Can we document it?

dorjesinpo (changes requested):
looks good as is and we could merge it.  but would be good to outline how we move forward

pniedzielski (approved):
Because this is going in the integration branch, let's merge.  We'll eventually make a PR into `main` with somethings from the integration branch, so it's not a big issue if this isn't fully fleshed out yet.

PR #747, opened 2025-05-14
fix: FBS::releaseRef w/o deletion on replica

dorjesinpo:
When purging App, replica must decrement refCount but should not attempt to write deletion record.
Calls leading to `releaseRef` now have `asPrimary` as an argument.
An alternative would be to use `FileStore::d_isPrimary` but it is interesting to trace all calls explicitly as primary or not.

### DIFF
--- a/src/groups/mqb/mqbi/mqbi_storage.h
+++ b/src/groups/mqb/mqbi/mqbi_storage.h
@@ -513,7 +513,6 @@ class Storage {
     /// `asPrimary` is `true`, delete the message data and record the event in
     /// the storage.
     /// Return one of the return codes from:
-    /// * e_SUCCESS          : success
     /// * e_GUID_NOT_FOUND      : `msgGUID` was not found
     /// * e_INVALID_OPERATION   : the value is invalid (already zero)
     /// * e_ZERO_REFERENCES     : message refCount has become zero
--- a/src/groups/mqb/mqbi/mqbi_storage.h
+++ b/src/groups/mqb/mqbi/mqbi_storage.h
@@ -509,8 +509,9 @@ class Storage {
                                         bool onReject = false) = 0;
 
     /// Decrement the reference count of the message identified by the
-    /// `msgGUID`.  If the resulting value is zero, delete the message data and
-    /// record the event in the storage.
+    /// `msgGUID`.  If the resulting value is zero and the specified
+    /// `asPrimary` is `true`, delete the message data and record the event in
+    /// the storage.
     /// Return one of the return codes from:
     /// * e_SUCCESS          : success
     /// * e_GUID_NOT_FOUND      : `msgGUID` was not found
@@ -521,8 +522,8 @@ class Storage {
     /// On CONFIRM, the caller of `confirm` is responsible to follow with
     /// `remove` call.  `releaseRef` is an alternative way to remove message in
     /// one call.
-    virtual StorageResult::Enum
-    releaseRef(const bmqt::MessageGUID& msgGUID) = 0;
+    virtual StorageResult::Enum releaseRef(const bmqt::MessageGUID& msgGUID,
+                                           bool asPrimary = true) = 0;
 
     /// Remove from the storage the message having the specified `msgGUID`
     /// and store it's size, in bytes, in the optionally specified `msgSize`.
--- a/src/groups/mqb/mqbs/mqbs_filebackedstorage.cpp
+++ b/src/groups/mqb/mqbs/mqbs_filebackedstorage.cpp
@@ -62,7 +62,8 @@ const int k_GC_MESSAGES_BATCH_SIZE = 1000;  // how many to process in one run
 // -----------------------
 
 // PRIVATE MANIPULATORS
-void FileBackedStorage::purgeCommon(const mqbu::StorageKey& appKey)
+void FileBackedStorage::purgeCommon(const mqbu::StorageKey& appKey,
+                                    bool                    asPrimary)
 {
     // This method is common to both primary and replica nodes, when a queue or
     // a specified virtual storage is purged.  QueueEngine should not be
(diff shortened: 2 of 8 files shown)


src/groups/mqb/mqbi/mqbi_storage.h
```diff
@@ -509,8 +509,9 @@ class Storage {
                                         bool onReject = false) = 0;
 
     /// Decrement the reference count of the message identified by the
-    /// `msgGUID`.  If the resulting value is zero, delete the message data and
-    /// record the event in the storage.
+    /// `msgGUID`.  If the resulting value is zero and the specified
+    /// `asPrimary` is `true`, delete the message data and record the event in
+    /// the storage.
     /// Return one of the return codes from:
     /// * e_SUCCESS          : success
```
kaikulimu:
```suggestion
```
This is inaccurate. We never return `e_SUCCESS`. On success, we either return `e_ZERO_REFERENCES` or `e_NON_ZERO_REFERENCE`.

src/groups/mqb/mqbs/mqbs_filebackedstorage.h
```diff
@@ -460,8 +460,9 @@ class FileBackedStorage BSLS_KEYWORD_FINAL : public ReplicatedStorage {
             bool                     onReject = false) BSLS_KEYWORD_OVERRIDE;
 
     /// Decrement the reference count of the message identified by the
-    /// 'msgGUID'.  If the resulting value is zero, delete the message data and
-    /// record the event in the storage.
+    /// `msgGUID`.  If the resulting value is zero and the specified
+    /// `asPrimary` is `true`, delete the message data and record the event in
+    /// the storage.
     /// Return one of the return codes from:
     /// * e_SUCCESS          : success
```
kaikulimu:
```suggestion
```
This is inaccurate. We never return `e_SUCCESS`. On success, we either return `e_ZERO_REFERENCES` or `e_NON_ZERO_REFERENCE`.

src/groups/mqb/mqbs/mqbs_inmemorystorage.h
```diff
@@ -336,8 +336,9 @@ class InMemoryStorage BSLS_KEYWORD_FINAL : public ReplicatedStorage {
             bool                     onReject = false) BSLS_KEYWORD_OVERRIDE;
 
     /// Decrement the reference count of the message identified by the
-    /// 'msgGUID'.  If the resulting value is zero, delete the message data and
-    /// record the event in the storage.
+    /// `msgGUID`.  If the resulting value is zero and the specified
+    /// `asPrimary` is `true`, delete the message data and record the event in
+    /// the storage.
     /// Return one of the return codes from:
     /// * e_SUCCESS          : success
```
kaikulimu:
```suggestion
```

src/groups/mqb/mqbs/mqbs_filebackedstorage.cpp
```diff
@@ -610,7 +613,7 @@ FileBackedStorage::removeAll(const mqbu::StorageKey& appKey)
                                   DataStoreRecordHandle());
 
         if (mqbi::StorageResult::e_SUCCESS == rc) {
-            purgeCommon(mqbu::StorageKey::k_NULL_KEY);
+            purgeCommon(mqbu::StorageKey::k_NULL_KEY, true);
```
hallfox, line 616: nit: I find call sites like these to be a little difficult to read (i.e. what does `true` mean? nothing from the function call really tells me), I might prefer doing something like this for readability:

```cpp
void purgeCommonAsPrimary(const mqbu::StorageKey& appKey) {
  purgeCommon(appKey, true);
}
void purgeCommonAsReplica(const mqbu::StorageKey& appKey) {
  purgeCommon(appKey, false);
}
```

It's not a big deal and there might be a better way to make it easier to understand what's happening from the call-site.

src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.cpp
```diff
@@ -412,7 +413,7 @@ VirtualStorageCatalog::purge(const mqbu::StorageKey& appKey,
             }
         }
 
-        purgeImpl(vs, itData, numVirtualStorages());
+        purgeImpl(vs, itData, numVirtualStorages(), true);
```
kaikulimu, line 422: Is it guaranteed that `VirtualStorageCatalog::purge` is only called by primary?

src/groups/mqb/mqbs/mqbs_inmemorystorage.cpp
```diff
@@ -438,7 +440,7 @@ InMemoryStorage::removeAll(const mqbu::StorageKey& appKey)
     // Clear out the virtual storage associated with the specified 'appKey'.
     // Note that this cannot be done while iterating over the it in the above
     // 'while' loop for obvious reasons.
-    d_virtualStorageCatalog.removeAll(appKey);
+    d_virtualStorageCatalog.removeAll(appKey, true);
```
kaikulimu, line 447: Is it guaranteed that `InMemoryStorage::removeAll` is only called by primary?

kaikulimu (changes requested):

src/groups/mqb/mqbs/mqbs_virtualstoragecatalog.cpp
dorjesinpo: It is not obvious, but the answer is yes.
The caller is `FileBackedStorage::removeAll` where we do want to `FileBackedStorage::writeAppPurgeRecord`.
The caller of which is `StorageUtil::purgeQueueDispatched`.
Note that `InMemoryStorage` does not call `VirtualStorageCatalog::purge`

src/groups/mqb/mqbs/mqbs_inmemorystorage.cpp
dorjesinpo: > Is it guaranteed that `InMemoryStorage::removeAll` is only called by primary?

It is called in primary by `StorageUtil::purgeQueueDispatched` and in broadcast mode and in proxy.  All of which are supposed to proactively remove items.

kaikulimu (approved):
lgtm

PR #748, opened 2025-05-15
CI: print local variables and args in cores

678098:
1. Group gdb outputs from different core dumps:

Screenshot 2025-05-15 at 16 29 12

2. Print local variables and function args if possible (limitation: no pointed structs deduction)
 
Screenshot 2025-05-15 at 18 03 18


pniedzielski (approved):
:+1: for the GH Actions groups

PR #751, opened 2025-05-16
Feat[ClusterCatalog]: load proxy configs on demand

jll63:
**Description**

A proxy picks "new" domain configurations installed even after it has started.  However, it reads cluster proxy definitions from `cluster.json`, only once,  at startup. This causes an error when a "new" domain is backed by a cluster unknown at startup. This PR handles proxy cluster configs in the same way as domain configs. The entirety of `clusters.json` is read during startup, as before. However, if, during an `openQueue`, a domain refers to a cluster not yet available in `ClusterCatalog`, the broker looks for a file named after the cluster, with a `.json` extension,  in the `proxyclusters` subdirectory of the config directory. If found, and it can be parsed as a `ClusterProxyDefinition`, it added to the `ClusterCatalog`.

Cluster names may contain directory separators. They don't receive any special treatment. Thus, if the proxy does not yet have a configuration for, say, cluster `foo/bar`, it will look for a `<config_dir>/proxyclusters/foo/bar.json`.

Note that `ClusterProxyDefinition`s for reversed proxy connections must still be present in `clusters.json` from the get-go.

**Testing performed**
Added new integration tests (`src/integration-tests/test_create_new_domain.py`).

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
@@ -123,11 +123,13 @@ int ClusterCatalog::createCluster(bsl::ostream& errorDescription,
 
     enum RcEnum {
         // Value for the various RC error categories
-        rc_SUCCESS                    = 0,
-        rc_NAME_NOT_UNIQUE            = -1,
-        rc_FETCH_DEFINITION_FAILED    = -2,
-        rc_NETCLUSTER_CREATION_FAILED = -3,
-        rc_NETCLUSTER_INVALID_NODE_ID = -4
+        rc_SUCCESS                            = 0,
+        rc_NAME_NOT_UNIQUE                    = -1,
+        rc_FETCH_DEFINITION_FAILED            = -2,
+        rc_NETCLUSTER_CREATION_FAILED         = -3,
+        rc_NETCLUSTER_INVALID_NODE_ID         = -4,
+        rc_BROKER_CLUSTER_CONFIG_LOAD_FAILURE = -5,
+        rc_JSON_DECODE_ERROR                  = -6
     };
 
     int rc = rc_SUCCESS;
@@ -205,9 +207,60 @@ int ClusterCatalog::createCluster(bsl::ostream& errorDescription,
 
         if (proxyDefinitionIter ==
             d_clustersDefinition.proxyClusters().end()) {
-            errorDescription << "Fetch proxy definition failed for cluster: '"
-                             << name << "'";
-            return rc_FETCH_DEFINITION_FAILED;  // RETURN
+            // Maybe a proxy cluster JSON was added to the
+            // 'proxyclusters' directory since we started.
+
+            const mqbcfg::AppConfig& brkrCfg = mqbcfg::BrokerConfig::get();
+            bsl::string              configFilename = brkrCfg.etcDir();
+
+            bdls::PathUtil::appendRaw(&configFilename, "proxyclusters");
+            bdls::PathUtil::appendRaw(&configFilename,
+                                      (name + ".json").c_str());
+
+            bsl::ifstream      configStream(configFilename.c_str());
+            bmqu::MemOutStream configParameters;
+            configParameters << configStream.rdbuf();
+
+            if (!configStream || !configParameters) {
+                errorDescription
+                    << "Fetch proxy definition failed for cluster: '" << name
+                    << "', file: '" << configFilename
+                    << "' is not readable or empty";
+                return rc_BROKER_CLUSTER_CONFIG_LOAD_FAILURE * 10 +
+                       rc_FETCH_DEFINITION_FAILED;  // RETURN
+            }
+
+            configStream.close();
+
+            BALL_LOG_INFO
+                << "Dynamically loading cluster proxy definition from '"
+                << configFilename;
+
+            // Decode the JSON stream
+            mqbcfg::ClusterProxyDefinition proxyDefinition;
+            baljsn::Decoder                decoder;
+            baljsn::DecoderOptions         options;
+            bslstl::StringRef              jsonString = configParameters.str();
+            bdlsb::FixedMemInStreamBuf     jsonStreamBuf(jsonString.data(),
+                                                     jsonString.length());
+
+            options.setSkipUnknownElements(true);
+
+            rc = decoder.decode(&jsonStreamBuf, &proxyDefinition, options);
+            if (rc != 0) {
+                BALL_LOG_ERROR << "Error while decoding JSON from '"
+                               << configFilename << "' [rc: " << rc
+                               << ", error: '" << decoder.loggedMessages()
+                               << "']" << ", content:\n"
+                               << configParameters.str();
+                return rc_JSON_DECODE_ERROR * 10 +
+                       rc_NETCLUSTER_CREATION_FAILED;  // RETURN
+            }
+
+            d_clustersDefinition.proxyClusters().push_back(
+                proxyDefinition);  // Add the new cluster to the list
+            proxyDefinitionIter = d_clustersDefinition.proxyClusters().end() -
+                                  1;
         }
 
         const mqbcfg::ClusterProxyDefinition& clusterProxyDefinition =
@@ -407,10 +460,10 @@ int ClusterCatalog::loadBrokerClusterConfig(bsl::ostream&)
 
     enum RcEnum {
         // Value for the various RC error categories
-        rc_SUCCESS                           = 0,
-        rc_BROKER_CLUSTER_CONFIG_LOADFAILURE = -1,
-        rc_CONFIG_READ_ERROR                 = -1,
-        rc_JSON_DECODE_ERROR                 = -2
+        rc_SUCCESS                            = 0,
+        rc_BROKER_CLUSTER_CONFIG_LOAD_FAILURE = -1,
+        rc_CONFIG_READ_ERROR                  = -1,
+        rc_JSON_DECODE_ERROR                  = -2
     };
 
     int rc = rc_SUCCESS;
@@ -421,7 +474,7 @@ int ClusterCatalog::loadBrokerClusterConfig(bsl::ostream&)
     if (bdls::PathUtil::appendIfValid(&configFilename, "clusters.json")) {
         BALL_LOG_ERROR << "Invalid path '" << configFilename << "'\n";
         return rc_CONFIG_READ_ERROR * 10 +
-               rc_BROKER_CLUSTER_CONFIG_LOADFAILURE;  // RETURN
+               rc_BROKER_CLUSTER_CONFIG_LOAD_FAILURE;  // RETURN
     }
 
     bsl::ifstream      configStream(configFilename.c_str());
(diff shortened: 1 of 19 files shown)


src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
```diff
@@ -49,6 +49,14 @@
...
 namespace mqbblp {
 
+enum RcEnum {
+    // Value for the various RC error categories
+    rc_SUCCESS                           = 0,
+    rc_BROKER_CLUSTER_CONFIG_LOADFAILURE = -1,
+    rc_CONFIG_READ_ERROR                 = -1,
+    rc_JSON_DECODE_ERROR                 = -2
+};
+
```
hallfox: This should be put under a struct to keep the enums from floating in the namespace

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
```diff
@@ -205,9 +213,58 @@ int ClusterCatalog::createCluster(bsl::ostream& errorDescription,
 
         if (proxyDefinitionIter ==
             d_clustersDefinition.proxyClusters().end()) {
-            errorDescription << "Fetch proxy definition failed for cluster: '"
-                             << name << "'";
-            return rc_FETCH_DEFINITION_FAILED;  // RETURN
+            // Maybe a proxy cluster JSON was added to directory
+            // 'proxyclusters' since we started.
```
hallfox:
```suggestion
            // Maybe a proxy cluster JSON was added to the
            // 'proxyclusters' directory since we started.
```

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
```diff
@@ -205,9 +213,58 @@ int ClusterCatalog::createCluster(bsl::ostream& errorDescription,
...
+                               << "']" << ", content:\n"
+                               << configParameters.str();
+                return rc_JSON_DECODE_ERROR * 10 +
+                       rc_BROKER_CLUSTER_CONFIG_LOADFAILURE;  // RETURN
+            }
+
+            d_clustersDefinition.proxyClusters().push_back(
+                proxyDefinition);  // Add the new cluster to the list
+            proxyDefinitionIter = d_clustersDefinition.proxyClusters().end() -
+                                  1;
```
hallfox: This seems to inflate the size of this function quite a bit, can this be moved into a helper function?

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
```diff
@@ -205,9 +213,58 @@ int ClusterCatalog::createCluster(bsl::ostream& errorDescription,
...
+
+            options.setSkipUnknownElements(true);
+
+            rc = decoder.decode(&jsonStreamBuf, &proxyDefinition, options);
+            if (rc != 0) {
+                BALL_LOG_ERROR << "Error while decoding JSON from '"
+                               << configFilename << "' [rc: " << rc
+                               << ", error: '" << decoder.loggedMessages()
+                               << "']" << ", content:\n"
+                               << configParameters.str();
```
hallfox, line 255: Is it safe to print an arbitrary file's contents out to console?

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
```diff
@@ -205,9 +213,58 @@ int ClusterCatalog::createCluster(bsl::ostream& errorDescription,
...
-            errorDescription << "Fetch proxy definition failed for cluster: '"
-                             << name << "'";
-            return rc_FETCH_DEFINITION_FAILED;  // RETURN
+            // Maybe a proxy cluster JSON was added to directory
+            // 'proxyclusters' since we started.
+
+            const mqbcfg::AppConfig& brkrCfg = mqbcfg::BrokerConfig::get();
+            bsl::string              configFilename = brkrCfg.etcDir();
+
+            bdls::PathUtil::appendRaw(&configFilename, "proxyclusters");
```
hallfox, line 216: Are there other places we refer to the "proxyclusters" directory? What other code assumes that this is where the proxy cluster definitions are?

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
```diff
@@ -205,9 +213,58 @@ int ClusterCatalog::createCluster(bsl::ostream& errorDescription,
...
+            bdls::PathUtil::appendRaw(&configFilename,
+                                      (name + ".json").c_str());
```
hallfox, line 218: This is UB, `(name + ".json")` will not live long enough

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
```diff
@@ -205,9 +213,58 @@ int ClusterCatalog::createCluster(bsl::ostream& errorDescription,
...
+
+            bsl::ifstream      configStream(configFilename.c_str());
```
hallfox, line 220: Would it be better to reload the whole directory's cluster configs if there are any changes, or should we load only new ones? Can we batch load all the new ones so we don't redo this operation in case there are a sequence of new proxy clusters?

src/python/blazingmq/dev/it/process/client.py
```diff
@@ -138,7 +134,7 @@ def send(self, command):
         Send a command to the client and return immediately.  This function
         returns nothing.
         """
-        self._logger.info(f"send: command = {command}")
+        self._logger.info("send: command = %s", command)
```
hallfox, line 137: Why move to old style formatting? f-strings are supported in 3.8

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
```diff
@@ -49,6 +49,14 @@
 namespace BloombergLP {
 namespace mqbblp {
 
+enum RcEnum {
+    // Value for the various RC error categories
+    rc_SUCCESS                           = 0,
+    rc_BROKER_CLUSTER_CONFIG_LOADFAILURE = -1,
```
kaikulimu:
```suggestion
    rc_BROKER_CLUSTER_CONFIG_LOAD_FAILURE = -1,
```

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
kaikulimu: Agree with Taylor. This is super confusing, especially because `ClusterCatalog::createCluster()` already has its own set of `RcEnum`:

```
    enum RcEnum {
        // Value for the various RC error categories
        rc_SUCCESS                    = 0,
        rc_NAME_NOT_UNIQUE            = -1,
        rc_FETCH_DEFINITION_FAILED    = -2,
        rc_NETCLUSTER_CREATION_FAILED = -3,
        rc_NETCLUSTER_INVALID_NODE_ID = -4
    };
```

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
```diff
@@ -205,9 +213,58 @@ int ClusterCatalog::createCluster(bsl::ostream& errorDescription,
...
+            options.setSkipUnknownElements(true);
+
+            rc = decoder.decode(&jsonStreamBuf, &proxyDefinition, options);
+            if (rc != 0) {
+                BALL_LOG_ERROR << "Error while decoding JSON from '"
+                               << configFilename << "' [rc: " << rc
+                               << ", error: '" << decoder.loggedMessages()
+                               << "']" << ", content:\n"
+                               << configParameters.str();
+                return rc_JSON_DECODE_ERROR * 10 +
```
kaikulimu, line 256: Suggestion: Add a new rc to https://github.com/bloomberg/blazingmq/blob/8ecea0fd16a2952a946798d2954422abe736511e/src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp#L132-L139 and return it.

src/python/blazingmq/dev/it/process/client.py
jll63: It is not recommended to use f-strings for logging. Original code pays the cost of interpolation, even if the string is eventually not logged. New code allows `info` interpolate only as needed.

This is flagged by pylint (`W1203: Use lazy % formatting in logging functions (logging-fstring-interpolation)`).

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
jll63: No other C++ code. In Python: `blazingmq.dev.configurator`.

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
jll63: Are you sure? See [cppreference](https://en.cppreference.com/w/cpp/language/lifetime.html#:~:text=All%20temporary%20objects%20are%20destroyed,to%20the%20order%20of%20creation.) Isn't the "full expression" the call to `appendRaw`? I would expect the temp string to be destroyed between the last closing parenthesis and the semicolon. I also tried valgrind and ubsan on a small similar example, they don't object. Thinking about it, even `foo(name + ".json");` would fail then.

That being said, I don't understand why I bother to create a temporary for this, innstead of calling `appendRaw` twice. Changing it to:

```c++
            bdls::PathUtil::appendRaw(&configFilename, name.c_str());
            bdls::PathUtil::appendRaw(&configFilename, ".json");
```

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
jll63: Given that each dynamic `ClusterProxyDefinition` lives in its own file, I don't think there is any benefit in catching 'em all. They're not pokémons ;-) Also who knows how many more new files could be there? It's OK to spend a lot of time loading stuff at startup, but here we are processing an openQueue request.

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
jll63: As safe as in the other places where we do this ;-)

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
jll63: By ~30%. Hmmm. Not sure. I feel that the code "flows" quite naturally like that. I'm not a big fan of functions that are called only once (except maybe in Python because of its silly scoping rules) unless the main function gets really big. ~200 lines seems reasonable to me. @kaikulimu what do you think?

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
hallfox: You're right! 

> The Standard mandates that all temporary objects get destroyed as the last step of evaluating of the full-expression that contains the point where the temporaries were created.

https://lesleylai.info/en/temporaries/

I was under the impression that temporary lifetimes were bound by *sequence points*.


Though I also agree having multiple `appendRaw` calls feels better here.

jll63: Alas `appendRaw` is not raw enough, it adds a `/` before appending...

kaikulimu (approved):
lgtm

PR #758, opened 2025-05-23
IT: can run tests with sanitizers

bbpetukhov:
**Describe your changes**
Updated Dockerfile for building broker with sanitizers. Created script for running ITs with sanitizers.

**Testing performed**
- Manually ran ITs
- Ran ITs in github workflows
- Added

**Additional context**
NB: Running all ITs with sanitizers may take hours to complete.


.github/workflows/sanitize.yaml
```diff
@@ -42,3 +42,7 @@ jobs:
       - name: UT [c++,${{ github.job }}]
         if: ${{ matrix.fuzz == 'off' }}
         run: docker run --rm sanitizer-${{ github.job }}
+      - name: IT [c++,${{ github.job }}]
+        if: ${{ matrix.fuzz == 'off' }}
+        run: |
+          docker run --rm sanitizer-${{ github.job }} /blazingmq/cmake.bld/Linux/run-it.sh -k breathing --reruns=3 --bmq-tolerate-dirty-shutdown
```
678098, line 48: `--reruns 3` might take a lot of time if tests keep failing. Might be good to configure reruns for specific ITs instead in the future

docker/sanitizers/build_sanitizer.sh
```diff
@@ -92,6 +93,17 @@ DIR_BUILD_EXT="${DIR_EXTERNAL}/cmake.bld"
...
 
+# Create Python venv
+python3 -m venv venv
+if [ -f "venv/bin/activate" ]; then
+    # shellcheck disable=SC1091
+    source venv/bin/activate
+    pip install -r "${DIR_SRC_BMQ}/src/python/requirements.txt"
+else
+  echo "Virtual environment not found."
+  exit 1
```
678098:
```suggestion
    echo "Virtual environment not found."
    exit 1
```
Small indentation issue

.github/workflows/sanitize.yaml
bbpetukhov: Make sense. Will consider this while extending the test suite

PR #763, opened 2025-05-28
Feat: fuzz tests

waldgange:
This PR inroduces a new job to sanitizers workflow. The job builds Fuzz-tests and their dependencies instrumented with address sanitizer. This job is launched in parallel with other sanitizers after a PR is approved. The fuzz tests sources reside in `src/standalones/s_bmqfuzz`. Three tests are added so far:
- `s_bmqfuzz_eval.cpp` targets `BloombergLP::bmqeval::SimpleEvaluator::compile`
- `s_bmqfuzz_parseutil.cpp` targets `BloombergLP::mqbcmd::ParseUtil::parse`
- `s_bmqfuzz_bmqt_uri.cpp` targets `BloombergLP::bmqt::UriParser::parse`

There is also a script for building them locally [bin/build-fuzz.sh](https://github.com/bloomberg/blazingmq/pull/763/files#diff-01a551953a97b923f96a697fb9ba67db27f05e8b1e824f2e1e04a8c407880c89)

### DIFF
--- a/src/standalones/s_bmqfuzz/README.md
+++ b/src/standalones/s_bmqfuzz/README.md
@@ -2,7 +2,6 @@
 
 This folder contains fuzzers for blazingmq
 
-- `eval_fuzzer.cpp` targets `BloombergLP::bmqeval::SimpleEvaluator::compile`
-- `parseutil_fuzzer.cpp` targets `BloombergLP::mqbcmd::ParseUtil::parse`
-- `bmqt_uri_fuzzer.cpp` targets `BloombergLP::bmqt::UriParser::parse`
-
+- `s_bmqfuzz_eval.cpp` targets `BloombergLP::bmqeval::SimpleEvaluator::compile`
+- `s_bmqfuzz_parseutil.cpp` targets `BloombergLP::mqbcmd::ParseUtil::parse`
+- `s_bmqfuzz_bmqt_uri.cpp` targets `BloombergLP::bmqt::UriParser::parse`
--- a/src/standalones/s_bmqfuzz/s_bmqfuzz_bmqt_uri.fuzz.cpp
+++ b/src/standalones/s_bmqfuzz/s_bmqfuzz_bmqt_uri.fuzz.cpp
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// bmqt_uri_fuzzer.cpp                                                -*-C++-*-
+// s_bmqfuzz_bmqt_uri.fuzz.cpp                                        -*-C++-*-
 #include <stdio.h>
 #include <stdlib.h>
 #include <string>
--- a/src/standalones/s_bmqfuzz/s_bmqfuzz_eval.fuzz.cpp
+++ b/src/standalones/s_bmqfuzz/s_bmqfuzz_eval.fuzz.cpp
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// eval_fuzzer.cpp                                                    -*-C++-*-
+// s_bmqfuzz_eval.fuzz.cpp                                            -*-C++-*-
 #include <stdio.h>
 #include <stdlib.h>
 #include <string>
--- a/src/standalones/s_bmqfuzz/s_bmqfuzz_parseutil.fuzz.cpp
+++ b/src/standalones/s_bmqfuzz/s_bmqfuzz_parseutil.fuzz.cpp
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// parseutil_fuzzer.cpp                                               -*-C++-*-
+// s_bmqfuzz_parseutil.fuzz.cpp                                       -*-C++-*-
 #include <stdio.h>
 #include <stdlib.h>
 #include <string>


CMakeLists.txt
```diff
@@ -313,12 +337,14 @@ install(TARGETS bmqbrkr_plugins
...
 install(EXPORT BmqbrkrPluginsTargets NAMESPACE blazingmq:: DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/BmqbrkrPlugins" COMPONENT bmqbrkr_plugins)
 
+find_package(BdeBuildSystem REQUIRED)
 bbs_read_metadata(GROUP mqb SOURCE_DIR src/groups/mqb)
 bbs_read_metadata(GROUP bmq SOURCE_DIR src/groups/bmq)
 set(bmqbrkr_plugins_deps)
 list(APPEND bmqbrkr_plugins_deps ${mqb_DEPENDS} ${bmq_DEPENDS})
 list(REMOVE_DUPLICATES bmqbrkr_plugins_deps)
 list(REMOVE_ITEM bmqbrkr_plugins_deps bmq) # don't need this dependency
+include(CMakePackageConfigHelpers)
```
hallfox, line 340:
```suggestion

include(CMakePackageConfigHelpers)
```

bin/build-fuzz.sh
```diff
@@ -0,0 +1,79 @@
...
+    -DBDE_BUILD_TARGET_CPP20=ON \
+    -DCMAKE_BUILD_TYPE=Debug \
+    -DCMAKE_INSTALL_LIBDIR="lib" \
+    -DCMAKE_INSTALL_PREFIX="${DIR_INSTALL}" \
+    -DCMAKE_MODULE_PATH="${DIR_THIRDPARTY}/bde-tools/cmake;${DIR_THIRDPARTY}/bde-tools/BdeBuildSystem" \
+    -DCMAKE_PREFIX_PATH="${DIR_INSTALL}" \
+    -DCMAKE_TOOLCHAIN_FILE="${DIR_THIRDPARTY}/bde-tools/BdeBuildSystem/toolchains/linux/clang-default.cmake" \
+    -DCMAKE_CXX_FLAGS="-fsanitize=fuzzer-no-link" \
+    -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+    -DINSTALL_TARGETS=fuzztests)
```
hallfox: I think we should start moving these configs into CMakePresets.json, so we can manage all the commonalities between these build modes.

Can this be replaced with a preset called `fuzz-tests`?

src/fuzz-tests/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+
+# Create a custom target 'fuzztests' that builds all fuzz tests
+add_custom_target("fuzztests")
+
+find_package(BdeBuildSystem REQUIRED)
+bbs_read_metadata(PACKAGE fuzz-tests)
+
+file(GLOB sources "*fuzzer.cpp")
+foreach (source ${sources})
+  string(REGEX MATCH "\/([A-Za-z_]+)\.cpp$" _ ${source})
```
hallfox: Why regex match if we glob all the .cpp files in the directory?

src/fuzz-tests/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+  set(targetName ${CMAKE_MATCH_1})
+  add_executable(${targetName} ${source})
+
+  target_bmq_default_compiler_flags(${targetName})
+  target_include_directories(${targetName} PRIVATE ${fuzz-tests_INCLUDE_DIRS})
+  target_link_libraries(${targetName} PRIVATE ${fuzz-tests_PCDEPS})
+  target_link_options(${targetName} PRIVATE "-fsanitize=fuzzer")
```
hallfox: Usually we set sanitizer flags in the toolchain file, we may want to provide one based off of the one we'd be using from https://github.com/bloomberg/bde-tools/tree/main/cmake/toolchains/linux

src/fuzz-tests/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+
+  set_target_properties(${targetName}
+    PROPERTIES OUTPUT_NAME ${targetName})
+
+  # Add the current tutorial to the 'fuzztests' rule
+  add_dependencies("fuzztests" ${targetName})
+endforeach()
```
hallfox: I think this is a nontrivial enough bit of CMake that it should be put in a function and documented

CMakeLists.txt
```diff
@@ -189,6 +195,19 @@ else()
     set(BMQ_TARGET_MQB_NEEDED        YES)
     set(BMQ_TARGET_PROMETHEUS_NEEDED YES)
   endif()
+
+  if (installIT)
+    set(BMQ_TARGET_BMQ_NEEDED       YES)
+    set(BMQ_TARGET_MQB_NEEDED       YES)
+    set(BMQ_TARGET_BMQBRKR_NEEDED   YES)
+    set(BMQ_TARGET_BMQTOOL_NEEDED   YES)
+  endif()
```
hallfox: Is this IT install rules related to fuzz testing? Can we have a separate PR for it if not?

waldgange: Ok. It's not related directly to the fuzz tests

src/fuzz-tests/CMakeLists.txt
waldgange: done

src/fuzz-tests/CMakeLists.txt
waldgange: moved all the flags to [docker/sanitizers/clang-libcxx-sanitizer.cmake](https://github.com/bloomberg/blazingmq/blob/4f7b2bff09ee996e7c8bf93382278412602d2ae4/docker/sanitizers/clang-libcxx-sanitizer.cmake)

src/fuzz-tests/CMakeLists.txt
waldgange: simplified this

bin/build-fuzz.sh
waldgange: added a preset for non-parametrized flags

docker/sanitizers/build_sanitizer.sh
```diff
@@ -137,30 +144,38 @@ cd -
...
 
 TOOLCHAIN_PATH="${DIR_SCRIPTS}/clang-libcxx-sanitizer.cmake"
 export SANITIZER_NAME="${SANITIZER_NAME}"
 export CC="clang"
 export CXX="clang++"
-export CMAKE_CXX_STANDARD_INCLUDE_DIRECTORIES="/usr/include;/usr/include/clang/${LLVM_VERSION}/include"
+if [ "${FUZZER}" == "on" ]; then
+  export FUZZER_FLAG="fuzzer-no-link"
+else
+  export CMAKE_CXX_STANDARD_INCLUDE_DIRECTORIES="/usr/include;/usr/include/clang/${LLVM_VERSION}/include"
```
hallfox: Why is this in the `else` branch?

docker/sanitizers/build_sanitizer.sh
```diff
@@ -24,11 +24,18 @@ set -eux
...
 if [ -z "${1}" ]; then
     echo 'Error: Missing sanitizer name.' >&2
-    echo '  (Usage: build_sanitizer.sh <sanitizer-name>)' >&2
+    echo '  (Usage: build_sanitizer.sh <sanitizer-name> <fuzzer>)' >&2
+    exit 1
+fi
+
+if [[ -z "${2}" || ("${2}" != "on" && "${2}" != "off") ]]; then
+    echo 'Error: Wrong fuzzer argument. It can be either "on" or "off"' >&2
+    echo '  (Usage: build_sanitizer.sh <sanitizer-name> <fuzzer>)' >&2
```
hallfox, line 39: Can we add a section to the comment block at the top with a usage example?

docker/sanitizers/clang-libcxx-sanitizer.cmake
```diff
@@ -44,12 +42,33 @@ string(CONCAT TOOLCHAIN_C_FLAGS
...
+           "-lc++abi "
+           )
+else()
+    # Use non-instrumented libstdc++ (GNU) standard library for C++. The reason
+    # why we use this library is that preinstalled clang and libFuzzer depend
+    # on GNU library. If we try to link it to a fuzzer with other dependencies,
+    # built with LLVM libc++, it leads to a conflict. An alternative is to
+    # build clang and libFuzzer ourselves with LLVM libc++. Though it takes too
+    # much time. Also it contradicts BFuzz workflow that we are trying to
+    # imitate here.
```
hallfox:
```suggestion
    # Use the non-instrumented libstdc++ (GNU) standard library for C++. We
    # use this library because the preinstalled clang and libFuzzer depend on
    # GNU's standard library. If we try to link it to a fuzzer with other dependencies
    # built with LLVM's libc++, it leads to a conflict. The alternative is to build clang
    # and libFuzzer ourselves with LLVM's libc++, but this takes too much time.
```

docker/sanitizers/clang-libcxx-sanitizer.cmake
```diff
@@ -44,12 +42,33 @@ string(CONCAT TOOLCHAIN_C_FLAGS
        "${TOOLCHAIN_SHARED_FLAGS} "
       )
 
-# Use libc++ standard library for C++.
-string(CONCAT TOOLCHAIN_CXX_FLAGS
-       "${TOOLCHAIN_CXX_FLAGS} "
-       "-stdlib=libc++ "
-       "-I${LIBCXX_BUILD_PATH}/include/c++/v1 "
-      )
+if(DEFINED ENV{LIBCXX_BUILD_PATH})
```
hallfox, line 46: In what build configuration are we using libc++? Just when fuzzers are disabled? I think we should be consistent about which standard library we use. It seems like we'd lose some instrumentation coverage if we don't custom build LLVM here. My understanding is we have this kind of matrix:

| Sanitizer | Fuzz? | std choice |
|----------|-------|------------|
| ASAN | no | libc++ instrumented |
| MSAN | no | libc++ instrumented |
| TSAN | no | libc++ instrumented |
| UBSAN | no | libc++ instrumented |
| ASAN | yes | libstd++ |

That makes me kind of think that the fuzz test builds are kind of exceptional and should probably have their own build rules

src/fuzz-tests/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+
+if(NOT BMQ_TARGET_FUZZTESTS_NEEDED OR NOT "${CMAKE_CXX_COMPILER_ID}" MATCHES "(Apple)?Clang")
+  return()
+endif()
+
+# Create a custom target 'fuzztests' that builds all fuzz tests
+add_custom_target("fuzztests")
+
+find_package(BdeBuildSystem REQUIRED)
+bbs_read_metadata(PACKAGE fuzz-tests)
```
hallfox: This is not a valid BDE package name, but BDE also doesn't really define a clear way for organizing a separate package with dedicated test executables. I'm going to recommend we invent the following conventions:

1. Fuzz tests should have the suffix `.fuzz.cpp`
2. This specific package of fuzz tests should live in `src/standalones/s_bmqfuzz`

 (maybe we can consult with them on what their opinions about this would be?)

src/fuzz-tests/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+  target_link_libraries(${targetName} PRIVATE ${fuzz-tests_PCDEPS})
+
+  set_target_properties(${targetName}
+    PROPERTIES OUTPUT_NAME ${targetName})
+
+  # Add the current fuzzer to the 'fuzztests' rule
+  add_dependencies("fuzztests" ${targetName})
+endfunction()
+
+file(GLOB sources "*fuzzer.cpp")
```
hallfox: Probably should get this list from `${fuzz-tests_SOURCE_FILES}`

docker/sanitizers/build_sanitizer.sh
waldgange: The default standard library include paths contain `libstdc++` headers. We don't want to change it for fuzz-tests. I didn't want to brake anything for the old workflows, so I desided to leave it in the `else` branch. However this doesn't look like correct include paths for `libc++`.  We set this path properly later in the toolchain:
https://github.com/bloomberg/blazingmq/blob/e1b5b1b143495dc43f3d419c26f6b05a2a783a55/docker/sanitizers/clang-libcxx-sanitizer.cmake#L48-L52
So I've got rid of this branch at all and added a [flag](https://github.com/bloomberg/blazingmq/pull/763/commits/66792318d2f2e7eeab317251073def953447ed46#diff-75c3a4106e0f414dc87976557bc3fa9d164b9a51584b48f02f8b12941817e5bfR50) that removes default include paths for non-fuzz case

docker/sanitizers/build_sanitizer.sh
waldgange: done

docker/sanitizers/clang-libcxx-sanitizer.cmake
waldgange: Correct. But these two workflows also have a lot in common - all dependencies, except the standard library, instumented with sanitizers. So it's very convenient to reuse this toolchain. Do we really want another toolchain for that?

src/fuzz-tests/CMakeLists.txt
waldgange: done

src/fuzz-tests/CMakeLists.txt
waldgange: done

src/fuzz-tests/CMakeLists.txt
waldgange: I've asked BDE team about fuzz tests. They build them in a slightly different way. They put each fuzz test to the test driver of the corresponding component. Here is an [example](https://github.com/bloomberg/bde/blob/8ba3ae5f3b9d82eececcc80f3d2ef6985f8dafb5/groups/bdl/bdlt/bdlt_dateutil.t.cpp#L131-L184). However they didn't integrate neither with BFuzz nor with oss-fuzz. They run their fuzz tests locally instead. What do you think? Should we do the same with our tests?

docker/sanitizers/clang-libcxx-sanitizer.cmake
hallfox: Yeah I think that's fair

hallfox (approved):

PR #766, opened 2025-05-29
handle reinserting duplicate (on replica)

dorjesinpo:
Suppose primary garbage collect message in its history and replica does not
Then, duplicate arrives and primary does not detect it as a duplicate and replicates to replica
But replica still has the message in its history and it will assert on `d_isLive`

### DIFF
--- a/src/groups/bmq/bmqc/bmqc_orderedhashmapwithhistory.h
+++ b/src/groups/bmq/bmqc/bmqc_orderedhashmapwithhistory.h
@@ -617,8 +617,8 @@ OrderedHashMapWithHistory<KEY, VALUE, HASH, VALUE_TYPE>::insert(
     const SOURCE_TYPE& value,
     TimeType           timePoint)
 {
-    bsl::pair<gc_iterator, bool> result = d_impl.insert(
-        Value(value, d_timeout ? timePoint + d_timeout : 0));
+    TimeType                     time = d_timeout ? timePoint + d_timeout : 0;
+    bsl::pair<gc_iterator, bool> result = d_impl.insert(Value(value, time));
     // No need to keep track of element's timePoint if the map is not
     // maintaining any history (i.e., if d_timeout == 0).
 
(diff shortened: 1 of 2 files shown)


src/groups/bmq/bmqc/bmqc_orderedhashmapwithhistory.h
```diff
@@ -635,6 +635,21 @@ OrderedHashMapWithHistory<KEY, VALUE, HASH, VALUE_TYPE>::insert(
         }
         it->d_next = d_impl.end();
     }
+    else {
+        if (!it->d_isLive) {
+            // Corner case of (re)inserting undetected duplicate
```
678098, line 640: `bmqc_orderedhashmapwithhistory` is a general-purpose container, so the documentation for it should not mention one of its concrete usages from another level of application logics 
```suggestion
            // Corner case of (re)inserting an element that was previously
            // marked for deletion but still exists in history.
```

src/groups/bmq/bmqc/bmqc_orderedhashmapwithhistory.t.cpp
```diff
@@ -332,6 +332,37 @@ static void test6_eraseThenGc()
     obj.gc(now, BATCH_SIZE);
 }
 
+static void test7_gcThenInsert()
+{
+    // ------------------------------------------------------------------------
+    // HISTORY
+    //
+    // Concerns:
+    //   The same GUID gets inserted after 'erase' (as an undetected duplicate)
```
678098, line 341: Message GUID is from `bmqt` package. `bmqc` package doesn't depend on `bmqt`, so it's not good to mention it here. `bmqc_orderedhashmapwithhistory` is a general-purpose container, so it's better to rephrase the comment using general word "elements"

Secondly, the comment mentions "undetected duplicates". This instantly raises a question whether `bmqc::OrderedHashMapWithHistory` has a duplicate detection mechanism. In fact, it doesn't, and this comment refers to another layer of the broker logics, so it should not be mentioned here at all

```suggestion
    //   The same element gets inserted after 'erase'
```

src/groups/bmq/bmqc/bmqc_orderedhashmapwithhistory.t.cpp
```diff
@@ -332,6 +332,37 @@ static void test6_eraseThenGc()
...
+    //
+    // Plan:
+    //   Insert a GUID
+    //   Erase.
+    //   Insert the GUID.
```
678098, line 346:
```suggestion
    //   Insert an element.
    //   Erase this element.
    //   Insert the same element again.
```

678098 (approved):
Looks good

PR #768, opened 2025-06-02
Fix[MQB]: erase queuehandles in ClusterNodeSession::teardown

dorjesinpo:
- Primary shuts down.
- `ClusterNodeSession::teardown` calls `QueueHandle::drop` which releases last reference to the handle
- Note that `ClusterQueueHelper::onHandleReleased` is NOT called since the peer did not close the queue.
- Meanwhile, close queue request arrives
- `ClusterQueueHelper::processPeerCloseQueueRequest` looks up handle context which contains raw pointer to the handle
- Segfault

The solution is to remove the context in `ClusterNodeSession::teardown`

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_queue.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_queue.cpp
@@ -57,94 +57,6 @@
 namespace BloombergLP {
 namespace mqbblp {
 
-namespace {
-const char k_LOG_CATEGORY[] = "MQBBLP.QUEUE";
-
-/// This method performs the actual drop. It dispatches to the templated
-/// (local or remote) specified `queue`.  The specified `handle` will be
-/// released.
-///
-/// THREAD: This method is called from the Queue's dispatcher thread.
-template <typename QueueType>
-void doDropHandle(QueueType&         queue,
-                  mqbi::QueueHandle* handle,
-                  bool               doDeconfigure)
-{
-    // executed by the *QUEUE* dispatcher thread
-
-    // PRECONDITIONS
-    BSLS_ASSERT_SAFE(handle);
-    BSLS_ASSERT_SAFE(
-        handle->queue()->dispatcher()->inDispatcherThread(handle->queue()));
-
-    BALL_LOG_SET_CATEGORY(k_LOG_CATEGORY);
-
-    // Since the handle is being dropped (which typically occurs if a client is
-    // stopping without explicitly closing its queues, or if a client crashes),
-    // we need to configure the handle's stream with null
-    // (e.g. 'consumerPriority == k_CONSUMER_PRIORITY_INVALID') parameters
-    // before we release the handle.  'configureHandle' call is eventually
-    // processed by the queue engine, which may choose to simply ignore this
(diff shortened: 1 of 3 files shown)


src/groups/mqb/mqbc/mqbc_clusternodesession.cpp
```diff
@@ -82,11 +82,14 @@ void ClusterNodeSession::teardown()
...
-         ++it) {
-        mqbi::QueueHandle* handle = it->second.d_handle_p;
+    QueueHandleMapIter qit = d_queueHandles.begin();
+
+    while (qit != d_queueHandles.end()) {
+        mqbi::QueueHandle* handle = qit->second.d_handle_p;
+        BSLS_ASSERT_SAFE(handle);
+
         handle->drop();
+        d_queueHandles.erase(qit++);
```
678098: I think this variant is more clear. It matters because we are modifying a collection when we iterate over it
```suggestion
        qit = d_queueHandles.erase(qit);
```

src/groups/mqb/mqbc/mqbc_clusternodesession.cpp
```diff
@@ -82,11 +82,14 @@ void ClusterNodeSession::teardown()
...
 
     // Release all queue handles that were associated with this session
-    for (QueueHandleMapIter it = d_queueHandles.begin();
-         it != d_queueHandles.end();
-         ++it) {
-        mqbi::QueueHandle* handle = it->second.d_handle_p;
+    QueueHandleMapIter qit = d_queueHandles.begin();
+
+    while (qit != d_queueHandles.end()) {
+        mqbi::QueueHandle* handle = qit->second.d_handle_p;
```
678098:
```suggestion
        mqbi::QueueHandle* handle_p = qit->second.d_handle_p;
```

src/groups/mqb/mqbblp/mqbblp_queue.cpp
```diff
@@ -284,17 +196,66 @@ void Queue::dropHandleDispatched(mqbi::QueueHandle* handle, bool doDeconfigure)
...
+    // (e.g. 'consumerPriority == k_CONSUMER_PRIORITY_INVALID') parameters
+    // before we release the handle.  'configureHandle' call is eventually
+    // processed by the queue engine, which may choose to simply ignore this
+    // request if handle's stream parameters are already zero (which could
+    // occur if a client crashed immediately after sending a close-queue
+    // request).  Note that we must execute the 'configureHandle' &
+    // 'releaseHandle' operations seperately for each subStream of the
+    // 'handle'.
 
-    updateStats();
```
678098, line 297: Why this call is removed?
```
void QueueState::updateStats()
{
    stats()
        ->setReaderCount(handleParameters().readCount())
        .setWriterCount(handleParameters().writeCount());
}
```
Even if we close all substreams and have 0 0 reader and writer counts, we can still report this metric.

dorjesinpo: because `Queue::releaseHandleDispatched` calls `updateStats()`

678098 (approved):

PR #770, opened 2025-06-06
Perf[BMQ,MQB]: speed up idle GC

678098:
# Background

The number of `FileStore`-s is equal to the number of partitions.
`FileStore` is a dispatcher client. All dispatcher clients support idle flush.
We schedule `FileStore::flush` to be executed if any event happens with the corresponding `partitionId`.
Under the hood, `FileStore` keeps many `ReplicatedStorage`-s (`FileBackedStorage` or `InMemoryStorage`). During `FileStore::flush` all these replicated storages are being iterated over:
https://github.com/bloomberg/blazingmq/blob/9894f6172d0fa4311900df01d376bcd590862e9a/src/groups/mqb/mqbs/mqbs_filestore.cpp#L7160-L7173

This means, that it is enough to get 1 event from any queue, and we have to iterate over all queue replicated storages associated with the same partition id, sooner (if idle) or later.

Say, if we have 10000 queues divided across 10 partitions. We have only one working queue with throughput 1 msg/s. Then, on idle we will iterate over `10000/10 = 1000` replicated storages after any storage event or deletion event from this queue.

The proposed patch doesn't remove iteration, but rather reduces the performance impact from it. Idle processing is not that scary, and we have time to think of a better solution in the future.

# Changes

- Get the current timestamp only once per iteration.
- Do not touch history metrics at all if history was not changed.
- Replace binary return code for GC result with ternary return codes (0 for no GC, negative for uncompleted GC and positive for completed GC).


src/groups/bmq/bmqc/bmqc_orderedhashmapwithhistory.h
```diff
@@ -655,21 +659,17 @@ OrderedHashMapWithHistory<KEY, VALUE, HASH, VALUE_TYPE>::insert(
...
 OrderedHashMapWithHistory<KEY, VALUE, HASH, VALUE_TYPE>::gc(TimeType now,
                                                             unsigned batchSize)
 {
     // Try to advance to either a young item, or to the end of the batch, or to
     // the end of collection.
-    // Return 'true' if at the end of the batch to be called as soon as
-    // possible.  Otherwise, return 'false' to be called at the next timeout.
-    // In either case, resume iteration from where we stopped (or wrap around).
-    // 'erase' can set the iterator back to erase item if its expiration time
-    // is sooner than the current one.
```
678098, line 668: This comment is not easy to understand, and I replaced it with return values description in declaration. Also, for better comments we should update & merge this
https://github.com/bloomberg/blazingmq/pull/498

src/groups/bmq/bmqc/bmqc_orderedhashmapwithhistory.h
```diff
@@ -692,11 +692,11 @@ OrderedHashMapWithHistory<KEY, VALUE, HASH, VALUE_TYPE>::gc(TimeType now,
 
         if (--batchSize == 0) {
             // remember where we have stopped and resume from there next time
-            return true;  // RETURN
+            return -1;  // RETURN
```
678098: Alternatively I can return negative `d_historySize - initialHistorySize` here which is equal to the original `-batchSize` (we modify `batchSize` on the fly).

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -7330,19 +7332,38 @@ void FileStore::flush()
...
+
+    if (haveMore) {
+        // Explicitly schedule 'flush()'
+        dispatcher()->execute(bdlf::BindUtil::bind(&FileStore::flush, this),
+                              this,
+                              mqbi::DispatcherEventType::e_CALLBACK);
+    }
+}
+
+void FileStore::gcStorage()
```
678098, line 7346: Note that we have `FileStore::flush` that is called on idle, and there is `FileStore::gcStorage` that is called by scheduler every 30 seconds. `FileStore::flush` only expires messages by TTL. `FileStore::gcStorage` expires messages by TTL and also GCs history.

src/groups/bmq/bmqc/bmqc_orderedhashmapwithhistory.h
```diff
@@ -277,9 +277,13 @@ class OrderedHashMapWithHistory {
...
     /// Iterate up to the specified `batchSize` of elements (both historical
     /// and live) and erase expired historical records according to the
-    /// specified `now` time.  Return `true`, if there are expired items
-    /// unprocessed because of the `batchSize` limit.
-    bool gc(TimeType now, unsigned batchSize = 0);
+    /// specified `now` time.
+    /// Return rc == 0, if no elements were GCed.
+    /// Return rc > 0, if all the needed elements were GCed and there is
+    ///                nothing more to do now.
+    /// Return rc == -1, if the maximum batch of elements was GCed, but there
```
dorjesinpo: Minor.
the motivation behind the replacing `bool` with `int` is to avoid doing `e_UPDATE_HISTORY` if `d_historySize` does not change as I understand it.  Still, it looks possible for `gc` to return `-1` _without_ changing the history (`if (it->d_isLive)`).
Overall, this seems like something unnecessary.

src/groups/mqb/mqbs/mqbs_filestore.cpp
dorjesinpo: > called by scheduler every 30 seconds

Where is the call to `scheduleEvent`/ `scheduleRecurringEvent` ?

src/groups/mqb/mqbs/mqbs_filestore.h
```diff
@@ -717,10 +717,16 @@ class FileStore BSLS_KEYWORD_FINAL : public DataStore {
...
         BSLS_KEYWORD_OVERRIDE;
 
-    /// Called by the dispatcher to flush any pending operation.. mainly
-    /// used to provide batch and nagling mechanism.
+    /// Called by the dispatcher to flush any pending operation.
     void flush() BSLS_KEYWORD_OVERRIDE;
 
+    /// Iterate over all storages and run garbage collection:
+    /// - On primary: expire messages by TTL and notify other nodes.
+    /// - On replica/primary: remove confirmed messages from deduplication
```
dorjesinpo: not necessarily confirmed messages.  Anything that got `erase`d for whatever reason

src/groups/mqb/mqbs/mqbs_filestore.h
```diff
@@ -717,10 +717,16 @@ class FileStore BSLS_KEYWORD_FINAL : public DataStore {
...
+    ///   history.
+    /// See also: gcExpiredMessages, gcHistory
+    void gcStorage();
```
dorjesinpo, line 724: The motivation for `gcStorage` is to combine `gcExpiredMessages` with `gcHistory` as I understand it.  But do we really need the latter - we already gc history when being idle _and_  when we insert.  Is there a need to `gcHistory` also on timeout (supposedly) or whenever we need to `flush` (for example when `forceFlushFileStores`)?
Logically, we may make `FileStore` agnostic of  the history which then becomes the business of `FileBackedStorage` only.  Then, there is no need for `gcStorage` (the name does not suit `gcHistory` well.  History is not a storage at the `FileStore` layer.

src/groups/mqb/mqbs/mqbs_filestore.cpp
678098: > > called by scheduler every 30 seconds
> 
> Where is the call to `scheduleEvent`/ `scheduleRecurringEvent` ?

This is the only change in `src/groups/mqb/mqbc/mqbc_storageutil.cpp`:

```
fs->execute(bdlf::BindUtil::bind(&mqbs::FileStore::gcStorage, fs));
```

src/groups/bmq/bmqc/bmqc_orderedhashmapwithhistory.h
678098: I fixed this by returning `-historyChange`
```
if (--batchSize == 0) {
    // Remember where we have stopped and resume from there next time

    const int historyChange = initialHistorySize - d_historySize;
    // Note that we return either a negative value or 0 here:
    return -historyChange;  // RETURN
}
```

So it's not possible to return non-zero value if history has not changed

src/groups/mqb/mqbs/mqbs_filestore.h
678098: Removed misleading part of the comment

src/groups/mqb/mqbs/mqbs_filestore.h
678098: > and when we insert

`gc` on insert is in another PR and is not merged yet. There can be gc call only if we run `flush` for `FileBackedStorage`.
Scheduled `gcStorage` is a safeguard against stale history memory leak. I think it's better to have a mechanism to free the memory than not to have it

678098:
@dorjesinpo made some fixes, left some comments

src/groups/mqb/mqbi/mqbi_storage.h
```diff
@@ -562,10 +562,14 @@ class Storage {
...
-    /// expired items unprocessed because of the batch limit.
-    virtual bool gcHistory() = 0;
+    /// Garbage collect a batch of expired messages from the deduplication
+    /// history, using the specified `now` as the current timestamp.
+    /// Return rc == 0, if no messages were GCed.
+    /// Return rc > 0, if all the needed messages were GCed and there is
+    ///                nothing more to do now.
+    /// Return rc < 0, if the maximum batch of elements was GCed, but there
+    ///                are more messages to GC.
+    virtual int gcHistory(bsls::Types::Int64 now) = 0;
```
dorjesinpo, line 572: after merging all PRs, do we even need to check the return codes?
we gcHistory on
1) queue `flush`
2) PUT insert, if needed
3) timeout `k_GC_MESSAGES_BATCH_SIZE`
as I understand, none of those need to reschedule?

src/groups/mqb/mqbs/mqbs_filestore.cpp
dorjesinpo: It starts in `StorageManager::onPartitionRecovery`
```
d_clusterData_p->scheduler().scheduleRecurringEvent(
            &d_gcMessagesEventHandle,
            bsls::TimeInterval(k_GC_MESSAGES_INTERVAL_SECONDS),
            bdlf::BindUtil::bind(&StorageManager::forceFlushFileStores, this));
```

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -7330,19 +7332,38 @@ void FileStore::flush()
...
-    // than the batchSize (1000), it returns 'true'.  In this case, re-enable
-    // flush client to call it again next Idle time.
-    // If it returns 'false', there is no immediate work.  Wait for the
-    // next k_GC_MESSAGES_INTERVAL_SECONDS.
-
     if (haveMore || haveMoreHistory) {
-        // Explicitly schedule 'flush()' instead of relying on idleness
-        dispatcher()->execute(bdlf::BindUtil::bind(&FileStore::flush, this),
+        // Explicitly schedule 'gcStorage()'
+        dispatcher()->execute(bdlf::BindUtil::bind(&FileStore::gcStorage,
```
dorjesinpo, line 7365: as I understand, this execution goes away when we merge the PR where we gc on insert?

dorjesinpo (approved):
delegating pending changes to https://github.com/bloomberg/blazingmq/pull/498

src/groups/mqb/mqbs/mqbs_filestore.cpp
678098: Yes

src/groups/mqb/mqbi/mqbi_storage.h
678098: Might be useful if we want to avoid extra queue history metrics updates.
Need to evaluate performance for this before making this change.

PR #771, opened 2025-06-09
Feat: Always handle SIGTERM and SIGQUIT gracefully

chrisbeard:
**Describe your changes**
SIGTERM is the standard signal to politely tell a running program to terminate.

Previously, SIGTERM (and SIQUIT) only resulted in gracefully shutdown when running the broker in a TTY. This is not ideal behavior.

Now, the broker always performs a graceful shutdown upon receiving these signals. This behavior now aligns with the intended use of these signals.

**Testing performed**
The testing framework uses signals for orchestration, need to validate this change does not impact that logic.


hallfox (dismissed):

src/python/blazingmq/dev/fuzztest/__main__.py
```diff
@@ -57,25 +57,25 @@ def launch_broker(broker_cmd: str, broker_dir: str, time_limit: float):
...
+        # should not exit with a non-zero exit code during graceful shutdown.
+        #
+        # Note on return codes:
         # When the broker is launched via 'run' script, intermediate bash shell
         # is being opened.  So the process tree looks like:
         # fuzzscript -> bash -> bmqbrkr.tsk
         # Return codes are received from the bash process, not from the
         # 'bmqbrkr.tsk'.  There are conventional exit codes for linux utility
         # applications.  In this case:
         # 128+n - Fatal error signal "n"
```
678098: If we don't need to explain where did we get `143` in `ex.returncode == 143`, we might as well remove this block of comments

src/python/blazingmq/dev/fuzztest/__main__.py
```diff
@@ -57,25 +57,25 @@ def launch_broker(broker_cmd: str, broker_dir: str, time_limit: float):
...
     except subprocess.TimeoutExpired:
         stop_broker(Path(broker_dir), BROKER_TERMINATE_TIMEOUT)
 
-    # boofuzz has an inner loop for handling exceptions.
-    # Just calling 'exit()' or 'sys.exit()' will generate SystemExit exception
-    # which will be handled - this will prevent the application from exit.
-    # That is why 'os._exit()' is used instead.
-    os._exit(-1)  # pylint: disable=W0212
+    # Broker exited gracefully (e.g. due to a stop_broker() call)
+    return
```
678098: Unnecessary `return` in the end looks a bit strange. It takes the role of the last command so the comment can be placed before it.
We can as well replace it with a meaningful print and put the comment there:
```suggestion
    print("Broker exited gracefully")  # e.g. due to a stop_broker() call
```

678098 (approved):

PR #772, opened 2025-06-09
Feat[BMQIMP]: SDK logs session instance #

dorjesinpo:
When an application creates more than one session, it is hard to filter the log.
Adding `# <N>` log prefix where `N` is the session instance number advertised in negotiation message and GUIDs.
With the possibility to modify the prefix in one place.

### DIFF
--- a/src/groups/bmq/bmqimp/bmqimp_application.h
+++ b/src/groups/bmq/bmqimp/bmqimp_application.h
@@ -276,6 +276,8 @@ class Application {
     /// Return `true` if the application is started, `false` otherwise.
     bool isStarted() const;
 
+    const SessionId& id() const;
+
     // MANIPULATORS
 
     /// Return a pointer to the blob shared pointer pool used by this instance.
@@ -339,6 +341,11 @@ inline bool Application::isStarted() const
     // the session is started
     return (state == bmqimp::BrokerSession::State::e_STARTED ||
             state == bmqimp::BrokerSession::State::e_RECONNECTING);
+}
+
+inline const SessionId& Application::id() const
+{
+    return d_brokerSession.id();
 }
 
 inline Application::BlobSpPool* Application::blobSpPool()
--- a/src/groups/bmq/bmqimp/bmqimp_application.h
+++ b/src/groups/bmq/bmqimp/bmqimp_application.h
@@ -230,6 +230,9 @@ class Application {
                         const bsl::shared_ptr<bmqp::HeartbeatMonitor>& monitor);
     void stopHeartbeat();
 
+    // PRIVATE ACCESSORS
+    const SessionId& id() const;
+
   private:
     // NOT IMPLEMENTED
     Application(const Application& other) BSLS_CPP11_DELETED;
@@ -275,8 +278,6 @@ class Application {
 
     /// Return `true` if the application is started, `false` otherwise.
     bool isStarted() const;
-
-    const SessionId& id() const;
 
     // MANIPULATORS
 
--- a/src/groups/bmq/bmqimp/bmqimp_eventqueue.h
+++ b/src/groups/bmq/bmqimp/bmqimp_eventqueue.h
@@ -133,6 +133,7 @@
 // BMQ
 
 #include <bmqimp_event.h>
+#include <bmqimp_sessionid.h>
 
 #include <bmqc_monitoredqueue_bdlccsingleproducerqueue.h>
 #include <bmqc_multiqueuethreadpool.h>
@@ -282,6 +283,8 @@ class EventQueue {
     // SpinLock to synchronize
     // 'pushBack'
 
+    const SessionId d_sessionId;
+
   private:
     // NOT IMPLEMENTED
     EventQueue(const EventQueue& other) BSLS_CPP11_DELETED;
@@ -332,6 +335,7 @@ class EventQueue {
                int                         highWatermark,
                const EventHandlerCallback& eventHandler,
                int                         numProcessingThreads,
+               const SessionId&            sessionId,
                bslma::Allocator*           allocator);
 
     /// Destructor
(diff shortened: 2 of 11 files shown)


src/groups/bmq/bmqimp/bmqimp_eventqueue.h
```diff
@@ -393,6 +397,8 @@ class EventQueue {
     /// last print.  The behavior is undefined unless the statistics were
     /// initialized by a call to `initializeStats`.
     void printStats(bsl::ostream& stream, bool includeDelta) const;
+
+    const SessionId& id() const;
```
678098: Can make this `private`

src/groups/bmq/bmqimp/bmqimp_sessionid.h
```diff
@@ -0,0 +1,82 @@
...
+#include <bmqp_ctrlmsg_messages.h>
+
+// BDE
+#include <bslim_printer.h>
+
+namespace BloombergLP {
+
+namespace bmqimp {
+
+struct SessionId {
```
678098, line 41:
```suggestion
struct SessionId BSLS_KEYWORD_FINAL {
```

src/groups/bmq/bmqimp/bmqimp_sessionid.h
```diff
@@ -0,0 +1,82 @@
...
+    // VST to hold the instance number (and more if needed)
+
+    const int d_sessionId;
+
+    SessionId();
+
+    SessionId(const bmqp_ctrlmsg::NegotiationMessage& negotiationMessage);
```
678098:
```suggestion
    explicit SessionId(const bmqp_ctrlmsg::NegotiationMessage& negotiationMessage);
```

src/groups/bmq/bmqimp/bmqimp_sessionid.h
```diff
@@ -0,0 +1,82 @@
...
+};
+
+/// Format the specified `rhs` to the specified output `stream` and return a
+/// reference to the modifiable `stream`.
+inline bsl::ostream& operator<<(bsl::ostream&            stream,
+                                const bmqimp::SessionId& rhs)
+{
+    if (rhs.d_sessionId) {
+        stream << "[#" << rhs.d_sessionId << "] ";
```
678098, line 58: 1. Extra space in the end is convenient but it assumes that something will always be printed in the `stream` after `SessionId`. It's not perfect from the design's perspective.
2. The alternative is to save `bsl::string` field with cached value of `"[#" << rhs.d_sessionId << "] "`, to avoid integer formatting every time. However, it will make `SessionId` more complex.

Can keep the current version for simplicity.

src/groups/bmq/bmqimp/bmqimp_sessionid.h
```diff
@@ -0,0 +1,82 @@
...
+
+inline SessionId::SessionId()
+: d_sessionId(0)
+{
+    // NOTHING
+}
+
+inline SessionId::SessionId(
+    const bmqp_ctrlmsg::NegotiationMessage& negotiationMessage)
+: d_sessionId(negotiationMessage.isClientIdentityValue()
```
678098, line 72: Setting a default value `0` if this is not a client identity is a robust solution. However, isn't it a problem if someone provides a negotiation message of another type to this constructor? What about adding a warning log in such a case?
Alternatively, is it possible to remove `#include <bmqp_ctrlmsg_messages.h>` and accept `int sessionId` argument in constructor?

src/groups/bmq/bmqimp/bmqimp_sessionid.h
```diff
@@ -0,0 +1,82 @@
...
+//@DESCRIPTION: Overload operator<< as a single point for logging the number.
+//
+/// Thread Safety
+///-------------
+// Thread safe.
+
+#include <bmqp_ctrlmsg_messages.h>
+
+// BDE
+#include <bslim_printer.h>
```
678098: The component declared in this header allows to print with indentation. Its features are not used in this class. Possible to replace with more specific include that declares `bsl::ostream`

src/groups/bmq/bmqimp/bmqimp_application.h
```diff
@@ -276,6 +276,8 @@ class Application {
     /// Return `true` if the application is started, `false` otherwise.
     bool isStarted() const;
 
+    const SessionId& id() const;
```
678098: Can make it private, to have a smaller public interface

678098 (commented):
A few comments and ideas

src/groups/bmq/bmqimp/bmqimp_sessionid.h
dorjesinpo: Yes, but I want to avoid serialization of the session id into `string`.

src/groups/bmq/bmqimp/bmqimp_sessionid.h
dorjesinpo: > Setting a default value `0` if this is not a client identity is a robust solution. However, isn't it a problem if someone provides a negotiation message of another type to this constructor? What about adding a warning log in such a case? Alternatively, is it possible to remove `#include <bmqp_ctrlmsg_messages.h>` and accept `int sessionId` argument in constructor?

The `SessionId` is not much of log client, too small.
The `bmqp_ctrlmsg::NegotiationMessage` as a ctor argument explicitly denotes the case when we construct it from the `bmqp_ctrlmsg::NegotiationMessage`.  What we extract and log is incapsulated inside `SessionId`.

src/groups/bmq/bmqimp/bmqimp_sessionid.h
dorjesinpo: What is the reason?  What if we want to extend it?

PR #774, opened 2025-06-10
Fix[MQB]: remove reverse connection

emelialei88:
- Remove reverse connection in integration tests and broker code.
- Add a `DummyType` for `NegotiationMessage` in case there's another type in the future and conflicts with the original `ReverseConnectionRequest` type.


src/applications/bmqstoragetool/integration-tests/data/detail_csl_result.txt
```diff
@@ -26,7 +26,8 @@
           queues = [
             [
               uri = "bmq://bmq.test.persistent.priority/my-first-queue"
- key = [ 26DACDC974 ]              partitionId = 0
+key = [ 26DACDC974 ]
```
dorjesinpo, line 29: everything in `src/applications/bmqstoragetool` seems unrelated and probably was not meant to be included

src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
```diff
@@ -859,63 +812,30 @@ int SessionNegotiator::initiateOutboundNegotiation(
...
-    // 'ClusterCatalog::initiateReversedClusterConnections' and is of type
-    // 'mqbblp::ClusterCatalog::NegotiationUserData').
     const mqbblp::ClusterCatalog::NegotiationUserData* userData =
         reinterpret_cast<mqbblp::ClusterCatalog::NegotiationUserData*>(
             context->d_initialConnectionContext_p->userData());
     BSLS_ASSERT_SAFE(userData);
 
     int rc = 0;
 
-    if (userData->d_isClusterConnection) {
```
dorjesinpo, line 886: can we assert `userData->d_isClusterConnection`?

src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp
```diff
@@ -149,26 +149,23 @@ bool TransportManager::processSession(
     if (state) {
         bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);  // d_mutex LOCK
 
-        if (state->d_isClusterConnection) {
```
dorjesinpo, line 152: is the logic based on the fact that if there is `state`, then it is outgoing connection and it cannot therefore be a `ClientSession`? 
If so, it is worth a comment

src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp
```diff
@@ -182,11 +179,7 @@ bool TransportManager::processSession(
                                  bsl::string(channelDescription.str()),
                                  state));
 
-        if (state->d_isClusterConnection) {
```
dorjesinpo, line 185: Can this be ClientSession in which case we should not call `read`?

Probably, better to restructure this methid

src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp
```diff
@@ -315,18 +308,12 @@ void TransportManager::onClose(const bsl::string& channelDescription,
         return;  // RETURN
     }
 
-    if (state->d_isClusterConnection) {
```
dorjesinpo, line 318: `state->d_node_p->resetChannel();` does not segfault because we install `onClose` callback only when there is `d_node_p`.  If that is correct, it is worth a comment.

Still, it is better to restructure the code so there is no chance we call `state->d_node_p->resetChannel();` without making sure there is `d_node_p`.

src/applications/bmqstoragetool/integration-tests/data/detail_csl_result.txt
emelialei88: Since my change include updating `bmqp_ctrlmsg.xsd`, the newly generated `.h` and `.cpp` are not compatible with these few tests.

src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
emelialei88: Just realized this `d_isClusterConnection` is not really used anymore. Maybe I should remove it.

src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp
emelialei88: Looks like it's set either during `SessionNegotiator::createSession` (incoming connection) and the connection type is not admin or client; or during `TransportManager::createCluster` (outgoing connection).

src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp
emelialei88: In case of a ClientSession, there's no `state`.

src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp
emelialei88: It seems like `TransportManager::createCluster` is the only place a `ConnectionState` is created, and `d_node_p` is always asserted to make sure it's existence before set [here](https://github.com/bloomberg/blazingmq/blob/c0ef570f0420502a8eedbc268fa9e196ca2a1a5d/src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp#L559-L571).

src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp
emelialei88: Also this is my note when tracking `state`:
*outgoing*: `TransportManager::connectLocked`: passed in argument
  - called in `TransportManager::createCluster`
  - both proxy and cluster nodes

*incoming*: `TransportManager::startListening`: passed in when calling the callback (`TransportManager::sessionResult`)
  - 1. get from `initialConnectionContext->resultState()`
    - set from
      - A. `TCPSessionFactory::handleInitialConnection` - `context->d_resultState_p` (0)
      - B. `ClusterCatalog::onNegotiationForClusterSession` - called for inter cluster node connection
  - 2. or `context->d_resultState_p` when CONNECT_FAILED

src/applications/bmqstoragetool/integration-tests/data/detail_csl_result.txt
dorjesinpo: > Since my change include updating `bmqp_ctrlmsg.xsd`, the newly generated `.h` and `.cpp` are not compatible with these few tests.

The 3 files, `detail_csl_result.txt`, `summary_csl_result.txt`, `m_bmqstoragetool_cslprinter.t.cpp`.
How are they related to the `bmqp_ctrlmsg.xsd`?
Changes seem to be formatting only and unnecessary

src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp
dorjesinpo: My understanding is that `state` exists only when the connection is outgoing.
Overall, this calls for refactoring (later).  Good indication is the presense of `reinterpret_cast`

src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp
```diff
@@ -146,29 +146,28 @@ bool TransportManager::processSession(
         negoMsg.isClientIdentityValue()
             ? negoMsg.clientIdentity()
             : negoMsg.brokerResponse().brokerIdentity();
+
+    // This is an outgoing connection to a cluster node
```
dorjesinpo: should this be under `if (state) {`?
also, can an outgoing connection be to a proxy (not a node).

src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp
dorjesinpo: Please add `BSLS_ASSERT_SAFE(d_node_p)` and a comment explaining that `TransportManager::createCluster` is the only place a ConnectionState is created and `onClose` is called

dorjesinpo (changes requested):
few very minor comments

src/applications/bmqstoragetool/integration-tests/data/detail_csl_result.txt
emelialei88: The current version generate a different printer function in `bmqp_ctrlmsg.cpp`, thus the output is different from before. Those files are hardcoded to compare against the previous format. If I don't change them the tests would fail.

src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp
emelialei88: Yeah I agree. The use of `ConnectionType` is sometime confusing as well.

src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp
emelialei88: But isn't the if statement in this function enough? Also in `TransportManager::onClusterReleased`, `d_node_p` is set to 0 when a node is deleted.

dorjesinpo (approved):
thank you

PR #775, opened 2025-06-11
Perf[MQBA]: get rid of tmp bmqt::Uri copy on every ACK

678098:
Screenshot 2025-06-11 at 17 36 03


678098:
Similar PR: https://github.com/bloomberg/blazingmq/pull/631

PR #779, opened 2025-06-13
Fix: Clang warnings related to signed-unsigned conversions

pniedzielski:
This PR fixes warnings that Clang emits due to signed-unsigned conversions.

### DIFF
--- a/src/groups/bmq/bmqa/bmqa_message.t.cpp
+++ b/src/groups/bmq/bmqa/bmqa_message.t.cpp
@@ -67,7 +67,6 @@ void generateSubQueueInfos(bmqp::Protocol::SubQueueInfosArray* subQueueInfos,
                            size_t numSubQueueInfos)
 {
     BSLS_ASSERT_SAFE(subQueueInfos);
-    BSLS_ASSERT_SAFE(numSubQueueInfos >= 0);
 
     subQueueInfos->clear();
 
@@ -75,8 +74,6 @@ void generateSubQueueInfos(bmqp::Protocol::SubQueueInfosArray* subQueueInfos,
         const unsigned int subQueueId = generateRandomInteger(0, 120);
         subQueueInfos->push_back(bmqp::SubQueueInfo(subQueueId));
     }
-
-    BSLS_ASSERT_SAFE(subQueueInfos->size() == numSubQueueInfos);
 }
 }
 


src/groups/bmq/bmqa/bmqa_message.t.cpp
```diff
@@ -64,20 +64,19 @@ int generateRandomInteger(int min, int max)
 /// `numSubQueueInfos` number of randomly generated SubQueueInfos. Note that
 /// `subQueueInfos` will be cleared.
 void generateSubQueueInfos(bmqp::Protocol::SubQueueInfosArray* subQueueInfos,
-                           int numSubQueueInfos)
+                           size_t numSubQueueInfos)
 {
     BSLS_ASSERT_SAFE(subQueueInfos);
     BSLS_ASSERT_SAFE(numSubQueueInfos >= 0);
```
678098:
```suggestion
```
Let's remove this, it's always true

src/groups/bmq/bmqa/bmqa_message.t.cpp
```diff
@@ -64,20 +64,19 @@ int generateRandomInteger(int min, int max)
...
 
-    for (int i = 0; i < numSubQueueInfos; ++i) {
+    for (size_t i = 0; i < numSubQueueInfos; ++i) {
         const unsigned int subQueueId = generateRandomInteger(0, 120);
         subQueueInfos->push_back(bmqp::SubQueueInfo(subQueueId));
     }
 
-    BSLS_ASSERT_SAFE(subQueueInfos->size() ==
-                     static_cast<unsigned int>(numSubQueueInfos));
+    BSLS_ASSERT_SAFE(subQueueInfos->size() == numSubQueueInfos);
```
678098:
```suggestion
```
Is there a reason to have this precondition here? It looks trivially true

678098 (commented):
Have a few ideas

src/groups/bmq/bmqa/bmqa_message.t.cpp
pniedzielski: I guess it's a post-condition guard, just in case the implementation changes.  I'm not sure it's doing anything important even in that way, though, so we can remove it.

src/groups/bmq/bmqa/bmqa_message.t.cpp
pniedzielski: Fixed with d5e6b4ac8

src/groups/bmq/bmqa/bmqa_message.t.cpp
pniedzielski: Fixed with d5e6b4ac8

hallfox (approved):

PR #780, opened 2025-06-13
Fix: Clang warnings related to switch statements

pniedzielski:
This PR fixes warnings that Clang emits due to missing `default` cases in switch statements.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_recoverymanager.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_recoverymanager.cpp
@@ -3443,6 +3443,16 @@ void RecoveryManager::processRecoveryEvent(
 
         case bmqp::RecoveryFileChunkType::e_UNDEFINED:
             BSLS_ASSERT_SAFE(false && "Unreachable by design.");
+            break;  // BREAK
+
+        default:
+            BMQTSK_ALARMLOG_ALARM("RECOVERY")
+                << d_clusterData_p->identity().description()
+                << ": For Partition [" << partitionId
+                << "], received unknown file chunk type: "
+                << header.fileChunkType()
+                << ", from: " << source->nodeDescription() << "."
+                << BMQTSK_ALARMLOG_END;
         }
 
         BSLS_ASSERT_SAFE(isData || isJournal || isQlist);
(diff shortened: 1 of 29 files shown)


src/groups/bmq/bmqp/bmqp_crc32c.cpp
```diff
@@ -573,20 +576,25 @@ static inline unsigned int crc32cSoftware(const unsigned char* data,
...
     if (adj > length) {
         // the specified length to process is less than the distance from
         // 'data' to the next alignment boundary.
         adj = length;
     }
 
+    // This code relies on the platform-dependent assumption that the distance
+    // from 'data' to next alignment boundary <= 7.  This is true on machines
+    // with 64-bit word sizes or smaller, but to be safe we assert this.
+    BSLS_ASSERT(adj < 8);
```
chrisbeard: This is probably correct, but may be the only "major" breaking change in this PR from what I see. To be clear, I'm fine with it. We'll revisit this code in the immediate near future with #688 (or similar, based on BDE's implementation).

chrisbeard (approved):
lgtm, a second look wouldn't hurt though

hallfox (dismissed):

src/groups/mqb/mqbblp/mqbblp_recoverymanager.cpp
```diff
@@ -1525,8 +1525,7 @@ int RecoveryManager::sendFile(RequestContext*                   context,
         break;  // BREAK
 
     case bmqp::RecoveryFileChunkType::e_UNDEFINED:
-        BSLS_ASSERT_SAFE(false);
-        break;  // BREAK
+    default: BSLS_ASSERT_SAFE(false); break;  // BREAK
```
hallfox: Should this be `BSLA_UNREACHABLE()` as well?

pniedzielski: @hallfox I don't think that the `bmqp::RecoveryFileChunkType::e_UNDEFINED` case is unreachable here, but `default` is.  Good catch.

hallfox (approved):

PR #778, opened 2025-06-13
Fix: Clang warnings related to unused or potentially unused variables

pniedzielski:
This PR fixes warnings that Clang emits due to unused entities, or entities that may be unused, depending on whether assertions are enabled or not.

Please see commit messages for more details.

PR #781, opened 2025-06-13
Fix: Miscellaneous Clang warnings

pniedzielski:
This PR fixes miscellaneous warnings that Clang emits, which are not covered by PRs #778, #779, #780, or #681.


src/groups/bmq/bmqst/bmqst_statcontexttableinfoprovider.cpp
```diff
@@ -492,9 +492,7 @@ int StatContextTableInfoProvider::getValueSize(int row, int column) const
         }
         else if (colInfo.d_doubleFunc) {
             double funcValue = colInfo.d_doubleFunc(value);
-            return bmqst::PrintUtil::printedValueLength(
-                       (bsls::Types::Int64)funcValue) +
-                   d_precision + 1;
```
hallfox, line 497: Did we lose `d_precision + 1` or is this redundant?

src/applications/bmqstoragetool/m_bmqstoragetool_filemanager.h
```diff
@@ -191,7 +191,7 @@ inline FileManagerImpl::FileHandler<ITER>::FileHandler(
 }
 
 template <typename ITER>
-inline FileManagerImpl::FileHandler<ITER>::~FileHandler()
+inline FileManagerImpl::FileHandler<ITER>::FileHandler::~FileHandler()
```
hallfox, line 194: I think this syntax doesn't compile on C++03, I believe I've been burned by this before.

src/groups/bmq/bmqst/bmqst_statcontexttableinfoprovider.cpp
pniedzielski: Hm, yeah, we want to pass it as a second argument here.  This code is reimplementing one of the overloads of `bmqst::PrintUtil::printedValueLength`, so I'm trying to call that overload instead.  Good catch.

src/applications/bmqstoragetool/m_bmqstoragetool_filemanager.h
pniedzielski: I'll give this a shot in C++03 then; if it doesn't compile, we'll have to turn that warning off.

I see in the git log that we also tried:
```
inline FileManagerImpl::FileHandler<ITER>::~FileHandler<ITER>()
```
which also doesn't compile in C++03.

src/groups/bmq/bmqst/bmqst_statcontexttableinfoprovider.cpp
pniedzielski: Fixed with commit a527c5641.

src/applications/bmqstoragetool/m_bmqstoragetool_filemanager.h
pniedzielski: We shouldn't need this to compile in C++03: the storage tool is written in C++11, and won't compile in C++03.  But our check for this only makes sure that we're not building on Solaris: building `all` in C++03 on Linux will fail with some C++11-isms in the storage tool.  I'll fix the check to make sure `CMAKE_CXX_STANDARD` is C++11 or better.

But either way, clang doesn't complain about this line in C++03.

678098: Can we solve this issue completely by placing the nested class declaration to the outer scope?

hallfox (approved):

PR #794, opened 2025-06-25
Fix[IT]: remove reverse_proxy in pytest.fixture

dorjesinpo:
removing anything reverse connection in the test_restart.py


678098:
I need this so I pressed the button

PR #659, opened 2025-03-14
Feat[MQB]: collect queue stats on replicas

678098:

PR #796, opened 2025-06-25
Refactor[MQBMOCK]: do not require buffer factory arg

678098:
Do not require `BlobBufferFactory` arg when constructing `mqbmock::Cluster`, use its own instance stored in a field instead and expose it in manipulator.

### DIFF
--- a/src/groups/mqb/mqbnet/mqbnet_multirequestmanager.t.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_multirequestmanager.t.cpp
@@ -119,33 +119,29 @@ struct Caller {
 
 class TestContext {
   private:
-    bdlbb::PooledBlobBufferFactory d_blobBufferFactory;
-    // Buffer factory provided to the various builders
+    /// Allocator to use
+    bslma::Allocator* d_allocator_p;
+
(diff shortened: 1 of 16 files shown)


678098:
Required by this PR https://github.com/bloomberg/blazingmq/pull/659

src/groups/mqb/mqbnet/mqbnet_multirequestmanager.t.cpp
```diff
@@ -235,48 +228,34 @@ class TestContext {
...
+, d_requestManager(
+      bsl::make_shared<ReqManagerType>(bmqp::EventType::e_CONTROL,
+                                       d_blobSpPool_sp.get(),
+                                       &d_cluster_mp->_scheduler(),
+                                       false,  // lateResponseMode
+                                       d_allocator_p))
+, d_multiRequestManager(
+      bsl::make_shared<MultiReqManagerType>(d_requestManager.get(),
+                                            d_allocator_p))
+, d_requestContextSp(d_multiRequestManager->createRequestContext())
```
678098, line 257: Here I reordered fields so they are constructed in the order of their dependencies and destructed in the reversed order.

PR #797, opened 2025-06-26
CI: print cores in fuzz test and add more PR prefixes

678098:


678098:
The example of cores printing is here:
https://github.com/bloomberg/blazingmq/actions/runs/15903865456/job/44854122246

PR #800, opened 2025-06-27
Fix[MQB]: pass allocator correctly

678098:


678098:
Required by this PR: https://github.com/bloomberg/blazingmq/pull/659

src/groups/mqb/mqbstat/mqbstat_clusterstats.cpp
```diff
@@ -251,7 +251,8 @@ void ClusterStats::initialize(const bsl::string&  name,
             bsl::shared_ptr<bmqst::StatContext>(
                 d_statContext_mp->addSubcontext(
                     bmqst::StatContextConfiguration(partitionName,
-                                                    &localAllocator))));
+                                                    &localAllocator)),
+                allocator));
```
678098, line 255: `shared_ptr` constructor from `ManagedPtr` also expects an allocator arg or uses a default one.

waldgange (approved):

PR #812, opened 2025-07-08
Fix[mqba::ClientSession]: improve thread safety

678098:
1. Keep `bmqu::SharedResource<ClientSession>` under `shared_ptr`. Pass this `shared_ptr` to all callbacks and don't worry if `ClientSession` was destructed in the meantime. Since we keep a copy of this `shared_ptr`, the shared resource is a correct object.
2. Do not schedule of dispatch `ClientSession` methods directly (with a few exceptions). If `ClientSession` was destructed, it is UB to call a function that has incorrect `this`, even if we don't access any of the `ClientSession` fields. Instead, we schedule events via static `ClientSessionDestructGuard` and pass a `shared_ptr` to `bmqu::SharedResource<ClientSession>`. We first try to acquire a resource, and only if we are successful we execute something on `ClientSession` while holding this resource.

### DIFF
--- a/src/groups/mqb/mqba/mqba_clientsession.cpp
+++ b/src/groups/mqb/mqba/mqba_clientsession.cpp
@@ -339,6 +339,479 @@ struct BuildAckOverflowFunctor {
 
 }  // close unnamed namespace
 
+class ClientSessionDestructGuard {
+  public:
+    // PUBLIC CLASS METHODS
+
+    static void checkUnconfirmed(
+        const ClientSession::SharedClientSessionSp& sharedSession_sp,
+        const ClientSession::ShutdownContextSp&     shutdownCtx,
+        const ClientSession::VoidFunctor&           completionCb)
+    {
+        // executed by *ANY* thread
+
+        bsl::shared_ptr<ClientSession> ptr = sharedSession_sp->acquire();
+        if (!ptr) {
+            // Session was already destroyed, just call a callback
+            completionCb();
+            return;  // RETURN
+        }
+
+        ptr->dispatcher()->execute(
+            bdlf::BindUtil::bindS(
+                ptr->d_state.d_allocator_p,
+                &ClientSessionDestructGuard::checkUnconfirmedDispatched,
+                sharedSession_sp,
+                shutdownCtx,
+                completionCb),
+            ptr.get());
+    }
+
+    static void checkUnconfirmedDispatched(
+        const ClientSession::SharedClientSessionSp& sharedSession_sp,
+        const ClientSession::ShutdownContextSp&     shutdownCtx,
+        const ClientSession::VoidFunctor&           completionCb)
+    {
+        // executed by the *CLIENT* dispatcher thread
+
+        bsl::shared_ptr<ClientSession> ptr = sharedSession_sp->acquire();
+        if (!ptr) {
+            // Session was already destroyed, just call a callback
+            completionCb();
+            return;  // RETURN
+        }
+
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqba/mqba_clientsession.cpp
```diff
@@ -3016,23 +3258,13 @@ void ClientSession::onLowWatermark()
 {
     // executed by the *IO* thread
 
-    if (!d_self.isValid()) {
```
678098, line 3019: `d_self_sp` is always valid now because we call `ClientSession::onLowWatermark` when we acquire it

src/groups/mqb/mqba/mqba_clientsession.cpp
```diff
@@ -2914,11 +3338,12 @@ void ClientSession::tearDown(const bsl::shared_ptr<void>& session,
...
-                                               this,
-                                               &semaphore,
-                                               session,
-                                               isBrokerShutdown),
+    dispatcher()->execute(bdlf::BindUtil::bindS(d_state.d_allocator_p,
+                                                &ClientSession::tearDownImpl,
+                                                this,
+                                                &semaphore,
+                                                session,
+                                                isBrokerShutdown),
```
678098, line 3350: `tearDown` sequence is one place where we don't use `ClientSessionDestructGuard`, because `d_self_sp` is already invalidated. We cannot check it anymore in `ClientSessionDestructGuard`

src/groups/mqb/mqba/mqba_clientsession.cpp
```diff
@@ -2979,27 +3408,14 @@ void ClientSession::invalidate()
     BSLS_ASSERT_SAFE(!dispatcher()->inDispatcherThread(this));
 
     dispatcher()->execute(
-        bdlf::BindUtil::bind(&ClientSession::invalidateDispatched, this),
+        bdlf::BindUtil::bindS(d_state.d_allocator_p,
+                              &ClientSession::invalidateDispatched,
+                              this),
```
678098, line 3413: This is another place where we don't use `ClientSessionDestructGuard`. We cannot acquire a shared resource to prevent destruction and invalidate it at the same time.

src/groups/mqb/mqba/mqba_clientsession.cpp
```diff
@@ -339,6 +339,479 @@ struct BuildAckOverflowFunctor {
...
+        const ClientSession::SharedClientSessionSp& sharedSession_sp,
+        const ClientSession::ShutdownContextSp&     shutdownCtx,
+        const ClientSession::VoidFunctor&           completionCb)
+    {
+        // executed by *ANY* thread
+
+        bsl::shared_ptr<ClientSession> ptr = sharedSession_sp->acquire();
+        if (!ptr) {
+            // Session was already destroyed, just call a callback
+            completionCb();
```
678098, line 356: The need to call `completionCb` according to the contract prevents from using `WeakMemFn`

src/groups/mqb/mqba/mqba_clientsession.cpp
```diff
@@ -339,6 +339,479 @@ struct BuildAckOverflowFunctor {
 
 }  // close unnamed namespace
 
+class ClientSessionDestructGuard {
```
hallfox, line 342: I don't think this class is a guard, it's not really protecting anything. I think we can document that it's a collection of helper functions for avoiding the destruction bug though.

```suggestion
struct ClientSessionUtil {
```

678098:
Decided to apply a simpler solution for now. Closing this PR so we don't lose it.

PR #813, opened 2025-07-08
Feat[bmqtool]: add autoPubSubModulo parameter to test subscriptions

dorjesinpo:
auto Producer generates message property with the name of "AUTO_PUBSUB_PROPERTY" and the value `autoIncrementedValue % d_parameters.autoPubSubModulo()` for each message 

auto Consumer generates `d_parameters.autoPubSubModulo()` subscriptions, each with expression "AUTO_PUBSUB_PROPERTY==`n`", where `n` is in the range [`0`, `d_parameters.autoPubSubModulo() - 1`]


src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -321,6 +321,40 @@ bool InputUtil::populateSubscriptions(bmqt::QueueOptions*              out,
...
+        equality += "==";
+        equality += bsl::to_string(i);
+
+        bmqt::SubscriptionExpression expression(
+            equality,
+            bmqt::SubscriptionExpression::e_VERSION_1);
+
+        to.setExpression(expression);
+
+        bsl::string error;
```
678098:
```suggestion
        bsl::string error(allocator);
```

src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -321,6 +321,40 @@ bool InputUtil::populateSubscriptions(bmqt::QueueOptions*              out,
...
+        bmqt::CorrelationId      correlationId(i);
+        bmqt::SubscriptionHandle handle(correlationId);
+
+        bsl::string equality(autoPubSubPropertyName, allocator);
+        equality += "==";
+        equality += bsl::to_string(i);
+
+        bmqt::SubscriptionExpression expression(
+            equality,
+            bmqt::SubscriptionExpression::e_VERSION_1);
```
678098, line 344: Both `bmqt::SubscriptionExpression`, `bmqt::Subscription` should support optional allocator arg, this is our overlook. I will make this change in the next PR

src/applications/bmqtool/m_bmqtool_inpututil.h
```diff
@@ -75,6 +75,11 @@ struct InputUtil {
     static bool populateSubscriptions(bmqt::QueueOptions*              out,
                                       const bsl::vector<Subscription>& in);
 
+    static bool populateSubscriptions(bmqt::QueueOptions* out,
+                                      int                 autoPubSubModulo,
+                                      const char*       autoPubSubPropertyName,
```
678098, line 84:
```suggestion
                                      const bslstl::StringRef&       autoPubSubPropertyName,
```

src/applications/bmqtool/bmqtoolcmd.xsd
```diff
@@ -246,6 +246,7 @@
       <element name='sequentialMessagePattern' type='string'  default=""/>
       <element name='messageProperties'        type='tns:MessageProperty' maxOccurs='unbounded'/>
       <element name='subscriptions'            type='tns:Subscription'    maxOccurs='unbounded'/>
+      <element name='autoPubSubModulo'         type='int'     default="0"/>
```
678098, line 249: What about making this property unsigned?
Alternatively, check that it is `>= 0`.

src/applications/bmqtool/m_bmqtool_inpututil.cpp
```diff
@@ -321,6 +321,40 @@ bool InputUtil::populateSubscriptions(bmqt::QueueOptions*              out,
...
 
+bool InputUtil::populateSubscriptions(bmqt::QueueOptions* out,
+                                      int                 autoPubSubModulo,
+                                      const char*       autoPubSubPropertyName,
+                                      bslma::Allocator* allocator)
+{
+    BSLS_ASSERT_SAFE(out);
+
+    bool failed = false;
+    for (int i = 0; i < autoPubSubModulo; ++i) {
```
678098, line 333: Do you think it's worth adding a reasonable upper limit for `autoPubSubModulo`, so we don't use all RAM? 1kk?

dorjesinpo: No, this is for testing and we want to be able to try any value

src/applications/bmqtool/bmqtoolcmd.xsd
dorjesinpo: `balcl::TypeInfo` does not support `unsigned`

678098 (approved):

PR #816, opened 2025-07-08
CI: Convert CI to use CMake Presets where possible

pniedzielski:


hallfox (dismissed):

.github/workflows/build.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
     uses: ./.github/workflows/dependencies.yaml
 
   build_ubuntu:
-    name: Build [ubuntu]
+    name: Build and Unit Test [ubuntu]
```
678098, line 23: We specially separated Build and UTs builds to speed up the CI.
The longest workflow in CI is ITs run. If ITs wait for UTs build, it becomes even slower, and it's not really necessary.

Compare these 2 workflow runs.
The old workflow: `Build [ubuntu]` takes just 8 minutes and ITs can start right after. At the same time, `UT [c++]` build and run takes 14-15 minutes in parallel.
Screenshot 2025-07-09 at 09 55 49

The new workflow. Now ITs have to wait for UTs build/run. As a result, the entire workflow run time increased from 29-30 minutes to 36 minutes.
Screenshot 2025-07-09 at 09 55 58

I think that saving 6 minutes of developer's time is worth doing.

678098: Another thing, if UTs are failing, will ITs run be skipped now?
Sometimes UTs failures are independent from ITs and do not represent a bug that must be fixed right now. Also we want to know all the errors to fix in advance. If there is an UT failure, we still want to know if there are IT failures. Re-running everything multiple times just to get to ITs is very slow.

678098: And looking at the build history:
https://github.com/bloomberg/blazingmq/actions/workflows/build.yaml?query=is%3Asuccess

In many cases the old workflow worked even faster and took just 26-28 minutes.

Screenshot 2025-07-09 at 10 09 58

PR #820, opened 2025-07-09
Fix[mqba::ClientSession]: use WeakMemFn to avoid crashes on teardown

678098:
Replace https://github.com/bloomberg/blazingmq/pull/812


dorjesinpo (approved):
thank you

PR #823, opened 2025-07-10
Feat[bmqt]: Limit queue names to 64 characters

chrisbeard:
This PR formalizes a 64 character limit on queue names.

Previously, queues that exceed this limit would warn. Now, queues that exceed this limit will be treated as invalid from the broker and therefore cannot be opened, used, etc.

### DIFF
--- a/src/groups/bmq/bmqt/bmqt_uri.cpp
+++ b/src/groups/bmq/bmqt/bmqt_uri.cpp
@@ -262,12 +262,13 @@ int UriParser::parse(Uri*                     result,
 {
     enum RcEnum {
         // Value for the various RC error categories
-        rc_SUCCESS        = 0,
-        rc_INVALID_FORMAT = -1,
-        rc_BAD_QUERY      = -2,
-        rc_MISSING_DOMAIN = -3,
-        rc_MISSING_QUEUE  = -4,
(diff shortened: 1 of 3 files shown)


src/groups/bmq/bmqt/bmqt_uri.cpp
```diff
@@ -382,20 +383,11 @@ int UriParser::parse(Uri*                     result,
...
-            bmqu::MemOutStream os;
-            os << "The queue name part of '" << uriString << "' is exceeding "
-               << "the maximum size limit of " << Uri::k_QUEUENAME_MAX_LENGTH
-               << ".  This is only a warning at the moment, but this limit "
-               << "will soon be enforced and  queue URI will be rejected!";
-            bsl::cerr << "BMQALARM [INVALID_QUEUE_NAME]: " << os.str() << '\n'
-                      << bsl::flush;
-            BALL_LOG_ERROR << os.str();
+        if (errorDescription) {
+            *errorDescription = "queue name exceeds 64 characters";
```
678098, line 387:
```suggestion
            *errorDescription = "queue name exceeds " +
                                bsl::to_string(Uri::k_QUEUENAME_MAX_LENGTH) +
                                " characters";
```

PR #829, opened 2025-07-14
Fix[Sanitizers]: turn off recovery mode for ubsan

alexander-e1off:
By default `Undefined Behaviour` sanitizer (ubsan) works in `recovery` mode: prints a verbose error report and continues execution. Since UTs are run via `ctest` with `--output-on-failure` option, we don't see ubsan errors and tests passed.
So, we need to turn off ubsan `recovery` mode to get `ctest` failed.


alexander-e1off:
> Should we see tests failing if this works?

Yes, some tests will fail now. Sanitizers are running only at PR approve, so we can skip existing ubsan failures for now and check that new ones are not appeared. I will fix/disable ubsan failures in further PRs.

hallfox (dismissed):
approving to run sanitizers

hallfox:
The expected UBSAN failures are now appearing. Can you go over a quick run down of the plan to skip the failures before we merge this?

hallfox:
Moving to #935

PR #837, opened 2025-07-17
Perf[BMQ,MQB]: simplify evaluator

678098:
Changes

1. `bmqeval::EvaluationContext`: get rid of the bool flag `d_stop` and use `d_lastError` to exit evaluation early. This also speeds up the reset of evaluation context. The idea is that if we have an error stored in the field already we must stop early, and we don't need a special bool flag for this.
2. `mqbblp::Routers`: remove unnecessary check `if (!d_evaluator.isCompiled())`. The same behaviour can be achieved by only checking `d_evaluator.isValid()`. Look at the table in the comments that explains return values of `bool Routers::Expression::evaluate()`.
3. `mqbblp::Routers`: remove unnecessary check `if (d_evaluationContext_p->hasError())`. The same can be checked in evaluator itself.
4. Remove `mutable int bmqp::MessageProperties::d_lastError`
5. `bmqp::MessageProperties::Property`: reorder fields and put `bdlb::Variant7` in the beginning to improve field packing and alignment. `sizeof` of this structure reduced from 88 bytes to 80. This struct has to be aligned by 16 byte offset, so the old version used 96 bytes if places continuously in memory, now these structures can be placed in memory without gaps because `80 = 16*5`.
6. `bmqp::MessageProperties::findProperty`: do not construct next property on stack if not needed.

### DIFF
--- a/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.cpp
+++ b/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.cpp
@@ -135,14 +135,12 @@ bool SimpleEvaluator::evaluate(EvaluationContext& context) const
 
     bdld::Datum value = d_expression->evaluate(context);
 
-    if (context.d_stop) {
+    if (context.hasError()) {
         return false;  // RETURN
     }
 
     if (!value.isBoolean()) {
-        context.d_stop      = true;
-        context.d_lastError = ErrorType::e_TYPE;
-
+        context.setError(ErrorType::e_TYPE);
         return false;  // RETURN
     }
 
@@ -182,15 +180,14 @@ SimpleEvaluator::Property::evaluate(EvaluationContext& context) const
                                                         context.d_allocator);
 
     if (value.isError()) {
-        context.d_stop = true;
-        int rc         = value.theError().code();
+        const int rc = value.theError().code();
 
-        if (rc >= ErrorType::e_EVALUATION_FIRST &&
-            ErrorType::e_EVALUATION_LAST <= rc) {
-            context.d_lastError = static_cast<ErrorType::Enum>(rc);
+        if (ErrorType::e_EVALUATION_FIRST <= rc &&
+            rc <= ErrorType::e_EVALUATION_LAST) {
+            context.setError(static_cast<ErrorType::Enum>(rc));
         }
         else {
-            context.d_lastError = ErrorType::e_UNDEFINED;
+            context.setError(ErrorType::e_UNDEFINED);
         }
     }
 
@@ -228,8 +225,7 @@ bdld::Datum
 SimpleEvaluator::UnaryMinus::evaluate(EvaluationContext& context) const
 {
     bdld::Datum expr = d_expression->evaluate(context);
-
-    if (context.d_stop) {
+    if (context.hasError()) {
         return bdld::Datum::createNull();  // RETURN
     }
 
@@ -242,9 +238,8 @@ SimpleEvaluator::UnaryMinus::evaluate(EvaluationContext& context) const
         value = expr.theInteger();
     }
     else {
-        context.d_lastError = ErrorType::e_TYPE;
-
-        return context.stop();  // RETURN
+        context.setError(ErrorType::e_TYPE);
+        return bdld::Datum::createNull();  // RETURN
     }
 
     return bdld::Datum::createInteger64(-value, context.d_allocator);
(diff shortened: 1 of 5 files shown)


src/groups/bmq/bmqeval/bmqeval_simpleevaluator.cpp
```diff
@@ -135,14 +135,12 @@ bool SimpleEvaluator::evaluate(EvaluationContext& context) const
 
     bdld::Datum value = d_expression->evaluate(context);
 
-    if (context.d_stop) {
+    if (context.hasError()) {
         return false;  // RETURN
     }
```
678098, line 140: This ensures the same behaviour as the code that was removed from `mqbblp::Routers` without adding extra performance cost:
```
        if (d_evaluationContext_p->hasError()) {
            return false;  // RETURN
        }
```

src/groups/bmq/bmqp/bmqp_messageproperties.h
```diff
@@ -805,11 +802,11 @@ MessageProperties::findProperty(const bsl::string& name) const
         if (rc) {
             // REVISIT: there are no means to report the error other than
             //          returning 'end()'
-            d_lastError = rc;
             return d_properties.end();  // RETURN
         }
 
         if (index < (d_originalNumProps - 1)) {
+            Property next;
```
678098, line 810: I am curious why do we even have this branch `if (index < (d_originalNumProps - 1))`? Is it just for validation that we can read the next property? It looks like a double work.

src/groups/bmq/bmqp/bmqp_messageproperties.h
```diff
@@ -196,9 +196,10 @@ class MessageProperties {
     // DATA
     bslma::Allocator* d_allocator_p;
 
+    /// Rb-tree containing property name->value pairs.
+    /// Note: when the number of properties is small map works faster than
+    ///       unordered_map. Also, map takes less space.
     mutable PropertyMap d_properties;
```
678098, line 202: https://github.com/bloomberg/blazingmq/blob/546fd677c4f04e66b2d5f0c09ac629362c45c60a/src/groups/bmq/bmqp/bmqp_messageproperties.h#L162

1. This is a `map`, not `unordered_map`.
2. I tried switching it to `unordered_map`, and it works much slower than `map` when we only have 1-2 properties. String hashing takes a lot of time in such cases.

The same benchmark, this function takes 4% of CPU frames when `map` is used:
```
4.08% BloombergLP::bmqp::MessageProperties::findProperty
```

If we switch it to `unordered_map`, it uses 9% of CPU frames and mostly it is string hashing.

src/groups/bmq/bmqeval/bmqeval_simpleevaluator.cpp
```diff
@@ -284,29 +279,27 @@ SimpleEvaluator::StringLiteral::evaluate(EvaluationContext& context) const
...
     if (!left.isBoolean()) {
-        context.d_lastError = ErrorType::e_TYPE;
-        return context.stop();  // RETURN
+        context.setError(ErrorType::e_TYPE);
+        return bdld::Datum::createNull();  // RETURN
     }
 
     if (left.theBoolean()) {
-        return bdld::Datum::createBoolean(true);  // RETURN
+        return left;  // RETURN
```
dorjesinpo, line 292: the return type is not `const`, are we sure we can use the same instance?
(wee seem to be ok returning `right`)

src/groups/bmq/bmqeval/bmqeval_simpleevaluator.h
```diff
@@ -711,30 +711,27 @@ class CompilationContext {
...
-    // Where to read properties from.
-    PropertiesReader* d_propertiesReader;
-
-    // The allocator to use during evaluation.
+    /// The allocator to use during evaluation.
     bslma::Allocator* d_allocator;
 
-    // Set by `Expression::evaluate` functions when a property does not
-    // have the type deduced during compilation. Stop evaluation and return
-    // `true`.
```
dorjesinpo, line 722: was the comment wrong and we return `false`?

src/groups/bmq/bmqp/bmqp_messageproperties.h
dorjesinpo: Because we calculate 'theProperty' length from two offsets unless this is the last property.

src/groups/mqb/mqbblp/mqbblp_routers.cpp
```diff
@@ -178,22 +178,20 @@ void Routers::MessagePropertiesReader::next(
...
-        return true;  // RETURN
-    }
+    /// |============|=========|========================|
+    /// | isCompiled | isValid | evaluate() = ?         |
+    /// |============|=========|========================|
+    /// | false      | false   | true                   |
+    /// | false      | true    | IMPOSSIBLE             |
+    /// | true       | false   | true                   |
+    /// | true       | true    | d_evaluator.evaluate() |
+    /// |============|=========|========================|
```
dorjesinpo, line 188: thank you for documenting this.  we could also document the result when a property is missing

src/groups/bmq/bmqeval/bmqeval_simpleevaluator.cpp
678098: We keep `left` by value, not by reference. We also return it by value. `Datum` is a simply copyable type, so value copy is safe.

There is no need to construct another `Datum` if we can simply return `left` that 1)owned by the local scope 2)is already constructed 3)has the correct `true` value.

src/groups/bmq/bmqeval/bmqeval_simpleevaluator.h
678098: The comment was wrong indeed. That is the problem with comments describing implementation in a distant part of the code.

src/groups/mqb/mqbblp/mqbblp_routers.cpp
678098: Added a comment below

dorjesinpo (approved):
thank you

PR #840, opened 2025-07-22
Fix[IT]: execute tests that do not have a consistency mark

jll63:
This fixes how CI runs the integration tests. Currently, a test that does _not_ use one of the `domain_urls` fixtures is not executed, because it does not have a consistency pytest mark. To work around this, some tests (e.g. in `test_admin_command_routing.py`) artificially "use" `domain_urls`, and do nothing with it (because they don't open any queue).

This PR adds a mark expression to `strategy.matrix.consistency` to run the tests that don't have the mark.


678098:
Can we make tests without marks to run on any mode instead, without adding a new matrix value?

678098 (approved):
Looks good. CI is also good.

pniedzielski:
#1166

Issue #749, opened 2025-05-16
Data race in `mqbblp_queueenginetester`


678098:
UT: `static void test44_throttleRedeliveryCancelledDelay()`
`QueueEngineTester` is modified from the main thread.
At the same time, it is modified from the scheduler thread (cause by handle drop, reject and throttled delivery).

PR #849, opened 2025-07-24
Protecting mock dispatched functor with mutex

dorjesinpo:
Canceling throttling in QueueEngines is done by rescheduling at 'bsls::TimeInterval()' _before_ removing the message so there is thread contention in the mock dispatcher environment.  
Adding mock::Dispatcher::d_mutex for such case(s)


678098 (approved):

678098:
https://github.com/bloomberg/blazingmq/issues/749

src/groups/mqb/mqbmock/mqbmock_dispatcher.cpp
```diff
@@ -35,6 +36,7 @@ namespace mqbmock {
 Dispatcher::Dispatcher(bslma::Allocator* allocator)
 : d_inDispatcherThread(false)
 , d_eventsForClients(allocator)
+, d_mutex()
```
678098, line 39:
```suggestion
, d_mutex()
, d_queue(d_allocator_p)
```

src/groups/mqb/mqbmock/mqbmock_dispatcher.cpp
```diff
@@ -95,25 +97,56 @@ void Dispatcher::execute(const mqbi::Dispatcher::VoidFunctor& functor,
...
+    bool run = false;
+
+    {
+        bslmt::LockGuard<bslmt::Mutex> lock(&mutex());
+
+        d_queue.push(functor);
+
+        if (1 == d_queue.size()) {
+            run = true;
+        }
```
678098:
```suggestion
        if (d_queue.empty()) {
            // The thread that pushes the first functor to the queue
            // will try to process this queue.
            run = true;
        }
        d_queue.push(functor);
```

src/groups/mqb/mqbmock/mqbmock_dispatcher.cpp
```diff
@@ -95,25 +97,56 @@ void Dispatcher::execute(const mqbi::Dispatcher::VoidFunctor& functor,
...
+        {
+            bslmt::LockGuard<bslmt::Mutex> lock(&mutex());
+
+            next = d_queue.front();
+
+            if (1 == d_queue.size()) {
+                run = false;
+            }
+
+            d_queue.pop();
```
678098:
```suggestion
            d_queue.pop();
            if (d_queue.empty()) {
                // There is nothing more to process, this thread can stop
                // after calling the last functor.
                run = false;
            }
```

src/groups/mqb/mqbmock/mqbmock_dispatcher.cpp
```diff
@@ -95,25 +97,56 @@ void Dispatcher::execute(const mqbi::Dispatcher::VoidFunctor& functor,
...
+        }
+        next();
```
678098, line 152: There is a thread race here.
`next()` call is not protected by mutex. Before we call `line 149: next();`, some other thread might call `void Dispatcher::_execute` and might call its own `next()` earlier than this one is called. It is true if `d_queue` becomes `empty()` after getting `next`.
It means that `mqbmock::Dispatcher::_execute` doesn't have FIFO guarantee.

678098 (changes requested):
A few things to think about

src/groups/mqb/mqbmock/mqbmock_dispatcher.cpp
dorjesinpo: That is correct.  There was no guarantee before and we do not add any.
Essentially, we need to run a dedicated thread

PR #850, opened 2025-07-24
CI: Modify CMakePresets to include from etc/presets

hallfox:
This is to help manage CMake preset buckets for different domains.


pniedzielski (dismissed):
This is nice; we can maybe eventually move part of `build-ubuntu` into this too.

hallfox:
Also includes presets from #816

pniedzielski (approved):

PR #851, opened 2025-07-24
IT: add raw client for IT

emelialei88:
Created an `RawClient` class to send raw requests. It took the idea from `AdminClient` and made it more extensible.


src/python/blazingmq/dev/it/process/rawclient.py
```diff
@@ -0,0 +1,214 @@
...
+        return json.loads(response_body)
+
+    def send_heartbeat_response(self) -> None:
+        """
+        Send a heartbeat response to the broker.
+        This is used to acknowledge the heartbeat request.
+        """
+        assert self._channel is not None
+        heartbeat_response = broker.HEARTBEAT_RESPONSE_SCHEMA
+        self._send_raw(self._wrap_control_event(heartbeat_response))
```
pniedzielski: I'm not sure this is right.  In the schema, `HeartbeatResponse` is a control message that is sent back in response to a `LeaderHeartbeat` as part of our Raft implementation.  It's a different sort of heartbeat, configured in docker/cluster/config/clusters.json:79 as `heartbeatBroadcastPeriodMs`, `heartbeatCheckPeriodMs`, and `heartbeatMissCount`.

The heartbeats you're detecting above in the read loop are heartbeats that we send over every TCP connection, configured in src/applications/bmqbrkr/etc/bmqbrkrcfg.json:75 as `heartbeats`.

They're named the same, unfortunately, but they're different concepts.  The first one is some sort of control message and would use `_wrap_control_event`.  But the second one, and I think they one you're interested in, has a different event type (src/groups/bmq/bmqp/bmqp_protocol.h:476).  From what I can tell here src/groups/bmq/bmqp/bmqp_protocolutil.cpp:128 and here src/groups/bmq/bmqp/bmqp_protocolutil.t.cpp:414, it has no body.

src/python/blazingmq/dev/it/process/rawclient.py
```diff
@@ -0,0 +1,214 @@
...
+            event_type = header[4] & 0b00111111
+            if not any(event_type == et.value for et in broker.EventType):
+                raise ValueError(
+                    f"Unknown event type: {event_type}, "
+                    "expected one of the EventType values"
+                )
+            if event_type != broker.EventType.HEARTBEAT_REQ:
+                print("Received event with type: ", broker.EventType(event_type).name)
+                break
+            print("Received heartbeat request.")
```
pniedzielski: This ignores heartbeats, right?  Should we respond to them?

pniedzielski (changes requested):
This is a good PR and will be very useful to us.

Only comments about the heartbeats.  I think either we take them out for the moment or fix them (correct response and then send the response too).  Otherwise good to go.

src/python/blazingmq/dev/it/process/rawclient.py
```diff
@@ -63,6 +63,25 @@ def _wrap_control_event(payload: Union[str, dict]) -> bytes:
             + padding
         )
 
+    @staticmethod
+    def _wrap_heartbeat_res_event() -> bytes:
+        """
+        Wraps the specified 'payload' with EventHeader and adds padding to the
```
pniedzielski: What `payload`?

src/python/blazingmq/dev/it/process/rawclient.py
```diff
@@ -109,10 +128,12 @@ def _receive_event(self) -> tuple[bytes, bytes]:
                     f"Unknown event type: {event_type}, "
                     "expected one of the EventType values"
                 )
-            if event_type != broker.EventType.HEARTBEAT_REQ:
+            if event_type == broker.EventType.HEARTBEAT_REQ:
+                print("Received heartbeat request.")
+                self._send_raw(self._wrap_heartbeat_res_event())
```
pniedzielski, line 129: I think this should be good.  This way, we can have a long-lived `RawClient`, and heartbeats are handled transparently.

pniedzielski (approved):

PR #854, opened 2025-07-28
Fix[MQB]: fix leader/primary divergence

waldgange:


src/integration-tests/test_connection_loss.py
```diff
@@ -155,27 +159,61 @@ def test_force_leader_primary_divergence(
...
-    leader = cluster.wait_leader()
-    assert leader.name == "west2"
+    new_leader = cluster.wait_leader()
+    assert new_leader.name == "west2"
+
+    # Now "east1" detects the leader / primary divergence. It is not a leader anymore but
+    # still primary. Hence, it is expected to shutdown itself gracefully.
+    rc = old_leader.wait()
+    assert rc == 0
+    cluster.delete_node(old_leader.name)
```
kaikulimu: Why do you need to delete `old_leader` i.e. `east1` from the cluster? `east1` is still part of the cluster.

src/integration-tests/test_connection_loss.py
```diff
@@ -155,27 +159,61 @@ def test_force_leader_primary_divergence(
...
-    for line in res.splitlines():
-        m = re.search(r"Primary Node.*\[(.+), \d+\]", line)
-        if m:
-            assert m.group(1) == "east1"
-            primaries.append(m.group(1))
-    assert len(primaries) == 4
+    admin.connect(new_leader.config.host, new_leader.config.port)
+    # Assigning primaries can take time, so we give the cluster 15 seconds for this
+    test_logger.info("Try to detect new primaries...")
+    attempts = 15
```
kaikulimu, line 196: Instead of attempting 15 times, a better way would be to sleep for 15 seconds then attempt one time.

src/python/blazingmq/dev/it/cluster.py
```diff
@@ -348,6 +348,20 @@ def stop_nodes(self, prevent_leader_bounce=False):
 
         self.last_known_leader = None
 
+    def delete_node(self, broker: Union[cfg.Broker, str]):
```
kaikulimu: Not sure if this function is needed at all. See comment above.

kaikulimu (changes requested):
Some comments

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -3457,10 +3457,20 @@ void Cluster::onClusterLeader(mqbnet::ClusterNode*                node,
...
-                : mqbstat::ClusterStats::LeaderStatus::e_FOLLOWER);
+        if (d_clusterData.membership().selfNode() == node) {
+            d_clusterData.stats().setIsLeader(
+                mqbstat::ClusterStats::LeaderStatus::e_LEADER);
+        }
+        else {
+            d_clusterData.stats().setIsLeader(
+                mqbstat::ClusterStats::LeaderStatus::e_FOLLOWER);
+            if (d_state.isSelfPrimary()) {
+                // We encountered the leader / primary divergence.
```
kaikulimu, line 3468: Print an error log before exiting

src/integration-tests/test_connection_loss.py
waldgange: Now we have a chance to complete the test earlier than in 15 seconds (usually 5 - 7 sec). If we sleep 15 seconds it will always take maximum time

src/integration-tests/test_connection_loss.py
waldgange: I've got rid of it

PR #855, opened 2025-07-28
Refactor[MQBU]: Remove unnecessary buffer from `mqbu_storagekey`

pniedzielski:
This patch removes a statically allocated buffer `k_NULL_KEY_BUFFER`
containing five null bytes from `mqbu_storagekey`.  This buffer is only
used in the default constructor for `StorageKey`, where we copy the five
null bytes into the `StorageKey`'s internal buffer.  In place of this,
we `memset` the internal buffer to `\0`.

Note that elsewhere we rely on `k_NULL_KEY_BUFFER[0] == '\0'` (see
src/groups/mqb/mqbu/mqbu_storagekey.h:312), so we can't freely change
this constant later on.

### DIFF
--- a/src/groups/mqb/mqbu/mqbu_storagekey.cpp
+++ b/src/groups/mqb/mqbu/mqbu_storagekey.cpp
@@ -29,7 +29,6 @@ BSLMF_ASSERT(StorageKey::e_KEY_LENGTH_BINARY == sizeof(StorageKey));
 // class StorageKey
 // ----------------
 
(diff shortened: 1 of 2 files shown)


pniedzielski:
Really the first byte of `d_buffer` is what tells us whether the `StorageKey` is non-null, and it should always be either `\0` or `0`, unless we use `fromHex` or `fromBinary`.  The former is only used in bmqtool and in unit tests.  The latter, though, is crucially used in src/groups/mqb/mqbs/mqbs_storageutil.cpp:63, so this is the simplest safest change.

PR #857, opened 2025-07-29
Fix[Storagetool] Add exact match filter for searching any record type by offset or sequence number

alexander-e1off:
Currently, when searching by offset or sequence number, `CONFIRM` and `DELETION` records are not found, because they are aggregated with `MESSAGE` record. This behaviour is confusing, and it was decided to search any record type  by offset or sequence number. In this case if `MESSAGE` record is found, there is no aggregation of  `CONFIRM` and `DELETION` records, only `MESSAGE` record is added to result.
Example of search result when searched  `CONFIRM` and `DELETION` records by offset or sequence number:
```
==============================

    RecordType      : CONFIRM
    Index           : 5
    Offset          : 344
    PrimaryLeaseId  : 1
    SequenceNumber  : 6
    Timestamp       : 29OCT2024_14:02:21.000000
    Epoch           : 1730210541
    QueueKey        : 26DACDC974
    QueueUri        : bmq://bmq.test.persistent.priority/test-queue
    AppKey          : 3000000000
    AppId           : ** NULL **
    GUID            : 400000000002B471F5B3AC11AA7D7DAB

==============================

    RecordType      : DELETION
    Index           : 6
    Offset          : 404
    PrimaryLeaseId  : 1
    SequenceNumber  : 7
    Timestamp       : 29OCT2024_14:02:21.000000
    Epoch           : 1730210541
    QueueKey        : 26DACDC974
    QueueUri        : bmq://bmq.test.persistent.priority/test-queue
    DeletionFlag    : NONE
    GUID            : 400000000002B471F5B3AC11AA7D7DAB
```


src/applications/bmqstoragetool/m_bmqstoragetool_searchresult.cpp
```diff
@@ -678,6 +678,181 @@ const bsl::shared_ptr<Printer>& SearchDetailResult::printer() const
...
+                                     d_printedDeletionCount,
+                                     d_printedQueueOpCount,
+                                     d_printedJournalOpCount,
+                                     d_processRecordTypes);
+}
+
+void SearchExactMatchResult::outputResult(
+    BSLA_UNUSED const GuidsList& guidFilter)
+{
+    BSLS_ASSERT_SAFE(false && "NOT SUPPORTED!");
```
waldgange, line 831: This method can only be called from SearchPartiallyConfirmedDecorator. Since this decorator is not applicable for this class, let's also reflect it in SearchResultFactory::createSearchResult()
https://github.com/bloomberg/blazingmq/blob/2ef4ce9f102d4b1a7c26de42fcbc8a0a39f7bebf/src/applications/bmqstoragetool/m_bmqstoragetool_searchresultfactory.cpp#L126-L132
let's check that `exactMatch` is false here

waldgange (changes requested):

waldgange (approved):

kaikulimu (approved):

PR #862, opened 2025-07-30
CI: Run sanitizers as part of Tests workflow

bbpetukhov:
Change trigger for Sanitizers workflow. Make it run on Integration Tests success as part of Tests workflow which in turn runs on every change to an open PR or submitting a new PR.

This change makes Sanitizer workflow results visible on the PR page.


678098 (changes requested):
CI is broken somehow in this PR. It shows only 6 CI steps, it should be 30+.

bbpetukhov:
> CI is broken somehow in this PR. It shows only 6 CI steps, it should be 30+.

Indeed there was a bug in build.yaml
Now it's fixed

PR #868, opened 2025-08-05
fix[mqbblp]: position VirtualPushStreamIterator

dorjesinpo:
Fixing bug in `VirtualPushStreamIterator` when `RelayQueueEngine` catches up with `RelayQueueEngine::d_storageIter_mp`, it may crass the latter, resulting in message processing twice.  If all other apps are gone, it will crash because the same iterator gets erased twice.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.t.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.t.cpp
@@ -1760,6 +1760,130 @@ static void test18_throttleRedeliveryNoMoreHandles()
     BMQTST_ASSERT_EQ(C3->_numMessages(), 0);
 }
 
+static void test19_redeliveryAndResume()
+// ------------------------------------------------------------------------
+// REDELIVERY AND RESUME
+//
+// Concerns:
+//   1. Verifying resume for a fanout appId..
+//
+// Plan:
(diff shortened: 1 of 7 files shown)


src/groups/mqb/mqbblp/mqbblp_relayqueueengine.t.cpp
```diff
@@ -1760,6 +1760,130 @@ static void test18_throttleRedeliveryNoMoreHandles()
...
+// ------------------------------------------------------------------------
+// REDELIVERY AND RESUME
+//
+// Concerns:
+//   1. Verifying resume for a fanout appId..
+//
+// Plan:
+//  Make the data stream of 1   2   3   4   5
+//                          |   |   |   |
+//  where   "a" state is    RL PAL  RP  |
```
678098, line 1773:
```suggestion
//  where   "a" state is    RL PAL  RL  |
```

678098 (approved):
A small comment, might merge as-is
There are also some formatter failures

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.t.cpp
dorjesinpo: `RP - Resume Point`, updated

PR #870, opened 2025-08-07
Fix[mqbnet::Channel]: flush secondary buffer on idle

678098:


src/groups/mqb/mqbnet/mqbnet_channel.h
```diff
@@ -317,13 +317,31 @@ class Channel {
...
+
+        /// Flush all builders, and transition to the next state.
+        /// - High priority items are flushed to IO.
+        e_IDLE_PRIMARY,
+
+        /// Call 'tryPopFront' on the secondary buffer until this buffer is
+        /// exhausted, and transition to the next state.
+        /// During this process:
+        /// - Low priority items are written to message builders directly.
+        e_SECONDARY,
```
678098: Alternatively, we can remove the secondary buffer that is used only for ACKs. However, it might remove back pressure on producers: they will receive ACKs faster and post new messages faster.

678098: Removing the secondary buffer:
+ Simplify the code and state management.
+ Makes impossible a situation when we never IDLE and never deliver ACKs when a node is under pressure. It means that under certain conditions the SECONDARY buffer can grow indefinitely.
- Removes back pressure on producers. Producers will receive ACKs sooner and, if they are waiting for ACK to post next, they will post more messages.
- Make a cluster less efficient during message bursts, because some of CPU time will be spent on low priority items that can be processed later.

hallfox: The back pressure point seems like a pretty major behavioral change, and this doesn't seem like too crazy of a complexity addition.

src/groups/mqb/mqbnet/mqbnet_channel.cpp
```diff
@@ -842,45 +844,67 @@ void Channel::threadFn()
                 if (d_buffer.popFront(&item) == 0) {
```
678098, line 840: It's possible to use `timedPopFront` with a timeout here and re-enter the loop every couple of seconds or so even if there is no activity. It might prevent bugs related to indefinite BLOCKs in the future.

src/groups/mqb/mqbnet/mqbnet_channel.cpp
```diff
@@ -842,45 +844,67 @@ void Channel::threadFn()
...
+            case e_IDLE_PRIMARY: {
+                // Primary buffer is exhausted, flush high priority items that
+                // are written to message builders.
                 if (flushAll(channel) == bmqio::StatusCategory::e_SUCCESS) {
-                    // Then, drain the secondary buffer
+                    // Then, drain the secondary buffer.
                     mode = e_SECONDARY;
+                    BALL_LOG_TRACE << "'" << d_description
+                                   << "': going to SECONDARY mode";
                 }
```
678098: We can also skip going to SECONDARY / IDLE_SECONDARY if secondary buffer is empty:
```
                if (flushAll(channel) == bmqio::StatusCategory::e_SUCCESS) {
                    // Then, drain the secondary buffer, if it's not empty,
                    // or circle back to BLOCK mode.
                    if (d_secondaryBuffer.isEmpty()) {
                        mode = e_BLOCK;
                        BALL_LOG_TRACE << "'" << d_description
                                       << "': going to BLOCK mode";
                    } else {
                        mode = e_SECONDARY;
                        BALL_LOG_TRACE << "'" << d_description
                                       << "': going to SECONDARY mode";
                    }
                }
```

However, if it's a normal workflow, we might expect that ACKs will always be there.

src/groups/mqb/mqbnet/mqbnet_channel.cpp
```diff
@@ -842,45 +844,67 @@ void Channel::threadFn()
...
             } break;
             case e_SECONDARY: {
                 if (d_secondaryBuffer.tryPopFront(&item)) {
```
678098: Also, we only use `d_secondaryBuffer` from one thread. It is a thread safe `MonitoredQueue`, and we can use a standard non thread safe queue or vector here to make it faster.

src/groups/mqb/mqbnet/mqbnet_channel.cpp
```diff
@@ -836,51 +835,32 @@ void Channel::threadFn()
...
                 }
             } break;
             }
         }
         else if (item->d_type == bmqp::EventType::e_UNDEFINED) {
-            // Enqueued by 'resetChannel' to wake us up.  Ignore.
+            // Enqueued by 'wakeUp', ignore this item
             item.reset();
-            // If this was 'flush', drain secondary items.
-            mode = e_IDLE;
```
678098, line 883: I think this mode transition is not really necessary.

dorjesinpo (commented):
Can we remove `isSecondary()` now?

src/groups/mqb/mqbnet/mqbnet_channel.h
dorjesinpo: I doubt, anyone uses ACKs for backpressure.  Some day, our library will implement flow control

dorjesinpo (approved):
thank you

PR #874, opened 2025-08-08
Fix[MQB]: Add missing const specifier to Item views

hallfox:
This was like a weird missing const correctness thing I noticed while reviewing another PR.


678098 (approved):
This looks good.
I also thought about these classes and I think we should eventually turn `mqbnet::Channel::Item` into a union of separate types. It's a bit weird how we keep all possible fields in `Item` now.

hallfox:
Yes, I had initially tried to do this, but the patterns of access were a bit weird. I will have to spend a bit more time on this to refactor it, but this was an easy win.

678098:
```
FAILED test_connection_loss.py::test_broker_client[multi_node_fsm-strong_consistency] - AssertionError: assert None
 +  where None = capture('CONNECTED', 5)
 +    where capture = <blazingmq.dev.it.process.client.Client object at 0x7f12f83863d0>.capture
```
The IT failure is unrelated, it is a flaky test that should be investigated and fixed

PR #876, opened 2025-08-11
CI: add manual trigger of git workflow

emelialei88:
Enable a workflow to be triggered manually

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -9,6 +9,7 @@ on:
     branches:
       - main
       - 'integration/**'
+  workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -9,6 +9,7 @@ on:
     branches:
       - main
       - 'integration/**'
+  workflow_dispatch:
```
pniedzielski, line 12: Is this enough to let you run this workflow for any commit?  A quick search of the docs makes me think it can run for different branches: https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow

PR #877, opened 2025-08-11
CI: Fix syntax error in publish workflow

pniedzielski:
Stray fancy-quotes got into this workflow, when they should have been ASCII quotes, giving a syntax error.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -14,7 +14,7 @@ env:
 jobs:
   create-github-prerelease:
     runs-on: ubuntu-latest
-    if: github.event_name == ’push’ && contains(github.ref_name, “rc”)
+    if: github.event_name == 'push' && contains(github.ref_name, "rc")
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -14,7 +14,7 @@ env:
 jobs:
   create-github-prerelease:
     runs-on: ubuntu-latest
-    if: github.event_name == ’push’ && contains(github.ref_name, “rc”)
+    if: github.event_name == 'push' && contains(github.ref_name, "rc")
```
emelialei88, line 17: Maybe do the same thing for line 35?

emelialei88 (changes requested):

.github/workflows/publish.yaml
pniedzielski: Good catch

emelialei88 (approved):

PR #880, opened 2025-08-13
Feat[CI]: python tests coverage report

waldgange:
This PR ads Github Action workflow for generating pytest coverage reports. It's launched every night according to a schedule or manually. Manual workflow also generates html and xml with report and uploads as an artifact for further inspection.


.github/workflows/pytest-coverage.yaml
```diff
@@ -0,0 +1,50 @@
...
+        run: |
+          pip install -r ${{ github.workspace }}/src/python/requirements.txt
+          cd ${{ github.workspace }}/src/python
+          pytest --cov blazingmq \
+            --cov-report term \
+            --cov-report html:python-coverage \
+            --cov-report xml:python-coverage.xml
+
+      - name: Archive code coverage results
+        if: ${{ github.event.workflow_dispatch && inputs.uploadArtifact }}
```
chrisbeard: Maybe we just do this all the time? Otherwise lgtm

waldgange: Done

chrisbeard (approved):
lgtm

PR #882, opened 2025-08-13
extending the wait on startup for followers

dorjesinpo:
Fixing the case with `STORAGE_OUT_OF_SYNC` exit in the legacy mode.

When a node does not receive sync points during the random 
`d_clusterConfig.partitionConfig().syncConfig().startupWaitDurationMs() + bsl::rand() % 5000` 
timeout and if there are no available nodes, the node assumes the storage is empty.  
When it finally receives a sync point later, it complains about `journal offset mismatch` and goes with `Ignoring this message`.  
This results in `Received storage msg JOURNAL_OP with missing sequence num.` and  `#EXIT Terminating BMQBRKR with error 'STORAGE_OUT_OF_SYNC'`

A solution could be to increase the timeout but that can negatively affect the cluster startup time since the leader does not need to wait that long.

The proposed solution is to keep the same timeout and the same timeout handling for leader.  For other (follower) nodes, however, extend the wait (up to 10 times) so that followers can process late sync points.


src/groups/mqb/mqbblp/mqbblp_recoverymanager.cpp
```diff
@@ -504,7 +507,31 @@ void RecoveryManager::recoveryStartupWaitPartitionDispatched(
...
+        }
+        else if (recoveryCtx.onStartupWaitRetry() > k_STARTUP_WAIT_RETRIES) {
+            BALL_LOG_WARN << d_clusterData_p->identity().description()
+                          << ": Partition [" << partitionId
+                          << "], not extending the wait for sync points.";
+
+            onPartitionRecoveryStatus(partitionId, 0 /* status */);
+        }
+        else {
+            BALL_LOG_WARN << d_clusterData_p->identity().description()
```
kaikulimu, line 501: I believe the logic from Line 482 to Line 508 should also be delayed for replica.

src/groups/mqb/mqbblp/mqbblp_recoverymanager.cpp
```diff
@@ -504,7 +507,31 @@ void RecoveryManager::recoveryStartupWaitPartitionDispatched(
...
         }
 
-        onPartitionRecoveryStatus(partitionId, 0 /* status */);
+        if (d_clusterData_p->electorInfo().isSelfLeader()) {
+            onPartitionRecoveryStatus(partitionId, 0 /* status */);
+        }
+        else if (recoveryCtx.onStartupWaitRetry() > k_STARTUP_WAIT_RETRIES) {
+            BALL_LOG_WARN << d_clusterData_p->identity().description()
+                          << ": Partition [" << partitionId
+                          << "], not extending the wait for sync points.";
```
kaikulimu:
```suggestion
                          << "], not extending the wait for sync points after retrying " << k_STARTUP_WAIT_RETRIES << " times.";
```

src/groups/mqb/mqbblp/mqbblp_recoverymanager.h
```diff
@@ -327,6 +332,8 @@ class RecoveryManager_RecoveryContext {
 
     void clear();
 
+    unsigned int onStartupWaitRetry();
```
kaikulimu, line 336: Document

kaikulimu (changes requested):

src/groups/mqb/mqbblp/mqbblp_recoverymanager.cpp
dorjesinpo: ok

kaikulimu (approved):
LGTM

PR #886, opened 2025-08-15
FIX[BMQ]: send nego message before create session

emelialei88:
Previously, we call `reset` on a session when `sendNegotiationMessage` fails. This will call destructor of a `Session` and assert on `!d_self.isValid()`. But this session hasn't been invalidated. We can't solve this problem by simply calling `invalidate()` since the dispatcher thread hasn't ben set up yet.

The approach here is to send the response message before a session is created.

### DIFF
--- a/src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
+++ b/src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
@@ -604,10 +604,21 @@ int SessionNegotiator::populateNegotiationContext(
     bsl::ostream&               errorDescription,
     const NegotiationContextSp& context)
 {
+    enum RcEnum {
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
```diff
@@ -683,17 +661,9 @@ void SessionNegotiator::createSession(bsl::ostream& errorDescription,
             peerIdentity.clusterNodeId());
 
         if (!clusterNode) {
-            return;  // RETURN
+            return -1;  // RETURN
```
dorjesinpo: minor, we could have an enum for return codes

dorjesinpo (approved):
thank you

PR #356, opened 2024-07-15
CI: add workflow with sanitizers check

alexander-e1off:
When PR is approved, run sanitizers check.

PR #891, opened 2025-08-21
UT: remove special handling of BDE < 4.9

678098:
We have updated the used BDE version a long time ago and now we use `4.23`.
We have a few outdated checks for `BDE >= 4.9` that are true now.


678098:
Related PR: https://github.com/bloomberg/blazingmq/pull/356

alexander-e1off (approved):
LGTM!

PR #892, opened 2025-08-21
IT[test_app_id_stats]: handle non-synchronous metrics

678098:
1. Simplify IT metrics data by compressing missing values
2. Allow `push` metrics to be any value. We don't have synchronization with pushes in the affected IT, so the value might indeed be anything and the value is not important to us.


src/python/blazingmq/dev/it/data/data_metrics.py
```diff
@@ -209,701 +237,327 @@ def check(self, value: Any) -> bool:
...
+                }
+            },
+            "foo": {
+                "values": {
+                    "queue_bytes_current": 0,
+                    "queue_confirm_time_avg": GreaterThan(0),
+                    "queue_confirm_time_max": GreaterThan(0),
+                    "queue_content_bytes": 96,
+                    "queue_content_msgs": 32,
+                    "queue_msgs_current": 0,
```
678098, line 336: Some of the values here are explicitly set to 0, because there are other `appIds` where these values are not 0

PR #893, opened 2025-08-21
Fix[MQB]: Enable/disable heartbeats under lock

dorjesinpo:
For the upcoming authentication feature.
`TCPSessionFactory::negotiationComplete` could be called from a thread (Authenticator) different from `TCPSessionFactory::onClose` (IO).
Therefore, an instantaneous disconnect could result in inconsistent heartbeat state.
Enforcing consistency by using the lock

### DIFF
--- a/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
@@ -559,7 +559,7 @@ void TCPSessionFactory::negotiationComplete(
         ++d_nbSessions;
 
         info.createInplace(d_allocator_p,
-                           channel.get(),
+                           channel,
                            monitoredSession,
                            initialConnectionContext_p->negotiationContext()
                                ->d_eventProcessor_p,
@@ -1525,10 +1525,10 @@ bool TCPSessionFactory::isEndpointLoopback(const bslstl::StringRef& uri) const
 
 TCPSessionFactory::ChannelInfo::ChannelInfo(
     const bsl::shared_ptr<bmqio::Channel>& channel_sp,
-    const bsl::shared_ptr<Session>& monitoredSession,
-    SessionEventProcessor*          eventProcessor,
-    int                             maxMissedHeartbeats,
-    int                             initialMissedHeartbeatCounter)
+    const bsl::shared_ptr<Session>&        monitoredSession,
+    SessionEventProcessor*                 eventProcessor,
+    int                                    maxMissedHeartbeats,
+    int                                    initialMissedHeartbeatCounter)
 : d_channel_sp(channel_sp)
 , d_session_sp(monitoredSession)
 , d_eventProcessor_p(eventProcessor)
--- a/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.h
+++ b/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.h
@@ -190,7 +190,7 @@ class TCPSessionFactory {
 
         bmqp::HeartbeatMonitor d_monitor;
 
-        /// @param channel The channel
+        /// @param channel_sp The channel
         /// @param session The session associated with this channel.
         /// @param eventProcessor The event processor of Events received on
         /// this channel.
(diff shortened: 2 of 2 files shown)


src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -755,6 +759,24 @@ void TCPSessionFactory::onClose(const bsl::shared_ptr<bmqio::Channel>& channel,
             channelInfo = it->second;
             d_channels.erase(it);
         }
+
+        // Synchronously remove from heartbeat monitored channels
+        if (channelInfo->d_monitor.isHearbeatEnabled() &&
```
emelialei88: This `if` block should be put inside the previous `if` block, otherwise `channelInfo` could be NULL.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -837,26 +843,30 @@ void TCPSessionFactory::onHeartbeatSchedulerEvent()
...
 
-void TCPSessionFactory::disableHeartbeat(
-    const bsl::shared_ptr<ChannelInfo>& channelInfo)
+void TCPSessionFactory::disableHeartbeat(const bmqio::Channel* channel_p)
 {
     // executed by the *SCHEDULER* thread
-    BSLS_ASSERT_SAFE(channelInfo);
-    BSLS_ASSERT_SAFE(channelInfo->d_session_sp);
+
+    const bsl::shared_ptr<ChannelInfo>& channelInfo_sp =
```
emelialei88: Instead of accessing directly, maybe we should use `find`?

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -809,11 +814,12 @@ void TCPSessionFactory::onHeartbeatSchedulerEvent()
 {
     // executed by the *SCHEDULER* thread
 
-    for (bsl::unordered_map<bmqio::Channel*, ChannelInfo*>::const_iterator it =
+    for (bsl::unordered_map<const bmqio::Channel*,
```
emelialei88: Maybe use a typedef

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
dorjesinpo: this is the case when `onHeartbeatSchedulerEvent` has already removed the context

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
dorjesinpo: it is interesting that `d_heartbeatChannels` now become `ChannelMap`

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -837,26 +842,36 @@ void TCPSessionFactory::onHeartbeatSchedulerEvent()
...
 
-void TCPSessionFactory::disableHeartbeat(
-    const bsl::shared_ptr<ChannelInfo>& channelInfo)
+void TCPSessionFactory::disableHeartbeat(const bmqio::Channel* channel_p)
 {
     // executed by the *SCHEDULER* thread
-    BSLS_ASSERT_SAFE(channelInfo);
-    BSLS_ASSERT_SAFE(channelInfo->d_session_sp);
+
+    ChannelMap::const_iterator cit = d_heartbeatChannels.find(channel_p);
```
678098, line 860: There is a thread race that might make `d_heartbeatChannels` or `ChannelMap::const_iterator cit` invalid in the middle of execution.

This sequence of events:

**IO thread**: start executing `void TCPSessionFactory::onClose`
**IO thread**: lock the mutex `d_mutex`
**IO thread**: schedule an event `TCPSessionFactory::disableHeartbeat` with `time == 0`
**IO thread**: unlock the mutex `d_mutex` and finish processing `void TCPSessionFactory::onClose`
**Main thread**: start executing `TCPSessionFactory::stop`
**Main thread**: start executing `TCPSessionFactory::stopHeartbeats`
**Main thread**: lock the mutex `d_mutex`
**Scheduler thread**: start late executing `TCPSessionFactory::disableHeartbeat`
**Scheduler thread**: assign iterator `ChannelMap::const_iterator cit = d_heartbeatChannels.find(channel_p); ` **!!!**
**Main thread**: clear the collection `d_heartbeatChannels.clear()` **!!!**

These 2 last actions might be done at the same time, so we might even try to get an iterator while we clear this collection from another thread.

dorjesinpo: Valid (and old) concern.  Refactoring to access `d_heartbeatChannels` in one thread only

PR #897, opened 2025-08-25
Fix[BMQ]: handling early channel close when authenticating

dorjesinpo:
Since `TCPSessionFactory::negotiationComplete` can happen on authentication thread and `TCPSessionFactory::onClose` - on IO thread, we need to handle early channel close (before `TCPSessionFactory::negotiationComplete`)

### DIFF
--- a/src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
+++ b/src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
@@ -295,18 +295,7 @@ void InitialConnectionHandler::complete(
     const bsl::string&                      error,
     const bsl::shared_ptr<mqbnet::Session>& session)
 {
-    // check if the channel is not closed (we can be in authentication
-    // thread)
-
-    if (!context->channel()) {
-        BALL_LOG_WARN
-            << "#TCP_UNEXPECTED_STATE InitialConnectionHandler '"
-            << context->name()
-            << "got an already closed channel at the end of negotiation ]";
-    }
-    else {
-        context->complete(rc, error, session);
-    }
+    context->complete(rc, error, session);
 }
 
 InitialConnectionHandler::InitialConnectionHandler(
--- a/src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
+++ b/src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
@@ -295,7 +295,18 @@ void InitialConnectionHandler::complete(
     const bsl::string&                      error,
     const bsl::shared_ptr<mqbnet::Session>& session)
 {
-    context->complete(rc, error, session);
+    // check if the channel is not closed (we can be in authentication
+    // thread)
+
+    if (!context->channel()) {
+        BALL_LOG_WARN
+            << "#TCP_UNEXPECTED_STATE InitialConnectionHandler '"
+            << context->name()
+            << "got an already closed channel at the end of negotiation ]";
+    }
+    else {
+        context->complete(rc, error, session);
+    }
 }
 
 InitialConnectionHandler::InitialConnectionHandler(
--- a/src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
@@ -25,12 +25,14 @@ namespace mqbnet {
 // class InitialConnectionContext
 // -------------------------------------
 
-InitialConnectionContext::InitialConnectionContext(bool isIncoming,
-                                                   const bsl::string& name)
-: d_isIncoming(isIncoming)
-, d_name(name)
-, d_resultState_p(0)
+InitialConnectionContext::InitialConnectionContext(bool isIncoming)
+: d_resultState_p(0)
 , d_userData_p(0)
+, d_channelSp()
+, d_initialConnectionCompleteCb()
+, d_negotiationCtxSp()
+, d_isIncoming(isIncoming)
+, d_isClosed(false)
 {
     // NOTHING
 }
@@ -73,12 +75,9 @@ InitialConnectionContext& InitialConnectionContext::setNegotiationContext(
     return *this;
 }
 
-void InitialConnectionContext::reset()
+void InitialConnectionContext::onClose()
 {
-    d_channelSp.reset();
-    d_initialConnectionCompleteCb = InitialConnectionCompleteCb();
-
-    // keep 'NegotiationContext'
+    d_isClosed = true;
 }
 
 bool InitialConnectionContext::isIncoming() const
@@ -118,9 +117,9 @@ InitialConnectionContext::negotiationContext() const
     return d_negotiationCtxSp;
 }
 
-const bsl::string& InitialConnectionContext::name() const
+bool InitialConnectionContext::isClosed() const
 {
-    return d_name;
+    return d_isClosed;
 }
 
 }  // close package namespace
--- a/src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
@@ -25,8 +25,10 @@ namespace mqbnet {
 // class InitialConnectionContext
 // -------------------------------------
 
-InitialConnectionContext::InitialConnectionContext(bool isIncoming)
+InitialConnectionContext::InitialConnectionContext(bool isIncoming,
+                                                   const bsl::string& name)
 : d_isIncoming(isIncoming)
+, d_name(name)
 , d_resultState_p(0)
 , d_userData_p(0)
 {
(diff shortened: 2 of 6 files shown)


src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
```diff
@@ -295,7 +295,18 @@ void InitialConnectionHandler::complete(
...
 {
-    context->complete(rc, error, session);
+    // check if the channel is not closed (we can be in authentication
+    // thread)
+
+    if (!context->channel()) {
+        BALL_LOG_WARN
+            << "#TCP_UNEXPECTED_STATE InitialConnectionHandler '"
+            << context->name()
+            << "got an already closed channel at the end of negotiation ]";
```
678098:
```suggestion
            << "#TCP_UNEXPECTED_STATE InitialConnectionHandler '"
            << context->name()
            << "' got an already closed channel at the end of negotiation";
```

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -730,16 +731,22 @@ void TCPSessionFactory::channelStateCallback(
     }
 }
 
-void TCPSessionFactory::onClose(const bsl::shared_ptr<bmqio::Channel>& channel,
-                                const bmqio::Status&                   status)
+void TCPSessionFactory::onClose(
+    bsl::shared_ptr<InitialConnectionContext> initialConnectionContext,
```
emelialei88: Why did we decide to pass the argument not as a const reference?

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -25,8 +25,10 @@ namespace mqbnet {
 // class InitialConnectionContext
 // -------------------------------------
 
-InitialConnectionContext::InitialConnectionContext(bool isIncoming)
+InitialConnectionContext::InitialConnectionContext(bool isIncoming,
+                                                   const bsl::string& name)
 : d_isIncoming(isIncoming)
+, d_name(name)
 , d_resultState_p(0)
 , d_userData_p(0)
```
678098, line 30: https://github.com/bloomberg/blazingmq/issues/916
I made an issue with my proposals how to refactor this class
Not necessary to do it in this PR

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -25,8 +25,10 @@ namespace mqbnet {
 // class InitialConnectionContext
 // -------------------------------------
 
-InitialConnectionContext::InitialConnectionContext(bool isIncoming)
+InitialConnectionContext::InitialConnectionContext(bool isIncoming,
+                                                   const bsl::string& name)
```
678098: Philosophical question, should we probably start with a `name`, and then continue with specific `isIncoming`?

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -730,16 +731,22 @@ void TCPSessionFactory::channelStateCallback(
...
+    const bmqio::Status&                      status)
 {
```
678098, line 763: IO thread?
Let's add a note (short one is good)
`// Thread: IO`

src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
```diff
@@ -295,7 +295,18 @@ void InitialConnectionHandler::complete(
...
+    }
+    else {
+        context->complete(rc, error, session);
```
678098: There is still a thread race here

The following sequence:

**AUTHN thread**: start executing `InitialConnectionHandler::complete`
**AUTHN thread**: line 301 `if (!context->channel())` => assume `channel` still exists => `if (false)` => go to `else` branch
**IO thread**: start executing `TCPSessionFactory::onClose`
**IO thread**: start executing `initialConnectionContext->reset();`
**IO thread**: in `InitialConnectionContext::reset` reset the channel: `d_channelSp.reset();`
**AUTHN thread**: in the `else` branch call `context->complete(rc, error, session);`
**AUTHN thread**: call `d_initialConnectionCompleteCb(rc, error, session, channel(), this);` **!!!** this will call the stored callback with empty `channel()` shared pointer that we cleared from IO thread **!!!**

There is another possibility.
If IO thread in this sequence manages to clear `d_initialConnectionCompleteCb`, we will crash on assertion `BSLS_ASSERT_SAFE(d_initialConnectionCompleteCb);` in `InitialConnectionContext::complete`

emelialei88: It's not shown here in this PR, but I have a mutex in `InitialConnectionHandler` protecting the caller of `InitialConnectionHandler::complete` which also guards the `initialConnectionContext->reset()` function. So this concern shouldn't be an issue.

dorjesinpo: Change of plans.
This can be safe without involving `InitialConnectionContext` lock.
We only want to make sure `TCPSessionFactory` does not insert already closed channel into `d_channels`.  Otherwise,  it is going to be stuck there.
Since both endings - `negotiationComplete` and `onClose` are in `TCPSessionFactory`, we can set/check `isClosed` flag inside `TCPSessionFactory` critical section.

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.h
```diff
@@ -119,11 +114,19 @@ class InitialConnectionContext {
     /// The NegotiationContext updated upon receiving a negotiation message.
     bsl::shared_ptr<NegotiationContext> d_negotiationCtxSp;
 
+    /// True if the session being negotiated originates
+    /// from a remote peer (i.e., a 'listen'); false if
+    /// it originates from us (i.e., a 'connect).
```
678098, line 119:
```suggestion
    /// from a remote peer (i.e., a `listen`); false if
    /// it originates from us (i.e., a `connect`).
```

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.h
```diff
@@ -140,6 +143,9 @@ class InitialConnectionContext {
     InitialConnectionContext&
     setNegotiationContext(const bsl::shared_ptr<NegotiationContext>& value);
 
+    /// Called by the IO upon `onCLose` signal
```
678098, line 146:
```suggestion
    /// Called by the IO upon `onClose` signal
```

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.h
```diff
@@ -148,6 +154,8 @@ class InitialConnectionContext {
     void*                                      resultState() const;
     const bsl::shared_ptr<bmqio::Channel>&     channel() const;
     const bsl::shared_ptr<NegotiationContext>& negotiationContext() const;
+    const bsl::string&                         name() const;
```
678098:
```suggestion
```

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -735,11 +747,17 @@ void TCPSessionFactory::channelStateCallback(
...
-                                const bmqio::Status&                   status)
+void TCPSessionFactory::onClose(
+    const bsl::shared_ptr<InitialConnectionContext>& initialConnectionContext,
+    const bmqio::Status&                             status)
 {
+    // Executed by one of the IO threads.
+
     --d_nbActiveChannels;
 
+    bsl::shared_ptr<bmqio::Channel> channel =
```
678098:
```suggestion
    bsl::shared_ptr<bmqio::Channel> &channel =
```

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -365,6 +365,14 @@ void TCPSessionFactory::handleInitialConnection(
             bdlf::PlaceHolders::_5,  // initialConnectionContext
             context));
 
+    // Register as observer of the channel to get the 'onClose'
```
678098, line 369:
```suggestion
    // Register as observer of the channel to get the `onClose`
```

src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
emelialei88: This fix doesn't work cuz previously, with the protection of the mutex for `InitialConnectionContext`, `onClose` can't reach the part where it tries to find a `ChannelInfo` until `initialConnectionComplete` finishes. And it makes sure `ClientSession::tearDown` would be called. With the change, `tearDown` won't be called and we will have a segfault `!d_self.isValid()`.

dorjesinpo: I see.
`WARN mqbnet_tcpsessionfactory.cpp:574 #TCP_UNEXPECTED_STATE TCPSessionFactory 'TCPInterface' got an already closed channel after negotiation.`
Maybe, call `tearDown` in this case?

678098 (approved):
Looks good, thanks for addressing refactor comments

PR #884, opened 2025-08-14
Build[Solaris]: replace nullptr_t() with null ptr

678098:

PR #900, opened 2025-08-26
Queue::confgiure deadlocks with QueueHandle::deconfigure

dorjesinpo:
Two threads.  Cluster and Queue.

1) OpenQueue. `Queue::configure` (executed by the cluster *DISPATCHER* thread) calls `dispatcher()->synchronize(this)`.  This enqueues `event1` on the queue dispatcher thread.  

2) It then waits for `event1` to complete.

3) ConfigureQueue. `RelayQueueEngine::configureApp` (executed by the queue *DISPATCHER* thread) calls `Cluster::configureQueue`. This enqueues `event2` on the cluster dispatcher thread.

4) `QueueHandle::~QueueHandle` (executed by the queue *DISPATCHER* thread) waits for `event2` to complete

The following sequence - 1), 3), 4), 2) - results in 1) waiting for `event1` on the queue thread and blocking the cluster thread and 4) waiting for `event2` on the cluster thread and blocking the queue thread.  Deadlock.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_queue.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_queue.cpp
@@ -62,9 +62,10 @@ namespace mqbblp {
 // class Queue
 // -----------
 
-void Queue::configureDispatched(int*          result,
-                                bsl::ostream* errorDescription,
-                                bool          isReconfigure)
+void Queue::configureDispatched(int*              result,
+                                bsl::ostream*     errorDescription,
+                                bool              isReconfigure,
+                                bslmt::Semaphore* sync)
 {
     // executed by the *QUEUE* dispatcher thread
 
@@ -87,7 +88,12 @@ void Queue::configureDispatched(int*          result,
     }
 
     if (result) {
+        BSLS_ASSERT_SAFE(sync);
         *result = rc;
+    }
+
+    if (sync) {
+        sync->post();
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/groups/mqb/mqbblp/mqbblp_queue.cpp
```diff
@@ -584,18 +590,29 @@ int Queue::configure(bsl::ostream* errorDescription_p,
...
+
+        sync.wait();
+    }
+    else {
+        dispatcher()->execute(bdlf::BindUtil::bind(&Queue::configureDispatched,
+                                                   this,
+                                                   bsl::nullptr_t(),
+                                                   bsl::nullptr_t(),
+                                                   isReconfigure,
+                                                   bsl::nullptr_t()),
```
678098: Will not compile on `Solaris`. You might want to use `static_cast<Type*>(0)`
See also https://github.com/bloomberg/blazingmq/pull/884

dorjesinpo: No other way but `static_cast`?
Anyway, this can be refactored

src/groups/mqb/mqbblp/mqbblp_queue.cpp
```diff
@@ -62,33 +62,58 @@ namespace mqbblp {
...
     else if (d_remoteQueue_mp) {
-        rc = d_remoteQueue_mp->configure(errStream, isReconfigure);
+        rc = d_remoteQueue_mp->configure(*errorDescription, isReconfigure);
     }
     else {
         BSLS_ASSERT_OPT(false && "Uninitialized queue");
     }
 
     if (result) {
+        BSLS_ASSERT_SAFE(sync);
```
pniedzielski: Duplicate of line 76 above.

src/groups/mqb/mqbblp/mqbblp_queue.cpp
```diff
@@ -62,33 +62,58 @@ namespace mqbblp {
...
         *result = rc;
     }
+
+    if (sync) {
```
pniedzielski: According to the precondition in line 76 this will be true, no need for `if`.

src/groups/mqb/mqbblp/mqbblp_queue.cpp
678098: 1. `bsl::nullptr_t` will not compile with `Bind` on Solaris
2. `NULL` will not compile with `Bind` on Linux

We have to specify a type for a pointer to help deduce binding so we either use

1. `static_cast<TYPE*>(0)`
2. Named constant like `static const TYPE* emptyType = NULL;`

pniedzielski (approved):

PR #901, opened 2025-08-27
Feat[STATS]: Report csl replication time

waldgange:


src/groups/mqb/mqbstat/mqbstat_clusterstats.cpp
```diff
@@ -68,6 +68,10 @@ struct ClusterStatsIndex {
         // Value: Leader status of cluster, non-zero (1) implies leader
         //        and 0 implies follower
         ,
+        e_CSL_REPLICATION_TIME_MCS
+        // Value: time in microseconds it took for replication of a new entry
+        // in CSL file.
```
chrisbeard: I believe all of our other latency/time-based stats use nanoseconds for the unit

kaikulimu (commented):
I verify those are the correct times to calculate CSL replication time

kaikulimu (approved):

PR #902, opened 2025-08-27
Fix: Accept greater leaseId primary unconditionally

dorjesinpo:
Fixing a race in legacy mode by accepting primary status advisory if it has greater lease id unconditionally

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp
@@ -1573,25 +1573,38 @@ void ClusterOrchestrator::processPrimaryStatusAdvisory(
         // 'StorageMgr::processPrimaryStatusAdvisoryDispatched' as well.
 
         if (pinfo.primaryNode()) {
-            if (pinfo.primaryNode() != source) {
+            if (pinfo.primaryLeaseId() > primaryAdv.primaryLeaseId()) {
+                BALL_LOG_ERROR
+                    << d_clusterData_p->identity().description()
+                    << ": Partition [" << primaryAdv.partitionId()
+                    << "]: received primary status advisory: " << primaryAdv
+                    << " from perceived primary: " << source->nodeDescription()
+                    << ", but with a lesser leaseId. Self perceived leaseId: "
+                    << pinfo.primaryLeaseId() << ". Will ignore the advisory.";
+                return;  // RETURN
+            }
+            else if (pinfo.primaryLeaseId() < primaryAdv.primaryLeaseId()) {
+                BALL_LOG_WARN
+                    << d_clusterData_p->identity().description()
+                    << ": Partition [" << primaryAdv.partitionId()
+                    << "]: received primary status advisory: " << primaryAdv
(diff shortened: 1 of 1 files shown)


src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp
```diff
@@ -1573,25 +1573,38 @@ void ClusterOrchestrator::processPrimaryStatusAdvisory(
...
+                    << ": Partition [" << primaryAdv.partitionId()
+                    << "]: received primary status advisory: " << primaryAdv
+                    << " from perceived primary: " << source->nodeDescription()
+                    << ", with a greater leaseId. Self perceived leaseId: "
+                    << pinfo.primaryLeaseId()
+                    << ". Will follow the advisory unconditionally.";
+            }
+            else if (pinfo.primaryLeaseId() != primaryAdv.primaryLeaseId()) {
+                BSLS_ASSERT_SAFE(pinfo.primaryLeaseId() ==
+                                 primaryAdv.primaryLeaseId());
```
kaikulimu: Logical fallacy: `else if (pinfo.primaryLeaseId() != primaryAdv.primaryLeaseId())` does not make any sense here. Do you mean `else if (pinfo.primary()->nodeId() != source->nodeId())`?

I prefer comparing node Ids over ClusterNode pointers.

kaikulimu (changes requested):

src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp
dorjesinpo: yes, I meant `pinfo.primaryNodeId() == source->nodeId()`. thank you

src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp
```diff
@@ -1573,25 +1573,38 @@ void ClusterOrchestrator::processPrimaryStatusAdvisory(
...
+                BALL_LOG_WARN
+                    << d_clusterData_p->identity().description()
+                    << ": Partition [" << primaryAdv.partitionId()
+                    << "]: received primary status advisory: " << primaryAdv
+                    << " from perceived primary: " << source->nodeDescription()
+                    << ", with a greater leaseId. Self perceived leaseId: "
+                    << pinfo.primaryLeaseId()
+                    << ". Will follow the advisory unconditionally.";
+            }
+            else if (pinfo.primaryNodeId() == source->nodeId()) {
```
kaikulimu: So in the case where `pinfo.primaryNode() != 0` and  `pinfo.primaryLeaseId() ==  primaryAdv.primaryLeaseId()` but `pinfo.primaryNodeId() != source->nodeId()`, this logic will go ahead and update primary node within the same leaseId. It seems wrong. We should reject in that case.

kaikulimu (changes requested):

dorjesinpo:
Interversion test [https://github.com/bloomberg/blazingmq/actions/runs/17303868402](https://github.com/bloomberg/blazingmq/actions/runs/17303868402)

PR #906, opened 2025-09-02
Fix[mqbs_inmemorystorage.cpp]: remove empty WARN log

678098:
`BALL_LOG_WARN_BLOCK` unconditionally makes a log record even if empty

```
02SEP2025_16:01:17.959 (140361930364672) WARN mqbs_inmemorystorage.cpp:121
```

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_inmemorystorage.cpp
+++ b/src/groups/mqb/mqbs/mqbs_inmemorystorage.cpp
@@ -118,12 +118,9 @@ int InMemoryStorage::configure(BSLA_UNUSED bsl::ostream& errorDescription,
 
 void InMemoryStorage::setConsistency(const mqbconfm::Consistency& value)
 {
-    BALL_LOG_WARN_BLOCK
-    {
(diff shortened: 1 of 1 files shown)


src/groups/mqb/mqbs/mqbs_inmemorystorage.cpp
```diff
@@ -118,12 +118,9 @@ int InMemoryStorage::configure(BSLA_UNUSED bsl::ostream& errorDescription,
 
 void InMemoryStorage::setConsistency(const mqbconfm::Consistency& value)
 {
-    BALL_LOG_WARN_BLOCK
-    {
-        if (value.isStrongValue()) {
-            BALL_LOG_OUTPUT_STREAM << "Trying to configure strong consistency "
-                                   << "for in-memory storage";
-        }
+    if (value.isStrongValue()) {
```
678098, line 121: This is a rare code path and I don't think it's worth to add PREDICT_UNLIKELY hints

dorjesinpo (approved):

PR #909, opened 2025-09-03
Feat[STATS]: Partition replication time

waldgange:


dorjesinpo (commented):
Few thoughts:
1.  Need to double check if `e_PARTITION_REPLICATION` metric for the weak consistency justifies the price of extra Receipts traffic.
2.  Rather than introducing extra `StorageMessageAttributes` field, (suppose the answer to the 1. is yes), Replicas can just send Receipt unconditionally.  Primary can add PUTs to `d_unreceipted` unconditionally (with a new flag, maybe).  Primary then can skip calling `onReplicatedBatch` for weak consistency messages.
3.  Backward compatibility.  Looks like (but need to double-check and test) extra Receipts do not hurt (older versions)
4.  There is `mqbs::ReplicatedStorage::isStrongConsistency` which _may_ be useful.

Wrt 1. the consensus is to NOT use the new metric for weak consistency yet.  That should reduce the scope of this PR

src/integration-tests/test_admin_client.py
```diff
@@ -375,6 +376,9 @@ def test_app_id_stats(multi_node: Cluster, domain_urls: tc.DomainUrls) -> None:
 
     _ = _change_leader(cluster)
 
+    # wait for at least one snapshot to initialize stats
+    time.sleep(1)
```
waldgange, line 380: This is not related to the partition replication time. It fixes the flaky `test_admin_client.py::test_app_id_stats` that sometimes fails with error: `RuntimeError: Path: after-confirm.values.queue_consumers_count, 0 != 3 (expected)`

dorjesinpo: @678098 fyi

src/groups/mqb/mqbblp/mqbblp_localqueue.cpp
```diff
@@ -484,24 +484,16 @@ void LocalQueue::postMessage(const bmqp::PutHeader&              putHeader,
...
-            ackMessage
-                .setStatus(bmqp::ProtocolUtil::ackResultToCode(
-                    mqbi::StorageResult::toAckResult(res)))
-                .setMessageGUID(putHeader.messageGUID());
-            // CorrelationId & QueueId are left unset as those fields will
-            // be filled downstream.
-
-            source->onAckMessage(ackMessage);
-        }
+    if (res != mqbi::StorageResult::e_SUCCESS || (haveReceipt && doAck)) {
```
dorjesinpo, line 487: 👍

PR #910, opened 2025-09-03
ci: Add problem matchers to catch CI issues

pniedzielski:
[Problem matchers][gh-actions-problem-matchers] are a poorly documented feature of GitHub Actions that allow a CI workflow to scan the output for given regex and generate annotations and log file decorations from it.  This pull request adds a problem matcher to catch issues from Clang and GCC while building BlazingMQ, and a problem matcher to catch issues from Doxygen while generating API documentation.

[gh-actions-problem-matchers]: https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md "Problem Matchers"

---

Annotations can appear inline in the diff:
Screenshot 2025-09-03 at 1 47 40 PM

The action run lists annotations, including warnings, up to 200 total (10 per job):
Screenshot 2025-09-03 at 1 47 06 PM

Errors and warnings are highlighted in the actions log:
Screenshot 2025-09-03 at 1 46 42 PM

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -345,7 +345,9 @@ jobs:
 
       - name: Build docs
         run: |
+          echo "::add-matcher::.github/workflows/problem-matchers/doxygen.json"
           doxygen Doxyfile
+          echo "::remove-matcher owner=doxygen"
 
   shellcheck:
     name: Shellcheck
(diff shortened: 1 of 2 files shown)


pniedzielski:
We can extend this later with more problem matchers, but this pull request just does GCC and Clang compilation.

678098 (approved):
Looks cool!

PR #911, opened 2025-09-03
Refactor[bmqu_memoutstream.t.cpp]: bdesb [169517258]

678098:


src/groups/bmq/bmqu/bmqu_memoutstream.t.cpp
```diff
@@ -38,12 +38,12 @@ static void test1_breathingTest()
     // remains working as efficiently as possible (over time as code
     // evolves).  However these strict numbers here are not the artifacts
     // of how this stream works, but how the BDE provided
-    // `bdesb_MemOutStreamBuf` works.  So it is possible that changes in
-    // `bdesb_MemOutStreamBuf` will break this test driver.  Said changes
+    // `bdlsb::MemOutStreamBuf` works.  So it is possible that changes in
```
pniedzielski: Could this be a Doxygen comment and use `\bbref`?

678098: Updated. Do we generate Doxygen for UTs?

pniedzielski: Not in CI right now (only API docs in CI), but I think we are able to locally.

pniedzielski (approved):

PR #913, opened 2025-09-04
Feat[BMQT]: better verbosity for bmqt::QueueId::isValid

678098:
Current logs do not give enough information why queueId is not valid
```
ERROR bmqa_session.cpp:1392 BMQA.SESSION Invalid QueueId 
```


src/groups/bmq/bmqa/bmqa_queueid.cpp
```diff
@@ -148,6 +148,7 @@ QueueId::print(bsl::ostream& stream, int level, int spacesPerLevel) const
     printer.start();
     printer.printAttribute("uri", uri());
     printer.printAttribute("correlationId", correlationId());
+    printer.printAttribute("state", d_impl_sp->state());
```
678098: This one is questionable, do we want to expose the implementation state?

PR #920, opened 2025-09-05
Fix[MQB]: fix possible int32 overflow in queue handle parameters

678098:
1. Add check `bool QueueState::canMerge` to prevent possible int32 overflow
2. `mqbblp::RootQueueEngine`, `mqbblp::RelayQueueEngine`: check that handle parameters can be merged without signed integer overflow before merging
3. `mqbblp::QueueEngineTester`: extend to support `adminCount`
4. UTs `mqbblp::RootQueueEngine`, `mqbblp::RelayQueueEngine`: remove repetitive disable default allocator checks and set it in one place. This should remove ~`(47+20) * 4 = 268` lines of code
5. UTs `mqbblp::RootQueueEngine`, `mqbblp::RelayQueueEngine`: add UTs to check reasonable and unreasonable read/write/admin counters setup

The added UTs are failing in the main branch and passing with this PR's fix

```
TEST /blazingmq/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.t.cpp CASE 20

Error /blazingmq/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp(1114): bmqp::QueueUtil::isValidSubset(queueHandle->handleParameters(), d_queueState_p->handleParameters())    (failed)
libc++abi: terminating due to uncaught exception of type BloombergLP::bsls::AssertTestException
zsh: abort      ./src/groups/mqb/mqbblp_relayqueueengine.t 20

TEST /blazingmq/src/groups/mqb/mqbblp/mqbblp_rootqueueengine.t.cpp CASE 47

Error /blazingmq/src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp(829): bmqp::QueueUtil::isValidSubset(queueHandle->handleParameters(), d_queueState_p->handleParameters())    (failed)
libc++abi: terminating due to uncaught exception of type BloombergLP::bsls::AssertTestException
zsh: abort      ./src/groups/mqb/mqbblp_rootqueueengine.t 47
```

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
@@ -1048,6 +1048,14 @@ mqbi::QueueHandle* RelayQueueEngine::getHandle(
         return 0;  // RETURN
     }
 
+    if (!d_queueState_p->canMerge(handleParameters)) {
+        CALLBACK(bmqp_ctrlmsg::StatusCategory::E_REFUSED,
+                 -1,
+                 "Reached maximum read/write/admin counters for a queue",
+                 0);
+        return 0;  // RETURN
+    }
+
     mqbi::QueueHandle* queueHandle =
         d_queueState_p->handleCatalog().getHandleByRequester(
             *clientContext,
--- a/src/groups/mqb/mqbblp/mqbblp_queueenginetester.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_queueenginetester.cpp
@@ -180,11 +180,11 @@ static void parseMessages(bsl::vector<bsl::string>* messages,
 /// attributes are populated with default values.  The format of attributes
 /// must be as follows:
 ///
-///   "[readCount=<N>] [writeCount=<M>] [isFinal=(true|false)]"
+///  "[readCount=<N>] [writeCount=<M>] [adminCount=<K>] [isFinal=(true|false)]"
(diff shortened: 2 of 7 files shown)


src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1048,6 +1048,14 @@ mqbi::QueueHandle* RelayQueueEngine::getHandle(
         return 0;  // RETURN
     }
 
+    if (!d_queueState_p->canMerge(handleParameters)) {
+        CALLBACK(bmqp_ctrlmsg::StatusCategory::E_REFUSED,
```
678098, line 1052: Strictly speaking, the handle parameters are not INVALID, so it is not correct to return invalid parameters return code. On the other hand, we expose some information about the current handle counters. Should we allow this?
It is possible to write a client that will probe the broker with different counters and depending on the result it will know the exact counters of the handle on the broker side

dorjesinpo: does not look like a concern.  would we want to address that (at least partially), we could add jitter and then refused return would mean that the counter is above certain threshold without known the exact value.

dorjesinpo (approved):
just nitpicking, we could do this without a macro (an the expense of more (explicit) code) and without `static_cast`.
I am fine with the way it is now too.

PR #921, opened 2025-09-05
IT test_rollover_csl.py: Prologue/swtich/epilogue structure; update apps

kaikulimu:
Enhancements to `test_rollover_csl.py`:
1. Instead of hardcoding to open 10 queues, open as many queues as needed to cause rollover.
2. Adopt a easier-to-follow Prologue/Swtich/Epilogue structure.
3. The tests now adds and removes appId before rolling over.

### DIFF
--- a/src/integration-tests/test_rollover_csl.py
+++ b/src/integration-tests/test_rollover_csl.py
@@ -50,16 +50,16 @@ class TestRolloverCSL:
         assert len(csl_files_before_rollover) == 1
 
         i = 0
-        while True:
+        # Open queues until rollover detected
+        while not leader.outputs_regex(r"Rolling over from log with logId", 0.01):
             producer.open(
                 f"bmq://{domain_priority}/q{i}", flags=["write,ack"], succeed=True
             )
             producer.close(f"bmq://{domain_priority}/q{i}", succeed=True)
             i += 1
-
-            # Rollover detected
-            if leader.outputs_regex(r"Rolling over from log with logId", 0.01):
-                break
+            assert i < 10000, (
+                "Failed to detect rollover after opening a reasonable number of queues"
+            )
         test_logger.info(f"Rollover detected after opening {i} queues")
 
         csl_files_after_rollover = glob.glob(
@@ -89,10 +89,10 @@ class TestRolloverCSL:
             Then, restart the cluster.
         3. EPILOGUE:
             - priority consumer gets the second message
-            - "quux" gets the third message
-            - "bar" gets 0 messages
             - "foo" gets 2 messages
+            - "bar" gets 0 messages
             - "baz" gets 3 messages
+            - "quux" gets the third message
         """
         du = domain_urls
         leader = cluster.last_known_leader
@@ -100,10 +100,8 @@ class TestRolloverCSL:
         producer = proxy.create_client("producer")
 
         # PROLOGUE
-        priority_queue, fanout_queue = (
-            f"bmq://{du.domain_priority}/q_in_use",
-            f"bmq://{du.domain_fanout}/q_in_use",
-        )
+        priority_queue = f"bmq://{du.domain_priority}/q_in_use"
+        fanout_queue = f"bmq://{du.domain_fanout}/q_in_use"
         for queue in [priority_queue, fanout_queue]:
             producer.open(queue, flags=["write,ack"], succeed=True)
             producer.post(
@@ -129,8 +127,8 @@ class TestRolloverCSL:
         consumer.close(priority_queue, succeed=True)
         consumer.close(fanout_queue + "?id=foo", succeed=True)
 
-        all_app_ids = default_app_ids + ["quux"]
-        cluster.set_app_ids(all_app_ids, du)
+        current_app_ids = default_app_ids + ["quux"]
+        cluster.set_app_ids(current_app_ids, du)
         producer.post(
             fanout_queue,
             ["msg3"],
@@ -138,11 +136,13 @@ class TestRolloverCSL:
             wait_ack=True,
         )
 
-        cluster.set_app_ids([a for a in all_app_ids if a not in ["bar"]], du)
+        current_app_ids.remove("bar")
+        cluster.set_app_ids(current_app_ids, du)
 
         # SWITCH
         i = 0
-        while True:
+        # Open queues until rollover detected
+        while not leader.outputs_regex(r"Rolling over from log with logId", 0.01):
             producer.open(
                 f"bmq://{du.domain_priority}/q_dummy_{i}",
                 flags=["write,ack"],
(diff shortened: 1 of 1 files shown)


src/integration-tests/test_rollover_csl.py
```diff
@@ -46,103 +44,183 @@ def test_csl_cleanup(self, cluster: Cluster, domain_urls: tc.DomainUrls):
...
 
         # Rollover and queueUnAssignmentAdvisory interleave
-        assert leader.outputs_regex(r"queueUnAssignmentAdvisory", timeout)
+        assert leader.outputs_regex(r"queueUnAssignmentAdvisory", timeout=5)
 
         cluster.restart_nodes()
         # For a standard cluster, states have already been restored as part of
         # leader re-election.
         if cluster.is_single_node:
             self.producer.wait_state_restored()
```
678098: `self.producer` -> `producer`
This one is causing ITs failure

kaikulimu: Fixed

src/integration-tests/test_rollover_csl.py
```diff
@@ -46,103 +44,183 @@ def test_csl_cleanup(self, cluster: Cluster, domain_urls: tc.DomainUrls):
...
-
         csl_files_before_rollover = glob.glob(
             str(cluster.work_dir.joinpath(leader.name, "storage")) + "/*csl*"
         )
         assert len(csl_files_before_rollover) == 1
 
-        # opening 10 queues would cause a rollover
-        for i in range(0, 10):
+        i = 0
+        while True:
```
678098:
```suggestion
        # Open queues until rollover detected
        while not leader.outputs_regex(r"Rolling over from log with logId", 0.01):
```

src/integration-tests/test_rollover_csl.py
```diff
@@ -46,103 +44,183 @@ def test_csl_cleanup(self, cluster: Cluster, domain_urls: tc.DomainUrls):
...
             producer.open(
                 f"bmq://{domain_priority}/q{i}", flags=["write,ack"], succeed=True
             )
             producer.close(f"bmq://{domain_priority}/q{i}", succeed=True)
+            i += 1
+
+            # Rollover detected
+            if leader.outputs_regex(r"Rolling over from log with logId", 0.01):
+                break
```
678098:
```suggestion
            assert i < 10000, "Failed to detect rollover after opening a reasonable number of queues"
```
If we refactor `FileStore` and change the logged line, this IT will hang up forever. We need to limit this loop somehow, either limit the number of iterations or limit the total time.

src/integration-tests/test_rollover_csl.py
```diff
@@ -46,103 +44,183 @@ def test_csl_cleanup(self, cluster: Cluster, domain_urls: tc.DomainUrls):
...
+            - remove "bar"
+        2. SWITCH:
+            By opening and GC'ing queues, cause the CSL file to rollover.
+            Then, restart the cluster.
+        3. EPILOGUE:
+            - priority consumer gets the second message
+            - "quux" gets the third message
+            - "bar" gets 0 messages
+            - "foo" gets 2 messages
+            - "baz" gets 3 messages
```
678098: Let's keep chronological order of appIds?
```suggestion
            - "foo" gets 2 messages
            - "bar" gets 0 messages
            - "baz" gets 3 messages
            - "quux" gets the third message
```

src/integration-tests/test_rollover_csl.py
```diff
@@ -46,103 +44,183 @@ def test_csl_cleanup(self, cluster: Cluster, domain_urls: tc.DomainUrls):
...
 
-        # Cause three QueueAssignmentAdvisories and QueueUnassignedAdvisories to be written to the CSL.  These records will be erased during rollover.
-        for i in range(0, 3):
-            self.producer.open(
-                f"bmq://{du.domain_priority}/q{i}", flags=["write,ack"], succeed=True
+        # PROLOGUE
+        priority_queue, fanout_queue = (
+            f"bmq://{du.domain_priority}/q_in_use",
+            f"bmq://{du.domain_fanout}/q_in_use",
+        )
```
678098: Might be simpler
```suggestion
        priority_queue = f"bmq://{du.domain_priority}/q_in_use"
        fanout_queue = f"bmq://{du.domain_fanout}/q_in_use"
```

src/integration-tests/test_rollover_csl.py
```diff
@@ -46,103 +44,183 @@ def test_csl_cleanup(self, cluster: Cluster, domain_urls: tc.DomainUrls):
...
+        all_app_ids = default_app_ids + ["quux"]
+        cluster.set_app_ids(all_app_ids, du)
+        producer.post(
+            fanout_queue,
+            ["msg3"],
+            succeed=True,
+            wait_ack=True,
+        )
+
+        cluster.set_app_ids([a for a in all_app_ids if a not in ["bar"]], du)
```
678098:
```suggestion
        all_app_ids.remove("bar")
        cluster.set_app_ids(all_app_ids, du)
```
Might as well remove `all_app_ids` -> `app_ids_ext` or `app_ids`

src/integration-tests/test_rollover_csl.py
```diff
@@ -46,103 +44,183 @@ def test_csl_cleanup(self, cluster: Cluster, domain_urls: tc.DomainUrls):
...
+
+        # SWITCH
+        i = 0
+        while True:
```
678098: Possible to put rollover regex search as the condition (like the comment above)

src/integration-tests/test_rollover_csl.py
```diff
@@ -46,103 +44,183 @@ def test_csl_cleanup(self, cluster: Cluster, domain_urls: tc.DomainUrls):
...
-        # do not wait for success, otherwise the following capture will fail
-        leader.force_gc_queues(block=False)
+            if i % 5 == 0:
+                # do not wait for success, otherwise the following capture will fail
+                leader.force_gc_queues(block=False)
 
-        assert leader.outputs_regex(r"Rolling over from log with logId", timeout)
+            # Rollover detected
+            if leader.outputs_regex(r"Rolling over from log with logId", 0.01):
+                break
```
678098: It's also preferable to have reasonable failure conditions in this loop

src/integration-tests/test_rollover_csl.py
```diff
@@ -46,103 +44,183 @@ def test_csl_cleanup(self, cluster: Cluster, domain_urls: tc.DomainUrls):
...
-            consumer.open(
-                f"bmq://{du.domain_fanout}/q0?id={app_id}",
-                flags=["read"],
-                succeed=True,
-            )
+        test_logger.info("Verify that removed app 'bar' cannot receive any message")
+        consumer.open(
+            fanout_queue + "?id=bar",
+            flags=["read"],
+            succeed=True,
```
678098, line 199: How can we open a `bar` consumer if `bar` was removed from appIds?

src/integration-tests/test_rollover_csl.py
kaikulimu: Renamed `all_app_ids` as `current_app_ids`

src/integration-tests/test_rollover_csl.py
kaikulimu: The implementation of BMQ is that consumer `bar` can still open the queue, but won't receive any messages. This stems from a common practical usage where an application team opens a consumer on an unauthorized app, then adds the app to domain config; that way, the newly authorized app won't have unconfirmed messages piling up and slowing down the whole queue.

If you have a better way to verify that the app is successfully removed, please let me know.

678098: It's possible to use admin client to ask for domain configuration, but it is too much for this PR. Can keep it simple.

PR #922, opened 2025-09-08
Feat[STATS]: Report partition offsets

waldgange:
This PR is introducing new partition metrics `e_PARTITION_DATA_OFFSET_BYTES` and `e_PARTITION_JOURNAL_OFFSET_BYTES` representing current offset in data and journal files respectively. This offset is literally the amount of bytes written to the file, as well as the position where the next record will be written


src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -3977,6 +3977,9 @@ int FileStore::issueSyncPointInternal(SyncPointType::Enum type,
         d_config.partitionId(),
         fs->d_outstandingBytesData,
         fs->d_outstandingBytesJournal);
+    d_clusterStats_p->setPartitionOffsetBytes(d_config.partitionId(),
```
bbpetukhov: Maybe it's more efficient to report everything all together? I mean to create a single method instead of 2: setPartitionOutstandingBytes and setPartitionOffsetBytes ?

src/groups/mqb/mqbstat/mqbstat_clusterstats.cpp
```diff
@@ -196,6 +202,18 @@ bsls::Types::Int64 ClusterStats::getValue(const bmqst::StatContext& context,
...
     }
+    case Stat::e_PARTITION_DATA_OFFSET: {
+        const bsls::Types::Int64 value =
+            STAT_RANGE(rangeMax, e_PARTITION_DATA_OFFSET_BYTES);
+        return value == bsl::numeric_limits<bsls::Types::Int64>::min() ? 0
+                                                                       : value;
+    }
+    case Stat::e_PARTITION_JOURNAL_OFFSET: {
+        const bsls::Types::Int64 value =
+            STAT_RANGE(rangeMax, e_PARTITION_JOURNAL_OFFSET_BYTES);
```
678098: Maybe we should report the latest `STAT_SINGLE` value here instead?
This situation:
`partition[0]: offset 1,000,000`
`partition[0]: rollover`
`partition[0]: offset 100`
And we will report offset `1,000,000` 5 minutes after rollover until old stats go away.

src/groups/mqb/mqbstat/mqbstat_clusterstats.cpp
```diff
@@ -415,6 +433,27 @@ ClusterStats::setPartitionOutstandingBytes(int                partitionId,
     return *this;
 }
 
+ClusterStats&
+ClusterStats::setPartitionOffsetBytes(int                partitionId,
+                                      bsls::Types::Int64 dataBytes,
+                                      bsls::Types::Int64 journalBytes)
```
678098: Looked at the code and opened this proposal, not necessary to make this refactor in your PR
https://github.com/bloomberg/blazingmq/issues/923

src/groups/mqb/mqbs/mqbs_filestore.cpp
waldgange: Efficiency is not affected here as precondition checks in `BSLS_ASSERT_SAFE(...)` macro are compiled and hence evaluated only in debug mode. Any way I combined the methods into one for readability

src/groups/mqb/mqbstat/mqbstat_clusterstats.cpp
waldgange: Changed it to report the last snapshot value

src/groups/mqb/mqbs/mqbs_filestore.cpp
678098: We compile with `SAFE` assert level on all modes, including the default `RelWithDebInfo`

waldgange: I double checked it. My apologies. Indeed, we compile with all asserts enabled even for release. Is it intended? I would say we are doing something wrong. What's the point of using different assert macros if all of them are always enabled? BDE [documentation](https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/group__bsls__assert.html#bsls_assert-selecting-which-assert-macro-to-use) says `BSLS_ASSERT_SAFE(...)` is for for heavy evaluations that can affect algorithmic complexity of a function, so it should be enabled only for tests and disabled for production.

kaikulimu:
@waldgange Can you put in the description what exact metric are you reporting?

kaikulimu:
This is reporting at each timestamp what is the write offset of journal and data files.

src/groups/mqb/mqbs/mqbs_filestore.cpp
678098: Yes, it is intended. We lose a bit of performance, but we can find inconsistent states (if any) early which is helpful for finding possible bugs. Initially we mostly used `SAFE` level for (somewhat) "unnecessary" paranoid checks. Sometimes, we can even check the same thing multiple times: for example `func1()` check for `inDispatcherThread()` and then calls `func2()` that also checks for `inDispatcherThread()`. We don't have heavy computations in `SAFE` checks, instead, we have these checks on busy paths which affects performance.

678098: We can just switch to `OPT` level everywhere, but it will generate a huge diff.

kaikulimu (approved):
LTGM, but please confirm with Evgenii before merging.

src/groups/mqb/mqbs/mqbs_filestore.cpp
waldgange: Thanks for the clarifications @678098 !

PR #926, opened 2025-09-10
Fix Partition FSM: Clear buffered live data upon ReplicaDataRequestPush

kaikulimu:
In FSM mode, as a healing replica, upon `e_REPLICA_DATA_RQST_PUSH` event, we *must* clear buffered live data because it is possible that the buffered data contains duplicate records as the recovery data primary is about to send us. This PR adds the buffer clearing logic and includes a corresponding integration test for replica late join scenario while live messages are flowing.

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
+++ b/src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
@@ -1646,6 +1646,21 @@ int RecoveryManager::loadBufferedStorageEvents(
     return rc_SUCCESS;
 }
 
+void RecoveryManager::clearBufferedStorageEvent(int partitionId)
+{
+    // executed by the *QUEUE DISPATCHER* thread associated with 'partitionId'
+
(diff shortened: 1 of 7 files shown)


src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
```diff
@@ -1646,6 +1646,21 @@ int RecoveryManager::loadBufferedStorageEvents(
...
+
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(0 <= partitionId);
+
+    RecoveryContext& recoveryCtx = d_recoveryContextVec[partitionId];
+    recoveryCtx.d_bufferedEvents.clear();
+
+    BALL_LOG_INFO << d_clusterData.identity().description() << " Partition ["
+                  << partitionId
+                  << "]: " << "Cleared all buffered storage events.";
```
dorjesinpo, line 1661: minor, may help logging the number of events cleared

PR #929, opened 2025-09-12
Fix [mqbc]: Fix FSM mode to synchronize node after missed rollover

alexander-e1off:
This PR fixes broker behaviour in `FSM` mode for the scenario if replica missed the rollover:

- Current behaviour: it is failed to synchronize storage, becomes out of sync with primary and aborted;
- New behaviour: it synchronizes storage with primary;

The following changes in behaviour are introduced:
- bmqp_ctrlmsg.xsd schema is enhanced for PrimaryStateRequest/PrimaryStateResponse and ReplicaStateRequest/ReplicaStateResponse messages with `firstSyncPointAfterRolloverSequenceNumber` field;
- `primary` uses `firstSyncPointAfterRolloverSequenceNumber` to detect that `replica` is missed rollover(s) and sends `REPLICA_DATA_RQST_DROP` event to it;
- `replica` on receiving of `REPLICA_DATA_RQST_DROP` drops its storage and requests from primary to synchronize storage;
- `primary` sends its storage content to replica

**Testing performed**
- new integration test is added to test scenario with missed rollover.


src/groups/bmq/bmqp/bmqp_ctrlmsg.xsd
```diff
@@ -697,32 +697,36 @@
...
         This type represents a request sent to the replica by the primary to
-        ask for replica's sequence number. The primary also sends its own
-        sequence number as part of this request.
+        ask for replica's sequence numbers. The primary also sends its own
+        sequence numbers as part of this request.
 
         partitionId:    partition id for corresponding partition.
-        sequenceNumber: Primary's sequence number for corresponding partition.
+        latestSequenceNumber: Primary's latest sequence number for corresponding partition.
+        firstSyncPointAfterRolloverSequenceNumber: Primary's first sync point after rollover sequence number for corresponding partition.
```
kaikulimu:
```suggestion
        firstSyncPointAfterRolloverSequenceNumber: Sequence number of primary's first sync point after rollover for corresponding partition.
```

src/groups/bmq/bmqp/bmqp_ctrlmsg.xsd
```diff
@@ -697,32 +697,36 @@
...
     <annotation>
       <documentation>
         This type represents a response sent by a replica to the primary along
-        with its sequence number.
+        with its sequence numbers.
 
         partitionId:    partition id for corresponding partition.
-        sequenceNumber: Replica's sequence number for corresponding partition.
+        latestSequenceNumber: Replica's latest sequence number for corresponding partition.
+        firstSyncPointAfterRolloverSequenceNumber: Replica's first sync point after rollover sequence number for corresponding partition.
```
kaikulimu:
```suggestion
        firstSyncPointAfterRolloverSequenceNumber: Sequence number of replica's first sync point after rollover for corresponding partition.
```

src/groups/bmq/bmqp/bmqp_ctrlmsg.xsd
```diff
@@ -731,31 +735,35 @@
       <documentation>
         This type represents a request sent to the primary by a replica to
         ask for primary's sequence number. The replica also sends it own
-        sequence number as part of this request.
+        sequence numbers as part of this request.
 
         partitionId:    partition id for corresponding partition.
-        sequenceNumber: Replica's sequence number for corresponding partition.
+        latestSequenceNumber: Replica's latest sequence number for corresponding partition.
+        firstSyncPointAfterRolloverSequenceNumber: Replica's first sync point after rollover sequence number for corresponding partition.
```
kaikulimu:
```suggestion
        firstSyncPointAfterRolloverSequenceNumber: Sequence number of replica's first sync point after rollover for corresponding partition.
```

src/groups/bmq/bmqp/bmqp_ctrlmsg.xsd
```diff
@@ -731,31 +735,35 @@
...
     <annotation>
       <documentation>
         This type represents a response sent by a primary to the replica along
-        with its sequence number.
+        with its sequence numbers.
 
         partitionId:    partition id for corresponding partition.
-        sequenceNumber: Primary's sequence number for corresponding partition.
+        latestSequenceNumber: Primary's latest sequence number for corresponding partition.
+        firstSyncPointAfterRolloverSequenceNumber: Primary's first sync point after rollover sequence number for corresponding partition.
```
kaikulimu:
```suggestion
        firstSyncPointAfterRolloverSequenceNumber: Sequence number of primary's first sync point after rollover for corresponding partition.
```

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -3615,12 +3615,20 @@ void Cluster::processResponse(const bmqp_ctrlmsg::ControlMessage& response)
     // A response must have an associated request Id
     BSLS_ASSERT_SAFE(!response.rId().isNull());
 
-    dispatcher()->execute(
-        bdlf::BindUtil::bind(&Cluster::processResponseDispatched,
-                             this,
-                             response,
-                             static_cast<mqbnet::ClusterNode*>(0)),  // source
-        this);
+    if (dispatcher()->inDispatcherThread(this)) {
```
kaikulimu: I checked the callers of `Cluster::processResponse`. They are **always** on the *cluster dispatcher* thread. Hence, there is no need to even redirect to `processResponseDispatched` at all. I think back in the day this function was executed by the *IO* thread, but it is no longer the case.

I suspect it is the same case for [`ClusterProxy::processResponse`](https://github.com/bloomberg/blazingmq/blob/c2c2d3fd38ac36330cb9748645d43cc755a82fef/src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp#L802). We can remove the indirection to `processResponseDispatched`.

You can add `BSLS_ASSERT_SAFE(dispatcher()->inDispatcherThread(this));` in both functions. Please double check.

bin/build-darwin.sh
```diff
@@ -97,10 +97,10 @@ mkdir -p "${DIR_INSTALL}"
 
 # :: Clone dependencies :::::::::::::::::::::::::::::::::::::::::::::::::::::::
 if [ ! -d "${DIR_THIRDPARTY}/bde-tools" ]; then
-    git clone --depth 1 --branch 4.23.0.0 https://github.com/bloomberg/bde-tools "${DIR_THIRDPARTY}/bde-tools"
+    git clone --depth 1 --branch 4.28.0.0 https://github.com/bloomberg/bde-tools "${DIR_THIRDPARTY}/bde-tools"
```
kaikulimu, line 100: Why are you bumping the versions here?

If it is needed, could you please make a separate small PR to do so?

src/groups/mqb/mqbc/mqbc_partitionfsm.h
```diff
@@ -94,6 +94,11 @@ class PartitionFSMEventData {
     /// associated partitionId.
     bmqp_ctrlmsg::PartitionSequenceNumber d_partitionSequenceNumber;
 
+    /// Partition first sync point after rollover sequence number as sent by
+    /// the `d_source_p` node for the associated partitionId.
```
kaikulimu:
```suggestion
    /// Sequence number of the first sync point after rollover as sent by
    /// the `d_source_p` node for the associated partitionId.
```

src/groups/mqb/mqbc/mqbc_storagemanager.h
```diff
@@ -163,17 +163,39 @@ class StorageManager BSLS_KEYWORD_FINAL
...
+    /// VST representing node's sequence number, first sync point after
+    /// rollover sequence number and flag of whether recovery data is in sync.
+    class NodeSeqNumContext {
+      public:
+        // DATA
+
+        /// Node's latest sequence number.
+        bmqp_ctrlmsg::PartitionSequenceNumber d_seqNum;
+
+        /// Node's latest sequence number.
```
kaikulimu:
```suggestion
        /// Sequence number of node's first sync point after rollover.
```

src/groups/mqb/mqbc/mqbc_recoverymanager.h
```diff
@@ -202,6 +202,10 @@ class RecoveryManager {
         /// Receive data context.
         ReceiveDataContext d_receiveDataContext;
 
+        /// First sync point after rollover sequence number.
+        bmqp_ctrlmsg::PartitionSequenceNumber
+            d_firstSyncPointAfterRolloverSeqNum;
```
kaikulimu, line 207: `mqbc::StorageManager::NodeSeqNumContext` already has the data member `d_firstSyncPointAfterRolloverSeqNum`; it looks redundant to have the same data member here. Is it possible to remove this data member, and retrieve the first sync point seq num freshly during [`RecoveryManager::recoverSeqNum`](https://github.com/alexander-e1off/blazingmq/blob/9d9139eca137125a30e04c6caf9f2c2f38ee4dbc/src/groups/mqb/mqbc/mqbc_recoverymanager.cpp#L1535)?

src/groups/mqb/mqbs/mqbs_filestore.h
```diff
@@ -936,10 +942,19 @@ class FileStore BSLS_KEYWORD_FINAL : public DataStore {
...
     void setIgnoreCrc32c(bool value);
 
-    // This will be used as Implicit Receipt
+    /// This will be used as Implicit Receipt
     void setLastStrongConsistency(unsigned int        primaryLeaseId,
                                   bsls::Types::Uint64 sequenceNum);
 
+    /// Set the first sync point after rollover sequence number from the
+    /// specified `seqNum` to d_firstSyncPointAfterRolloverSeqNum member;
+    void setFirstSyncPointAfterRolloverSeqNum(
```
kaikulimu: Unused. Please remove.

src/groups/mqb/mqbs/mqbs_filestore.h
```diff
@@ -936,10 +942,19 @@ class FileStore BSLS_KEYWORD_FINAL : public DataStore {
...
+        const bmqp_ctrlmsg::PartitionSequenceNumber& seqNum);
+
+    /// Set the first sync point after rollover offset from the specified
+    /// `offset` to JournalFileHeader.d_firstSyncPointOffset;
+    void setFirstSyncPointAfterRolloverOffset(bsls::Types::Uint64 offset);
```
kaikulimu: Unused. Please remove.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -1322,26 +1337,28 @@ void StorageManager::do_storeSelfSeq(const PartitionFSMArgsSp& args)
...
-    nodeSeqNumCtx.second = false;
+    nodeSeqNumCtx.d_firstSyncPointAfterRolloverSeqNum =
+        getSelfFirstSyncPointAfterRolloverSequenceNumber(partitionId);
+    nodeSeqNumCtx.d_isInSync = false;
 
     BALL_LOG_INFO << d_clusterData_p->identity().description()
                   << ": In Partition [" << partitionId << "]'s FSM, "
                   << "storing self sequence number as "
-                  << d_nodeToSeqNumCtxMapVec[partitionId][selfNode].first;
+                  << d_nodeToSeqNumCtxMapVec[partitionId][selfNode].d_seqNum;
```
kaikulimu:
```suggestion
                  << nodeSeqNumCtx.d_seqNum;
```

kaikulimu: Also, can print `nodeSeqNumCtx.d_firstSyncPointAfterRolloverSeqNum` here.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -1932,7 +1985,16 @@ void StorageManager::do_replicaDataRequestPush(const PartitionFSMArgsSp& args)
         if (cit->first->nodeId() == selfNode->nodeId()) {
             continue;  // CONTINUE
         }
-        if (cit->second.first <= selfSeqNum && !cit->second.second) {
+
+        // Skip node with empty storage and different first sync point after
```
kaikulimu:
```suggestion
        // Skip node with non-empty storage and different first sync point after
```

src/groups/mqb/mqbc/mqbc_storagemanager.h
```diff
@@ -163,17 +163,39 @@ class StorageManager BSLS_KEYWORD_FINAL
...
+        bmqp_ctrlmsg::PartitionSequenceNumber
+            d_firstSyncPointAfterRolloverSeqNum;
+
+        /// Flag of whether recovery data is in sync.
+        bool d_isInSync;
```
kaikulimu:
```suggestion
        /// Flag of whether recovery data is already sent to that node. It, however, does that not mean the node is already healed.
        bool d_isRecoveryDataSent;
```
It helps to clarify.

src/integration-tests/test_sync_after_rollover_fsm.py
```diff
@@ -0,0 +1,244 @@
...
+    tweak,
+    start_cluster,
+    fsm_multi_cluster,
+)
+
+from blazingmq.dev import paths
+
+
+def _run_storage_tool(journal_file: Path, mode: str) -> subprocess.CompletedProcess:
+    """Run storage tool on journal file in specified mode."""
```
kaikulimu:
```suggestion
    """Run storage tool on `journal_file` in the specified `mode`."""
```

src/integration-tests/test_sync_after_rollover_fsm.py
```diff
@@ -0,0 +1,244 @@
...
+        ],
+        capture_output=True,
+        check=True,
+    )
+
+
+def _compare_journal_files(
+    leader_name: str, replica_name: str, cluster: Cluster
+) -> None:
+    """Compare leader and replica journal files content."""
```
kaikulimu:
```suggestion
    """Compare leader and replica journal files content, and assert that they are equal."""
```

src/integration-tests/test_sync_after_rollover_fsm.py
```diff
@@ -0,0 +1,244 @@
...
+
+        # Run storage tool on replica journal file in "summary" mode to check journal file headers
+        replica_res = _run_storage_tool(replica_file, "summary")
+        assert replica_res.returncode == 0
+
+        # Check that content of leader and replica journal files is equal
+        assert leader_res.stdout == replica_res.stdout
+
+
+@tweak.cluster.partition_config.max_journal_file_size(884)
```
kaikulimu: `884` is a magic number. Please use a constant.

src/integration-tests/test_sync_after_rollover_fsm.py
```diff
@@ -0,0 +1,244 @@
...
+    assert replica.outputs_substr("Cluster (itCluster) is available", 10)
+
+    assert leader == cluster.last_known_leader
+
+    # Check that leader and replica journal files are equal
+    _compare_journal_files(leader.name, replica.name, cluster)
+
+
+@start_cluster(False)
+@tweak.cluster.partition_config.max_journal_file_size(884)
```
kaikulimu: `884` is a magic number. Please use a constant.

src/integration-tests/test_sync_after_rollover_fsm.py
```diff
@@ -0,0 +1,244 @@
...
+        # Run storage tool on replica journal file in "summary" mode to check journal file headers
+        replica_res = _run_storage_tool(replica_file, "summary")
+        assert replica_res.returncode == 0
+
+        # Check that content of leader and replica journal files is equal
+        assert leader_res.stdout == replica_res.stdout
+
+
+@tweak.cluster.partition_config.max_journal_file_size(884)
+def test_synch_after_missed_rollover(
```
kaikulimu:
```suggestion
def test_sync_after_missed_rollover(
```

src/integration-tests/test_sync_after_rollover_fsm.py
```diff
@@ -0,0 +1,244 @@
...
+
+    assert leader == cluster.last_known_leader
+
+    # Check that leader and replica journal files are equal
+    _compare_journal_files(leader.name, replica.name, cluster)
+
+
+@start_cluster(False)
+@tweak.cluster.partition_config.max_journal_file_size(884)
+def test_synch_after_missed_rollover_after_restart(
```
kaikulimu:
```suggestion
def test_sync_after_missed_rollover_after_restart(
```

src/integration-tests/test_sync_after_rollover_fsm.py
```diff
@@ -0,0 +1,244 @@
...
+        consumer.confirm(uri_priority, "*", succeed=True)
+
+    # Stop replica
+    replica.stop()
+    cluster.make_sure_node_stopped(replica)
+    replica.drain()
+
+    # Put more messages w/o confirm to initiate the rollover
+    for i in range(3, 8):
+        producer.post(uri_priority, [f"msg{i}"], succeed=True, wait_ack=True)
```
kaikulimu:
```suggestion
    i = 3
    while not leader.outputs_substr("Initiating rollover", 0.01):
        producer.post(uri_priority, [f"msg{i}"], succeed=True, wait_ack=True)
        i += 1
```
This way, we post as many messages as needed until rollover. Please do the same for the `test_synch_after_missed_rollover_after_restart` test.

src/integration-tests/test_sync_after_rollover_fsm.py
```diff
@@ -0,0 +1,244 @@
...
+        node.stop()
+        cluster.make_sure_node_stopped(node)
+        node.drain()
+
+    # Start all nodes, leader `east1` is the last to start
+    # to initiate `ReplicaStateRequest` to all replicas
+    for node in (east2, west1, west2):
+        node.start()
+        node.wait_until_started()
+        node.set_quorum(5)
```
kaikulimu, line 241: Not sure if you have observed a race where one of these nodes have transitioned to leader before its quorum is set to 5. In that case, it will continue to be the leader and `east1` will **not** become the leader.

If you never observe this race, then it's fine to leave as is.

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
alexander-e1off: Removed indirect call of `processResponseDispatched` and call it directly. `processResponseDispatched` checks that it is called from dispatcher thread.

src/groups/mqb/mqbc/mqbc_recoverymanager.h
alexander-e1off: This member is used for caching the value and avoid heavy [mqbs::FileStoreUtil::loadIterators()](https://github.com/alexander-e1off/blazingmq/blob/9d9139eca137125a30e04c6caf9f2c2f38ee4dbc/src/groups/mqb/mqbc/mqbc_recoverymanager.cpp#L1568) call in `RecoveryManager::recoverSeqNum()` which requires access to memory mapped file. Since [StorageManager::getSelfFirstSyncPointAfterRolloverSequenceNumber()](https://github.com/alexander-e1off/blazingmq/blob/it-rollover/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L4817) is called multiple times during `PrimaryStateRequest/ReplicaStateRequest`, it make sense to initialize value once and then use the cached value. 
Also, I noticed that sometimes `mqbs::FileStoreUtil::loadIterators()` caused bus error while file descriptors seem to be in correct state.

`mqbc::StorageManager::NodeSeqNumContext.d_firstSyncPointAfterRolloverSeqNum` is used in `StorageManager` to keep nodes info, which is then used in event processing. Self value is set in [StorageManager::do_storeSelfSeq()](https://github.com/alexander-e1off/blazingmq/blob/it-rollover/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L1364), other nodes values are retrieved from message events `PrimaryStateRequest/ReplicaStateRequest`.

Should I try to remove this cache and get fresh value every time by calling `mqbs::FileStoreUtil::loadIterators()`?

src/integration-tests/test_sync_after_rollover_fsm.py
alexander-e1off: So far I didn't observe such race condition neither locally nor on CI. I tried to run this test locally in cycle multiple times.

src/groups/mqb/mqbc/mqbc_recoverymanager.h
kaikulimu: What you are doing now is fine. Let's not change it.

PR #932, opened 2025-09-15
Fix[MQB]: always attempt to process pending open queue requests

dorjesinpo:
There is a race on Replica between 

1. storage recovery 
2. state-restore event for Partition [`N`]
3. state-restore event for Partition [ALL] 

Need to make sure, pending Open Queue request is not stuck as in the case 1. 2. 3.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
@@ -845,6 +845,8 @@ void ClusterQueueHelper::onQueueContextAssigned(
         }
     }
 
+    // REVISIT: 'processOpenQueueRequest' seems to do similar (possibly
+    // redundant) check for 'hasActiveAvailablePrimary',
     if (havePending && haveActivePrimary && isAvailable) {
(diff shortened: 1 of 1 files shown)


src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -3797,6 +3797,13 @@ void ClusterQueueHelper::restoreStateCluster(int partitionId)
                         // 'updateAppIds' which is asynchronous (CSL commit)
                         liveQInfo.d_pendingUpdates.push_back(park);
 
+                        park = bdlf::BindUtil::bind(
+                            &ClusterQueueHelper::processPendingContexts,
+                            this,
+                            queueContext);
+
+                        liveQInfo.d_pendingUpdates.push_back(park);
```
kaikulimu: Not needed because ClusterQueueHelper::onQueueUpdated already calls `processPendingContexts`. But, we need to fix it such that we `convertToLocal` before `processPendingContexts`, i.e. switch the order.

kaikulimu (approved):

PR #944, opened 2025-09-25
Fix[BMQ]: cancel heartbeat when `TCPSessionFactory` stops

emelialei88:
# Issue
This is a fix on a segfault occurred during shutting down. `onHeartbeatSchedulerEvent` fires after `TCPSessionFactory` gets destroyed. The map `d_heartbeatChannels` has became invalid, so there's a bad access in a late `onHeartbeatSchedulerEvent`.

# Fix
Make the scheduled event hold a weak pointer to `TCPSessionFactory`, and cancel the scheduled heartbeat event when `TCPSessionFactory` is stopping.

# Log
The log can be found through [github action](https://github.com/bloomberg/blazingmq/actions/runs/17980140590/job/51172731177?pr=773) -> fuzz test -> view raw log

### DIFF
--- a/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
@@ -1136,8 +1136,9 @@ int TCPSessionFactory::start(bsl::ostream& errorDescription)
         d_scheduler_p->scheduleRecurringEvent(
             &d_heartbeatSchedulerHandle,
             interval,
-            bdlf::BindUtil::bind(&TCPSessionFactory::onHeartbeatSchedulerEvent,
-                                 this));
+            bmqu::WeakMemFnUtil::weakMemFn(
+                &TCPSessionFactory::onHeartbeatSchedulerEvent,
+                d_self.acquireWeak()));
         d_heartbeatSchedulerActive = true;
     }
     else {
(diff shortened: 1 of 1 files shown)


src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -1291,6 +1292,9 @@ void TCPSessionFactory::stop()
         stopChannelFactory(d_tcpChannelFactory_mp.get());
     }
 
+    // Cancel the scheduled heartbeat event
+    d_scheduler_p->cancelEvent(&d_heartbeatSchedulerHandle);
```
dorjesinpo: `cancelEventAndWait`

dorjesinpo (changes requested):

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -1291,6 +1292,9 @@ void TCPSessionFactory::stop()
         stopChannelFactory(d_tcpChannelFactory_mp.get());
     }
 
+    // Cancel the scheduled heartbeat event
+    d_scheduler_p->cancelEventAndWait(&d_heartbeatSchedulerHandle);
```
dorjesinpo: I think, we need to check `d_heartbeatSchedulerHandle` first.
Also, originally, this was under the lock (got removed by mistake)

dorjesinpo (approved):
thank you

PR #946, opened 2025-09-25
CI[Fuzz]: print stack traces on deadlocks, reduce amount of logging to GH Actions

678098:
- Speed up fuzz test by reducing the time to wait for no response.
- Redirect fuzz test logs to a log file in CI and print at most 2000 latest lines on failure (the full identity request log has 900k+ lines in total). This prevents GitHub Actions web ui from clogging when opening fuzz logs.
- Print both the current thread stack trace and all thread stack traces on core dump, to allow debug both crashes and deadlocks.
- Remove custom SIGQUIT handler in bmqbrkr that prevents the broker from terminating immediately on deadlock.
- When stopping a broker during fuzz test, first try to gracefully terminate, after this try to SIGQUIT to get a stack trace of a possible deadlock, last try to kill immediately.
- Remove access to private `os._exit` function. Instead, pass return code from broker thread to the main thread to allow returning the correct exit code from the main thread.
- Do `join` broker thread when stopping fuzz test.
- Check that broker dir is correct and contains a broker binary before spawning broker thread.
- Simplify `stop_broker` API

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -179,7 +179,15 @@ jobs:
         run: |
           pip install -r ${{ github.workspace }}/src/python/requirements.txt
           cd src/python
-          python3 -m blazingmq.dev.fuzztest --broker-dir ${{ github.workspace }}/build/blazingmq/src/applications/bmqbrkr --request ${{ matrix.request }}
+          python3 -m blazingmq.dev.fuzztest --broker-dir ${{ github.workspace }}/build/blazingmq/src/applications/bmqbrkr --request ${{ matrix.request }} > ${{ matrix.request}}_fuzz.log
+
+        # Do not print everything, GitHub Actions is not able to handle 100MB+ of logs.
+        # Typically the error found with fuzzer is related to the latest requests and broker state.
+        # If this is not enough, consider running the test on your local machine and get full logs.
+      - name: Print Fuzz Log tail
+        if: failure()
+        run: |
+          cd src/python && tail -n 2000 ${{ matrix.request}}_fuzz.log
 
       - name: Print core information
         if: failure()
--- a/src/applications/bmqbrkr/bmqbrkr.m.cpp
+++ b/src/applications/bmqbrkr/bmqbrkr.m.cpp
@@ -683,11 +683,6 @@ int main(int argc, const char* argv[])
         bsl::cerr << "Failed to install SIGINT handler  (rc: " << rc << ")\n"
                   << bsl::flush;
     }
-    rc = ::sigaction(SIGQUIT, &sa, NULL);
-    if (rc != 0) {
-        bsl::cerr << "Failed to install SIGQUIT handler  (rc: " << rc << ")\n"
-                  << bsl::flush;
-    }
     rc = ::sigaction(SIGTERM, &sa, NULL);
     if (rc != 0) {
         bsl::cerr << "Failed to install SIGTERM handler  (rc: " << rc << ")\n"
--- a/src/python/blazingmq/dev/processtools.py
+++ b/src/python/blazingmq/dev/processtools.py
@@ -21,6 +21,7 @@ from pathlib import Path
 from typing import Optional
 
 import psutil
+import signal
 
 
 def stop_broker(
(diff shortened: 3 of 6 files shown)


src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -683,11 +683,6 @@ int main(int argc, const char* argv[])
         bsl::cerr << "Failed to install SIGINT handler  (rc: " << rc << ")\n"
                   << bsl::flush;
     }
-    rc = ::sigaction(SIGQUIT, &sa, NULL);
-    if (rc != 0) {
-        bsl::cerr << "Failed to install SIGQUIT handler  (rc: " << rc << ")\n"
-                  << bsl::flush;
-    }
```
678098, line 690: We should not "gracefully" handle SIGQUIT. This signal should immediately stop the application and generate a core dump file.
With the current SIGQUIT handler the broker doesn't generate a core dump on this signal and continues execution even if it is deadlocked:

```
Sending SIGQUIT to the broker...
Shutting down [received signal 'Quit' (3)]
Broker hasn't exited after the timeout
Sending SIGKILL to the broker...
./run: line 42:  2121 Killed                  ${debugCmd} "$@" ./bmqbrkr.tsk localBMQ/etc
```

src/python/blazingmq/dev/processtools.py
```diff
@@ -18,41 +18,40 @@
...
+        broker_process.terminate()
+        try:
+            return broker_process.wait(timeout=timeout)
+        except psutil.TimeoutExpired:
+            print("Broker hasn't exited after the timeout", flush=True)
+
+        print("Sending SIGQUIT to the broker...", flush=True)
+        broker_process.send_signal(signal.SIGQUIT)
+        try:
+            return broker_process.wait(timeout=timeout)
```
678098, line 54: SIGQUIT might still hang up on MAC

678098:
An output example if we introduce a deadlock
https://github.com/bloomberg/blazingmq/actions/runs/18048443792/job/51365535575

src/python/blazingmq/dev/processtools.py
```diff
@@ -18,41 +18,40 @@
...
 import psutil
 
 
-def stop_broker(
-    path: Path, timeout: int = None, error_ok: bool = True
-) -> Optional[int]:
+def stop_broker(path: Path, timeout: int = None) -> Optional[int]:
     """
     Stop the broker (using the pid file), first with 'terminate', then, after the
     specified 'timeout', with 'kill'.  In case when timeout is not specified it
```
pniedzielski: now it's "first with `SIGTERM`, then after the specified `timeout`, with `SIGQUIT`, then after the specified `timeout` once more, with `SIGKILL`."

.github/workflows/build.yaml
```diff
@@ -179,7 +179,15 @@ jobs:
...
-          python3 -m blazingmq.dev.fuzztest --broker-dir ${{ github.workspace }}/build/blazingmq/src/applications/bmqbrkr --request ${{ matrix.request }}
+          python3 -m blazingmq.dev.fuzztest --broker-dir ${{ github.workspace }}/build/blazingmq/src/applications/bmqbrkr --request ${{ matrix.request }} > ${{ matrix.request}}_fuzz.log
+
+        # Do not print everything, GitHub Actions is not able to handle 100MB+ of logs.
+        # Typically the error found with fuzzer is related to the latest requests and broker state.
+        # If this is not enough, consider running the test on your local machine and get full logs.
+      - name: Print Fuzz Log tail
+        if: failure()
+        run: |
+          cd src/python && tail -n 2000 ${{ matrix.request}}_fuzz.log
```
pniedzielski, line 190: This is a reasonable number of lines.

src/applications/bmqbrkr/bmqbrkr.m.cpp
pniedzielski: The POSIX documentation for SIGQUIT is annoyingly sparse.  At the very least, continuing on in a deadlock is bad.  This seems reasonable.

pniedzielski (approved):

src/python/blazingmq/dev/processtools.py
678098: Improved the docs

pniedzielski (approved):

PR #948, opened 2025-09-29
cancel read might cause a deadlock

emelialei88:
We observed a [deadlock](https://github.com/bloomberg/blazingmq/actions/runs/17980140590/attempts/1?pr=773) during a fuzz test ([Fuzz test [identity]](https://github.com/bloomberg/blazingmq/actions/runs/17980140590/job/51145700189?pr=773#logs)) that's hard to reproduce. Before the test hangs, we saw "tearDown" being printed but not "tearDownImpl", so the assumption for the deadlock falls under `cancelRead` function. Since read is cancelled in `NTC` during shutdown anyways, we decide to remove the `cancelRead` for our `ClientSession` and `AdminSession`.


dorjesinpo (approved):
thank you

PR #951, opened 2025-09-30
Fix mqbc: Cluster FSM must heal before starting Partition FSMs

kaikulimu:
Today in the Cluster FSM, leader becomes healed upon the [CSL commit success of the first leader advisory](https://github.com/bloomberg/blazingmq/blob/fe46fe4182435f6bb03a41e82bbceefe658f8de2/src/groups/mqb/mqbc/mqbc_clusterstatetable.h#L515-L518), where it will [initialize the queue key info map](https://github.com/bloomberg/blazingmq/blob/fe46fe4182435f6bb03a41e82bbceefe658f8de2/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L3771) on the cluster thread. Near identical logic exists at the [follower node](https://github.com/bloomberg/blazingmq/blob/fe46fe4182435f6bb03a41e82bbceefe658f8de2/src/groups/mqb/mqbc/mqbc_clusterstatetable.h#L398-L401).

At the same time, CSL commit callback fires and triggers the [`onPartitionPrimaryAssignment` observer](https://github.com/bloomberg/blazingmq/blob/fe46fe4182435f6bb03a41e82bbceefe658f8de2/src/groups/mqb/mqbc/mqbc_clusterstatemanager.cpp#L998), which [jumpstarts the Partition FSM](https://github.com/bloomberg/blazingmq/blob/fe46fe4182435f6bb03a41e82bbceefe658f8de2/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L4067-L4074). The Partition FSM will eventually attempt to open FileStore in the partition thread, which will [access the queue key info map](https://github.com/bloomberg/blazingmq/blob/fe46fe4182435f6bb03a41e82bbceefe658f8de2/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L2922). There is a slight chance of race condition, so let's fix it.

The fix consists of two parts:
1. We always update the partition information before triggering the corresponding transitions in the Partition FSMs, so that PFSMs always have access to latest info. As a result, the PFSM actions `do_storePartitionInfo` and `do_clearPartitionInfo` are removed.
2. We make sure PFSMs are only jumpstarted after queue key info map is initialized. We achieve this by having [the following transition](https://github.com/kaikulimu/blazingmq/blob/772c9372c99889553441d47da3e0124c8ae64e1e/src/groups/mqb/mqbc/mqbc_clusterstatetable.h#L525-L528) in the Cluster FSM:

```
        CST_CFG(LDR_HEALING_STG2,
                CSL_CMT_SUCCESS,
                stopWatchDog_initializeQueueKeyInfoMap_updatePrimaryInPFSMs,
                LDR_HEALED);
```
which guarantees the completion of `initializeQueueKeyInfoMap` before jumpstarting the PFSMs.

After the fix, we can observe that the events happen in the correct order -- Cluster FSM becomes healed before any Partition FSM can start:

```
east1            16Dec2025_22:55:29.741 (     6174732288) INFO     *qbc.clusterstatemanager mqbc_clusterstatemanager.cpp:1032 Cluster (itCluster): Committed advisory: [ rId = NULL choice = [ clusterMessage = [ choice = [ leaderAdvisory = [ sequenceNumber = [ electorTerm = 1 sequenceNumber = 1 ] partitions = [ [ partitionId = 0 primaryNodeId = 1 primaryLeaseId = 1 ] [ partitionId = 1 primaryNodeId = 1 primaryLeaseId = 1 ] [ partitionId = 2 primaryNodeId = 1 primaryLeaseId = 1 ] [ partitionId = 3 primaryNodeId = 1 primaryLeaseId = 1 ] ] queues = [ ] ] ] ] ] ], with status 'SUCCESS'
east1            16Dec2025_22:55:29.779 (     6174732288) INFO     *bmqbrkr.mqbc.clusterfsm mqbc_clusterfsm.cpp:93  Cluster FSM on Event 'CSL_CMT_SUCCESS', transition: State 'LDR_HEALING_STG2' =>  State 'LDR_HEALED'
east1            16Dec2025_22:55:29.842 (     6171291648) INFO     *rkr.mqbc.storagemanager mqbc_storagemanager.cpp:508 Cluster (itCluster) Partition [0]: Self Transition to Primary in the Partition FSM.
east1            16Dec2025_22:55:29.848 (     6171291648) INFO     *qbrkr.mqbc.partitionfsm mqbc_partitionfsm.cpp:78  Partition FSM for Partition [0] on Event 'DETECT_SELF_PRIMARY', transition: State 'UNKNOWN' =>  State 'PRIMARY_HEALING_STG1'
```

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_clusterstatemanager.cpp
+++ b/src/groups/mqb/mqbc/mqbc_clusterstatemanager.cpp
@@ -1124,7 +1124,7 @@ void ClusterStateManager::onCommit(
 
     const bmqp_ctrlmsg::ClusterMessage& clusterMessage =
         advisory.choice().clusterMessage();
-    bsl::vector<int> modifiedPartitions;
+    bsl::vector<int> modifiedPartitions(d_allocator_p);
     mqbc::ClusterUtil::apply(d_state_p,
                              clusterMessage,
                              *d_clusterData_p,


src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
```diff
@@ -606,12 +607,22 @@ void StorageManager::clearPrimaryForPartitionDispatched(
     mqbs::FileStore* fs    = d_fileStores[partitionId].get();
     PartitionInfo&   pinfo = d_partitionInfoVec[partitionId];
 
+    if (primary != pinfo.primary()) {
```
dorjesinpo, line 608: There seems to be the same check in `mqbc::StorageUtil::clearPrimaryForPartition` already (without logging)

src/groups/mqb/mqbc/mqbc_partitionfsm.h
```diff
@@ -204,32 +204,11 @@ class PartitionFSM {
                       bsl::vector<PartitionFSMEventData> >
         EventWithData;
 
-    struct PartitionFSMArgs {
```
dorjesinpo, line 207: Good simplification.  Can we make the same for the Cluster FSM?

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -317,18 +306,18 @@ void StorageManager::dispatchEventToPartition(mqbs::FileStore*          fs,
...
-    // successive events which may be generated as part of performing included
-    // actions as documented in the state transition table for PartitionFSM.
-    queueSp->emplace(event, eventDataVec);
-
-    fs->execute(bdlf::BindUtil::bind(&PartitionFSM::popEventAndProcess,
-                                     d_partitionFSMVec[partitionId].get(),
-                                     queueSp));
+    mqbs::FileStore* fs = d_fileStores[partitionId].get();
+    BSLS_ASSERT_SAFE(fs);
+    if (fs->inDispatcherThread()) {
```
dorjesinpo, line 311: Is there any motivation behind the `inDispatcherThread()` check other than optimization?  Does it fix some race, if so which one?

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -3959,13 +3678,7 @@ void StorageManager::initializeQueueKeyInfoMap(
     BSLS_ASSERT_SAFE(d_dispatcher_p->inDispatcherThread(d_cluster_p));
 
     if (d_isQueueKeyInfoMapVecInitialized) {
-        BALL_LOG_WARN << d_clusterData_p->identity().description()
```
dorjesinpo, line 3961: why removing this warning?  Should it be an assertion?

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -391,18 +380,96 @@ void StorageManager::setPrimaryStatusForPartitionDispatched(
...
+
+    if (!d_isQueueKeyInfoMapVecInitialized) {
+        // We must wait until queue key info map is initialized before
+        // processing primary detection in the Partition FSM.
+
+        return;  // RETURN
+    }
+
+    if (primaryNode->nodeId() ==
+        d_clusterData_p->membership().selfNode()->nodeId()) {
```
dorjesinpo: Should `mqbnet::ClusterImp::d_selfNodeId` be `const`?

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -424,26 +497,22 @@ void StorageManager::processPrimaryDetect(int                  partitionId,
                               primaryNode,
                               primaryLeaseId);
 
-    mqbs::FileStore* fs = d_fileStores[partitionId].get();
-    BSLS_ASSERT_SAFE(fs);
-
-    dispatchEventToPartition(fs,
-                             PartitionFSM::Event::e_DETECT_SELF_PRIMARY,
+    dispatchEventToPartition(PartitionFSM::Event::e_DETECT_SELF_PRIMARY,
```
dorjesinpo: It looks like `e_DETECT_SELF_PRIMARY` can be enqueued twice (depending on the race).  By `StorageManager::initializeQueueKeyInfoMap` and by `setPrimaryForPartitionDispatched`.
If so, this is undesirable.

Same about `e_DETECT_SELF_REPLICA`

dorjesinpo (commented):
Can't we call `initializeQueueKeyInfoMap` (unconditionally) before _any_ FSM event (and not as FSM event) to avoid any possibility for a race?

src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
kaikulimu: The same check used to be in `mqbc::StorageUtil::clearPrimaryForPartition`, but I removed it because I want to call this util method during [`mqbc::StorageManager::processShutdownEventDispatched`](https://github.com/kaikulimu/blazingmq/blob/f9d52c6d0565d6f114f030d56cfce8cce1d6a60d/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L1200-L1204) and [`mqbc::StorageManager::stop`](https://github.com/kaikulimu/blazingmq/blob/f9d52c6d0565d6f114f030d56cfce8cce1d6a60d/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L3633-L3638), neither of which has easy access to `mqbnet::ClusterNode* primary`.

Thus, I moved the check upwards to `mqbblp::StorageManager::clearPrimaryForPartitionDispatched`. And you reminded me to do the same check inside `mqbc::StorageManager::clearPrimaryForPartitionDispatched`. Will add.

src/groups/mqb/mqbc/mqbc_partitionfsm.h
kaikulimu: Yes, similar to Partition FSM, we should have a singleton internal queue to process FSM events, instead of [creating a new event queue every time](https://github.com/bloomberg/blazingmq/blob/5e8635e3e0e35b9764fecbb854ba87032a2c82fa/src/groups/mqb/mqbc/mqbc_clusterstatemanager.cpp#L1053-L1057) and passing it as FSM args. Will make the change.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: Before my PR, `dispatchEventToPartition` was only executed in the cluster dispatcher thread. I extended the method to be possibly executed by either cluster or queue dispatcher thread. I added this `inDispatcherThread()` check mostly for optimization, and to prevent unexpected logic when the PFSM is enqueueing events -- dispatching an event from queue thread to itself introduces a gap and is prone to race conditions. I did not encounter a specific bug while developing the code, but I was following a [similar code change by Alex](https://github.com/alexander-e1off/blazingmq/blob/6d66a1941432b377b9eb223fa3bc9693dcac773a/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L779-L793) in https://github.com/bloomberg/blazingmq/pull/929 in which he did report to fix a race condition. So I would say executing in place is more thread-safe in general.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: `initializeQueueKeyInfoMap` needs to be called exactly once when the Cluster FSM becomes healed for the first time (either [here](https://github.com/alexander-e1off/blazingmq/blob/6d66a1941432b377b9eb223fa3bc9693dcac773a/src/groups/mqb/mqbc/mqbc_clusterstatetable.h#L393-L396) or [here](https://github.com/alexander-e1off/blazingmq/blob/6d66a1941432b377b9eb223fa3bc9693dcac773a/src/groups/mqb/mqbc/mqbc_clusterstatetable.h#L508-L511)), and before Partition FSMs are jumpstarted. If there is a change in leader, Cluster FSM goes through unknown to healed cycle again, but this time `initializeQueueKeyInfoMap` should not be called. The `d_isQueueKeyInfoMapVecInitialized` flag prevents the method from being called again. Therefore, we have no need to warn or alarm if we return early -- it simply indicates a change in leader.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: Yes

kaikulimu: Changed `mqbnet::ClusterImp::d_name` to `const` too

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: This is an important design decision I made for this PR, and please suggest alternatives if you think there is a better solution. `processPrimaryDetect/processReplicaDetect` triggers the healing process in Partition FSMs by enqueueing an `e_DETECT_SELF_PRIMARY/e_DETECT_SELF_REPLICA` event. `processPrimaryDetect/processReplicaDetect` is called during `setPrimaryForPartitionDispatched`, which happens whenever we commit a CSL advisory. The first CSL advisory commit transitions us to a healed [leader](https://github.com/alexander-e1off/blazingmq/blob/6d66a1941432b377b9eb223fa3bc9693dcac773a/src/groups/mqb/mqbc/mqbc_clusterstatetable.h#L508-L511)/[follower](https://github.com/alexander-e1off/blazingmq/blob/6d66a1941432b377b9eb223fa3bc9693dcac773a/src/groups/mqb/mqbc/mqbc_clusterstatetable.h#L393-L396), and at this time we must do `initializeQueueKeyInfoMap` before jumpstarting the Partition FSMs -- this is the race you reported that I am trying to fix. Therefore, this PR's implementation [blocks `processPrimaryDetect/processReplicaDetect` from being called](https://github.com/kaikulimu/blazingmq/blob/51d45b3fe6c1c97e7d6218a544893f39e0801e5e/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L408-L422) during `setPrimaryForPartitionDispatched` if `d_isQueueKeyInfoMapVecInitialized` is false. Instead, `initializeQueueKeyInfoMap` [calls `processPrimaryDetect/processReplicaDetect`](https://github.com/kaikulimu/blazingmq/blob/51d45b3fe6c1c97e7d6218a544893f39e0801e5e/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L3729-L3747) near the end of the method. Future calls of `setPrimaryForPartitionDispatched` will not be blocked by `d_isQueueKeyInfoMapVecInitialized` and thus will trigger the Partition FSMs normally.

dorjesinpo: Superficially, this is still a race.  A boolean (atomic) is checked in one (cluster) thread while being set in another (partition).

More questions
`initializeQueueKeyInfoMap` seems to get triggered by a commit.  Do we ignore subsequent commits?
if the broker goes back to `UNKNOWN` state, do we not need to reset `d_isQueueKeyInfoMapVecInitialized`?
let's discuss a possibility to employ FSM to solve the challenge, since we already have FSMs

src/groups/mqb/mqbc/mqbc_clusterstatemanager.cpp
```diff
@@ -1034,10 +1124,16 @@ void ClusterStateManager::onCommit(
 
     const bmqp_ctrlmsg::ClusterMessage& clusterMessage =
         advisory.choice().clusterMessage();
-    mqbc::ClusterUtil::apply(d_state_p, clusterMessage, *d_clusterData_p);
+    bsl::vector<int> modifiedPartitions;
```
dorjesinpo: allocator

dorjesinpo (dismissed):
one minor comment

PR #961, opened 2025-10-07
Feat[mqb]: Add broker stats to stat log

chrisbeard:
This commit adds the # of clients and queues to the broker's stat log. These values can be derived from existing stats with some legwork; logging them in one line just makes it easier.

Note: These metrics show the number of connected clients and opened queues. If a queue exists in storage but has no clients, it does not count.

**Additional context**


chrisbeard:
Not immediately needed, this can be revisited in the context of #1180

PR #963, opened 2025-10-08
Refactor[MQB,BMQ]: keep the same order in stats enums

678098:


src/groups/mqb/mqbstat/mqbstat_queuestats.h
```diff
@@ -317,7 +317,7 @@ class QueueStatsClient {
     /// are monitored.
     struct EventType {
         // TYPES
-        enum Enum { e_PUT, e_PUSH, e_ACK, e_CONFIRM };
+        enum Enum { e_ACK, e_CONFIRM, e_PUT, e_PUSH };
```
678098, line 320: The corresponding enum in cpp file:
https://github.com/bloomberg/blazingmq/blob/0be90c4c4c1078e3d9af9d3f25be307332d7a8d9/src/groups/mqb/mqbstat/mqbstat_queuestats.cpp#L79-L97

src/groups/mqb/mqbstat/mqbstat_clusterstats.h
```diff
@@ -313,7 +313,7 @@ class ClusterNodeStats {
     /// are monitored.
     struct EventType {
         // TYPES
-        enum Enum { e_PUT, e_PUSH, e_ACK, e_CONFIRM };
+        enum Enum { e_ACK, e_CONFIRM, e_PUT, e_PUSH };
```
678098, line 316: Should be in sync with `ClientStats` enum

PR #964, opened 2025-10-08
Remove[MQB]: hostTags parameter

678098:
* `hostTags` is not used
* set correct namespace for `Expression`/`ExpressionVersion`: `tns` -> `mqbconfm`

### DIFF
--- a/src/groups/mqb/mqbconfm/mqbconf.xsd
+++ b/src/groups/mqb/mqbconfm/mqbconf.xsd
@@ -78,13 +78,11 @@
         Generic type to hold identification of a broker.
 
         hostName......: machine name
-        hostTags......: machine tags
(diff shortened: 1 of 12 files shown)


src/groups/mqb/mqbconfm/mqbconf.xsd
```diff
@@ -363,7 +361,7 @@
       </documentation>
     </annotation>
     <sequence>
-      <element name='version' type='tns:ExpressionVersion' default='E_UNDEFINED'/>
```
678098, line 366: Namespace `tns` is irrelevant and incorrect

`The QName value 'tns:ExpressionVersion' has no corresponding namespace declaration in scope.`

PR #966, opened 2025-10-08
Remove[bmqeval]: forbid chained expressions

678098:
The previous version of bmqeval parser allowed expressions like `x == y == z`, `x < y < z`, etc.
These expressions might be misleading.

Example
`x == y == z` might look like `x, y, z are equal`, but it's not the case.
Consider `x == 10`, `y == 0`, `z == false`. The result of evaluation is:
`10 == 0 == false` -> `(10 == 0) == false` -> `false == false` -> `true`

### DIFF
--- a/src/groups/bmq/bmqeval/bmqeval_simpleevaluatorparser.y
+++ b/src/groups/bmq/bmqeval/bmqeval_simpleevaluatorparser.y
@@ -65,30 +65,30 @@
 %token END 0 "end of expression"
 %token <char> INVALID "invalid character"
 %token <bsl::string> OVERFLOW "overflow"
-%token <bsl::string> PROPERTY "property";
-%token <bsl::string> STRING "string";
-%token <uint64_t> INTEGER "integer";
-%token TRUE "true";
-%token FALSE "false";
-%token LPAR "(";
-%token RPAR ")";
-%token PLUS "+" MINUS "-";
-%token TIMES "*" DIVIDES "/" MODULUS "%";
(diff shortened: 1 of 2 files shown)


src/groups/bmq/bmqeval/bmqeval_simpleevaluatorparser.y
```diff
@@ -65,30 +65,30 @@
...
-%left AND;
-%left EQ NE
-%left LT LE GE GT;
-%left PLUS MINUS;
-%left TIMES DIVIDES MODULUS;
-%precedence NOT;
-
-%type<bsl::string> variable;
-%type<SimpleEvaluator::ExpressionPtr> expression;
-%type<SimpleEvaluator::ExpressionPtr> predicate;
```
678098, line 91: `;` are not necessary

src/groups/bmq/bmqeval/bmqeval_simpleevaluatorparser.y
```diff
@@ -65,30 +65,30 @@
...
+%token <bsl::string> PROPERTY "property"
+%token <bsl::string> STRING "string"
+%token <uint64_t> INTEGER "integer"
+%token TRUE "true"
+%token FALSE "false"
+%token LPAR "("
+%token RPAR ")"
+%token NOT "!"
```
678098, line 75: `NOT` moved earlier.
Previously `!abc == cde` was evaluated like `!(abc == cde)`, now it is evaluated like `(!abc) == cde`

PR #974, opened 2025-10-13
Fix[MQB]: prevent possible crash by protecting PushStream from out-of-order

dorjesinpo:
1.  Push {msg1, app1}
2.  Push {msg2, app2}
3.  Push {msg1, app2}.  Now app2 has msg2 -> masg1
4.  Start delivery from msg1.  App2 has no capacity and sets the resume point at msg1
5.  App2 resumes.  It expects msg1 to be the fist but it is not.

3. must be marked as `isOutOfOrder`.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
@@ -1823,6 +1823,14 @@ bool RelayQueueEngine::subscriptionId2upstreamSubQueueId(
     return true;
 }
 
+
+//              InOrder             OutOfOrder
+//
+//  Replica     use PushStream      use RedeliveryList (per App)
+//
+//  Proxy       use PushStream      use RedeliveryList (per App)
+//              storePush           storePush
+
 void RelayQueueEngine::push(mqbi::StorageMessageAttributes*     attributes,
                             const bmqt::MessageGUID&            msgGUID,
                             const bsl::shared_ptr<bdlbb::Blob>& appData,
@@ -1835,10 +1843,10 @@ void RelayQueueEngine::push(mqbi::StorageMessageAttributes*     attributes,
     if (!isOutOfOrder) {
         if (!d_pushStream.findOrAddLast(&itGuid, msgGUID)) {
             BMQ_LOGTHROTTLE_WARN
-                << "#QUEUE_OUT_OF_ORDER_PUSH  Remote queue: "
+                << "#QUEUE_OUT_OF_ORDER_PUSH Remote queue: "
                 << d_queueState_p->uri() << " (id: " << d_queueState_p->id()
                 << ") treating PUSH message for guid " << msgGUID
-                << " as out-of-order ";
+                << " as out-of-order.";
 
             isOutOfOrder = true;
         }
@@ -1889,8 +1897,26 @@ void RelayQueueEngine::push(mqbi::StorageMessageAttributes*     attributes,
             ordinalPlusOne = 1 + app->ordinal();
             it->setId(ordinalPlusOne);
 
+            // Only Proxy needs to write to a storage.  Replica writes when it
+            // receives replication.  Proxies use 'InMemoryStorage' which sums
+            // up refCount if the guid is already there.
+            // Note that `isDuplicate` detects duplicates as existing
+            // {guid, app} pairs _before_ blindly adding to the existing
+            // refCount.
+            // 1.  We need 'StorageMessageAttributes::refCount' _before_
+            //      'storePush'.
+            // 2.  In the OutOfOrder case, we need 'storePush' _before_
+            //      'processAppRedelivery' which is per app
+            // 3.  This is different from the InOrder case, when we do not call
+            //      'processAppRedelivery' and use the PushStream to deliver to
+            //      multiple apps at once.
+            // Hence, the separate call to 'storePush' outside the loop for the
+            // InOrderInOrder case.
+
             attributes->setRefCount(1);
-            storePush(attributes, msgGUID, appData, subscriptions);
+            bmqp::Protocol::SubQueueInfosArray singleApp(d_allocator_p);
+            singleApp.push_back(*it);
+            storePushIfProxy(attributes, msgGUID, appData, singleApp);
 
             // Attempt to deliver
             processAppRedelivery(subQueueId, app);
@@ -1908,7 +1934,7 @@ void RelayQueueEngine::push(mqbi::StorageMessageAttributes*     attributes,
                         << "Remote queue: " << d_queueState_p->uri()
                         << " (id: " << d_queueState_p->id()
                         << ") discarding a PUSH message for guid " << msgGUID
-                        << ", with unknown App Id " << subscriptionId;
+                        << ", with unknown App Id " << subQueueId;
                     continue;  // CONTINUE
                 }
 
@@ -1948,7 +1974,7 @@ void RelayQueueEngine::push(mqbi::StorageMessageAttributes*     attributes,
     if (count) {
         // Pass correct ref count
         attributes->setRefCount(count);
-        storePush(attributes, msgGUID, appData, subscriptions);
+        storePushIfProxy(attributes, msgGUID, appData, subscriptions);
     }
 }
 
@@ -1984,11 +2010,11 @@ bool RelayQueueEngine::isDuplicate(const App_State*         app,
     return false;
 }
 
-void RelayQueueEngine::storePush(
+void RelayQueueEngine::storePushIfProxy(
     mqbi::StorageMessageAttributes*           attributes,
     const bmqt::MessageGUID&                  msgGUID,
     const bsl::shared_ptr<bdlbb::Blob>&       appData,
-    const bmqp::Protocol::SubQueueInfosArray& subscriptions)
+    const bmqp::Protocol::SubQueueInfosArray& subQueueIds)
 {
     if (d_queueState_p->domain()->cluster()->isRemote()) {
         // Save the message along with the subIds in the storage.  Note that
@@ -2018,8 +2044,8 @@ void RelayQueueEngine::storePush(
 
             // Reusing previously cached ordinals.
             for (bmqp::Protocol::SubQueueInfosArray::const_iterator cit =
-                     subscriptions.begin();
-                 cit != subscriptions.end();
+                     subQueueIds.begin();
+                 cit != subQueueIds.end();
                  ++cit) {
                 if (cit->id() > 0) {
                     dataStreamMessage->app(cit->id() - 1).setPushState();
--- a/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.h
+++ b/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.h
@@ -350,10 +350,11 @@ class RelayQueueEngine BSLS_KEYWORD_FINAL : public mqbi::QueueEngine {
     bool isDuplicate(const App_State*         app,
                      const bmqt::MessageGUID& msgGUID) const;
 
-    void storePush(mqbi::StorageMessageAttributes*           attributes,
-                   const bmqt::MessageGUID&                  msgGUID,
-                   const bsl::shared_ptr<bdlbb::Blob>&       appData,
-                   const bmqp::Protocol::SubQueueInfosArray& subscriptions);
+    void
+    storePushIfProxy(mqbi::StorageMessageAttributes*           attributes,
+                     const bmqt::MessageGUID&                  msgGUID,
+                     const bsl::shared_ptr<bdlbb::Blob>&       appData,
+                     const bmqp::Protocol::SubQueueInfosArray& subQueueIds);
 
     void beforeOneAppRemoved(unsigned int upstreamSubQueueId);
 


src/groups/mqb/mqbblp/mqbblp_clusterstatemanager.cpp
```diff
@@ -810,7 +810,8 @@ void ClusterStateManager::markOrphan(const bsl::vector<int>& partitions,
     BSLS_ASSERT_SAFE(dispatcher()->inDispatcherThread(d_cluster_p));
     BSLS_ASSERT_SAFE(primary);
 
-    for (int i = 0; i < static_cast<int>(partitions.size()); ++i) {
+    for (int i = static_cast<int>(partitions.size()); i > 0;) {
+        --i;
```
678098: This one should be fixed in this PR https://github.com/bloomberg/blazingmq/pull/973

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -215,7 +221,7 @@ class PushStream {
 
     /// Introduce the specified `guid` to the Push Stream if it is not present.
     /// Return an iterator pointing to the `guid`.
-    iterator findOrAppendMessage(const bmqt::MessageGUID& guid);
+    bool findOrAddLast(iterator* itGuid, const bmqt::MessageGUID& guid);
```
678098, line 225: Docs require update

src/groups/mqb/mqbblp/mqbblp_queueenginetester.cpp
```diff
@@ -149,20 +149,19 @@ static void dummyHandleConfiguredCallback(
 ///
 /// The behavior is undefined unless `messages` is formatted as above.  Note
 /// that `messages` will be cleared.
-static void parseMessages(bsl::vector<bsl::string>* messages,
-                          bsl::string               messagesStr)
+static void parseList(bsl::vector<bsl::string>* out, bsl::string in)
```
678098: Docs require update

src/groups/mqb/mqbblp/mqbblp_queueenginetester.cpp
```diff
@@ -1459,6 +1457,54 @@ void QueueEngineTester::dropHandles()
...
+                             const bslstl::StringRef& listApps,
+                             const bslstl::StringRef& listMessages,
+                             bool                     isOutOfOrder)
+{
+    // PRECONDITIONS
+    BSLS_ASSERT_OPT(d_queueEngine_mp &&
+                    "'createQueueEngine()' was not called");
+
+    bsl::vector<bsl::string> apps(d_allocator_p);
+    parseList(&apps, listApps);
```
678098: Here an intermediate string is being constructed with default allocator

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1823,28 +1823,49 @@ bool RelayQueueEngine::subscriptionId2upstreamSubQueueId(
...
+                            bool                                isOutOfOrder)
 {
-    if (isOutOfOrder) {
-        BSLS_ASSERT_SAFE(subscriptions.size() == 1);
+    PushStream::iterator itGuid;  // uninitialized if out of order
+    unsigned int         count = 0;
+
+    if (!isOutOfOrder) {
+        if (!d_pushStream.findOrAddLast(&itGuid, msgGUID)) {
+            BMQ_LOGTHROTTLE_WARN << "#QUEUE_UNKNOWN_SUBSCRIPTION_ID "
```
```suggestion
            BMQ_LOGTHROTTLE_WARN << "#QUEUE_OUT_OF_ORDER_PUSH "
```

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1855,108 +1876,84 @@ RelayQueueEngine::push(mqbi::StorageMessageAttributes*     attributes,
...
-                    << "') discarding a duplicate PUSH for guid " << msgGUID;
-
-                continue;  // CONTINUE
-            }
+                    continue;  // CONTINUE
+                }
 
-            if (!checkForDuplicate(app.d_app.get(), msgGUID)) {
-                continue;  // CONTINUE
+                if (!checkForDuplicate(app.d_app.get(), msgGUID)) {
```
```suggestion
                if (!hasDuplicate(app.d_app.get(), msgGUID)) {
```
`checkForDuplicate` name is not good, the function return value might indicate anything, I can imagine `true` value meaning either `has duplicate` or `everything is ok, no duplicates`.
Should rename, possibly in another PR

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1855,108 +1876,84 @@ RelayQueueEngine::push(mqbi::StorageMessageAttributes*     attributes,
...
+                continue;  // CONTINUE
             }
 
             app->putForRedelivery(msgGUID);
 
-            attributes->setRefCount(1);
-
             // Reusing 'subscriptions' to 'setPushState()' below.
             ordinalPlusOne = 1 + app->ordinal();
             subscriptions.begin()->setId(ordinalPlusOne);
```
678098: Previously, this code was in `if (isOutOfOrder)` branch, together with precondition `BSLS_ASSERT_SAFE(subscriptions.size() == 1);`
Now this precondition is gone, and there can be any number of subscriptions configured. `subscriptions.begin()` is possibly unrelated to the current subscription
It should be `it->setId(ordinalPlusOne);` I think

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -675,10 +681,24 @@ PushStream::create(const bmqp::RdaInfo&  rda,
...
+    }
+    else if (existing == --d_stream.end()) {
+        *itGuid = existing;
+    }
+    else {
+        // Anything other than the last is out of order
+        return false;
+    }
+
+    return true;
```
678098:
```suggestion
    if (existing == d_stream.end()) {
        *itGuid = d_stream.insert(bsl::make_pair(guid, Elements())).first;
        return true;
    }
    if (existing == --d_stream.end()) {
        *itGuid = existing;
        return true;
    }

    // Anything other than the last is out of order
    return false;
```
I think this variant is simpler, we don't have long paths to `return`s anymore

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1855,108 +1876,84 @@ RelayQueueEngine::push(mqbi::StorageMessageAttributes*     attributes,
...
-        // Reusing 'subscriptions' to 'setPushState()' below.
-        ordinalPlusOne = 1 + itApp->second.d_app->ordinal();
-        it->setId(ordinalPlusOne);
+            // Reusing 'subscriptions' to 'setPushState()' below.
+            ordinalPlusOne = 1 + itApp->second.d_app->ordinal();
+            it->setId(ordinalPlusOne);
 
-        d_pushStream.add(element);
-        ++count;
+            d_pushStream.add(element);
```
678098: I am curious if there is a reason to build `PushStream::Element` (line 1938) separately from adding it to a `PushStream` (line 1947).
Are lines 1943-1945 affecting the push stream element? If so, the connection is not clear, it might be more explicit.
If lines 1943-1945 do not affect the push stream/element, consider simpler API
```
d_pushStream.add(it->rdaInfo(),
                 subscriptionId,
                 itGuid,
                 itApp);
```
Might be done in another PR

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.t.cpp
```diff
@@ -1872,6 +1872,110 @@ static void test20_handleParametersLimits()
     }
 }
 
+static void test21_deliverOutOfOrder()
+// ------------------------------------------------------------------------
+// REDELIVERY AND RESUME
+//
+// Concerns:
+//   1. Verifying resume for a fanout appId..
```
678098:
```suggestion
//   1. Verifying resume for a fanout appId.
```
I am also fine with removing the last dot at all to save 1 byte (and to have less readable characters) in similar cases
```suggestion
//   1. Verifying resume for a fanout appId
```

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.t.cpp
```diff
@@ -1872,6 +1872,110 @@ static void test20_handleParametersLimits()
...
+//  Then trigger 'processAppRedelivery' for "a".
+//  It should 'deliverMessages' from 3 to 4 in the first phase and then attempt
+//  to advance all apps.  If the first phase fails to stop at 4, then 4 (and 5)
+//  will get processed twice.
+//
+// Testing:
+//   'processAppRedelivery'.
+// ------------------------------------------------------------------------
+{
+    bmqtst::TestHelper::printTestName("ROUND-ROBIN AND REDELIVERY");
```
678098:
```suggestion
    bmqtst::TestHelper::printTestName("DELIVER OUT OF ORDER");
```
Note that we also have `test19` where we copy-pasted this test name incorrectly

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.t.cpp
```diff
@@ -1872,6 +1872,110 @@ static void test20_handleParametersLimits()
...
+//   1. Verifying resume for a fanout appId..
+//
+// Plan:
+//  Make the data stream of 1   2   3   4   5
+//                          |   |   |   |
+//  where   "a" state is    RL PAL  RP  |
+//          "b" state is    QH  QH PAL  |
+//                                      iterator
+//  RL - RedeliveryList, PAL - PutAsideList, QH queue handle, RP - Resume Point
+//  Then remove "b".
```
678098: Do we actually remove `"b"`? Where does it happen?

678098: Nvm, the comment is copy-pasted.
I am fine with removing the header at all, replacing it with a brief description of what is being tested. Not necessary to duplicate the code in the comment, just placing a few more comments in the code below is enough

src/groups/mqb/mqbblp/mqbblp_queueenginetester.cpp
dorjesinpo: let's pass an allocator

dorjesinpo: heavy change

678098: Can do in another PR if not convenient, my deal is only to find possible issues

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -662,27 +665,16 @@ inline const PushStream::Element* PushStream::App::last() const
...
-inline PushStream::iterator
-PushStream::findOrAppendMessage(const bmqt::MessageGUID& guid)
-{
-    return d_stream.insert(bsl::make_pair(guid, Elements())).first;
-}
-
-inline void PushStream::add(Element* element)
-{
     // Add to the GUID
     BSLS_ASSERT_SAFE(element);
```
678098, line 679: Can remove this precondition now

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1823,69 +1823,28 @@ bool RelayQueueEngine::subscriptionId2upstreamSubQueueId(
...
-            subscriptions.begin()->setId(ordinalPlusOne);
-
-            storePush(attributes, msgGUID, appData, subscriptions, true);
-
-            // Attempt to deliver
-            processAppRedelivery(subQueueId, app);
+    if (!isOutOfOrder) {
+        if (!d_pushStream.findOrAddLast(&itGuid, msgGUID)) {
+            BMQ_LOGTHROTTLE_WARN
+                << "#QUEUE_OUT_OF_ORDER_PUSH  Remote queue: "
```
```suggestion
                << "#QUEUE_OUT_OF_ORDER_PUSH Remote queue: "
```

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1823,69 +1823,28 @@ bool RelayQueueEngine::subscriptionId2upstreamSubQueueId(
...
+                << d_queueState_p->uri() << " (id: " << d_queueState_p->id()
+                << ") treating PUSH message for guid " << msgGUID
+                << " as out-of-order ";
```
```suggestion
                << " as out-of-order";
```

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1904,63 +1863,97 @@ RelayQueueEngine::push(mqbi::StorageMessageAttributes*     attributes,
...
+                continue;  // CONTINUE
+            }
+
+            app->putForRedelivery(msgGUID);
+
+            // Reusing 'subscriptions' to 'setPushState()' below.
+            ordinalPlusOne = 1 + app->ordinal();
+            it->setId(ordinalPlusOne);
+
+            attributes->setRefCount(1);
```
678098, line 1916: I think it should be similar to `else` branch and should inc the counter, so we set up the correct ref count before exiting this function
```suggestion
            ++count;
```

If we don't increment `count` here, there will be a bug with ref counting, right?

dorjesinpo: hmm...
we need the refCount _before_ `storePush`.  In the case of `isOutOfOrder`, we need `storePush` before `processAppRedelivery` which is per `app` (vs the one for all apps).

678098: Do we need to look at all subscriptions to init `count` before calling all `storePush`?
Can we `storePush` twice now during one function call?

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1904,63 +1863,97 @@ RelayQueueEngine::push(mqbi::StorageMessageAttributes*     attributes,
...
                     << ") discarding a PUSH message for guid " << msgGUID
-                    << ", with unknown App Id " << subscriptionId;
+                    << ", with unknown App Id " << subQueueId;
+
                 continue;  // CONTINUE
             }
 
-            itApp =
-                d_pushStream.d_apps.emplace(subQueueId, app_cit->second).first;
+            if (isDuplicate(app, msgGUID)) {
```
678098, line 1890: I am also concerned about duplicates.
We have this continue, in this case we do not update `count`.
```
if (isDuplicate(app, msgGUID)) {
    continue;
}
```
Imagine a situation, we have 3 apps [`app1`, `app2`, `app3`]
We have a message `msg1` only for `app1`: `{app1, msg1}`
In this case, refCount is 1, right?

Now, let's assume message `msg1` is delivered again for all apps.
It is a duplicate for `app1`, so we continue here.
We increment `count` twice (`app2`, `app3`) and set `refCount == 2` in the end.
So we have `msg1` not confirmed for all 3 apps but have `refCount == 2` (should be 3)

Do you think there is a bug here?

678098: Should we init `count` from `attributes` in the beginning?

dorjesinpo: > I am also concerned about duplicates. We have this continue, in this case we do not update `count`.
> 
> ```
> if (isDuplicate(app, msgGUID)) {
>     continue;
> }
> ```
> 
> Imagine a situation, we have 3 apps [`app1`, `app2`, `app3`] We have a message `msg1` only for `app1`: `{app1, msg1}` In this case, refCount is 1, right?
> 
> Now, let's assume message `msg1` is delivered again for all apps. It is a duplicate for `app1`, so we continue here. We increment `count` twice and set `refCount == 2` in the end. So we have `msg1` not confirmed for all 3 apps but have `refCount == 2` (should be 3)
> 
> Do you think there is a bug here?

Consider Proxy and Replica cases separately.
For Replica, there is no call to the storage upon push, so refCount is not important.  The storage has refCount 3.

For Proxy, there call to the storage upon push, unless it is duplicate.  If `msg1` is already gone, we store it again.  Otherwise, it is either duplicate, in which case we do not store it.  Or, it is there for another app, in which case we need to increment the refCount by 1.

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -1904,63 +1863,97 @@ RelayQueueEngine::push(mqbi::StorageMessageAttributes*     attributes,
...
+            if (itApp == d_pushStream.d_apps.end()) {
+                AppsMap::const_iterator app_cit = d_apps.find(subQueueId);
+
+                if (app_cit == d_apps.end()) {
+                    BMQ_LOGTHROTTLE_ERROR
+                        << "#QUEUE_UNKNOWN_SUBSCRIPTION_ID "
+                        << "Remote queue: " << d_queueState_p->uri()
+                        << " (id: " << d_queueState_p->id()
+                        << ") discarding a PUSH message for guid " << msgGUID
+                        << ", with unknown App Id " << subscriptionId;
```
```suggestion
                        << ", with unknown App Id " << subQueueId;
```

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
dorjesinpo: > Should we init `count` from `attributes` in the beginning?

No, `count` must be `0` and change only when 1) this is inOrder; 2) there is at least one known app

PR #976, opened 2025-10-14
Fix: GCC warnings in SDK

pniedzielski:
This PR fixes a handful of warnings from GCC when building our SDK, broken out from an incoming PR.  It's a bit of a grab-bag; please see the commit messages for explanations.

### DIFF
--- a/src/groups/bmq/bmqimp/bmqimp_queue.cpp
+++ b/src/groups/bmq/bmqimp/bmqimp_queue.cpp
@@ -50,6 +50,10 @@ const char k_STAT_NAME[] = "queues";
 /// time and scale it back down at print time.
 const double k_COMPRESSION_RATIO_PRECISION_FACTOR = 10000.;
 
+/// Inverse of `k_COMPRESSION_RATIO_PRECISION_FACTOR` for fast multiplication.
+const double k_COMPRESSION_RATIO_PRECISION_FACTOR_INV =
+    1 / k_COMPRESSION_RATIO_PRECISION_FACTOR;
+
 enum {
     k_STAT_IN = 0  // value = bytes received ; increments =
                    // messages received
@@ -77,7 +81,8 @@ calculateCompressionRatio(const bmqst::StatValue&                   value,
     const double ratioSum = static_cast<double>(
         bmqst::StatUtil::value(value, start));
 
-    return (ratioSum / messageCount) / k_COMPRESSION_RATIO_PRECISION_FACTOR;
+    return (ratioSum / messageCount) *
+           k_COMPRESSION_RATIO_PRECISION_FACTOR_INV;
 }
 
 double
@@ -97,7 +102,8 @@ calculateCompressionRatio(const bmqst::StatValue&                   value,
     const double ratioSum = static_cast<double>(
         bmqst::StatUtil::valueDifference(value, start, endPlus));
 
-    return (ratioSum / messageCount) / k_COMPRESSION_RATIO_PRECISION_FACTOR;
+    return (ratioSum / messageCount) *
+           k_COMPRESSION_RATIO_PRECISION_FACTOR_INV;
 }
 
 }  // close unnamed namespace


src/groups/bmq/bmqa/bmqa_message.cpp
```diff
@@ -96,8 +96,15 @@ Message& Message::setDataRef(const char* data, size_t length)
                      "'MessageEventBuilder::startMessage' to get one");
     BSLS_ASSERT_SAFE(d_impl.d_event_p->putEventBuilder() &&
                      "message not editable");
+    BSLS_ASSERT_SAFE(length > 0 && "length must be greater than zero");
```
678098:
```suggestion
    BSLS_ASSERT_SAFE(0 < length && "length must be greater than zero");
```
Left-to-right comparisons might make the codebase more uniform and simple (predictable)

src/groups/bmq/bmqa/bmqa_message.h
```diff
@@ -218,7 +218,7 @@ class Message {
 
     /// Set the payload of this message to the specified `length` bytes
     /// starting at the specified `data` address.  The behavior is undefined
-    /// unless `data` is non-null and `length` is greater than zero.  Note that
+    /// unless `data` is non-null and `0 < length <= INT_MAX`.  Note that
```
678098, line 221: What is interesting, we allow to set data size up to `INT_MAX` in this setter but forbid it on a protocol level. There, we have much smaller limit on a message size. The problem is that we mention one limitation here and use another limitation in another place.

src/groups/bmq/bmqimp/bmqimp_application.cpp
```diff
@@ -680,9 +683,13 @@ Application::Application(
     bmqst::StatValue::SnapshotLocation start;
     bmqst::StatValue::SnapshotLocation end;
     if (d_sessionOptions.statsDumpInterval() != bsls::TimeInterval()) {
+        // statsDumpInterval().seconds() must be < 60 minutes, which can safely
+        // be stored in an `int`.
+        int statsDumpSeconds = static_cast<int>(
```
678098:
```suggestion
        const int statsDumpSeconds = static_cast<int>(
```

src/groups/bmq/bmqimp/bmqimp_application.cpp
```diff
@@ -85,15 +85,18 @@ statContextConfiguration(const bmqt::SessionOptions& options,
 {
     bmqst::StatContextConfiguration config("stats", allocator);
     if (options.statsDumpInterval() != bsls::TimeInterval()) {
+        // statsDumpInterval().seconds() must be < 60 minutes, which can safely
+        // be stored in an `int`.
+        int statsDumpSeconds = static_cast<int>(
```
678098:
```suggestion
        const int statsDumpSeconds = static_cast<int>(
```

src/groups/bmq/bmqa/bmqa_message.cpp
```diff
@@ -96,8 +96,15 @@ Message& Message::setDataRef(const char* data, size_t length)
...
+    BSLS_ASSERT_SAFE(length <= bsl::numeric_limits<int>::max() &&
+                     "length must be less than or equal to INT_MAX");
 
-    d_impl.d_event_p->putEventBuilder()->setMessagePayload(data, length);
+    // By the above, `length` can be represented by `int`.
+    int narrowedLength = static_cast<int>(length);
```
678098:
```suggestion
    const int narrowedLength = static_cast<int>(length);
```

src/groups/bmq/bmqimp/bmqimp_queue.cpp
```diff
@@ -65,34 +65,39 @@ double
 calculateCompressionRatio(const bmqst::StatValue&                   value,
                           const bmqst::StatValue::SnapshotLocation& start)
 {
-    const bsls::Types::Int64 messageCount = bmqst::StatUtil::increments(value,
-                                                                        start);
-    if (messageCount == 0) {
+    // All arithmetic here is done in `double`s, so we'll convert the `Int64`s
+    // from the stats layer into `doubles` explicitly.
```
678098:
```suggestion
    // from the stats layer into `double`s explicitly.
```

src/groups/bmq/bmqimp/bmqimp_queue.cpp
```diff
@@ -65,34 +65,39 @@ double
...
         return 0.0;  // RETURN
     }
 
-    const bsls::Types::Int64 ratioSum = bmqst::StatUtil::value(value, start);
+    const double ratioSum = static_cast<double>(
+        bmqst::StatUtil::value(value, start));
 
-    return (static_cast<double>(ratioSum) / messageCount) /
-           k_COMPRESSION_RATIO_PRECISION_FACTOR;
+    return (ratioSum / messageCount) / k_COMPRESSION_RATIO_PRECISION_FACTOR;
```
678098:
```suggestion
    return k_COMPRESSION_RATIO_PRECISION_FACTOR * (ratioSum / messageCount);
```
If we are modifying this code, we might as well replace division with multiplication. Division is expensive, can avoid it.

Multiplication requires inverting the constant (`lambda x: x -> 1/x`)
```
const double k_COMPRESSION_RATIO_PRECISION_FACTOR = 1e-4;
```

src/groups/bmq/bmqt/bmqt_sessionoptions.h
```diff
@@ -438,7 +438,7 @@ SessionOptions::setStatsDumpInterval(const bsls::TimeInterval& value)
                     "value must be a multiple of 30s");
     BSLS_ASSERT_OPT(value.seconds() <
                         bdlt::TimeUnitRatio::k_SECONDS_PER_HOUR &&
-                    "value must be < 10 min");
+                    "value must be < 60 min");
```
678098, line 442: Might be good to add a missing check that `0 <= value`

src/groups/bmq/bmqa/bmqa_message.h
pniedzielski: Interesting---let me take a look at how we handle that error.  I wonder if we do that error checking after we convert this unsigned to signed, meaning the behavior is:
  1. Well defined, okay: [0, max message size]
  2. Well defined, error: (max message size, int max]
  3. Undefined: (int max, uint max]

Will investigate more.  It may be better to have a wide contract here, so there's no undefined behavior.

src/groups/bmq/bmqimp/bmqimp_queue.cpp
pniedzielski: @678098 As it stands right now, we divide by `k_COMPRESSION_RATIO_PRECISION_FACTOR` in both overloads of `calculateCompressionRatio` (functions changed by this patch), and multiply by it in `Queue::statReportCompressionRatio`.  The latter is called whenever a new message has been compressed.  It doesn't seem obvious which is more frequent, and if we keep only the one constant, there will be both multiplication and division no matter what.  Thoughts?

src/groups/bmq/bmqa/bmqa_message.h
pniedzielski: Looking into this more:
  1. We need to convert `length` from `size_t` to `int` before `src/groups/bmq/bmqp/bmqp_puteventbuilder.cpp:367`, because `BlobUtil::append` takes an `int`.
  2. After that we optionally compress the payload blob at `src/groups/bmq/bmqp/bmqp_puteventbuilder.cpp:376`, which may decrease its size.
  3. Maybe it did, maybe it didn't; either way we call `packMessageInternal` at `src/groups/bmq/bmqp/bmqp_puteventbuilder.cpp:416`.
  4. Here we validate the payload size.  `src/groups/bmq/bmqp/bmqp_puteventbuilder.cpp:91`.

So it's what I suspected above: this undefined behavior was always here, well before fwe hit the well-defined error case.  But, the error case is hard to check early, because a payload that is too large uncompressed could be compressed to a small enough size while packing it.

src/groups/bmq/bmqimp/bmqimp_queue.cpp
678098: 2 constants? `<NAME>` and `<NAME>_INV`?

678098: @pniedzielski

pniedzielski: Sounds good, added in a new commit.

678098 (approved):

PR #978, opened 2025-10-15
Fix[bmqio]: Use largest connection backlog allowed by OS

pniedzielski:
When listening on a new socket, `bmqio_ntcchannel` sets the connection backlog size to be ten connections.  This allows at most ten connections to the socket to be waiting to be accepted before they are rejected by the OS for us.  Ten is quite a low number, and seems like a mistake. This patch changes the backlog to be determined by NTF, which makes it as large as the OS allows.

### DIFF
--- a/src/groups/bmq/bmqio/bmqio_ntcchannel.cpp
+++ b/src/groups/bmq/bmqio/bmqio_ntcchannel.cpp
@@ -1579,7 +1579,9 @@ int NtcListener::listen(bmqio::Status*              status,
     int backlog;
     if (!options.properties().load(&backlog,
                                    NtcListenerUtil::listenBacklogProperty())) {
-        backlog = 10;
+        // The magic number 0 means the backlog will be the maximum allowed by
+        // the OS.
+        backlog = 0;
     }
 
     ntsa::Endpoint endpoint;


src/groups/bmq/bmqio/bmqio_ntcchannel.cpp
```diff
@@ -1579,7 +1579,9 @@ int NtcListener::listen(bmqio::Status*              status,
     int backlog;
     if (!options.properties().load(&backlog,
                                    NtcListenerUtil::listenBacklogProperty())) {
-        backlog = 10;
+        // The magic number 0 means the backlog will be the maximum allowed by
+        // the OS.
+        backlog = 0;
```
678098, line 1584: The default value will be 65535
https://github.com/bloomberg/ntf-core/blob/0bb993b1920129d1acf91e35ae80bbdb708d303c/groups/nts/ntsu/ntsu_socketutil.cpp#L1322
What is interesting, we define `backlog` as `int`, NTF api expects `size_t`, and NTF implementation casts it to `int`

pniedzielski: Note that we won't reach 65535 on most systems, as the OS will clamp it to the value of `sysctl net.core.somaxconn` as per https://linux.die.net/man/2/listen (that's how we found this).

PR #988, opened 2025-10-17
Fix[MQB]: handle `e_DELAY` case in `catchUp` (Assertion failed: result == Routers::e_INVALID)

678098:


src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -959,27 +959,34 @@ size_t QueueEngineUtil_AppState::catchUp(bsls::TimeInterval*          delay,
...
+            case Routers::e_DELAY: BSLA_FALLTHROUGH;
+            case Routers::e_INVALID: {
+                // e_DELAY: poisoned message possibly, have to wait
+                // e_INVALID: the {GUID, App} is not valid anymore
+                shouldBreak = true;
+            } break;
+            default: {
+                BSLS_ASSERT_OPT(false && "Unreachable by design");
+            } break;
             }
```
678098, line 988: Using `switch/case` is good because there are warnings if some enums are not explicitly handled in it. Also it is just faster than a chain of `if else`s.
The only bad thing is having to pass break status out of the `switch`

src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -942,6 +942,7 @@ size_t QueueEngineUtil_AppState::catchUp(bsls::TimeInterval*          delay,
     // 'end' is never CONFIRMed, so the 'VirtualStorageIterator' cannot skip it
 
     d_resumePoint = bmqt::MessageGUID();
+    bool shouldBreak = false;
```
678098, line 945: Enough to init this once before running the loop

src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -959,27 +960,35 @@ size_t QueueEngineUtil_AppState::catchUp(bsls::TimeInterval*          delay,
...
-            else if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(
-                         result == Routers::e_NO_CAPACITY_ALL)) {
+            } break;
+            case Routers::e_NO_CAPACITY_ALL: BSLA_FALLTHROUGH;
+            case Routers::e_DELAY: {
+                // e_NO_CAPACITY_ALL: have to wait
+                // e_DELAY: poisoned message possibly, have to wait
                 d_resumePoint = start->guid();
-                break;
+                shouldBreak   = true;
```
678098, line 978: This is the main change, `e_DELAY` is handled now.
We set up a resume point and break after this.
The behaviour is in sync with `processDeliveryList`

PR #990, opened 2025-10-20
Feat[MQB]: Add authn plugin

emelialei88:
This PR includes:
- Updated plugin interface in `mqbplug`
- Configurations for plugins in `mqbcfg`
- `mqbauthn` package that controls the plugin and includes built-in plugins

### DIFF
--- a/src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
+++ b/src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
@@ -32,6 +32,9 @@
 namespace BloombergLP {
 namespace mqbauthn {
 
+const char* AnonAuthenticator::k_NAME      = "AnonAuthenticator";
+const char* AnonAuthenticator::k_MECHANISM = "ANONYMOUS";
+
 // ------------------------------
 // class AnonAuthenticationResult
 // ------------------------------
@@ -205,8 +208,9 @@ AnonAuthenticatorPluginFactory::create(bslma::Allocator* allocator)
         mqbplug::AuthenticatorUtil::findAuthenticatorConfig(
             AnonAuthenticator::k_NAME);
 
-    return bslma::ManagedPtrUtil::allocateManaged<AnonAuthenticator>(allocator,
-                                                                     config);
+    return bslma::ManagedPtr<mqbplug::Authenticator>(
+        new (*allocator) AnonAuthenticator(config, allocator),
+        allocator);
 }
 
 }  // close package namespace
--- a/src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.h
+++ b/src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.h
@@ -108,10 +108,8 @@ class AnonAuthenticationResult : public mqbplug::AuthenticationResult {
 class AnonAuthenticator : public mqbplug::Authenticator {
   public:
     // PUBLIC CLASS DATA
-    static BSLS_KEYWORD_INLINE_CONSTEXPR bsl::string_view k_NAME =
-        "AnonAuthenticator";
-    static BSLS_KEYWORD_INLINE_CONSTEXPR bsl::string_view k_MECHANISM =
-        "ANONYMOUS";
+    static const char* k_NAME;
+    static const char* k_MECHANISM;
 
   private:
     // CLASS-SCOPE CATEGORY
--- a/src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
+++ b/src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
@@ -36,6 +36,9 @@
 namespace BloombergLP {
 namespace mqbauthn {
 
+const char* BasicAuthenticator::k_NAME      = "BasicAuthenticator";
+const char* BasicAuthenticator::k_MECHANISM = "BASIC";
+
 namespace {
 
 const int k_AUTHN_DURATION_SECONDS = 600;
@@ -212,9 +215,11 @@ BasicAuthenticatorPluginFactory::create(bslma::Allocator* allocator)
         mqbplug::AuthenticatorUtil::findAuthenticatorConfig(
             BasicAuthenticator::k_NAME);
 
-    return bslma::ManagedPtrUtil::allocateManaged<BasicAuthenticator>(
-        allocator,
-        config);
+    allocator = bslma::Default::allocator(allocator);
+
+    return bslma::ManagedPtr<mqbplug::Authenticator>(
+        new (*allocator) BasicAuthenticator(config, allocator),
+        allocator);
 }
 
 }  // close package namespace
--- a/src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.h
+++ b/src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.h
@@ -95,10 +95,8 @@ class BasicAuthenticationResult : public mqbplug::AuthenticationResult {
 class BasicAuthenticator : public mqbplug::Authenticator {
   public:
     // PUBLIC CLASS DATA
-    static BSLS_KEYWORD_INLINE_CONSTEXPR bsl::string_view k_NAME =
-        "BasicAuthenticator";
-    static BSLS_KEYWORD_INLINE_CONSTEXPR bsl::string_view k_MECHANISM =
-        "BASIC";
+    static const char* k_NAME;
+    static const char* k_MECHANISM;
 
   private:
     // CLASS-SCOPE CATEGORY


src/groups/mqb/mqbauthn/mqbauthn_anonfailauthenticator.cpp
```diff
@@ -0,0 +1,121 @@
...
+    // NOTHING
+}
+
+bslma::ManagedPtr<mqbplug::Authenticator>
+AnonFailAuthenticatorPluginFactory::create(bslma::Allocator* allocator)
+{
+    bslma::ManagedPtr<mqbplug::Authenticator> result(
+        new (*allocator) AnonFailAuthenticator(),
+        allocator);
+    return result;
```
678098:
```suggestion
    return bslma::ManagedPtrUtil::allocateManaged<mqbauthn::AnonFailAuthenticator>(allocator);
```
Might be simpler, if pointer cast is fine

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.h
```diff
@@ -0,0 +1,173 @@
...
+    AnonFailAuthenticator(const AnonFailAuthenticator& other)
+        BSLS_KEYWORD_DELETED;
+    AnonFailAuthenticator&
+    operator=(const AnonFailAuthenticator& other) BSLS_KEYWORD_DELETED;
+
+  public:
+    // TRAITS
+    BSLMF_NESTED_TRAIT_DECLARATION(AnonFailAuthenticator,
+                                   bslma::UsesBslmaAllocator)
+
```
678098, line 136:
```suggestion
```
It's not using bslma::Allocator

src/groups/mqb/mqbplug/mqbplug_statconsumer.h
```diff
@@ -79,7 +79,7 @@ class StatConsumer {
     // ACESSORS
 
     /// Return the name of the plugin.
-    virtual bslstl::StringRef name() const = 0;
+    virtual bsl::string_view name() const = 0;
```
678098, line 82: This is a breaking change.
All other plugins still use `bslstl::StringRef` and cannot inherit from `bsl::string_view` one
Even if we update them, we will lose compatibility with older versions of the plugins. It might result in strange crashes due to incompatible API

src/groups/mqb/mqbplug/mqbplug_statconsumer.h
```diff
@@ -140,7 +140,7 @@ struct StatConsumerUtil {
 
     /// Find consumer config with the specified `name`
     static const mqbcfg::StatPluginConfig*
-    findConsumerConfig(bslstl::StringRef name);
+    findConsumerConfig(bsl::string_view name);
```
678098, line 143: This is also a breaking change for the same reasons

src/groups/mqb/mqbplug/mqbplug_statconsumer.h
678098: To deal with this, I propose to
1. Keep the original `bslstl::StringRef` declaration
2. Add a new one virtual `virtual bsl::string_view name() const {
    return static_cast<bslstl::StringRef>(name());
}`
With the default implementation, it will be compatible with older plugins.
We are free to gradually add `bsl::string_view` implementation in plugins and eventually remove `bslstl::StringRef` one.

Alternatively we can revert this change to remove possible risks
@hallfox what do you think?

src/groups/mqb/mqbplug/mqbplug_statconsumer.h
678098: Same idea, might declare the second one `findConsumerConfig(bsl::string_view name);` and eventually remove `bslstl::StringRef` one

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+AnonAuthenticator::AnonAuthenticator(
+    const mqbcfg::AuthenticatorPluginConfig* config,
+    bslma::Allocator*                        allocator)
+: d_allocator_p(allocator)
+, d_isStarted(false)
+, d_shouldPass(true)
+{
+    if (!config) {
+        return;
```
678098, line 80: Will be nice to log something here, to notify that we are trying to run without config
```suggestion
        BALL_LOG_INFO << "No configuration provided, configuring AnonAuthenticator with the default shouldPass = true";
        return;
```

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    // Load the configured `shouldPass` argument
+    bsl::vector<mqbcfg::PluginConfigKeyValue>::const_iterator it =
+        config->configs().cbegin();
+    for (; it != config->configs().cend(); ++it) {
```
678098: If I read this code for the first time without reading the rest, I would assume that `config->configs()` are different configs for different plugins, while `config` is the app config
Maybe renaming it to `config->settings()` is more clear

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
```diff
@@ -0,0 +1,186 @@
...
+        if (it->key() == "shouldPass" && it->value().isBoolValValue()) {
+            d_shouldPass = it->value().boolVal();
+        }
+    }
```
678098: I think it's worth adding a `bool isShouldPassFound = false;` before the loop, and we can also add logs for interesting cases.

Logs might be useful. Example: what if someone assumes that `int` values `0`, `1` are good to enable/disable passing of this plugin. The current code will always silently use the default value `true`.


```suggestion
        if (it->key() == "shouldPass") {
            if (!it->value().isBoolValValue()) {
                BALL_LOG_WARN << "Expected bool for 'shouldPass' setting...";
                continue;
            }
            d_shouldPass = it->value().boolVal();
            isShouldPassFound = true;
        }
    }
    if (isShouldPassFound) {
        BALL_LOG_INFO << "... setting found, using the default shouldPass =<VAL>";
    } else {
        BALL_LOG_INFO << "... setting not found, using the default shouldPass = true";
    }
```

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include <mqbauthn_anonauthenticator.h>
+
+// MQB
+#include <mqbcfg_messages.h>
+#include <mqbplug_authenticator.h>
+
+// BDE
+#include <bsl_iostream.h>
+#include <bsl_memory.h>
+#include <bsl_optional.h>
```
678098:
```suggestion
```
Possibly not used

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
```diff
@@ -0,0 +1,186 @@
...
+{
+    return k_MECHANISM;
+}
+
+int AnonAuthenticator::authenticate(
+    bsl::ostream&                                   errorDescription,
+    bsl::shared_ptr<mqbplug::AuthenticationResult>* result,
+    BSLA_UNUSED const mqbplug::AuthenticationData& input) const
+{
+    if (d_shouldPass) {
```
678098, line 139:
```suggestion
    BSLS_ASSERT_SAFE(result);

    if (d_shouldPass) {
```

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
```diff
@@ -0,0 +1,186 @@
...
+{
+    const mqbcfg::AuthenticatorPluginConfig* config =
+        mqbplug::AuthenticatorUtil::findAuthenticatorConfig(
+            AnonAuthenticator::k_NAME);
+
+    bslma::ManagedPtr<mqbplug::Authenticator> result =
+        bslma::ManagedPtrUtil::allocateManaged<AnonAuthenticator>(allocator,
+                                                                  config);
+
+    return result;
```
678098:
```suggestion
    return bslma::ManagedPtrUtil::allocateManaged<AnonAuthenticator>(allocator,
                                                                     config);
```
Can simplify a bit

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.h
```diff
@@ -0,0 +1,189 @@
...
+    // TRAITS
+    BSLMF_NESTED_TRAIT_DECLARATION(AnonAuthenticator,
+                                   bslma::UsesBslmaAllocator)
+
+    // CREATORS
+
+    /// Construct an AnonAuthenticator with the specified `config`.  If
+    /// `shouldPass` in `config` is set to true, authentication will always
+    /// succeed; if false, authentication will always fail.  If no config is
+    /// set, authentication defaults to always succeed.
```
678098:
```suggestion
    /// succeed; if false, authentication will always fail.  If no config is
    /// provided, or `shouldPass` setting is missing, authentication
    /// defaults to always succeed.
```
Might be good to mention NULL config case

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.h
```diff
@@ -0,0 +1,189 @@
...
+    /// Authenticate using the data provided in the specified `input`.
+    /// Behavior depends on the `shouldPass` configuration:
+    /// - If `shouldPass` is true:
+    ///   Return `0` on success, and populate the specified `result` with
+    ///   client identity and its remaining lifetime if it has a fixed
+    ///   duration.
+    /// - If `shouldPass` is false:
+    ///   Return a non-zero plugin-specific return code upon failure, and
+    ///   populate the specified `errorDescription` with a brief reason for
+    ///   logging purposes. The `result` will not be populated.
```
678098: `shouldPass` usage is mentioned in two places.
The alternative is to describe the configuration format on top of this file in common docs just once, and keep shorter docs for these functions

678098: It can be done similarly to https://github.com/bloomberg/blazingmq/blob/2c19b3d673870cdab725f303492ca146e18590a1/src/groups/bmq/bmqsys/bmqsys_statmonitorsnapshotrecorder.h#L28-L29

Probably not worth to provide concrete code examples

src/groups/mqb/mqbplug/mqbplug_pluginmanager.h
```diff
@@ -129,7 +129,7 @@ class PluginManager BSLS_CPP11_FINAL {
     ///
     /// See `PluginLibrary::loadPluginLibrary()` comments for more detail.
     void enableRequiredPlugins(
-        const bslstl::StringRef&                         pluginPath,
+        bsl::string_view                                 pluginPath,
```
678098, line 132: This is a broker-only change, so it is safe

src/groups/mqb/mqbplug/mqbplug_plugininfo.h
```diff
@@ -94,9 +94,9 @@ class PluginInfo {
 
     // MANIPULATORS
     PluginInfo& setType(mqbplug::PluginType::Enum value);
-    PluginInfo& setName(bslstl::StringRef value);
-    PluginInfo& setDescription(bslstl::StringRef value);
-    PluginInfo& setVersion(bslstl::StringRef value);
+    PluginInfo& setName(bsl::string_view value);
+    PluginInfo& setDescription(bsl::string_view value);
+    PluginInfo& setVersion(bsl::string_view value);
```
678098, line 99: These are breaking changes, not compatible with older versions of plugins

src/groups/mqb/mqbcfg/mqbcfg.xsd
```diff
@@ -303,6 +305,120 @@
...
+  </complexType>
+
+  <complexType name="AuthenticatorPluginConfig">
+    <annotation>
+      <documentation>
+        The configuration for an authenticator plugin.
+
+        name....:
+            The name of the authenticator plugin.
+        configs.:
```
678098: I still think that `settings` is a better name here. Currently:
1. There is a general broker config
2. There is a plugin config
3. Each plugin config might have multiple configs (here)

src/groups/mqb/mqbcfg/mqbcfg.xsd
```diff
@@ -303,6 +305,120 @@
...
+            Plugin-specific configurations.
+      </documentation>
+    </annotation>
+    <sequence>
+      <element name='name'    type='string'/>
+      <element name='configs' type='tns:PluginConfigKeyValue' minOccurs='0' maxOccurs='unbounded'/>
+    </sequence>
+  </complexType>
+
+  <complexType name="PluginConfigKeyValue">
```
678098:
```suggestion
  <complexType name="PluginSettingKeyValue">
```

src/groups/mqb/mqbcfg/mqbcfg.xsd
```diff
@@ -303,6 +305,120 @@
...
+      </documentation>
+    </annotation>
+    <sequence>
+      <element name='key' type='string'/>
+      <element name='value' type="tns:PluginConfigValue">
+      </element>
+    </sequence>
+  </complexType>
+
+  <complexType name="PluginConfigValue">
```
678098:
```suggestion
  <complexType name="PluginSettingValue">
```

src/groups/mqb/mqbauthn/mqbauthn_pluginlibrary.cpp
```diff
@@ -0,0 +1,75 @@
...
+// limitations under the License.
+
+#include <mqbauthn_pluginlibrary.h>
+
+// MQB
+#include <mqbauthn_anonauthenticator.h>
+#include <mqbauthn_basicauthenticator.h>
+
+// BDE
+#include <bslma_default.h>
```
678098:
```suggestion
```
Not used

src/groups/mqb/mqbplug/mqbplug_authenticator.h
```diff
@@ -173,53 +164,42 @@ class AuthenticatorPluginFactory : public PluginFactory {
...
+// ------------------------
+// class AuthenticationData
+// ------------------------
 
-inline void AuthenticationResult::reset()
+inline AuthenticationData::AuthenticationData(
+    const bsl::vector<char>& authnPayload,
+    bsl::string_view         clientIpAddress)
+: d_authnPayload(authnPayload)
+, d_clientIpAddress(clientIpAddress)
```
678098:
```suggestion
, d_clientIpAddress(clientIpAddress, allocator)
```
Missing allocator here

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.h
```diff
@@ -0,0 +1,185 @@
...
+#include <mqbplug_authenticator.h>
+
+// BDE
+#include <ball_log.h>
+#include <bsl_map.h>
+#include <bsl_memory.h>
+#include <bsl_optional.h>
+#include <bsl_string.h>
+#include <bsl_string_view.h>
+#include <bsla_annotations.h>
```
678098:
```suggestion
```

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
```diff
@@ -0,0 +1,212 @@
...
+bslma::ManagedPtr<mqbplug::Authenticator>
+BasicAuthenticatorPluginFactory::create(bslma::Allocator* allocator)
+{
+    const mqbcfg::AuthenticatorPluginConfig* config =
+        mqbplug::AuthenticatorUtil::findAuthenticatorConfig(
+            BasicAuthenticator::k_NAME);
+
+    bslma::ManagedPtr<mqbplug::Authenticator> result(
+        new (*allocator) BasicAuthenticator(config, allocator),
+        allocator);
```
678098: Possible to simplify with `allocateManaged`

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+void BasicAuthenticator::stop()
+{
+    if (!d_isStarted) {
+        return;  // RETURN
+    }
+
+    d_isStarted = false;
+
+    BALL_LOG_INFO << "BasicAuthenticator stopped";
```
678098, line 194: Basic Authn has start/stop logs, Anon Authn doesn't have it.
I think they should either both log or both not log

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
```diff
@@ -0,0 +1,212 @@
...
+{
+    if (d_isStarted) {
+        errorDescription << "start() can only be called once on this object";
+        return -1;
+    }
+
+    if (!d_authenticatorConfig_p) {
+        BALL_LOG_WARN << "No config provided for BasicAuthenticator '"
+                      << name() << "' with mechanism '" << mechanism();
+    }
```
678098: Similar to Anon Pass, this can be moved to constructor

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
```diff
@@ -0,0 +1,212 @@
...
+    bsl::shared_ptr<mqbplug::AuthenticationResult>* result,
+    const mqbplug::AuthenticationData&              input) const
+{
+    BALL_LOG_INFO << "BasicAuthenticator: "
+                  << "authentication using mechanism '" << mechanism() << "'.";
+
+    const bsl::vector<char>& payload = input.authnPayload();
+    bsl::string_view payloadView(reinterpret_cast<const char*>(payload.data()),
+                                 payload.size());
+    bsl::size_t      colonPos = payloadView.find(':');
```
678098, line 142: What are the allowed symbols for the password? Do we have limitations or requirements?
Do we allow to have `:` as a part of username/pass?
This line `user:pass1:pass2` parses with user = `user`, password = `pass1:pass2`

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
```diff
@@ -0,0 +1,212 @@
...
+    if (!config) {
+        return;
+    }
+
+    // Load the configured key-values as username and password
+    bsl::vector<mqbcfg::PluginConfigKeyValue>::const_iterator it =
+        config->configs().cbegin();
+    for (; it != config->configs().cend(); ++it) {
+        if (it->value().isStringValValue()) {
+            d_credentials[it->key()] = it->value().stringVal();
```
678098: Same with Anon Pass, can print something if there is a type mismatch or if there are no values(?)

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
```diff
@@ -0,0 +1,212 @@
...
+BasicAuthenticator::BasicAuthenticator(
+    const mqbcfg::AuthenticatorPluginConfig* config,
+    bslma::Allocator*                        allocator)
+: d_allocator_p(allocator)
+, d_authenticatorConfig_p(config)
+, d_credentials(allocator)
+, d_isStarted(false)
+{
+    if (!config) {
+        return;
```
678098, line 89: Put a comment below, the warning might be here

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
```diff
@@ -0,0 +1,212 @@
...
+#include <bslma_allocator.h>
+#include <bslma_managedptr.h>
+#include <bsls_assert.h>
+
+namespace BloombergLP {
+namespace mqbauthn {
+
+namespace {
+
+const int k_TIMEOUT_SECONDS = 600;
```
678098:
```suggestion
const int k_AUTHN_DURATION_SECONDS = 600;
```
Possible alternative name. The reason for this is that `timeout` word is more often used in a context of failure, e.g. timeout on establishing a connection, timeout on getting a resource etc

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
```diff
@@ -0,0 +1,212 @@
...
+#include <mqbplug_authenticator.h>
+
+// BDE
+#include <bdlt_timeunitratio.h>
+#include <bsl_iostream.h>
+#include <bsl_memory.h>
+#include <bsl_optional.h>
+#include <bsl_string.h>
+#include <bsl_string_view.h>
+#include <bsla_annotations.h>
```
678098:
```suggestion
```

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.h
```diff
@@ -0,0 +1,175 @@
...
+// limitations under the License.
+
+// mqbauthn_authenticationcontroller.h                          -*-C++-*-
+#ifndef INCLUDED_MQBAUTHN_AUTHENTICATIONCONTROLLER
+#define INCLUDED_MQBAUTHN_AUTHENTICATIONCONTROLLER
+
+/// @file mqbauthn_authenticationcontroller.h
+///
+/// @brief Provide a small utility class that orchestrates authenticator
+/// plugins.
```
678098:
```suggestion
/// @brief Provide an utility class that orchestrates authenticator
/// plugins.
```
2 ideas:
1. Less text to read -> faster reading the code.
We can remove/rephrase some comments sometimes to simplify
2. Some details are not always relevant or important. A small class might become big one day, a new algorithm might become legacy one day etc

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.h
```diff
@@ -0,0 +1,175 @@
...
+///
+/// @bbref{AuthenticationController} provides a small utility class that
```
678098:
```suggestion
/// @bbref{AuthenticationController} provides an utility class that
```

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.h
```diff
@@ -0,0 +1,175 @@
...
+    /// True if this component is started.
+    bool d_isStarted;
+
+    /// Allocator to use.
+    bslma::Allocator* d_allocator_p;
+
+  private:
+    // NOT IMPLEMENTED
+    AuthenticationController(const AuthenticationController& other)
+        BSLS_CPP11_DELETED;
```
678098:
```suggestion
        BSLS_KEYWORD_DELETED;
```

Deprecated, from BDE docs:
```
//@PURPOSE: Provide macros for C++11 forward compatibility.
//
//@DEPRECATED: Use `bsls_keyword` instead.
```

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.h
```diff
@@ -0,0 +1,175 @@
...
+    AuthenticationController&
+    operator=(const AuthenticationController& other) BSLS_CPP11_DELETED;
```
678098:
```suggestion
    operator=(const AuthenticationController& other) BSLS_KEYWORD_DELETED;
```

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -0,0 +1,483 @@
...
+
+    bmqu::MemOutStream errorStream(d_allocator_p);
+
+    // If authenticators are explicitly configured, only create those
+    const bsl::vector<mqbcfg::AuthenticatorPluginConfig>& authenticators =
+        authenticatorConfig.authenticators();
+    if (!authenticators.empty()) {
+        BALL_LOG_INFO << "Creating " << authenticators.size()
+                      << " explicitly configured authenticator(s)";
+
```
678098: Good to use an early return trick here to reduce indentation level of a big chunk of code following


```suggestion
    if (authenticators.empty()) {
        BALL_LOG_INFO << "No configured authenticator(s)";
        return rc_SUCCESS;
    }

    BALL_LOG_INFO << "Creating " << authenticators.size()
                  << " explicitly configured authenticator(s)";
```

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -0,0 +1,483 @@
...
+                errorDescription << "Duplicate authenticator mechanism '"
+                                 << normMech << "' for plugin '" << pluginName
+                                 << "'. Each mechanism can only "
+                                 << "have one active authenticator";
+                return rc_DUPLICATE_MECHANISM;  // RETURN
+            }
+
+            // Start the authenticator
+            if (int status = authenticator->start(errorStream)) {
+                BMQTSK_ALARMLOG_ALARM("#AUTHENTICATION")
```
678098: If Authn Controller makes decision to print an alarm, should it also print it on other failures in this function?

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.h
emelialei88: What does it mean? Isn't it constructed with allocator at the end?

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
emelialei88: It's used for `lifetimeMs`.

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
emelialei88: Do you mean if we should ALARM on other errors in this function? I'm actually not fully sure what are the scenarios where we trigger an ALARM.

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
emelialei88: colon (":") should be forbidden for username but permitted for password according to [HTTP RFC](https://datatracker.ietf.org/doc/html/rfc7617#section-2.1). I added one line comment at the top of the header file. Lmk if there's a better place for documentation.

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.h
678098: It is an old comment, `AnonFailAuthenticator` didn't use allocator at all. The only existing `AnonAuthenticator` uses it now, so the comment is unrelated

src/groups/mqb/mqbplug/mqbplug_statconsumer.h
678098: Updated in another PR https://github.com/bloomberg/blazingmq/pull/993

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
678098: It is included in the corresponding header already

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
678098: Can move forward with this for now, can discuss the alarms separately.

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
```diff
@@ -0,0 +1,209 @@
...
+    const mqbcfg::AuthenticatorPluginConfig* config,
+    bslma::Allocator*                        allocator)
+: d_allocator_p(allocator)
+, d_isStarted(false)
+, d_shouldPass(true)
+{
+    if (!config) {
+        BALL_LOG_INFO
+            << "No configuration provided, configuring AnonAuthenticator with "
+               "the default shouldPass = true";
```
678098:
```suggestion
            << "No configuration provided, using the default shouldPass = true";
```
I just understood that we already have `MQBAUTHN.ANONAUTHENTICATOR` as a log prefix, this means that we can omit naming this class in this log. Shorter logs are easier to read, and they take less space

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    // Load the configured `shouldPass` argument
+    bool isShouldPassFound = false;
+    bsl::vector<mqbcfg::PluginSettingKeyValue>::const_iterator it =
+        config->settings().cbegin();
+    for (; it != config->settings().cend(); ++it) {
+        if (it->key() == "shouldPass") {
+            if (!it->value().isBoolValValue()) {
+                BALL_LOG_WARN << "Expected bool for 'shouldPass' setting...";
```
678098:
```suggestion
                BALL_LOG_WARN << "Expected bool for 'shouldPass' setting, got type id = " << it->value().selectionId();
```
Possible to add more verbosity here and print type index. It might help to debug one day if this warning is triggered

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
```diff
@@ -0,0 +1,209 @@
...
+                continue;
+            }
+            d_shouldPass      = it->value().boolVal();
```
678098, line 99:
```suggestion
            if (isShouldPassFound) {
                BALL_LOG_WARN << "Encountered duplicating setting 'shouldPass', overriding";
            }
            d_shouldPass      = it->value().boolVal();
```
Another case. I think the current settings format is a list of key-value items, right? So no one stops us from having the same setting multiple times

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
```diff
@@ -0,0 +1,209 @@
...
+            isShouldPassFound = true;
+        }
+    }
+    if (isShouldPassFound) {
+        BALL_LOG_INFO << "... setting found, using shouldPass = "
+                      << d_shouldPass;
```
678098:
```suggestion
        BALL_LOG_INFO << "Setting found in configuration: using shouldPass = "
                      << (d_shouldPass ? "true" : "false");
```
1. I meant to replace `...` with something meaningful.
This variant is good
2. Note that `<< (bool)` will print 0 or 1. It might be confusing because we forbid to use `0` or `1` in the config itself, we only expect bool `"true"`/`"false"` there. For this reason it is better to also log `true`/`false`

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
```diff
@@ -0,0 +1,209 @@
...
+    }
+    else {
+        BALL_LOG_INFO
+            << "... setting not found, using the default shouldPass = "
+            << d_shouldPass;
```
678098:
```suggestion
            << "Setting not found in configuration: using the default shouldPass = "
            << (d_shouldPass ? "true" : "false");
```

src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp
```diff
@@ -0,0 +1,209 @@
...
+{
+    if (d_isStarted) {
+        errorDescription << "start() can only be called once on this object";
+        return -1;  // RETURN
+    }
+
+    d_isStarted = true;
+
+    BALL_LOG_INFO << "AnonAuthenticator started with shouldPass = "
+                  << d_shouldPass;
```
678098:
```suggestion
                  << (d_shouldPass ? "true" : "false");
```

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
```diff
@@ -0,0 +1,218 @@
...
+    return k_MECHANISM;
+}
+
+int BasicAuthenticator::authenticate(
+    bsl::ostream&                                   errorDescription,
+    bsl::shared_ptr<mqbplug::AuthenticationResult>* result,
+    const mqbplug::AuthenticationData&              input) const
+{
+    BALL_LOG_INFO << "BasicAuthenticator: "
+                  << "authentication using mechanism '" << mechanism() << "'.";
```
678098:
```suggestion
    BALL_LOG_INFO << "Authenticating using mechanism '" << mechanism() << "'.";
```
Can omit the class name because we have log prefix `MQBAUTHN.BASICAUTHENTICATOR`

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
```diff
@@ -0,0 +1,218 @@
...
+BasicAuthenticator::BasicAuthenticator(
+    const mqbcfg::AuthenticatorPluginConfig* config,
+    bslma::Allocator*                        allocator)
+: d_allocator_p(allocator)
+, d_credentials(allocator)
+, d_isStarted(false)
+{
+    if (!config) {
+        BALL_LOG_WARN << "No config provided for BasicAuthenticator '"
+                      << name() << "' with mechanism '" << mechanism() << "'.";
```
678098: The current log looks like

`MQBAUTHN.BASICAUTHENTICATOR No config provided for BasicAuthenticator 'BasicAuthenticator' with mechanism 'BASIC'`

Can be simpler:

```suggestion
        BALL_LOG_WARN << "No configuration provided, have no credentials";
```

I think `name()` and `mechanism()` should mostly be used by external classes that operate authenticators in a generic way without knowing the concrete authenticator type

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
```diff
@@ -0,0 +1,218 @@
...
+        config->settings().cbegin();
+    for (; it != config->settings().cend(); ++it) {
+        if (!it->value().isStringValValue()) {
+            BALL_LOG_WARN << "Expected string for 'key-value' setting...";
+            continue;
+        }
+        d_credentials[it->key()] = it->value().stringVal();
+        isValueFound             = true;
+    }
+    if (isValueFound) {
```
678098: Can get rid of `isValueFound`, we have information about the number of added credentials to print a log:
```suggestion
    if (!d_credentials.empty()) {
```

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
```diff
@@ -0,0 +1,218 @@
...
+        return;
+    }
+
+    // Load the configured key-values as username and password
+    bool isValueFound = false;
+    bsl::vector<mqbcfg::PluginSettingKeyValue>::const_iterator it =
+        config->settings().cbegin();
+    for (; it != config->settings().cend(); ++it) {
+        if (!it->value().isStringValValue()) {
+            BALL_LOG_WARN << "Expected string for 'key-value' setting...";
```
678098:
```suggestion
            BALL_LOG_WARN << "Expected string for credential, got type id = " << it->value().selectionId();
```
Possible to add more verbosity here

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
```diff
@@ -0,0 +1,218 @@
...
+            continue;
+        }
+        d_credentials[it->key()] = it->value().stringVal();
+        isValueFound             = true;
+    }
+    if (isValueFound) {
+        BALL_LOG_INFO << "... setting found.  Loaded " << d_credentials.size()
+                      << " credential(s).";
```
678098:
```suggestion
        BALL_LOG_INFO << "Loaded " << d_credentials.size()
                      << " credential(s) from configuration";
```

src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp
```diff
@@ -0,0 +1,218 @@
...
+    }
+    else {
+        BALL_LOG_INFO << "... setting not found.  Make sure this is expected.";
```
678098:
```suggestion
        BALL_LOG_WARN << "No credentials found in configuration. Please double check if this is intended."
```
Can log a similar message to https://github.com/bloomberg/blazingmq/blob/157313364b9eb131df3022f1462834c0f609fcf3/src/groups/mqb/mqbplug/mqbplug_pluginmanager.cpp#L320

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -0,0 +1,483 @@
...
+    };
+
+    const bdlb::NullableValue<mqbcfg::AnonymousCredential>& anonCredential =
+        authenticatorConfig.anonymousCredential();
+
+    if (anonCredential.isNull()) {
+        bsl::string anonMechanism = bsl::string(AnonAuthenticator::k_MECHANISM,
+                                                d_allocator_p);
+
+        if (d_authenticators.find(anonMechanism) != d_authenticators.cend()) {
```
678098, line 309:
```suggestion
        if (d_authenticators.find(AnonAuthenticator::k_MECHANISM) != d_authenticators.cend()) {
```
Possible to avoid constructing tmp string

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -0,0 +1,483 @@
...
+
+    return rc_SUCCESS;
+}
+
+int AuthenticationController::ensureDefaultAuthenticator(
+    bsl::ostream& errorDescription)
+{
+    // If no authenticators are configured, use AnonAuthenticator for
+    // anonymous pass authentication
+    if (d_authenticators.empty()) {
```
678098:
```suggestion
    if (!d_authenticators.empty()) {
        // Have authenticators, nothing to do
        return 0;
    }

    // If no authenticators are configured, use AnonAuthenticator for
    // anonymous pass authentication
```
Not strictly necessary to do, but I would make an early return here to reduce indentation of the following code.
Your original option is also good: it only has 1 return, not 2.

If the body of `if-then` has more lines or if there are even more layers of indentation, early return looks better.

678098: How it might help the reader.
Consider the following code:
```
void foo(bool flag) {
    if (flag) {
        // 200 lines here
    }
    // nothing in the end
}
```

If a reader sees `if` like this, they must scroll down 200 lines to see the end of this function just to know if there is any more processing if `flag == false`.

Compared to this:
```
void foo(bool flag) {
    if (!flag) return;
    // 200 lines here
}
```

The reader knows that after the first line (with early return) there are no instructions for `flag == false`


Another case:
```
void foo(bool flag) {
    if (flag) {
        // 100 lines here
    }
    if (somethingElse()) {
        // 100 lines here
    }
    // nothing in the end
}
```
In this case, it is not even enough to scroll down 200 lines, the reader has to notice the end of the current `if` structure and start of another structure not related to `flag`. This might be solved by folding structures in IDE, however, these are extra mouse clicks

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -0,0 +1,483 @@
...
+    d_isStarted = true;
+    return rc_SUCCESS;
+}
+
+void AuthenticationController::stop()
+{
+    if (!d_isStarted) {
+        return;  // RETURN
+    }
+
```
678098, line 421:
```suggestion
    BALL_LOG_INFO << "Stopping AuthenticationController";

```
If we log starting, we might log stopping as well

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -0,0 +1,483 @@
...
+                             << normMech << "'. (rc = " << rc
+                             << "). Detailed error: " << errorStream.str();
+            return (rc * 10 + rc_AUTHENTICATION_FAILED);
+        }
+    }
+    else {
+        errorDescription
+            << "AuthenticationController: authentication mechanism '"
+            << normMech << "' not supported.";
+        return (rc * 10 + rc_MECHANISM_NOT_SUPPORTED);
```
678098: `rc == e_SUCCESS` here, might simplify:

```suggestion
        return rc_MECHANISM_NOT_SUPPORTED;
```

We already have functions where we mix `const * rc + enum` with just `rc` or `enum`, this for example: https://github.com/bloomberg/blazingmq/blob/157313364b9eb131df3022f1462834c0f609fcf3/src/groups/mqb/mqbc/mqbc_clusterstateledgerutil.cpp#L512

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -0,0 +1,483 @@
...
+
+    enum RcEnum {
+        // Enum for the various RC error categories
+        rc_SUCCESS                 = 0,
+        rc_AUTHENTICATION_FAILED   = -1,
+        rc_MECHANISM_NOT_SUPPORTED = -2
+    };
+
+    int                rc = rc_SUCCESS;
+    bmqu::MemOutStream errorStream(d_allocator_p);
```
678098:
```suggestion
```
Possible to move closer to `authenticate` below.
This is a "minimal scope" principle for definition of variables.
Minimal scope means that a reader has to keep in mind less variables at a point of time, making reading the code easier. Also, sometimes (rarely) defining variables in the least scope possible prevents others from misusing these variables later

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -0,0 +1,483 @@
...
+
+    const bsl::string normMech = normalizeMechanism(mechanism, d_allocator_p);
+    AuthenticatorMap::const_iterator cit = d_authenticators.find(normMech);
+    if (cit != d_authenticators.cend()) {
+        const AuthenticatorMp& authenticator = cit->second;
+        BALL_LOG_DEBUG << "AuthenticationController: "
+                       << "authenticating with mechanism '" << normMech << "'"
+                       << " (authenticator: '" << authenticator->name()
+                       << "')";
+        rc = authenticator->authenticate(errorStream, result, input);
```
678098:
```suggestion
    bmqu::MemOutStream errorStream(d_allocator_p);
    const int          rc = authenticator->authenticate(errorStream, result, input);
```

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
emelialei88: It would cause a compiler error since `bsl::unordered_map<bsl::string, AuthenticatorMp>` doesn't accept `bsl::string_view` as the parameter to `find`.

678098 (approved):

PR #991, opened 2025-10-20
Fix[MQB]: drop unknown Rejects (`Assertion failed: result == Routers::e_INVALID`)

dorjesinpo:
When switching primaries, if the new Primary did not get a chance to PUSH the 'msgGUID' (which the old Primary did) _before_ the Downstream generated the Reject, drop the Reject 

The logic of QueueEngine(s) throttles (re)delivery for messages in redelivery lists only and asserts otherwise.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
@@ -195,7 +195,7 @@ void QueueHandle::confirmMessageDispatched(const bmqt::MessageGUID& msgGUID,
             << "Received a CONFIRM message from client '"
             << d_clientContext_sp->client()->description() << "' for queue '"
             << d_queue_sp->description()
-            << "' which doesn't have the subQueue: " << "downstreamSubQueueId";
+            << "' which doesn't have the subQueue: " << downstreamSubQueueId;
         return;
     }
     const bsl::shared_ptr<Downstream>& subStream = downstream(


src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
```diff
@@ -252,22 +252,37 @@ void QueueHandle::rejectMessageDispatched(const bmqt::MessageGUID& msgGUID,
...
-        updateMonitor(subStream, msgGUID, bmqp::EventType::e_REJECT);
-        // That erases 'msgGUID' from 'd_unconfirmedMessages'
+        // This can happen when switching primaries and the new Primary did not
+        // get a chance to PUSH the 'msgGUID' (which the old Primary did)
+        // _before_ the Downstream generated the Reject.
+
+        // The logic of QueueEngine(s) throttles (re)delivery for messages in
+        // redelivery lists only and asserts otherwise.
+
+        BALL_LOG_WARN << "#CLIENT_IMPROPER_BEHAVIOR "
```
678098: I think it's not correct to blame the client here, client doesn't know anything about the current primary and it's not its responsibility to wait for the right moment to crash

What about
```suggestion
        BALL_LOG_WARN << "#UNEXPECTED_REJECT "
```
?

Secondly, do you think it's worth to throttle this log? Is it possible to have thousands or tens of thousands unexpected rejects?

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
```diff
@@ -252,22 +252,37 @@ void QueueHandle::rejectMessageDispatched(const bmqt::MessageGUID& msgGUID,
...
+                      << "Received a REJECT message from client '"
+                      << d_clientContext_sp->client()->description()
+                      << "' for queue '" << d_queue_sp->description()
+                      << "' for a message not in pending unconfirmed.";
```
678098: Log will be more helpful with `msgGUID` logged

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
678098: Also, need to revisit the logs above (lines 244, 229), is it client improper behaviour indeed or not?

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.t.cpp
```diff
@@ -5074,6 +5074,69 @@ static void test47_handleParametersLimits()
     }
 }
 
+static void test48_unknownReject()
+// ------------------------------------------------------------------------
+// UNKNOWN REJECT
```
678098: I have a proposal to log `#UNEXPECTED_REJECT`, we can use the same terminology everywhere. We can put UNEXPECTED REJECT here too to make it consistent, or we can log `#UNKNOWN_REJECT` above

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.t.cpp
```diff
@@ -5074,6 +5074,69 @@ static void test47_handleParametersLimits()
...
+//
+// Concerns:
+//   When switching Primary, the new Primary can receive a Reject before
+//   sending PUSH.
+//
+// Plan:
+//   1) Configure 3 handles, C1, C2, C3  each with distinct app ids.
```
678098:
```suggestion
//   1) Configure 3 handles, C1, C2, C3 each with distinct app ids.
```

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
dorjesinpo: > Also, need to revisit the logs above (lines 244, 229), is it client improper behaviour indeed or not?

I guess by "client" we mean downstream.  Could be either  ClientSession or ClusterNode

678098: My main point is, if it's not physically possible to ensure proper behaviour on the client (downstream) side, we cannot blame it in logs, because it adds confusion

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
```diff
@@ -241,33 +248,50 @@ void QueueHandle::rejectMessageDispatched(const bmqt::MessageGUID& msgGUID,
...
-                      << d_clientContext_sp->client()->description()
-                      << "' for queue '" << d_queue_sp->description()
-                      << "' which doesn't have the subQueue: "
-                      << downstreamSubQueueId;
+        BMQ_LOGTHROTTLE_WARN
+            << "#CLIENT_IMPROPER_BEHAVIOR "
+            << "Received a REJECT message from client '"
+            << d_clientContext_sp->client()->description() << "' for queue '"
+            << d_queue_sp->description()
+            << "' which doesn't have the subQueue: " << downstreamSubQueueId;
```
678098, line 255: What do you think about additionally logging `msgGUID` here as well and on line 236?

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
```diff
@@ -183,12 +190,12 @@ void QueueHandle::confirmMessageDispatched(const bmqt::MessageGUID& msgGUID,
...
-                      << d_clientContext_sp->client()->description()
-                      << "' for queue '" << d_queue_sp->description()
-                      << "' which doesn't have the subQueue: "
-                      << "downstreamSubQueueId";
+        BMQ_LOGTHROTTLE_WARN << "#CLIENT_IMPROPER_BEHAVIOR "
+                             << "Received a CONFIRM message from client '"
+                             << d_clientContext_sp->client()->description()
+                             << "' for queue '" << d_queue_sp->description()
+                             << "' which doesn't have the subQueue: "
+                             << "downstreamSubQueueId";
```
678098:
```suggestion
                             << downstreamSubQueueId;
```
This looks like an error in the original code

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
678098: I can do this separately if you don't want to increase the scope of this PR

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
```diff
@@ -183,12 +190,12 @@ void QueueHandle::confirmMessageDispatched(const bmqt::MessageGUID& msgGUID,
...
-                      << d_clientContext_sp->client()->description()
-                      << "' for queue '" << d_queue_sp->description()
-                      << "' which doesn't have the subQueue: "
-                      << "downstreamSubQueueId";
+        BMQ_LOGTHROTTLE_WARN
+            << "#CLIENT_IMPROPER_BEHAVIOR "
+            << "Received a CONFIRM message from client '"
+            << d_clientContext_sp->client()->description() << "' for queue '"
+            << d_queue_sp->description()
+            << "' which doesn't have the subQueue: " << "downstreamSubQueueId";
```
678098:
```suggestion
            << "' which doesn't have the subQueue: " << downstreamSubQueueId;
```

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
678098: @dorjesinpo

PR #992, opened 2025-10-21
IT[test_fsm_partition_sync.py] Add tests for FSM mode to cover partition synchronization scenarios

alexander-e1off:
Add integration tests to test cluster synchronization in FSM mode:
- when replica missed records;
- when primary missed records;

### DIFF
--- a/src/integration-tests/test_fsm_partition_sync.py
+++ b/src/integration-tests/test_fsm_partition_sync.py
@@ -369,7 +369,7 @@ def test_sync_if_leader_missed_records(
     for node in sorted_nodes:
         node.start()
         node.wait_until_started()
-        quorum = 1 if node == next_leader else 5
+        quorum = 3 if node == next_leader else 5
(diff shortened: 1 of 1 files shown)


src/integration-tests/test_fsm_partition_sync.py
```diff
@@ -253,3 +253,131 @@ def test_sync_after_missed_rollover_after_restart(
...
+    cluster.drain()
+
+    # Start all nodes, `next_leader` is the first, force it to be a leader
+    sorted_nodes = sorted(
+        cluster.nodes(), key=lambda node: 0 if node == next_leader else 1
+    )
+    for node in sorted_nodes:
+        node.start()
+        node.wait_until_started()
+        quorum = 1 if node == next_leader else 5
```
kaikulimu:
```suggestion
        quorum = 3 if node == next_leader else 5
```
I will do this change in a commit myself

PR #996, opened 2025-10-22
Fix[bmqimp]: Close channel on open, reopen, and close timeouts

pniedzielski:
This PR causes the client to close the channel to the broker when opening, reopening, or closing a queue times out, rather than marking the queue as expired.

### DIFF
--- a/src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
+++ b/src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
@@ -3174,13 +3174,18 @@ void BrokerSession::asyncRequestNotifier(
     }
     else if (context->isError()) {
         bmqt::GenericResult::Enum result = context->result();
-
-        enqueueSessionEvent(eventType,
-                            result,
-                            context->response().choice().status().message(),
-                            correlationId,
-                            queue,
-                            eventCallback);
+        // Don't deliver error on e_QUEUE_REOPEN_RESULT.  This is to avoid
+        // existing applications incorrectly handling e_QUEUE_REOPEN_RESULT
+        // failures.  It's hard for an application to handle this correctly.
+        if (bmqt::SessionEventType::e_QUEUE_REOPEN_RESULT != eventType) {
+            enqueueSessionEvent(
+                eventType,
+                result,
+                context->response().choice().status().message(),
+                correlationId,
+                queue,
+                eventCallback);
+        }
     }
     else {
         enqueueSessionEvent(eventType,
--- a/src/integration-tests/test_client_timeout.py
+++ b/src/integration-tests/test_client_timeout.py
@@ -122,6 +122,12 @@ def test_client_timeout_reopen(multi_node: Cluster, domain_urls: tc.DomainUrls):
         == Client.e_SUCCESS
     )
 
+    # Make sure the producer has not seen a QUEUE_REOPEN_RESULT failure.
+    assert not producer.outputs_substr(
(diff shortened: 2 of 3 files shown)


src/integration-tests/test_client_timeout.py
```diff
@@ -0,0 +1,169 @@
...
+def test_client_timeout_open(multi_node: Cluster, domain_urls: tc.DomainUrls):
+    """
+    If opening a queue times out on the client, the client should kill the
+    channel, dropping all queue connections, and mark the queue as closed.
+    """
+    uri_priority = domain_urls.uri_priority
+    brokers = multi_node.nodes()
+    leader = multi_node.last_known_leader
+
+    producer_broker = brokers[1]
```
dorjesinpo: Does this rely on the leader being always `brokers[0]`?  We do not want `producer_broker == leader`, do we?

src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
```diff
@@ -2405,82 +2375,104 @@ void BrokerSession::QueueFsm::handleResponseTimeout(
...
     } break;
     case QueueState::e_CLOSING_CFG: {
         // Expect 'CloseQueue' request and 'Status' response.
         BSLS_ASSERT_SAFE(context->request().choice().isCloseQueueValue());
         BSLS_ASSERT_SAFE(bmqt::QueueFlagsUtil::isReader(queue->flags()));
 
-        // Set EXPIRED state
-        setQueueState(queue, QueueState::e_CLOSING_CFG_EXPIRED, event);
+        // Set CLOSED state
+        setQueueState(queue, QueueState::e_CLOSED, event);
```
dorjesinpo, line 2458: do we not need to `actionRemoveQueue(queue);`?  Does not look that `e_CHANNEL_DOWN` will do that in `e_CLOSED` state.  Maybe, it should?

src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
```diff
@@ -2405,82 +2375,104 @@ void BrokerSession::QueueFsm::handleResponseTimeout(
...
+        d_session.d_channel_sp->close();
     } break;
     case QueueState::e_CLOSING_CLS: {
         // Expect 'CloseQueue' request and 'Status' response.
         BSLS_ASSERT_SAFE(context->request().choice().isCloseQueueValue());
 
-        // Set EXPIRED state
-        setQueueState(queue, QueueState::e_CLOSING_CLS_EXPIRED, event);
+        // Set CLOSED state
+        setQueueState(queue, QueueState::e_CLOSED, event);
```
dorjesinpo, line 2476: same comment about `actionRemoveQueue`.

src/integration-tests/test_client_timeout.py
pniedzielski: I'll make this explicit.  I think we do this in other ITs too, so TODO for a different PR to fix more broadly.

src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
pniedzielski: fixed

src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
pniedzielski: fixed

PR #999, opened 2025-10-23
Fix[STATS]: report domain queue count after recovery

waldgange:
This PR fixes the bug described below:
After restart broker recovers the queues containing outstanding messages and starts reportng queue-related metrics. However domain-related metric `domain_queue_count` is not reported. Moreover, it is not reported until a client opens a new queue that has not been recovered before.
The problem:
`ClusterState` class contains a map of `DomainState`s which contain pointers to their `Domain`s. These pointers don't get initialized if `Domains` are created during recovery, though do get initialized if the `Domains` are created upon `QueueOpen` request.
The solution:
Initialise DomainStates and give them pointers on their domains after recovery.

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_storageutil.cpp
+++ b/src/groups/mqb/mqbc/mqbc_storageutil.cpp
@@ -1441,6 +1441,7 @@ void StorageUtil::recoveredQueuesCb(
     BSLS_ASSERT_SAFE(fs);
     BSLS_ASSERT_SAFE(unrecognizedDomainsLock);
     BSLS_ASSERT_SAFE(unrecognizedDomains && unrecognizedDomains->empty());
+    BSLS_ASSERT_SAFE(clusterState);
     BSLS_ASSERT_SAFE(0 <= partitionId);
     BSLS_ASSERT_SAFE(fs->inDispatcherThread());
 
@@ -1604,13 +1605,13 @@ void StorageUtil::recoveredQueuesCb(
 
     // Notify 'ClusterState' about the recognized domains to initialize the
     // corresponding `DomainStates`. So it can start reporting domain-related
-    // metrics immediately after recovery.
+    // metrics almost immediately after recovery.
     if (!recognizedDomains.empty()) {
         clusterState->cluster()->dispatcher()->execute(
             bdlf::BindUtil::bindS(allocator,
                                   &ClusterState::onDomainsCreated,
-                                 clusterState,
-                                 recognizedDomains),
+                                  clusterState,
+                                  recognizedDomains),
             clusterState->cluster());
     }
 


src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -208,7 +208,7 @@ void Cluster::startDispatched(bsl::ostream* errorDescription, int* rc)
                       d_clusterData.clusterConfig(),
                       this,
                       &d_clusterData,
-                      d_state,
+                      &d_state,
```
678098, line 211: The lifetime requirement for `d_state` remains the same: safe to refactor

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -1592,9 +1596,23 @@ void StorageUtil::recoveredQueuesCb(
                     dit->first,
                     mqbs::StorageUtil::QueueMessagesCountMap()));
             }
+            else {
+                recognizedDomains.insert(*dit);
+            }
         }
     }
 
+    // Notify 'ClusterSate' about the recognized domains to initialize the
```
678098:
```suggestion
    // Notify 'ClusterState' about the recognized domains to initialize the
```

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -1592,9 +1596,23 @@ void StorageUtil::recoveredQueuesCb(
...
+    // corresponding `DomainStates`. So it can start reporting domain-related
+    // metrics immediately after recovery.
+    if (!recognizedDomains.empty()) {
+        clusterState->cluster()->dispatcher()->execute(
+            bdlf::BindUtil::bind(&ClusterState::onDomainsCreated,
```
678098:
```suggestion
            bdlf::BindUtil::bindS(allocator,
                                  &ClusterState::onDomainsCreated,
```
Not strictly necessary

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -1592,9 +1596,23 @@ void StorageUtil::recoveredQueuesCb(
...
             }
+            else {
+                recognizedDomains.insert(*dit);
+            }
         }
     }
 
+    // Notify 'ClusterSate' about the recognized domains to initialize the
+    // corresponding `DomainStates`. So it can start reporting domain-related
+    // metrics immediately after recovery.
```
678098: Strictly speaking, there is a small time window when the metrics are not initialized yet and might be reported wrong: `ClusterState::onDomainsCreated` might take some time to execute. However, nothing more can be done here

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -1428,9 +1428,11 @@ void StorageUtil::recoveredQueuesCb(
     mqbi::DomainFactory*         domainFactory,
     bslmt::Mutex*                unrecognizedDomainsLock,
     DomainQueueMessagesCountMap* unrecognizedDomains,
+    mqbc::ClusterState*          clusterState,
```
678098, line 1431:
```
    BSLS_ASSERT_SAFE(clusterState);
```
Might add a precondition check for a new arg

chrisbeard:
With this change, what are the semantics of the metric? Is it "the number of queues per domain in storage" or "the number of currently opened queues per domain"?

I think we'll eventually want both of those metrics, but they should be different metrics.

src/groups/mqb/mqbc/mqbc_storageutil.cpp
waldgange: edited the comment

src/groups/mqb/mqbc/mqbc_storageutil.cpp
waldgange: added the check

678098 (approved):

PR #1000, opened 2025-10-23
Feat[MQB]: add authentication message schema and event

emelialei88:


src/groups/bmq/bmqp/bmqp_event.h
```diff
@@ -197,6 +201,13 @@ class Event {
     template <class TYPE>
     int loadElectorEvent(TYPE* message) const;
 
+    /// Load into the specified `message`, the decoded message contained in
+    /// this event.  The behavior is undefined unless `isAuthenticationEvent()`
+    /// returns true.  Return 0 on success, and a non-zero return code on
+    /// error.
+    template <class TYPE>
+    int loadAuthenticationEvent(TYPE* message) const;
```
678098, line 209: I am not sure if this function is really needed, if we can check `isAuthenticationEvent()` and call `loadSchemaEvent` directly

678098 (approved):

src/groups/bmq/bmqp/bmqp_event.h
emelialei88: I did this to stay consistent with what we did for `loadElectorEvent` and `loadControlEvent`.

678098: This makes sense

PR #1004, opened 2025-10-24
Feat[MQB]: Add Authenticator and InitialConnectionContext

emelialei88:
This PR introduces several refactors and enables authentication:
- Adds an `Authenticator` class to encapsulate authentication logic.
- Migrates connection initialization logic from `InitialConnectionHandler` to `InitialConnectionContext`.

### DIFF
--- a/src/integration-tests/test_authn.py
+++ b/src/integration-tests/test_authn.py
@@ -139,9 +139,7 @@ def test_authenticate_concurrent(
     def auth_worker(idx):
         client = RawClient()
         client.open_channel(*single_node.admin_endpoint)
-        auth_resp = client.authenticate(
-            "Basic", f"user{idx}:password{idx}"
-        )
+        auth_resp = client.authenticate("Basic", f"user{idx}:password{idx}")
         results[idx] = auth_resp["authenticationResponse"]["status"]["code"]
         client.negotiate()
         client.stop()


src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -347,31 +347,44 @@ void TCPSessionFactory::handleInitialConnection(
...
     // Register as observer of the channel to get the 'onClose'
+    bsl::weak_ptr<InitialConnectionContext> context_wp =
+        initialConnectionContext;
     channel->onClose(
         bdlf::BindUtil::bindS(d_allocator_p,
                               &TCPSessionFactory::onClose,
                               this,
-                              initialConnectionContext,
+                              context_wp,
+                              channel,
```
dorjesinpo, line 398: The comment below is unclear.  What mutex is meant by "we can't have mutex?

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -42,44 +228,497 @@ InitialConnectionContext::~InitialConnectionContext()
...
+
+    bsl::shared_ptr<mqbnet::Session> session;
+
+    if (rc == rc_SUCCESS && d_state == InitialConnectionState::e_NEGOTIATED) {
+        rc = d_negotiator_p->createSessionOnMsgType(errStream, &session, this);
+        BALL_LOG_INFO << "Created a session with " << channel()->peerUri();
+    }
+
+    if (rc != rc_SUCCESS || d_state == InitialConnectionState::e_NEGOTIATED) {
+        BALL_LOG_INFO << "Finished initial connection with rc = " << rc
```
dorjesinpo, line 692: if `rc != rc_SUCCESS` then `d_state` can remain `InitialConnectionState::e_NEGOTIATED`.  Probably, not a problem since the object is about to get destructed

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -347,31 +347,44 @@ void TCPSessionFactory::handleInitialConnection(
...
+                this,
+                bdlf::PlaceHolders::_1,  // status
+                bdlf::PlaceHolders::_2,  // errorDescription
+                bdlf::PlaceHolders::_3,  // session
+                bdlf::PlaceHolders::_4,  // channel
+                bdlf::PlaceHolders::_5,  // initialConnectionContext
+                context));
 
     // Register as observer of the channel to get the 'onClose'
+    bsl::weak_ptr<InitialConnectionContext> context_wp =
```
dorjesinpo: What is the motivation behind the use of `weak_ptr`?

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,501 @@
...
+                                                       context);
+    if (buildResponseRc != rc_SUCCESS) {
+        rc    = (buildResponseRc * 10) + rc_BUILD_RESPONSE_FAILED;
+        error = buildResponseErrStream.str();
+        return;  // RETURN
+    }
+
+    // For anonymous authentication, skip sending the response and proceed
+    // directly to the next negotiation step.
+    if (isDefaultAuthn) {
```
dorjesinpo, line 277: minor, if `isReauthn`, there is no need for `authenticationResponse`

src/groups/mqb/mqbnet/mqbnet_authenticationcontext.cpp
```diff
@@ -0,0 +1,261 @@
...
+{
+    // executed by an *AUTHENTICATION* thread
+
+    // PRECONDITION
+    BSLS_ASSERT_SAFE(scheduler_p);
+
+    bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);  // LOCKED
+
+    if (d_state != AuthenticationState::e_AUTHENTICATING) {
+        errorDescription << "State not AUTHENTICATING (was " << d_state << ")";
```
dorjesinpo: `was` -> `is`

dorjesinpo (changes requested):
interim review.
we have pairs of the following:
`decodeInitialConnectionMessage`, 
`scheduleRead`, 
`readCallback`
`readBlob`,
`processBlob`,
`complete`
... in both `InitialConnectionHandler` and `InitialConnectionContext`.

The point of FSM-like state management is that everything happens in `handleEvent` and there is only one place where we call `complete`
Otherwise, we can't rely on controlling actions by the state 

Is it possible to engage `handleEvent` immediately in `InitialConnectionHandler::handleInitialConnection` and call `InitialConnectionContext::complete` from `handleEvent` only.

We seem to have `InitialConnectionContext::handleInitialConnection` but we call `InitialConnectionHandler::handleInitialConnection` instead which does not seem to support auth?

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
emelialei88: I'm not sure what you meant. This comment was made prior to my change and has been removed.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
emelialei88: Previously we have `InitialConnectionContext` bind to `onClose`, which means it won't be destroyed until the channel closes. It's works but I was hoping to clean it up once the initial connection part finishes, to make the lifetime match with the semantic.

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
emelialei88: Do you mean `InitialConnectionState::e_FAILED`?

src/groups/mqb/mqbmock/mqbmock_cluster.cpp
```diff
@@ -33,6 +33,7 @@
 // BDE
 #include <bdlbb_pooledblobbufferfactory.h>
 #include <bsl_iostream.h>
+#include <bsl_memory.h>
```
dorjesinpo, line 36: just double checking if this is necessary

src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -320,6 +334,13 @@ int Application::start(bsl::ostream& errorDescription)
     }
 
     // Start the transport manager
+    bslma::ManagedPtr<mqbnet::Authenticator> authenticatorMp(
+        new (*d_allocator_p) Authenticator(d_authenticationController_mp.get(),
+                                           &d_blobSpPool,
+                                           d_scheduler_p,
+                                           d_allocators.get("Authenticator")),
+        d_allocator_p);
+
```
dorjesinpo, line 341: why not creating `SessionNegotiator` below in the same way (vs.  using raw pointer and then `bslma::ManagedPtr`)?

src/groups/mqb/mqba/mqba_authenticator.h
```diff
@@ -0,0 +1,265 @@
...
+#include <bsl_string_view.h>
+#include <bslma_allocator.h>
+#include <bslma_usesbslmaallocator.h>
+#include <bslmf_nestedtraitdeclaration.h>
+
+namespace BloombergLP {
+
+// FORWARD DECLARATION
+namespace mqbblp {
+class ClusterCatalog;
```
dorjesinpo: is it needed?

src/groups/mqb/mqba/mqba_authenticator.h
```diff
@@ -0,0 +1,265 @@
...
+}
+namespace mqbi {
+class Dispatcher;
```
dorjesinpo: is it needed?

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,501 @@
...
+/// for connections.  For incoming connections, it authenticates using the
+/// received AuthenticationRequest and responds with an AuthenticationResponse.
+/// For outgoing connections, it initiates authentication by sending an
+/// AuthenticationRequest and awaits an AuthenticationResponse.  Upon
+/// successful authentication during initial connection, the negotiation
+/// process continues.  All authentication operations are performed
+/// asynchronously.
+
+// MQB
+#include <mqbblp_clustercatalog.h>
```
dorjesinpo: is this needed?

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,501 @@
...
+        bsl::allocate_shared<mqbnet::AuthenticationContext>(
+            d_allocator_p,
+            context.get(),      // initialConnectionContext
+            authenticationMsg,  // authenticationMessage
+            context
+                ->authenticationEncodingType(),  // authenticationEncodingType
+            mqbnet::AuthenticationState::e_AUTHENTICATING  // state
+        );
+
+    context->setAuthenticationContext(authenticationContext);
```
dorjesinpo, line 96: does `BSLS_ASSERT_SAFE(!context->authenticationContext());` make sense?  reauth does not call `onAuthenticationRequest` or does it?

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,501 @@
...
+    enum RcEnum {
+        // Value for the various RC error categories
+        rc_SUCCESS                             = 0,
+        rc_BUILD_RESPONSE_FAILED               = -1,
+        rc_AUTHENTICATION_FAILED               = -2,
+        rc_SEND_AUTHENTICATION_RESPONSE_FAILED = -3,
+    };
+
+    const bmqp_ctrlmsg::AuthenticationRequest& authenticationRequest =
+        context->authenticationMessage().authenticationRequest();
```
dorjesinpo: A comment somewhere would help about how the `d_state` protects `d_authenticationMessage`.
No thread (IO) can call `setAuthenticationMessage` if the state is not `AuthenticationState::e_AUTHENTICATED` and no thread calls  `authenticationMessage` if the state is not `AuthenticationState::e_AUTHENTICATING`.  Hence, no need for the mutex in either.
(please check this reasoning)

src/groups/mqb/mqbnet/mqbnet_authenticationcontext.cpp
```diff
@@ -0,0 +1,273 @@
...
+    const bsl::shared_ptr<bmqio::Channel>& channel)
+{
+    // executed by an *AUTHENTICATION* thread
+
+    // PRECONDITION
+    BSLS_ASSERT_SAFE(scheduler_p);
+
+    bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);  // LOCKED
+
+    if (d_state != AuthenticationState::e_AUTHENTICATING) {
```
dorjesinpo, line 170: A comment would help explaining in which scenario this is possible.  Is it the close scenario?

src/groups/mqb/mqbnet/mqbnet_authenticationcontext.cpp
```diff
@@ -0,0 +1,273 @@
...
+void AuthenticationContext::setAuthenticationEncodingType(
+    bmqp::EncodingType::Enum value)
+{
+    d_encodingType = value;
+}
+
+int AuthenticationContext::setAuthenticatedAndScheduleReauthn(
+    bsl::ostream&                          errorDescription,
+    bdlmt::EventScheduler*                 scheduler_p,
+    const bsl::optional<int>&              lifetimeMs,
```
dorjesinpo: type mismatch for the `lifetimeMs`: `bsl::optional<int>` vs `bsl::optional<bsls::Types::Int64>` (`AuthenticationResult::lifetimeMs()`)

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,501 @@
...
+                                                       authenticationRequest,
+                                                       channel,
+                                                       context);
+    if (buildResponseRc != rc_SUCCESS) {
+        rc    = (buildResponseRc * 10) + rc_BUILD_RESPONSE_FAILED;
+        error = buildResponseErrStream.str();
+        return;  // RETURN
+    }
+
+    // For reauthentication, skip sending the response
```
dorjesinpo: why no response for re-authentication?  Any request should get a response, we may build some client logic on it.

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,501 @@
...
+        rc_AUTHENTICATION_FAILED               = -2,
+        rc_SEND_AUTHENTICATION_RESPONSE_FAILED = -3,
+    };
+
+    const bmqp_ctrlmsg::AuthenticationRequest& authenticationRequest =
+        context->authenticationMessage().authenticationRequest();
+
+    bmqp_ctrlmsg::AuthenticationMessage   authenticationResponse;
+    bmqp_ctrlmsg::AuthenticationResponse& response =
+        authenticationResponse.makeAuthenticationResponse();
```
dorjesinpo: maybe, building response can move to `sendAuthenticationMessage` so there is unnecessary `makeAuthenticationResponse()` call when `isDefaultAuthn==true`

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,501 @@
...
+            bsl::ref(rc),
+            "reauthenticationError",
+            bsl::ref(error),
+            channel));
+    }
+    else {
+        // For initial authentication: setup state machine transition
+        input.emplace(InitialConnectionEvent::e_ERROR);
+        scopeGuard.emplace(bdlf::BindUtil::bind(
+            &mqbnet::InitialConnectionContext::handleEvent,
```
dorjesinpo, line 245: Better place for `handleEvent ` could be higher - in `Authenticator::handleAuthentication`.

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -42,44 +228,501 @@ InitialConnectionContext::~InitialConnectionContext()
...
+            }
+        }
+        else {
+            errStream << "Unexpected event received: " << oldState << " -> "
+                      << input;
+        }
+        break;
+    }
+    case InitialConnectionEvent::e_AUTHN_REQUEST: {
+        if (!bsl::holds_alternative<bmqp_ctrlmsg::AuthenticationMessage>(
```
dorjesinpo: Looks like this can be `BSLS_ASSERT_SAFE` instead.  A bit simpler.

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -42,44 +228,501 @@ InitialConnectionContext::~InitialConnectionContext()
...
+                                                         authenticationMsg);
+        }
+        else {
+            errStream << "Unexpected event received: " << oldState << " -> "
+                      << input;
+        }
+        break;
+    }
+    case InitialConnectionEvent::e_NEGOTIATION_MESSAGE: {
+        if (!bsl::holds_alternative<bmqp_ctrlmsg::NegotiationMessage>(
```
dorjesinpo: Looks like this can be `BSLS_ASSERT_SAFE` instead.  A bit simpler.

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
dorjesinpo: my mistake

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -42,44 +228,501 @@ InitialConnectionContext::~InitialConnectionContext()
...
+        bmqu::MemOutStream errStream;
+        const int          rc = scheduleRead(errStream);
+        if (rc != 0) {
+            handleEvent(rc, errStream.str(), InitialConnectionEvent::e_ERROR);
+        }
+    }
+}
+
+void InitialConnectionContext::handleEvent(
+    int                                                   statusCode,
```
dorjesinpo: the meaning of `statusCode` is not clear.  It seems that all errors are communicated as `InitialConnectionEvent::e_ERROR` and `TCPSessionFactory::initialConnectionComplete` just needs `statusCode != 0` (which is not enforced by `InitialConnectionEvent::e_ERROR` case).
Unless we want to communicate exact value to `bmqio::Channel::close` for some reason, the code can be simplified by removing `statusCode` argument.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
dorjesinpo: since there is `d_initialConnectionContextCache` now, maybe we do not need to bind `bsl::shared_ptr<InitialConnectionContext>` anymore?

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -342,46 +344,77 @@ void TCPSessionFactory::handleInitialConnection(
...
+        // call `authenticationOutbound` here instead before calling
+        // `negotiateOutbound`.
+        initialConnectionContext->handleEvent(
+            rc_SUCCESS,
+            bsl::string(),
+            mqbnet::InitialConnectionEvent::e_OUTBOUND_NEGOTATION);
+    }
+    else {
+        bmqu::MemOutStream errStream;
+        const int rc = initialConnectionContext->scheduleRead(errStream);
```
dorjesinpo: minor, this can be `handleEvent(mqbnet::InitialConnectionEvent::e_INCOMING)` and then `InitialConnectionContext::scheduleRead` can be private.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -730,9 +778,6 @@ void TCPSessionFactory::channelStateCallback(
                     bmqsys::Time::highResolutionTimer();
             }  // close mutex lock guard // UNLOCK
 
-            // Keep track of active channels, for logging purposes
-            ++d_nbActiveChannels;
```
dorjesinpo, line 734: moving `++d_nbActiveChannels` changes the semantics of it.  It used to track all channels, including incomplete ones.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -780,13 +825,17 @@ void TCPSessionFactory::onClose(
...
+        if (context) {
+            context->onClose();
+        }
 
         ChannelMap::const_iterator it = d_channels.find(channel.get());
         if (it != d_channels.end()) {
             channelInfo = it->second;
             d_channels.erase(it);
 
+            --d_nbActiveChannels;
```
dorjesinpo: is already decremented on line 814

dorjesinpo (changes requested):
interim review

src/groups/mqb/mqbmock/mqbmock_cluster.cpp
emelialei88: Indirectly included in `mqbmock_cluster.h` but maybe we shouldn't depend on indirect includes. Needed for `shared_ptr`.

src/groups/mqb/mqba/mqba_application.cpp
emelialei88: `setAdminCommandEnqueueCallback` is a function for `SessionNegotiator`, but not exposed for `mqbnet::Negotiator`.

src/groups/mqb/mqba/mqba_authenticator.cpp
emelialei88: Not sure what you mean. `onAuthenticationRequest` is not called for reauthn. And for initial connection, `authenticationContext` is not set until `onAuthenticationRequest`.

src/groups/mqb/mqba/mqba_authenticator.cpp
emelialei88: `Authenticator::handleAuthentication` is in IO thread, and the caller is `handleEvent` so error cases are already  handled by `handleEvent`. `Authenticator::authenticate` is in authentication thread.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
emelialei88: I think this might be the race condition we discussed earlier, though I don’t recall the exact scenario. All the tests are passing now — possibly because the fuzz tests have sped up and no longer reproduce the issue. Definitely a good reminder for me to document these cases better next time.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
emelialei88: Merge issue.

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
emelialei88: Most of the errors that occur during initial connection don’t print the error code or logs until `initialConnectionComplete` is called. If we were to follow this convention, it still feels necessary to pass the error code through `handleEvent` to `initialConnectionComplete` so it can be properly logged.

src/groups/mqb/mqba/mqba_application.cpp
dorjesinpo: This is very minor (since we do not throw exceptions) but what I meant is
```
    bslma::ManagedPtr<mqbnet::Negotiator> negotiatorMp(
        new (*d_allocator_p)
        SessionNegotiator(&d_bufferFactory,
                          d_dispatcher_mp.get(),
                          d_statController_mp->clientsStatContext(),
                          &d_blobSpPool,
                          d_scheduler_p,
                          d_allocators.get("SessionNegotiator"),
        d_allocator_p);
    negotiatorMp
        ->setAdminCommandEnqueueCallback(
            bdlf::BindUtil::bind(&Application::enqueueCommand,
                                 this,
                                 bdlf::PlaceHolders::_1,    // source
                                 bdlf::PlaceHolders::_2,    // cmd
                                 bdlf::PlaceHolders::_3,    // onProcessedCb
                                 bdlf::PlaceHolders::_4));  // fromReroute

```

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -527,17 +560,22 @@ void TCPSessionFactory::negotiationComplete(
         return;  // RETURN
     }
 
+    BSLS_ASSERT_SAFE(
+        d_initialConnectionContextCache.contains(initialConnectionContext_p));
+    bsl::shared_ptr<InitialConnectionContext> initialConnectionContext_sp =
```
dorjesinpo: the purpose of this seems to be to keep the object alive until this method returns since there is no other reference left.  Is that correct? If so, it is worth a comment.

src/groups/mqb/mqbnet/mqbnet_authenticationcontext.h
```diff
@@ -0,0 +1,247 @@
...
+    /// Authentication State.
+    AuthenticationState::Enum d_state;
+
+    /// The initial connection context associated with this authentication
+    /// context.  It is set during the initial authentication, and is null for
+    /// reauthentication.
+    InitialConnectionContext* d_initialConnectionContext_p;
+
+    /// The authentication message received.
+    bmqp_ctrlmsg::AuthenticationMessage d_authenticationMessage;
```
chrisbeard: What is the motivation for storing this for the lifetime of the connection? We should not store this material once authentication completes. Can that be avoided?

src/groups/mqb/mqbnet/mqbnet_authenticationcontext.h
```diff
@@ -0,0 +1,247 @@
...
+
+    /// The encoding type used for sending a message.  It should match with the
+    /// encoding type of the received message.
```
chrisbeard: nit: this comment feels like it can be made clearer, perhaps something like:

```
// The encoding type used for sending authentication responses. The value matches
// the encoding type set by the client during authentication and ensures that all
// responses are sent using the same encoding.
```

src/groups/mqb/mqbnet/mqbnet_authenticationcontext.h
```diff
@@ -0,0 +1,247 @@
...
+        bmqp::EncodingType::Enum                   authenticationEncodingType,
+        AuthenticationState::Enum                  state,
+        bslma::Allocator*                          allocator = 0);
+
+    // MANIPULATORS
+    void setAuthenticationResult(
+        const bsl::shared_ptr<mqbplug::AuthenticationResult>& value);
+
+    // NOTE: AuthenticationMessage and encodingType are set only when
+    // AuthenticationState is e_AUTHENTICATED.  authenticationMessage() and
```
chrisbeard: Is the state mentioned in this comment correct? e_AUTHENTICATED feels wrong here

src/groups/mqb/mqbnet/mqbnet_authenticator.h
```diff
@@ -0,0 +1,109 @@
...
+        const bsl::shared_ptr<InitialConnectionContext>& context,
+        const bmqp_ctrlmsg::AuthenticationMessage& authenticationMsg) = 0;
+
+    /// Produce and send outbound authentication message with the specified
+    /// `context`.  Return 0 on success, or a non-zero error code and populate
+    /// the specified `errorDescription` with a description of the error
+    /// otherwise.
+    virtual int authenticationOutbound(
+        bsl::ostream&                                 errorDescription,
+        const bsl::shared_ptr<AuthenticationContext>& context) = 0;
```
chrisbeard, line 89: This method feels a little out of place. Maybe that means design-wise we should rethink this, leave it out for now, or documentation on the method and interface should better define its purpose.

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,465 @@
...
+#include <bslmt_threadutil.h>
+#include <bsls_timeinterval.h>
+
+namespace BloombergLP {
+namespace mqba {
+
+namespace {
+
+const int k_MIN_THREADS = 1;  // Minimum number of threads in the thread pool
+const int k_MAX_THREADS = 3;  // Maximum number of threads in the thread pool
```
chrisbeard: This should probably be configuration-driven.

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,465 @@
...
+{
+    // executed by one of the *IO* threads
+
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(authenticationMsg.isAuthenticationRequestValue());
+    BSLS_ASSERT_SAFE(context->isIncoming());
+
+    BALL_LOG_DEBUG << "Received authentication message from '"
+                   << context->channel()->peerUri()
+                   << "': " << authenticationMsg;
```
chrisbeard: Authentication material is sensitive, it should not be logged.

src/groups/mqb/mqba/mqba_authenticator.h
```diff
@@ -0,0 +1,244 @@
...
+    /// specified `context` and `channel`.  The specified `isAnonAuthn` is set
+    /// to true when this is for anonymous authentication.  The specified
+    /// `isReauthn` is set to true when this is for re-authentication.  Return
+    /// 0 on success, or a non-zero error code and populate the specified
+    /// `errorDescription` with a description of the error otherwise.
+    int authenticateAsync(bsl::ostream&                  errorDescription,
+                          const AuthenticationContextSp& context,
+                          const bsl::shared_ptr<bmqio::Channel>& channel,
+                          bool                                   isAnonAuthn,
+                          bool isReauthn) BSLS_KEYWORD_OVERRIDE;
```
chrisbeard: Should this really be exposed in the interface? This looks more like an implementation detail that leaks to the TcpSessionFactory. Maybe an alternative is a `handleReauthentication` method that forwards appropriate to this (internal) async method.

src/groups/mqb/mqba/mqba_clientsession.cpp
```diff
@@ -2487,6 +2487,9 @@ void ClientSession::processEvent(const bmqp::Event& event,
 {
     // executed by the *IO* thread
 
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(!event.isAuthenticationEvent());
+
```
chrisbeard, line 2490: Should this logic also exist in `mqba::AdminSession`?

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -935,19 +983,83 @@ void TCPSessionFactory::stopHeartbeats()
...
+        BALL_LOG_ERROR << "#CORRUPTED_EVENT " << description
+                       << ": Received invalid authentication message "
+                          "from client [reason: 'failed to decode', rc: "
+                       << rc << "]:\n"
+                       << bmqu::BlobStartHexDumper(event.blob());
+        return;  // RETURN
+    }
+
+    BALL_LOG_INFO << description << ": Received authentication message: "
+                  << authenticationMessage;
```
chrisbeard: Similarly to elsewhere, authentication material should not be logged.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -935,19 +983,83 @@ void TCPSessionFactory::stopHeartbeats()
...
+    BSLS_ASSERT_SAFE(channelInfo->d_session_sp);
+
+    const bsl::shared_ptr<AuthenticationContext>& context =
+        channelInfo->d_authenticationCtx_sp;
+    const bsl::string& description = channelInfo->d_session_sp->description();
+
+    if (!context->tryStartReauthentication()) {
+        BALL_LOG_ERROR << "#CLIENT_IMPROPER_BEHAVIOR " << description
+                       << ": Dropping Authentication event since "
+                          "authentication is in progress: "
```
chrisbeard: nit: Does this mean authentication is _already_ in progress?

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -935,19 +983,83 @@ void TCPSessionFactory::stopHeartbeats()
...
+                       << bmqu::BlobStartHexDumper(event.blob());
+        return;  // RETURN
+    }
+
+    BALL_LOG_INFO << description << ": Received authentication message: "
+                  << authenticationMessage;
+
+    context->setAuthenticationMessage(authenticationMessage);
+    context->setAuthenticationEncodingType(
+        event.authenticationEventEncodingType());
```
chrisbeard, line 1030: Similar comment here, we should prefer to not store the authentication message for the lifetime of the connection.

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,465 @@
...
+        rc_SUCCESS                             = 0,
+        rc_AUTHENTICATION_FAILED               = -1,
+        rc_SCHEDULE_REAUTHN_FAILED             = -2,
+        rc_SEND_AUTHENTICATION_RESPONSE_FAILED = -3,
+    };
+
+    int         rc = rc_SUCCESS;
+    bsl::string error;
+
+    // Setup error handler based on whether this is initial authn or reauthn
```
chrisbeard:
```suggestion
    // Set up error handler based on whether this is initial authn or reauthn
```

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,465 @@
...
+    bsl::optional<mqbnet::InitialConnectionEvent::Enum> input;
+    bsl::optional<bdlb::ScopeExitAny>                   scopeGuard;
+
+    if (isReauthn) {
+        // For reauthentication: setup error guard to handle failures
```
chrisbeard:
```suggestion
        // For reauthentication: set up error guard to handle failures
```

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,465 @@
...
+        scopeGuard.emplace(bdlf::BindUtil::bind(
+            &mqbnet::AuthenticationContext::onReauthenticateErrorOrTimeout,
+            context.get(),
+            bsl::ref(rc),
+            "reauthenticationError",
+            bsl::ref(error),
+            channel));
+    }
+    else {
+        // For initial authentication: setup state machine transition
```
chrisbeard:
```suggestion
        // For initial authentication: set up state machine transition
```

src/groups/mqb/mqbnet/mqbnet_authenticationcontext.cpp
```diff
@@ -0,0 +1,289 @@
...
+        if (lifetime < 0) {
+            BALL_LOG_WARN
+                << "Authenticator returned negative remaining lifetime: "
+                << bsl::to_string(lifetime)
+                << ". Schedule reauthentication timer with lifetime set to 0.";
+            lifetime = 0;
+        }
+
+        bmqu::MemOutStream errorStream;
+        errorStream << "Authentication timeout after " << lifetime << " ms";
```
chrisbeard: This error could be made clearer. If I understand correctly, it's trying to convey that the client did not reauthenticate within before the lifetime expired?

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,465 @@
...
+        rc_AUTHENTICATION_FAILED               = -1,
+        rc_SCHEDULE_REAUTHN_FAILED             = -2,
+        rc_SEND_AUTHENTICATION_RESPONSE_FAILED = -3,
+    };
+
+    int         rc = rc_SUCCESS;
+    bsl::string error;
+
+    // Setup error handler based on whether this is initial authn or reauthn
+    bsl::optional<mqbnet::InitialConnectionEvent::Enum> input;
```
dorjesinpo: minor, this is `event`

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,465 @@
...
+                  << "' with mechanism '" << authenticationRequest.mechanism()
+                  << "'";
+
+    // Authenticate
+    bmqu::MemOutStream                             authnErrStream;
+    bsl::shared_ptr<mqbplug::AuthenticationResult> result;
+    const bsl::vector<char>&    data = authenticationRequest.data().isNull()
+                                           ? bsl::vector<char>()
+                                           : authenticationRequest.data().value();
+    mqbplug::AuthenticationData authenticationData(data, channel->peerUri());
```
dorjesinpo:
```
    mqbplug::AuthenticationData authenticationData(
        authenticationRequest.data().isNull()
            ? bsl::vector<char>() : authenticationRequest.data().value(), channel->peerUri());
```

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,465 @@
...
+        context->setAuthenticationResult(result);
+        const int scheduleRc = context->setAuthenticatedAndScheduleReauthn(
+            scheduleErrStream,
+            d_scheduler_p,
+            result->lifetimeMs(),
+            channel);
+        if (scheduleRc != 0) {
+            rc    = (scheduleRc * 10) + rc_SCHEDULE_REAUTHN_FAILED;
+            error = scheduleErrStream.str();
+            return;  // RETURN
```
dorjesinpo, line 320: it is better still to respond (with an error)

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -42,44 +231,521 @@ InitialConnectionContext::~InitialConnectionContext()
...
+{
+    // executed by an *AUTHENTICATION* or one of the *IO* threads
+
+    enum RcEnum {
+        // Value for the various RC error categories
+        rc_SUCCESS = 0,
+        rc_ERROR   = -1
+    };
+
+    bsl::shared_ptr<InitialConnectionContext> self = shared_from_this();
```
dorjesinpo: strictly speaking, we do not need this, since the `TCPSessionFactory::d_initialConnectionContextCache` and if everything ends up in `initialConnectionComplete`.  Could simplify the code a bit

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
dorjesinpo: Understood.  But `d_state` is primary and `statusCode` is secondary.  On the line 710, there is no check for `statusCode`, if it is `0`, there will be no `complete` call.

Maybe, we should rely on `InitialConnectionState::e_FAILED` and `InitialConnectionState::e_NEGOTIATED` being the only final states and call `complete` as soon as we reach them.  Assuming, any other state is transitory and awaits for more event(s).

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -42,44 +231,521 @@ InitialConnectionContext::~InitialConnectionContext()
...
+    if (rc == rc_SUCCESS && d_state == InitialConnectionState::e_NEGOTIATED) {
+        rc = d_negotiator_p->createSessionOnMsgType(errStream, &session, this);
+        BALL_LOG_INFO << "Created a session with " << channel()->peerUri();
+    }
+
+    if (rc != rc_SUCCESS) {
+        setState(InitialConnectionState::e_FAILED);
+    }
+
+    if (rc != rc_SUCCESS || d_state == InitialConnectionState::e_NEGOTIATED) {
```
dorjesinpo: `if (d_state == InitialConnectionState::e_FAILED || d_state == InitialConnectionState::e_NEGOTIATED) {`

or `if (isFinalState(d_state)) {`

src/groups/mqb/mqba/mqba_authenticator.cpp
dorjesinpo: I meant something like `BSLS_ASSERT_SAFE(!d_authenticationCtxSp);` in `InitialConnectionContext::setAuthenticationContext`

src/integration-tests/test_authn.py
```diff
@@ -0,0 +1,672 @@
...
+    }
+)
+def test_reauthenticate_success(single_node: Cluster) -> None:
+    """Test successful reauthentication with same credentials."""
+    client = RawClient()
+    client.open_channel(*single_node.admin_endpoint)
+
+    # Initial authentication
+    auth_resp = client.send_authentication_request("Basic", "user1:password1")
+    assert auth_resp["authenticationResponse"]["status"]["code"] == 0
```
dorjesinpo, line 184: maybe, check `lifetimeMs`?

src/integration-tests/test_authn.py
```diff
@@ -0,0 +1,672 @@
...
+    }
+)
+def test_reauthenticate_failure(single_node: Cluster) -> None:
+    """Test reauthentication failure with invalid credentials."""
+    client = RawClient()
+    client.open_channel(*single_node.admin_endpoint)
+
+    # Initial authentication succeeds
+    auth_resp = client.send_authentication_request("Basic", "user1:password1")
+    assert auth_resp["authenticationResponse"]["status"]["code"] == 0
```
dorjesinpo, line 219: `lifetimeMs`

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
dorjesinpo: I don't recall having any problems with `d_nbActiveChannels`, we do not seem to check it, only log

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
emelialei88: Are you saying we don't need this line? How do we pass `InitialConnectionContext` to other functions (e.g., `Negotiator::negotiateOutbound`) then?

src/groups/mqb/mqbnet/mqbnet_authenticationcontext.h
emelialei88: These two functions are called only during reauthentication.  The state should be `e_AUTHENTICATED` at that time.

src/groups/mqb/mqba/mqba_authenticator.cpp
emelialei88: I feel like failing to schedule reauthentication is more an internal error than something to notify the client.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
emelialei88: There's a comment for `d_initialConnectionContextCache` in `mqbnet::tcpsessionfactory.h`.

src/groups/mqb/mqba/mqba_application.cpp
emelialei88: Yeah I understand. But `mqbnet::Negotiator` doesn't expose an interface `setAdminCommandEnqueueCallback`.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
emelialei88: Yes

src/groups/mqb/mqba/mqba_application.cpp
dorjesinpo: That is not a problem (we could do `bslma::ManagedPtr<mqbnet::Negotiator> negotiatorMp(impl);` where impl is `bslma::ManagedPtr<mqbnet::Negotiator>`).  The problem is `setClusterCatalog` below.
This is ok since we do not throw exceptions, do early returns

src/groups/mqb/mqba/mqba_authenticator.cpp
dorjesinpo: we should always respond to request, client logic may depend on it.  In this case, it looks like an internal error and the response shoudl indicate an error.

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
dorjesinpo: we do not need `bsl::enable_shared_from_this<InitialConnectionContext>` 
`negotiateOutbound` can take raw pointer because of `d_initialConnectionContextCache`
(provided we always end up in `initialConnectionComplete` which is the goal of FSM)

dorjesinpo (changes requested):
Still think we always need to respond to auth request even in the case of internal error.
A minor point - since `d_initialConnectionContextCache`, there no need for `bsl::enable_shared_from_this<InitialConnectionContext>`

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
emelialei88: I'll get rid of the other `shared_from_this` but I don't think we can fully remove `enable_shared_from_this`. What if a scheduled read happens after `initialConnectionComplete` is called? We still need to bind a shared pointer of `InitialConnectionContext` when calling `channel()->read` in `InitialConnectionContext::scheduleRead`.

dorjesinpo: > We still need to bind a shared pointer of `InitialConnectionContext` when calling `channel()->read` in `InitialConnectionContext::scheduleRead`.

If `d_initialConnectionContextCache` always holds a `shared_ptr` when `InitialConnectionContext` reads, then `readCallbackTrampoline` does not need to worry about `shared_ptr`.  In fact, there is no need for `readCallbackTrampoline`.

That is all valid only if we make sure, there is no read before `e_OUTBOUND_NEGOTATION`/`e_INCOMING` and after `e_NEGOTIATED` / `e_FAILED` which is the goal of the FSM

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,493 @@
...
+                              this,
+                              context,
+                              channel,
+                              isDefaultAuthn,
+                              isReauthn));
+
+    if (rc != 0) {
+        errorDescription << "Failed to enqueue authentication job for '"
+                         << channel->peerUri() << "' [rc: " << rc
+                         << ", message: " << context->authenticationMessage()
```
chrisbeard: Not sure we want to include the authentication material here.

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,493 @@
...
+    mqbplug::AuthenticationData                    authenticationData(
+        authenticationRequest.data().isNull()
+            ? bsl::vector<char>()
+            : authenticationRequest.data().value(),
+        channel->peerUri());
+    const int authnRc = d_authnController_p->authenticate(
+        authnErrStream,
+        &result,
+        authenticationRequest.mechanism(),
+        authenticationData);
```
chrisbeard, line 273: Potentially open question: it doesn't look like we log the result of authentication here or in the authenticationcontroller. Is the expectation that the plugin implementation should log if you want logs?

chrisbeard: On a related note, control plane requests (e.g. openQueue, configureQueue, etc) have timers that track how long the broker took to handle the request. Do we have a plan for supporting something similar for authN?

emelialei88: Logging added to authenticationcontroller including timing authentication.

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -335,8 +335,15 @@ void Authenticator::authenticate(
         return;  // RETURN
     }
 
-    // Transition to the next state
-    event.value() = InitialConnectionEvent::e_AUTHN_SUCCESS;
+    if (isReauthn) {
+        // For reauthentication, we're done
```
dorjesinpo, line 360: was this a bug that closed the channel upon reauth?

If so, this function may be restructured to better handle `isReauthn` fork.  For example, call `handleEvent` in the case of reauth with the transition `e_NEGOTIATED` -> `e_AUTHN_SUCCESS` -> `e_NEGOTIATED` if successful, `e_NEGOTIATED` -> `e_ERROR` -> `e_FAILED` otherwise.

dorjesinpo (approved):
one comment about possible improvement

src/groups/mqb/mqbnet/mqbnet_authenticationcontext.cpp
```diff
@@ -0,0 +1,305 @@
...
+    if (d_state != AuthenticationState::e_AUTHENTICATING) {
+        errorDescription << "State not AUTHENTICATING (is " << d_state << ")";
+        return -1;
+    }
+
+    d_state = AuthenticationState::e_AUTHENTICATED;
+
+    /*
+    // TODO: Reauthentication is disabled for now.  Uncomment once we're ready
+    // to enable reauthentication.
```
chrisbeard, line 179: If there are blockers, let's write them down here.

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -18,21 +18,200 @@
...
-// -------------------------------------
+// ------------------------------
 
-InitialConnectionContext::InitialConnectionContext(bool isIncoming)
-: d_resultState_p(0)
-, d_userData_p(0)
-, d_channelSp()
-, d_initialConnectionCompleteCb()
+// CREATORS
+InitialConnectionContext::InitialConnectionContext(
```
chrisbeard, line 191: General comment: we probably want to review the logs for this component to make sure it's always clear what the relevant channel (or similar) is for any given log line.

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -42,44 +221,482 @@ InitialConnectionContext::~InitialConnectionContext()
...
+    if (rc != rc_SUCCESS) {
+        setState(InitialConnectionState::e_FAILED);
+    }
+
+    if (hasFinalState()) {
+        BALL_LOG_INFO << "Finished initial connection with rc = " << rc
+                      << ", error = '" << errStream.str() << "'";
+        guard.release()->unlock();
+        complete(rc, errStream.str(), session);
+    }
```
chrisbeard, line 696: This logic sequence needs improvement in terms of logging.

e_NEGOTIATED is considered a "final" state, so both logs will be printed. Do we really need both?

If there's a failure (from multiple places it seems), the transition to e_FAILED isn't logged here.

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.h
```diff
@@ -45,20 +73,141 @@ namespace mqbnet {
...
+struct InitialConnectionState {
+    // TYPES
+    enum Enum {
+        e_INITIAL             = 0,  // Initial state.
+        e_AUTHENTICATING      = 1,  // First message is authentication Request.
+        e_AUTHENTICATED       = 2,  // Authentication success.
+        e_ANON_AUTHENTICATING = 3,  // First message is Negotiation Request.
+        e_NEGOTIATING_OUTBOUND = 4,  // Outbound negotiation.
+        e_NEGOTIATED           = 5,  // Negotiation success.  Final state.
+        e_FAILED               = 6   // Final state.
```
chrisbeard, line 92: Should we better document the semantics of these states, valid transitions, etc?

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -42,44 +221,482 @@ InitialConnectionContext::~InitialConnectionContext()
...
+    default:
+        errStream << "InitialConnectionContext: "
+                  << "unexpected event received: " << event;
+        BALL_LOG_ERROR << "#UNEXPECTED_STATE " << errStream.str()
+                       << " [peer: " << channel()->peerUri() << "]";
+    }
+
+    BALL_LOG_INFO << "In initial connection state transition: " << oldState
+                  << " -> (" << event << ") -> " << d_state;
```
chrisbeard, line 678: TODO: This log is not well-placed. Notably, it's printed when state transitions don't happen (e.g. `e_INITIAL`), and when they do, the log appears well after the state transition.

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,533 @@
...
+
+    // Authenticate
+    bmqu::MemOutStream                             authnErrStream;
+    bsl::shared_ptr<mqbplug::AuthenticationResult> result;
+    mqbplug::AuthenticationData                    authenticationData(
+        authenticationRequest.data().isNull()
+            ? bsl::vector<char>()
+            : authenticationRequest.data().value(),
+        channel->peerUri());
+    const int authnRc = d_authnController_p->authenticate(
```
chrisbeard, line 269: In addition to checking the return code, we need to ensure that the plugin returns a proper `mqbplug::AuthenticationResult` via `result`.

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -460,20 +467,36 @@ int AuthenticationController::authenticate(
...
+        errorDescription << "Authentication mechanism '" << normMech
+                         << "' not supported.";
         return rc_MECHANISM_NOT_SUPPORTED;
     }
 
+    // Log the successful authentication
+    const bsls::Types::Int64 elapsed = bmqsys::Time::highResolutionTimer() -
+                                       start;
+
+    bmqu::MemOutStream logStream(d_allocator_p);
```
chrisbeard, line 486: TODO: I don't think the principal/lifetime should be logged here.

src/groups/mqb/mqbcfg/mqbcfg.xsd
```diff
@@ -322,11 +322,17 @@
...
+            Minimum number of threads in the authentication thread pool.
+        maxThreads..............:
+            Maximum number of threads in the authentication thread pool.
       </documentation>
     </annotation>
     <sequence>
       <element name='authenticators'      type='tns:AuthenticatorPluginConfig' minOccurs="0" maxOccurs='unbounded'/>
       <element name='anonymousCredential' type='tns:AnonymousCredential' minOccurs='0'/>
+      <element name='minThreads' type='int' default='1'/>
+      <element name='maxThreads' type='int' default='50'/>
```
chrisbeard, line 335: TODO: review defaults

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -488,34 +502,58 @@ void TCPSessionFactory::readCallback(const bmqio::Status& status,
...
+        // because if close has been triggered by the peer already, erasing
+        // InitialConnectionContext would lose the last reference of channel.
+        // channel->close() would then cause a segfault.
+        bsl::shared_ptr<InitialConnectionContext> initialConnectionContext_sp;
+        {
+            bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
+            initialConnectionContext_sp = d_initialConnectionContextCache.at(
+                initialConnectionContext_p);
+            d_initialConnectionContextCache.erase(initialConnectionContext_p);
+        }
```
chrisbeard: TODO: From reading this, it feels like we can find a better solution for managing lifetimes here

src/integration-tests/test_authn.py
```diff
@@ -0,0 +1,867 @@
...
+    domain_urls: tc.DomainUrls,  # pylint: disable=unused-argument
+) -> None:
+    """
+    Test that broker fails at startup when a custom ANONYMOUS authenticator
+    is configured without an anonymousCredential.
+
+    This validates that validateAnonymousCredential() ensures custom ANONYMOUS
+    authenticators (non-default) must have an explicit credential configured.
+    """
+    check_fail_to_start(single_node)
```
chrisbeard, line 781: TODO: I'm not sure I agree with this behavior.

Implementing a custom plugin is independent of the value of the default credential. It's reasonable for the operator to want to use the default anonymous credentials with a custom plugin.

src/integration-tests/test_authn.py
chrisbeard, line 1: TODO: another test to consider: client sends authentication request & negotiation request before receiving a response in a single write.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -488,34 +502,58 @@ void TCPSessionFactory::readCallback(const bmqio::Status& status,
...
 
+        // Remove from cache before closing channel, but need to keep the
+        // InitialConnectionContext alive until after channel->close(). This is
+        // because if close has been triggered by the peer already, erasing
+        // InitialConnectionContext would lose the last reference of channel.
+        // channel->close() would then cause a segfault.
+        bsl::shared_ptr<InitialConnectionContext> initialConnectionContext_sp;
+        {
+            bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
+            initialConnectionContext_sp = d_initialConnectionContextCache.at(
```
678098: `.at`?

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -342,46 +345,57 @@ void TCPSessionFactory::handleInitialConnection(
...
+                this,
+                bdlf::PlaceHolders::_1,  // status
+                bdlf::PlaceHolders::_2,  // errorDescription
+                bdlf::PlaceHolders::_3,  // session
+                bdlf::PlaceHolders::_4,  // channel
+                bdlf::PlaceHolders::_5,  // initialConnectionContext
+                context));
+
+    // Cache the context.  It will be removed in 'initialConnectionComplete'.
+    d_initialConnectionContextCache[initialConnectionContext.get()] =
```
678098: mutex?

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -527,17 +565,22 @@ void TCPSessionFactory::negotiationComplete(
         return;  // RETURN
     }
 
+    BSLS_ASSERT_SAFE(
+        d_initialConnectionContextCache.contains(initialConnectionContext_p));
```
678098: mutex?

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -779,8 +826,13 @@ void TCPSessionFactory::onClose(
...
 
         // set the 'isClosed' flag under lock to be checked under lock in
-        // 'negotiationComplete'.
-        initialConnectionContext->onClose();
+        // 'initialConnectionComplete'.
+        bsl::shared_ptr<InitialConnectionContext> initialConnectionContext_sp;
+        if (d_initialConnectionContextCache.contains(context_p)) {
+            initialConnectionContext_sp = d_initialConnectionContextCache.at(
+                context_p);
+            initialConnectionContext_sp->onClose();
```
678098, line 832: why don't we erase here?

src/groups/mqb/mqba/mqba_adminsession.cpp
```diff
@@ -317,6 +317,9 @@ void AdminSession::processEvent(const bmqp::Event& event,
 {
     // executed by the *IO* thread
 
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(!event.isAuthenticationEvent());
+
```
678098: Don't understand why do we crash if this is an Authn event but just log error and return if this is any other unexpected event.

Maybe we should just remove this check and rely on the existing logics:
```suggestion
```

src/groups/mqb/mqba/mqba_application.h
```diff
@@ -27,6 +27,7 @@
 
 // MQB
 #include <mqba_commandrouter.h>
+#include <mqbauthn_authenticationcontroller.h>
```
678098, line 30: Can be removed from header and replaced with forward declaration

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,533 @@
...
+int Authenticator::onAuthenticationRequest(
+    bsl::ostream&                              errorDescription,
+    const bmqp_ctrlmsg::AuthenticationMessage& authenticationMsg,
+    mqbnet::InitialConnectionContext*          context)
+{
+    // executed by one of the *IO* threads
+
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(authenticationMsg.isAuthenticationRequestValue());
+    BSLS_ASSERT_SAFE(context->isIncoming());
```
678098, line 78:
```suggestion
    BSLS_ASSERT_SAFE(context_p);
    BSLS_ASSERT_SAFE(context_p->isIncoming());
```

Make sure the ptr is not null before accessing

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,533 @@
...
+namespace mqba {
+
+// -------------------
+// class Authenticator
+// -------------------
+
+int Authenticator::onAuthenticationRequest(
+    bsl::ostream&                              errorDescription,
+    const bmqp_ctrlmsg::AuthenticationMessage& authenticationMsg,
+    mqbnet::InitialConnectionContext*          context)
```
678098, line 72:
```suggestion
    mqbnet::InitialConnectionContext*          context_p)
```

For better readability, especially if we have a mixed usage of `_p` and `_sp` of `InitialConnectionContext`

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,533 @@
...
+{
+    // executed by one of the *IO* threads
+
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(authenticationMsg.isAuthenticationRequestValue());
+    BSLS_ASSERT_SAFE(context->isIncoming());
+
+    BALL_LOG_DEBUG << "Received authentication message from '"
+                   << context->channel()->peerUri();
```
678098, line 81:
```suggestion
                   << context->channel()->peerUri() << "'";
```

Quote is not closed

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,533 @@
...
+    bsl::shared_ptr<mqbnet::AuthenticationContext> authenticationContext =
+        bsl::allocate_shared<mqbnet::AuthenticationContext>(
+            d_allocator_p,
+            context,  // initialConnectionContext
+            authenticationMsg.authenticationRequest()
+                .mechanism(),   // mechanism
+            authenticationMsg,  // authenticationMessage
+            context
+                ->authenticationEncodingType(),  // authenticationEncodingType
+            mqbnet::AuthenticationState::e_AUTHENTICATING  // state
```
678098, line 93: Can we construct Authn Context in any state other than the base `AUTHENTICATING`? What would it mean in such a case?
If no, we might remove this from constructor args

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,533 @@
...
+
+    // Authenticate
+    int rc = authenticateAsync(
+        errorDescription,
+        authenticationContext,
+        context->channel(),
+        context->state() == InitialConnectionState::e_ANON_AUTHENTICATING,
+        false);
+
+    return rc;
```
678098, line 106:
```suggestion
    return authenticateAsync(
        errorDescription,
        authenticationContext,
        context->channel(),
        context->state() == InitialConnectionState::e_ANON_AUTHENTICATING,
        false);
```
An option

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,533 @@
...
+}
+
+int Authenticator::onAuthenticationResponse(
+    BSLA_UNUSED bsl::ostream& errorDescription,
+    BSLA_UNUSED const bmqp_ctrlmsg::AuthenticationMessage& authenticationMsg,
+    BSLA_UNUSED mqbnet::InitialConnectionContext* context)
```
678098, line 112: `BSLA_MAYBE_UNUSED` is the one we should use

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,533 @@
...
+        rc_WRITE_FAILURE = -2
+    };
+
+    // Build authentication response message
+    bmqp_ctrlmsg::AuthenticationMessage   message;
+    bmqp_ctrlmsg::AuthenticationResponse& response =
+        message.makeAuthenticationResponse();
+
+    response.status().code()    = authnRc;
+    response.status().message() = errorMsg;
```
678098, line 143: String assignment, and the base object constructed without allocator

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,533 @@
...
+}
+
+int Authenticator::onAuthenticationResponse(
+    BSLA_UNUSED bsl::ostream& errorDescription,
+    BSLA_UNUSED const bmqp_ctrlmsg::AuthenticationMessage& authenticationMsg,
+    BSLA_UNUSED mqbnet::InitialConnectionContext* context)
+{
+    // executed by one of the *IO* threads
+
+    BALL_LOG_ERROR << "Not Implemented";
```
678098, line 116: I think this one has to be implemented in the future, so we need to add a comment that it is planned

src/integration-tests/test_authn.py
```diff
@@ -0,0 +1,867 @@
...
+    num_threads = 8
+    results = [None] * num_threads
+    threads = []
+
+    def auth_worker(idx):
+        client = RawClient()
+        client.open_channel(*single_node.admin_endpoint)
+        auth_resp = client.authenticate(
+            "Basic", f"user{idx}:password{idx}"
+        )
```
chrisbeard:
```suggestion
        auth_resp = client.authenticate("Basic", f"user{idx}:password{idx}")
```

CMakeLists.txt
```diff
@@ -125,6 +125,9 @@ if (NOT DEFINED INSTALL_TARGETS)
   set(BMQ_TARGET_PROMETHEUS_NEEDED     NO)
   set(BMQ_TARGET_IT_NEEDED             YES)
   set(BMQ_TARGET_FUZZTESTS_NEEDED      NO)
```
pniedzielski, line 127: I think all the changes in this file are no longer necessary, since authnpass/authnfail are gone and authnbasic is builtin.

bin/build-darwin.sh
```diff
@@ -165,7 +165,7 @@ CMAKE_OPTIONS=(\
 
 PKG_CONFIG_PATH="${DIR_INSTALL}/lib/pkgconfig:${BREW_PKG_CONFIG_PATH}" \
 cmake -B "${DIR_BUILD}/blazingmq" -S "${DIR_ROOT}" "${CMAKE_OPTIONS[@]}"
-make -C "${DIR_BUILD}/blazingmq" -j 16
+cmake --build "${DIR_BUILD}/blazingmq" --parallel 16 --target bmqbrkr bmqtool all.it
```
pniedzielski: Not sure why this change is here.

src/groups/mqb/mqba/mqba_adminsession.cpp
```diff
@@ -317,6 +317,9 @@ void AdminSession::processEvent(const bmqp::Event& event,
 {
     // executed by the *IO* thread
 
+    // PRECONDITIONS
```
pniedzielski: This isn't a precondition of the function, and I think we should do the same strategy as below for error handling: log an error and return, rather than crash the broker if an admin client sends us an authn message after negotiating a session.

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -0,0 +1,533 @@
...
+    else {
+        response.status().category() = bmqp_ctrlmsg::StatusCategory::E_SUCCESS;
+        response.lifetimeMs()        = lifetimeMs;
+    }
+
+    // Send authentication response message
+    bdlma::LocalSequentialAllocator<2048> localAllocator(d_allocator_p);
+    bmqp::SchemaEventBuilder              builder(d_blobSpPool_p,
+                                     authenticationEncodingType,
+                                     &localAllocator);
```
678098, line 157: Will be safer for the future changes to use a long-living allocator `d_allocator_p`, because we have a `shared_ptr` to `Blob` and might want to use it. We don't want to construct a `shared_ptr` using a local allocator

src/groups/mqb/mqba/mqba_adminsession.cpp
pniedzielski: AuthenticationEvents are not control events, so we handle this automatically below.

bin/build-darwin.sh
pniedzielski: Removing change, it can go in another PR if we want.  We should migrate this script to CMake presets anyway.

src/groups/mqb/mqba/mqba_authenticator.h
```diff
@@ -0,0 +1,253 @@
...
+
+    /// True if this component is started.
+    bool d_isStarted;
+
+  private:
+    // NOT IMPLEMENTED
+
+    /// Copy constructor and assignment operator not implemented.
+    Authenticator(const Authenticator&);             // = delete
+    Authenticator& operator=(const Authenticator&);  // = delete
```
678098, line 119: KEYWORD DELETED

src/groups/mqb/mqba/mqba_sessionnegotiator.h
```diff
@@ -29,11 +29,12 @@
 ///
 /// Thread Safety                              {#mqba_sessionnegotiator_thread}
 /// =============
-/// This component is owned by `InitialConnectionHandler`, and its functions
+/// This component is held by `InitialConnectionContext`, and its functions
 /// are called only from there.  It is not thread safe.
 
 // MQB
 #include <mqbconfm_messages.h>
+#include <mqbnet_initialconnectioncontext.h>
```
678098, line 37: Can forward declare and remove this from header

src/groups/mqb/mqbcfg/mqbcfg.xsd
678098, line 335: 50 threads is too much?

678098 (approved):

src/groups/mqb/mqbcfg/mqbcfg.xsd
chrisbeard: I think I agree, at least by default it should be not that big.

678098: We can also use just 1 setting and always spawn `minThreads == maxThreads`
This is what we do for dispatchers

678098: I don't think we will ever configure `minThreads` or do any fine-tuning for this

PR #1005, opened 2025-10-24
Fix[MQB]: Change Negotiator to a class

emelialei88:


src/groups/mqb/mqba/mqba_initialconnectionhandler.cpp
```diff
@@ -102,21 +100,7 @@ void InitialConnectionHandler::readCallback(
...
     }
 
-    rc = processBlob(errStream, &session, &isContinueRead, outPacket, context);
-    if (rc != rc_SUCCESS) {
-        rc    = (rc * 10) + rc_PROCESS_BLOB_ERROR;
-        error = bsl::string(errStream.str().data(), errStream.str().length());
-        return;  // RETURN
-    }
-
-    if (isContinueRead) {
```
emelialei88, line 112: Got rid of this part since `isContinueRead` flag is not used.

src/groups/mqb/mqba/mqba_sessionnegotiator.h
```diff
@@ -29,10 +29,12 @@
 ///
 /// Thread Safety                              {#mqba_sessionnegotiator_thread}
 /// =============
-/// TODO
+/// This component is owned by `InitialConnectionHandler`, and its functions
+/// are called only from there.  It is not thread safe.
 
 // MQB
 #include <mqbconfm_messages.h>
+#include <mqbnet_initialconnectioncontext.h>
```
678098: As far as I can see, `mqbnet::InitialConnectionContext` is used accessed only by pointer in this header. This allows to move the include to `.cpp` file and keep only the forward declaration in this header

678098 (approved):

Issue #923, opened 2025-09-08
Refactor `mqbstat_clusterstats.cpp`: create partition views to avoid passing and checking `partitionId` every time


678098:
There are multiple setters in `ClusterStats` with the same pattern:

```
void ClusterStats::onPartitionEvent(PartitionEventType::Enum type,
                                    int                      partitionId,
                                    bsls::Types::Int64       value)
{
    // PRECONDITIONS
    BSLS_ASSERT_SAFE(partitionId >= 0 &&
                     partitionId <
                         static_cast<int>(d_partitionsStatContexts.size()));
    BSLS_ASSERT_SAFE(d_partitionsStatContexts[partitionId] &&
                     "initialize was not called");
```

1. Passed `partitionId` arg
2. Checks for `partitionId` range
3. Check that subcontext was initialized

I propose to add `PartitionStats` view to handle partition stats updates and access `d_partitionsStatContexts[partitionId]` without extra checks. It might look like this:

```
// Perform all checks during this `get` once and cache `pstats_sp` for later use:
PartitionStats pstats_sp = d_clusterStats_sp->getPartitionStats(partitionId);
...
// Update partition stats without extra checks
pstats_sp->onPartitionEvent(e_ROLLOVER, value);
```

PR #1007, opened 2025-10-31
Refactor[mqbstat]: add PartitionStats view

waldgange:
https://github.com/bloomberg/blazingmq/issues/923

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_filestore.cpp
+++ b/src/groups/mqb/mqbs/mqbs_filestore.cpp
@@ -4001,10 +4001,10 @@ int FileStore::issueSyncPointInternal(SyncPointType::Enum type,
 
     // Report cluster's partition stats
     d_partitionStats_sp->setPartitionBytes(fs->d_outstandingBytesData,
-                                          fs->d_outstandingBytesJournal,
-                                          fs->d_dataFilePosition,
-                                          fs->d_journalFilePosition,
-                                          d_sequenceNum);
+                                           fs->d_outstandingBytesJournal,
+                                           fs->d_dataFilePosition,
+                                           fs->d_journalFilePosition,
+                                           d_sequenceNum);
 
     return rc_SUCCESS;
 }
@@ -5243,10 +5243,10 @@ int FileStore::open(const QueueKeyInfoMap& queueKeyInfoMap)
     // Report cluster's partition stats
     const FileSet* fs = d_fileSets[0].get();
     d_partitionStats_sp->setPartitionBytes(fs->d_outstandingBytesData,
-                                          fs->d_outstandingBytesJournal,
-                                          fs->d_dataFilePosition,
-                                          fs->d_journalFilePosition,
-                                          d_sequenceNum);
+                                           fs->d_outstandingBytesJournal,
+                                           fs->d_dataFilePosition,
+                                           fs->d_journalFilePosition,
+                                           d_sequenceNum);
 
     return rc_SUCCESS;
 }


src/plugins/bmqprometheus/bmqprometheus_prometheusstatconsumer.cpp
```diff
@@ -704,15 +704,15 @@ void PrometheusStatConsumer::captureClusterPartitionsStats()
                  clusterIt->subcontextIterator();
              partitionIt;
              ++partitionIt) {
-            mqbstat::ClusterStats::PrimaryStatus::Enum primaryStatus =
-                static_cast<mqbstat::ClusterStats::PrimaryStatus::Enum>(
+            mqbstat::PartitionStats::PrimaryStatus::Enum primaryStatus =
```
waldgange, line 707: The same changes are required in GUTS plugin. So we have to eather release them simultaneously or leave this Enum in `ClusterStats` for backward compatibility.

src/groups/mqb/mqbs/mqbs_filestore.h
```diff
@@ -285,9 +285,9 @@ class FileStore BSLS_KEYWORD_FINAL : public DataStore {
     // Dispatcher client data associated
     // with this instance.
 
-    mqbstat::ClusterStats* d_clusterStats_p;
+    bsl::shared_ptr<mqbstat::PartitionStats> d_partitionStats_p;
```
678098:
```suggestion
    bsl::shared_ptr<mqbstat::PartitionStats> d_partitionStats_sp;
```

I was sure that it is a raw pointer when I was reading the previous code, `_sp` in the name gives more clarity

src/groups/mqb/mqbstat/mqbstat_clusterstats.cpp
```diff
@@ -283,9 +285,22 @@ bsls::Types::Int64 ClusterStats::getValue(const bmqst::StatContext& context,
 #undef STAT_SINGLE
 }
 
+bsl::shared_ptr<PartitionStats>
+ClusterStats::getPartitionStats(int partitionId) const
+{
+    // PRECONDITIONS
+    BSLS_ASSERT_OPT(partitionId >= 0 &&
```
678098:
```suggestion
    BSLS_ASSERT_OPT(0 <= partitionId &&
```
Nice to have but not necessary, keeping comparisons left-to-right helps with code consistency

src/groups/mqb/mqbstat/mqbstat_clusterstats.h
```diff
@@ -295,16 +264,68 @@ class ClusterStats {
...
+    /// Set the time in nanoseconds it took for the rollover operation to the
+    /// specified `value`.
+    PartitionStats& setRoloverTime(bsls::Types::Int64 value);
+
+    /// Set the time in nanoseconds it took for the replication of a new entry
+    /// in journal file to the specified `value`.
+    PartitionStats& setReplicationTime(bsls::Types::Int64 value);
+
+    /// Set the primary status of the partition to the specified `value`.
+    PartitionStats& setNodeRole(PrimaryStatus::Enum value);
```
678098: I think we never have a situation when we can report any 2 of these metrics at the same time. This means that builder-like setters are not useful - we never really use setter chains.

The second thing is introduced performance cost. I am not sure if unused return value will be optimized here or not.

Due to this, it is better to keep the class API simpler


```suggestion
    /// Set the time in nanoseconds it took for the rollover operation to the
    /// specified `value`.
    void setRoloverTime(bsls::Types::Int64 value);

    /// Set the time in nanoseconds it took for the replication of a new entry
    /// in journal file to the specified `value`.
    void setReplicationTime(bsls::Types::Int64 value);

    /// Set the primary status of the partition to the specified `value`.
    void setNodeRole(PrimaryStatus::Enum value);
```

src/groups/mqb/mqbstat/mqbstat_clusterstats.h
```diff
@@ -295,16 +264,68 @@ class ClusterStats {
...
+
+    /// Set the partition outstanding bytes of the partition data and journal
+    /// files to the corresponding specified `outstandingDataBytes`,
+    /// `outstandingJournalBytes`, `offsetDataBytes`, `offsetJournalBytes` and
+    /// `sequenceNumber` values.
+    PartitionStats&
```
678098:
```suggestion
    void
```

src/groups/mqb/mqbstat/mqbstat_clusterstats.cpp
```diff
@@ -442,61 +434,71 @@ ClusterStats::setPartitionCfgBytes(bsls::Types::Int64 dataBytes,
...
+}
+
+PartitionStats& PartitionStats::setRoloverTime(bsls::Types::Int64 value)
+{
+    d_statContext_sp->reportValue(ClusterStatsIndex::e_PARTITION_ROLLOVER_TIME,
+                                  value);
+    return *this;
+}
+
+PartitionStats& PartitionStats::setReplicationTime(bsls::Types::Int64 value)
```
678098: For better performance, you might move these setters to the header and make them inline. At least the ones that are called often (`setReplicationTime`?)

src/groups/mqb/mqbstat/mqbstat_clusterstats.h
waldgange: Agree here. What about `ClusterStats` setters? They also return self ref that we never use. Maybe it's worth simplifying them as well?

678098: > Agree here. What about `ClusterStats` setters? They also return self ref that we never use. Maybe it's worth simplifying them as well?

Yes, it's a good idea

678098 (approved):
Looks good

PR #1010, opened 2025-11-05
Fix mqbc: Construct merged state with uncommitted CSLs before send

kaikulimu:
Fix for Internal Ticket 181406010

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_clusterutil.cpp
+++ b/src/groups/mqb/mqbc/mqbc_clusterutil.cpp
@@ -1464,7 +1464,6 @@ void ClusterUtil::sendClusterState(
                 clusterData->clusterConfig().partitionConfig().numPartitions(),
                 true,  // isTemporary
                 allocator);
-            apply(&tempState, clusterMessage, *clusterData);
             const int rc = load(&tempState,
                                 ledger->getIterator().get(),
                                 *clusterData,


src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp
```diff
@@ -1761,6 +1770,17 @@ void ClusterOrchestrator::processLeaderPassiveNotification(
     BSLS_ASSERT_SAFE(notification.choice().isLeaderPassiveValue());
     BSLS_ASSERT_SAFE(notifier);
 
+    if (d_clusterConfig.clusterAttributes().isFSMWorkflow()) {
```
dorjesinpo, line 1773: Does this mean we should not send `LeaderPassive` notification in FSM mode?
If so, can stop sending it rather than stop processing it.

src/groups/mqb/mqbc/mqbc_clusterstate.h
```diff
@@ -567,17 +570,24 @@ class ClusterState {
     /// TODO (FSM); remove after switching to FSM
     Assignments d_doubleAssignments;
 
+    // PRIVATE ACCESSORS
+
+    /// Return the cluster name with temporary suffix if applicable.
+    const bsl::string clusterNameWithTempSuffix() const;
```
dorjesinpo: the name does not make it obvious that there is no temp suffix unless `d_isTemporary`.  Maybe, just name it `name`?

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -1438,8 +1453,96 @@ void ClusterUtil::sendClusterState(
...
+            // information from uncommitted CSL advisories, before sending over
+            // to follower(s).
+            //
+            // NOTE: This code path is *only* reachable in non-FSM mode.
+            ClusterState tempState(
+                &clusterData->cluster(),
+                clusterData->clusterConfig().partitionConfig().numPartitions(),
+                true,  // isTemporary
+                allocator);
+            apply(&tempState, clusterMessage, *clusterData);
```
dorjesinpo: is `clusterMessage` empty at this point?  If so, what does `apply` do?

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -1438,8 +1453,96 @@ void ClusterUtil::sendClusterState(
...
+                return;  // RETURN
+            }
+
+            ClusterStateLedger::ClusterMessageCRefList uncommittedAdvisories;
+            ledger->uncommittedAdvisories(&uncommittedAdvisories);
+            for (ClusterStateLedger::ClusterMessageCRefList::const_iterator
+                     cit = uncommittedAdvisories.begin();
+                 cit != uncommittedAdvisories.end();
+                 ++cit) {
+                apply(&tempState, *cit, *clusterData);
```
dorjesinpo, line 1486: Is the order important?  First, uncommitted, then committed or vice versa?

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -1438,8 +1453,96 @@ void ClusterUtil::sendClusterState(
...
+                            << clusterData->identity().description()
+                            << " Partition [" << pid
+                            << "]: Self leader sees that CSL has recorded a "
+                               "leaseId of "
+                            << leaseIdInCSL
+                            << ", while self was about to assign a leaseId of "
+                            << pit->primaryLeaseId() << ".  Bumping up to "
+                            << leaseIdInCSL + 1
+                            << " to avoid potential conflict.";
+                        pit->primaryLeaseId() = leaseIdInCSL + 1;
```
dorjesinpo, line 1518: if a new leader gets elected, would it get `leaseIdInCSL + 1`?  Why not using `leaseIdInCSL` at this point?

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -1438,8 +1453,96 @@ void ClusterUtil::sendClusterState(
...
+                    }
+                }
+            }
+            else {
+                loadPartitionsInfo(&advisory.partitions(), tempState);
```
dorjesinpo, line 1523: only when `newlyAssigned.empty()`?  Can there be assigned partitions before `newlyAssigned`? 
Or, is this all or nothing?

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -1438,8 +1453,96 @@ void ClusterUtil::sendClusterState(
         clusterData->electorInfo().nextLeaderMessageSequence(
             &advisory.sequenceNumber());
 
-        advisory.partitions() = partitions;
-        loadQueuesInfo(&advisory.queues(), clusterState);
+        if (trustCSL) {
```
dorjesinpo, line 1456: `trustCSL` seems to be `false` if there were no commits yet.  But, is it still possible to have uncommitted advisories to merge?

src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp
kaikulimu: K, in FSM mode, we will stop sending it. And if we receive it, the logging will be on error level saying "deprecated notification" instead.

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
kaikulimu: Applying the cluster message is no-op. Good catch. Will remove.

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
kaikulimu: The natural order is committed then uncommitted, which is what I do in this PR. If you do vice versa, it is error-prone. For example, we have a committed assignment of `q1` then an uncommitted unassignment of `q1` -- the correct order must be to process committed before uncommitted.

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
kaikulimu: In normal flow, new leader would assign primaries with `leaseIdInCSL + 1`. However, in case new leader has missed any uncommitted CSL advisory which has newer partition assignments, we put in a safety measure that ensures the new leaseId must be at least one greater than the `leaseIdInCSL`.

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
kaikulimu: In legacy mode, assigning partitions are _always_ followed by a `sendClusterState` call -- the two places are at [`mqbblp::ClusterStateManager::onSelfActiveLeader()`](https://github.com/bloomberg/blazingmq/blob/7a8722a62b96d86de27b298e601013170d20c209/src/groups/mqb/mqbblp/mqbblp_clusterstatemanager.cpp#L120) and [`ClusterOrchestrator::onNodeUnavailable()`](https://github.com/bloomberg/blazingmq/blob/7a8722a62b96d86de27b298e601013170d20c209/src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp#L523). In both cases, `newlyAssigned` will be non-empty. Hence, there **cannot** be newly assigned partitions before `newlyAssigned`.

src/groups/mqb/mqbc/mqbc_clusterutil.cpp
kaikulimu: In legacy mode, `trustCSL` is `false` at startup because CSL is not the source-of-truth at startup. Therefore, we don't care if there are any uncommitted advisories. We overwrite the CSL with cluster state constructed from journal/data/qlist files regardless.

In short, if we don't trust CSL, we don't trust its uncommitted advisories neither.

PR #1014, opened 2025-11-07
Perf[BMQP]: minimizing copies in MessageProperties

dorjesinpo:
1) Introducing `bmqp::MessageProperties ::d_doDeepCopy` (`true` by default).  When `false` , there is no `blob` copy.  Meaning, `MessageProperties` valid only when the original `blob` is alive. This helps `Routers::MessagePropertiesReader` when iterating subscriptions.

2) avoiding extra `bsl::string` copy when reading strings from blob.  Use `bsl::string_view` instead.  Note that `setProperty` still needs to copy strings.  And `getProperty` needs to convert `bsl::string_view` into `bsl::string`.  Again, `Routers::MessagePropertiesReader` does benefit from the use of `bsl::string_view`.

### DIFF
--- a/src/groups/bmq/bmqa/bmqa_messageproperties.h
+++ b/src/groups/bmq/bmqa/bmqa_messageproperties.h
@@ -101,7 +101,7 @@ class MessageProperties {
     /// Constant representing the maximum size of a
     /// `bmqp::MessageProperties` object, so that the below AlignedBuffer
     /// is big enough.
-    static const int k_MAX_SIZEOF_BMQP_MESSAGEPROPERTIES = 280;
+    static const int k_MAX_SIZEOF_BMQP_MESSAGEPROPERTIES = 184;
 
     // PRIVATE TYPES
     typedef bsls::AlignedBuffer<k_MAX_SIZEOF_BMQP_MESSAGEPROPERTIES>
--- a/src/groups/bmq/bmqp/bmqp_messageproperties.cpp
+++ b/src/groups/bmq/bmqp/bmqp_messageproperties.cpp
@@ -318,13 +318,12 @@ bool MessageProperties::streamInPropertyValue(const Property& p) const
     case bmqt::PropertyType::e_STRING: {
         // Try to avoid copying the string.  'd_blop' already keeps a copy.
         bmqu::BlobPosition end;
-        int                ret =
+        const int          ret =
             bmqu::BlobUtil::findOffset(&end, *d_blob_p, position, p.d_length);
         bool doCopy = true;
         if (ret == 0) {
             // Do not align
-            if (position.buffer() == end.buffer() ||
-                (position.buffer() + 1 == end.buffer() && end.byte() == 0)) {
+            if (bmqu::BlobUtil::isDataContinuous(position, end)) {
                 // Section is good
                 char* start = d_blob_p->buffer(position.buffer()).data() +
                               position.byte();
@@ -372,14 +371,14 @@ MessageProperties::MessageProperties(bslma::Allocator* basicAllocator)
 , d_originalSize(0)
 , d_blob()
 , d_blob_p(d_blob.address())
-, d_isBlobConstructed(false)
-, d_isDirty(true)  // by default, this should be true
 , d_mphSize(0)
 , d_mphOffset(0)
 , d_numProps(0)
 , d_dataOffset(0)
 , d_schema()
 , d_originalNumProps(0)
+, d_isBlobConstructed(false)
+, d_isDirty(true)  // by default, this should be true
 , d_doDeepCopy(true)
 {
 }
@@ -392,14 +391,14 @@ MessageProperties::MessageProperties(const MessageProperties& other,
 , d_originalSize(other.d_originalSize)
 , d_blob()
 , d_blob_p(d_blob.address())
-, d_isBlobConstructed(false)
-, d_isDirty(other.d_isDirty)
 , d_mphSize(other.d_mphSize)
 , d_mphOffset(other.d_mphOffset)
 , d_numProps(other.d_numProps)
 , d_dataOffset(other.d_dataOffset)
 , d_schema(other.d_schema)
 , d_originalNumProps(other.d_originalNumProps)
+, d_isBlobConstructed(false)
+, d_isDirty(other.d_isDirty)
 , d_doDeepCopy(other.d_doDeepCopy)
 {
     if (other.d_isBlobConstructed) {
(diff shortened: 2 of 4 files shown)


src/groups/bmq/bmqa/bmqa_messageproperties.h
```diff
@@ -101,7 +101,7 @@ class MessageProperties {
     /// Constant representing the maximum size of a
     /// `bmqp::MessageProperties` object, so that the below AlignedBuffer
     /// is big enough.
-    static const int k_MAX_SIZEOF_BMQP_MESSAGEPROPERTIES = 184;
+    static const int k_MAX_SIZEOF_BMQP_MESSAGEPROPERTIES = 280;
```
678098: Why 280? The current sizeof is 192.

678098: Secondly, it is not necessary to grow `sizeof(bmqp::MessageProperties)` at all
This class design has a flaw - random placement of fields.
If you move all bool fields to the end of the fields list it will be enough to get rid of extra alignment padding:

```
    mutable bool d_isBlobConstructed;
    // Flag indicating if an instance of
    // the blob has been constructed in
    // 'd_blob'.

    mutable bool d_isDirty;
    // Flag indicating if this instance has
    // been updated since the previous
    // invocation of 'streamOut()'.

    bool d_doDeepCopy;
```

And the sizeof will still be 184 like before

src/groups/bmq/bmqp/bmqp_messageproperties.cpp
```diff
@@ -300,19 +316,38 @@ bool MessageProperties::streamInPropertyValue(const Property& p) const
...
 
     case bmqt::PropertyType::e_STRING: {
-        bsl::string value(p.d_length, ' ');
-        rc = bmqu::BlobUtil::readNBytes(&value[0],
-                                        d_blob.object(),
-                                        position,
-                                        p.d_length);
+        // Try to avoid copying the string.  'd_blop' already keeps a copy.
+        bmqu::BlobPosition end;
+        int                ret =
```
678098:
```suggestion
        const int          ret =
```

src/groups/bmq/bmqp/bmqp_messageproperties.cpp
```diff
@@ -300,19 +316,38 @@ bool MessageProperties::streamInPropertyValue(const Property& p) const
...
+        if (ret == 0) {
+            // Do not align
+            if (position.buffer() == end.buffer() ||
+                (position.buffer() + 1 == end.buffer() && end.byte() == 0)) {
+                // Section is good
+                char* start = d_blob_p->buffer(position.buffer()).data() +
+                              position.byte();
+
+                p.d_value = bsl::string_view(start, p.d_length);
+                doCopy    = false;
```
678098, line 332: `doCopy` is unused, can be removed

dorjesinpo: used at the line 336

src/groups/bmq/bmqp/bmqp_messageproperties.cpp
```diff
@@ -300,19 +316,38 @@ bool MessageProperties::streamInPropertyValue(const Property& p) const
...
-                                        p.d_length);
+        // Try to avoid copying the string.  'd_blop' already keeps a copy.
+        bmqu::BlobPosition end;
+        const int          ret =
+            bmqu::BlobUtil::findOffset(&end, *d_blob_p, position, p.d_length);
+        bool doCopy = true;
+        if (ret == 0) {
+            // Do not align
+            if (position.buffer() == end.buffer() ||
+                (position.buffer() + 1 == end.buffer() && end.byte() == 0)) {
```
678098: I think it's worth extracting this to `bmqu::BlobUtil::isDataContinuous(start, end)`

src/groups/bmq/bmqp/bmqp_messageproperties.cpp
```diff
@@ -300,19 +316,38 @@ bool MessageProperties::streamInPropertyValue(const Property& p) const
...
+                // Section is good
+                char* start = d_blob_p->buffer(position.buffer()).data() +
+                              position.byte();
+
+                p.d_value = bsl::string_view(start, p.d_length);
+                doCopy    = false;
+            }
+        }
+        if (doCopy) {
```
678098, line 335: We can still get rid of `doCopy` if we just return when we read a property

src/groups/bmq/bmqp/bmqp_messageproperties.h
```diff
@@ -918,7 +950,14 @@ MessageProperties::getPropertyAsInt64(const bsl::string& name) const
 inline const bsl::string&
 MessageProperties::getPropertyAsString(const bsl::string& name) const
 {
-    return getProperty<bsl::string>(name);
+    PropertyMapIter it = findProperty(name);
+    BSLS_ASSERT((it != d_properties.end()) && "Property does not exist");
+
+    const PropertyVariant& value = getPropertyValueAsString(it->second);
+
+    BSLS_ASSERT(value.is<bsl::string>() && "Property data type mismatch");
```
678098, line 958: There was just 1 call in this function before, now there are at least 2 calls and 2 extra checks. Not sure if it doesn't harm the performance more than the benefit we get from this PR.

dorjesinpo: There were the same calls `findProperty` and `getPropertyValue` (in `getProperty`).  
The same check.
What is added is this:
```
    if (v.is<bsl::string_view>()) {
        v = bsl::string(v.the<bsl::string_view>(), d_allocator_p);
    }
```    
Consider the fact that `MessagePropertiesReader` does not make a copy of string anymore (it does not execute this code).  That is one of main benefits of this PR.  The second one being not copying blob.

src/groups/bmq/bmqp/bmqp_messageproperties.cpp
dorjesinpo: Sure, but is it that much better?

678098: Let's keep the current one

678098: Without micro-benchmarks we can only speculate if it is better or not

src/groups/bmq/bmqp/bmqp_messageproperties.h
```diff
@@ -684,6 +696,7 @@ MessageProperties::setProperty(const bsl::string& name, const TYPE& value)
 
     p.d_length  = newPropValueLen;
     p.d_value   = value;
+    // This cannot have bsl::string_view' type.
```
678098, line 699:
```suggestion
    // This cannot have `bsl::string_view` type.
```
Missing opening quote

678098 (approved):
Looks good, let's merge it.
Can fix the comment in another PR.

src/groups/bmq/bmqp/bmqp_messageproperties.h
678098: Fixed here https://github.com/bloomberg/blazingmq/pull/1021

PR #1015, opened 2025-11-07
Fix[mqbblp]: Fix build on Solaris

pniedzielski:
This patch fixes a build error on Solaris/C++03: the new `ElectorEventFunctor` constructor takes a `bslmf::MovableRef<bmqp::Event>`, which in C++11 is an alias for `bmqp::Event&&`.  In C++03, though, temporary objects are not automatically converted into `MovableRef`s, resulting in an invocation of the `createInplace` function template to fail instantiation.  We solve this by explicitly calling `bslmf::MovableRefUtil::move` on this temporary, which is a no-op in C++11 but a necessary call in C++03.

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_event.h
+++ b/src/groups/bmq/bmqp/bmqp_event.h
@@ -52,6 +52,7 @@
 #include <bsl_string.h>
 #include <bslma_allocator.h>
 #include <bslma_usesbslmaallocator.h>
+#include <bslmf_movableref.h>
 #include <bslmf_nestedtraitdeclaration.h>
 #include <bsls_assert.h>
 #include <bsls_performancehint.h>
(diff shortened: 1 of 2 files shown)


678098 (dismissed):

src/groups/bmq/bmqp/bmqp_event.h
```diff
@@ -390,6 +401,15 @@ inline Event::Event(const Event& src, bslma::Allocator* allocator)
     initialize(src.d_blob_p, false);
 }
 
+inline Event::Event(bslmf::MovableRef<Event> src, bslma::Allocator* allocator)
+: d_allocator_p(allocator)
+, d_clonedBlob_sp(0, allocator)
+, d_blob_p(0)
```
678098, line 407: Might be possible to avoid double assignment of the same field in constructor

src/groups/bmq/bmqp/bmqp_event.h
```diff
@@ -399,6 +419,15 @@ inline Event& Event::operator=(const Event& rhs)
     return *this;
 }
 
+inline Event& Event::operator=(bslmf::MovableRef<Event> rhs)
+{
+    if (this != &bslmf::MovableRefUtil::access(rhs)) {
```
678098, line 424: Might be useful to invert this condition and add PREDICT UNLIKELY macro and UNLIKELY hint

678098 (approved):
I checked that Solaris build passes with the current version of the branch. We can merge and address the review notes later

PR #1016, opened 2025-11-10
Feat[stats]: Add domain metric for opened queues count

waldgange:
In addition to `domain_queue_count` metric representing the number of assigned (e.g. present in the storage) queues we decided to add a metric for the number of queues that are currently open (e.g. have clients). The new metric's name is `domain_queue_count_open`. Both metrics are calculated per domain and reported only from primary node.

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_event.h
+++ b/src/groups/bmq/bmqp/bmqp_event.h
@@ -52,6 +52,7 @@
 #include <bsl_string.h>
 #include <bslma_allocator.h>
 #include <bslma_usesbslmaallocator.h>
+#include <bslmf_movableref.h>
 #include <bslmf_nestedtraitdeclaration.h>
 #include <bsls_assert.h>
 #include <bsls_performancehint.h>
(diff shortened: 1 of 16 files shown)


src/groups/mqb/mqbc/mqbc_clusterstate.cpp
```diff
@@ -807,6 +807,17 @@ void ClusterState::DomainState::adjustQueueCount(int by)
     }
 }
 
+void ClusterState::DomainState::adjustOpenedQueueCount(int by)
+{
```
678098, line 812:
```suggestion
{
    // Thread: CLUSTER dispatcher
```

or longer

```suggestion
{
    // Executed by the CLUSTER dispatcher thread
```

This comment is needed to justify safety of change of `int d_numOpenedQueues`

678098 (commented):
Just one small comment

src/groups/mqb/mqbc/mqbc_clusterstate.cpp
waldgange: I also added the same comment to the previous function

PR #1017, opened 2025-11-13
style: Improve ctors and assignment operators of `bmqp_event`

pniedzielski:
This patch addresses two small comments that were found while reviewing changes to `bmqp_event` in PR #1015:

  1. First, it avoids double assignment to `d_clonedBlob_sp` in both the copy and the move constructors.

  2. Second, it adds `BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY` hints to self-assignment in both the copy and the move assignment operators.


678098 (approved):
Looks good

PR #1021, opened 2025-11-18
Refactor[BMQ]: simplify bmqu::BlobUtil

678098:
- Remove special handling of the case `start.buffer() == blob.numDataBuffers() && length > 0` and remove `rc == -3` return code from `BlobUtil`
- Use left-to-right comparisons for consistency.
- Fix typo and docs.
- Update UTs.

### DIFF
--- a/src/groups/bmq/bmqu/bmqu_blob.cpp
+++ b/src/groups/bmq/bmqu/bmqu_blob.cpp
@@ -1,4 +1,4 @@
-// Copyright 2022-2023 Bloomberg Finance L.P.
+// Copyright 2014-2025 Bloomberg Finance L.P.
 // SPDX-License-Identifier: Apache-2.0
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -86,7 +86,7 @@ bool BlobUtil::isValidPos(const bdlbb::Blob& blob, const BlobPosition& pos)
         return pos.byte() == 0;  // RETURN
     }
 
-    return pos.byte() >= 0 && pos.byte() < bufferSize(blob, pos.buffer());
+    return 0 <= pos.byte() && pos.byte() < bufferSize(blob, pos.buffer());
 }
 
 int BlobUtil::findOffset(BlobPosition*       pos,
@@ -123,9 +123,7 @@ int BlobUtil::findOffset(BlobPosition*       pos,
         }
     }
 
-    // Looking past the end of the blob (return -3 since the 'safe' version
-    // already uses -1 for invalid 'start' and -2 for invalid 'offset').
-    return -3;
+    return -2;
 }
 
 bool BlobUtil::isValidSection(const bdlbb::Blob& blob,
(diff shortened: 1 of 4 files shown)


src/groups/bmq/bmqu/bmqu_blob.cpp
```diff
@@ -494,12 +492,6 @@ int BlobUtil::readUpToNBytes(char*               buf,
         return -1;  // RETURN
     }
 
-    if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(
-            start.buffer() == blob.numDataBuffers() && length > 0)) {
-        BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
-        return -2;  // RETURN
```
678098, line 500: `BlobPosition` is valid, we checked it already.
In this case, it is 100% valid to read 0 bytes in `readUpToNBytes` if we start reading from the rightmost blob position.
This allows to simplify and remove this extra check.

src/groups/bmq/bmqu/bmqu_blob.h
```diff
@@ -578,14 +584,6 @@ inline int BlobUtil::findOffsetSafe(BlobPosition*       pos,
         return -1;  // RETURN
     }
 
-    // Check if 'offset' has a proper value if 'start' position points to
-    // the buffer past the last one of the blob.
-    if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(
-            start.buffer() == blob.numDataBuffers() && offset > 0)) {
-        BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
-        return -2;  // RETURN
```
678098, line 586: This check should be triggered very, very rarely, it makes no much sense to rely on it. Instead, we can just rely on `int BlobUtil::findOffset` to find that we are trying to find an offset beyond the end of the blob.

src/groups/bmq/bmqu/bmqu_blob.t.cpp
```diff
@@ -1090,7 +1090,7 @@ static void test14_readUpToNBytes()
         {L_, bmqu::BlobPosition(3, 0), 3, 3, "ghi"},
         {L_, bmqu::BlobPosition(3, 0), 12, 3, "ghi"},
         {L_, bmqu::BlobPosition(4, 0), 0, 0, ""},
-        {L_, bmqu::BlobPosition(4, 0), 1, -2, ""},  // invalid length
+        {L_, bmqu::BlobPosition(4, 0), 1, 0, ""},
```
678098, line 1093: This is a valid case now.
It is okay to try to read `up to` 1 byte in the end of the blob and read 0 bytes.

src/groups/bmq/bmqu/bmqu_blob.t.cpp
```diff
@@ -1105,18 +1105,18 @@ static void test14_readUpToNBytes()
...
 
-        BMQTST_ASSERT_EQ_D("line " << test.d_line, rc, test.d_expectedRc);
+        BMQTST_ASSERT_EQ_D("line " << test.d_line, numRead, test.d_expectedRc);
 
         if (test.d_expectedRc == 0) {
             BMQTST_ASSERT_EQ_D(
                 "line " << test.d_line,
                 bsl::string(buffer,
-                            test.d_length,
+                            numRead,
```
678098, line 1119: There was a bug here before.
This UTs didn't assume that we can read less than `d_length` bytes. The updated test case now covers this scenario.

src/groups/bmq/bmqu/bmqu_blob.t.cpp
678098: + add comment

PR #1023, opened 2025-11-19
Feat[BMQ]: add reserved property name

emelialei88:
Reserve "bmq.*" property name for internal use only.

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_messageproperties.h
+++ b/src/groups/bmq/bmqp/bmqp_messageproperties.h
@@ -349,6 +349,10 @@ class MessageProperties {
     // there is only one property and property's name has maximum allowable
     // length, and also takes into consideration the protocol overhead.
 
+    /// Reserved property name prefix. Property names starting with this
+    /// prefix are reserved for internal use and cannot be set by users.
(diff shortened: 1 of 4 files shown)


src/groups/bmq/bmqp/bmqp_messageproperties.h
```diff
@@ -624,6 +628,13 @@ inline bool MessageProperties::isValidPropertyName(const bsl::string& name)
         return false;  // RETURN
     }
 
+    // Check if the name starts with the reserved prefix
+    const size_t reservedPrefixLen = bsl::strlen(k_RESERVED_PROPERTY_PREFIX);
+    if (name.length() >= reservedPrefixLen &&
+        name.compare(0, reservedPrefixLen, k_RESERVED_PROPERTY_PREFIX) == 0) {
+        return false;  // RETURN
+    }
+
```
hallfox, line 637: [`starts_with`](https://en.cppreference.com/w/cpp/string/basic_string/starts_with.html) is available in BDE via a backport

hallfox:
```suggestion
  if (name.starts_with(k_RESERVED_PROPERTY_PREFIX)) {
    return false;
  }
```

PR #1024, opened 2025-11-19
Refactor[bmqc::MQTP]: get rid of `QueueItem`

678098:
Before, there were 3 types of events enqueued to MQTP:
- User event (using `QueueItem` with `d_monitorEvent == false` and `d_event_p != NULL`)
- Monitor event (using `QueueItem` with `d_monitorEvent == true` and `d_event_p == NULL`)
- Zero event (using `QueueItem` with `d_monitorEvent == false` and `d_event_p == NULL`) - to notify queue processing thread that we need to stop

This PR combines monitor event and zero event, which allows to get rid of `QueueItem` class at all. Now, there are 2 types of events enqueued to MQTP:
- User event (`event != NULL`)
- Monitor event (`event == NULL`)

The new reference counting mechanism allows to use monitor events to signal the queue processing thread that we have to stop processing.

The performance of `bdlcc::SingleConsumerQueue::pushBack` is the same after replacing struct item type to pointer item type:
```
----------------------------------------------------------------------------------------
Benchmark                                              Time             CPU   Iterations
----------------------------------------------------------------------------------------
pushBack<char*> threads=1/iterations:1              21.9 ms        0.009 ms            1
pushBack<char*> threads=2/iterations:1               144 ms        0.011 ms            1
pushBack<char*> threads=4/iterations:1               484 ms        0.018 ms            1
pushBack<char*> threads=8/iterations:1              2412 ms        0.014 ms            1
pushBack<ComplexItem> threads=1/iterations:1        22.4 ms        0.014 ms            1
pushBack<ComplexItem> threads=2/iterations:1         146 ms        0.013 ms            1
pushBack<ComplexItem> threads=4/iterations:1         489 ms        0.015 ms            1
pushBack<ComplexItem> threads=8/iterations:1        2413 ms        0.017 ms            1

```

However, the performance of MQTP itself should improve a bit, because we removed one of the checks in the queue loop

### DIFF
--- a/src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
+++ b/src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
@@ -175,6 +175,7 @@
 #include <bslmt_lockguard.h>
 #include <bslmt_mutex.h>
 #include <bslmt_threadutil.h>
+#include <bslmt_timedsemaphore.h>
 #include <bsls_assert.h>
 #include <bsls_atomic.h>
 #include <bsls_keyword.h>
@@ -510,7 +511,10 @@ class MultiQueueThreadPool BSLS_KEYWORD_FINAL {
         /// as long as `stop()` was not called.
         bsls::AtomicInt d_processQueueRefCount;
 
-        bslmt::Semaphore d_finished;
+        /// A semaphore used to verify that a queue has stopped.
+        /// Note: shared_ptr is used because copy/move are not defined for
+        ///       `TimedSemaphore` and we still need to copy `QueueInfo`.
+        bsl::shared_ptr<bslmt::TimedSemaphore> d_finished_sp;
 
         bslmt::ThreadUtil::Handle d_exclusiveThreadHandle;
 
@@ -524,7 +528,10 @@ class MultiQueueThreadPool BSLS_KEYWORD_FINAL {
         , d_name(basicAllocator)
         , d_monitorState(e_MONITOR_PROCESSED)
         , d_processQueueRefCount(1)
-        , d_finished(0)
+        , d_finished_sp(bsl::allocate_shared<bslmt::TimedSemaphore>(
+              basicAllocator,
+              0,
+              bsls::SystemClockType::e_MONOTONIC))
         , d_exclusiveThreadHandle(bslmt::ThreadUtil::invalidHandle())
         {
             // NOTHING
@@ -538,7 +545,7 @@ class MultiQueueThreadPool BSLS_KEYWORD_FINAL {
         , d_monitorState(static_cast<int>(other.d_monitorState))
         , d_processQueueRefCount(
               static_cast<int>(other.d_processQueueRefCount))
-        , d_finished(other.d_finished.getValue())
+        , d_finished_sp(other.d_finished_sp)
         , d_exclusiveThreadHandle(other.d_exclusiveThreadHandle)
         {
             // NOTHING
@@ -557,6 +564,10 @@ class MultiQueueThreadPool BSLS_KEYWORD_FINAL {
   private:
     // CLASS-SCOPE CATEGORY
     BALL_LOG_SET_CLASS_CATEGORY("BMQC.MULTIQUEUETHREADPOOL");
+
+    // CLASS DATA
+    /// The timeout for `timedWait` when stopping the queues
+    static const int k_MAX_WAIT_SECONDS_AT_SHUTDOWN = 300;
 
     // DATA
     Config d_config;
@@ -967,7 +978,7 @@ inline void MultiQueueThreadPool<TYPE>::processQueue(int queue)
                 // Note: it is possible that another monitor event will be
                 //       enqueued right after the check is done, but it's okay.
                 //       We will skip it with any remainder events on `stop()`.
-                info.d_finished.post();
+                info.d_finished_sp->post();
                 return;  // RETURN
             }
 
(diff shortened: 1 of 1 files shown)


src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -339,14 +339,7 @@ class MultiQueueThreadPoolConfig {
...
 
-    struct QueueItem {
-        // PUBLIC DATA
-        bool d_monitorEvent;  // this is a monitor event
-
-        Event* d_event_p;  // Pointer to event
-    };
-
-    typedef MonitoredQueue<bdlcc::SingleConsumerQueue<QueueItem> > Queue;
+    typedef MonitoredQueue<bdlcc::SingleConsumerQueue<Event*> > Queue;
```
678098, line 343: This step is useful for the future refactor of MQTP

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -930,26 +939,38 @@ template <typename TYPE>
...
+            // Process monitor event
+            if (0 == info.d_processQueueRefCount.subtractRelaxed(1)) {
+                // 0 ref count means that:
+                // - `stop()` was called: it released the initial reference.
+                // - It is the last monitor event enqueued to the queue.
+                // No need to process this event, it is time to return.
+                // Note: it is possible that another monitor event will be
+                //       enqueued right after the check is done, but it's okay.
+                //       We will skip it with any remainder events on `stop()`.
+                info.d_finished.post();
```
678098: It is not possible to use only one atomic counter `d_processQueueRefCount` to notify the stopping thread that we have processed everything. This leads to tricky multi-threaded issues.
Using a second variable, semaphore `d_finished`, is a more robust solution free from these issues.

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -960,13 +981,6 @@ inline void MultiQueueThreadPool<TYPE>::processQueue(int queue)
             continue;  // CONTINUE
         }
 
-        Event* event = item.d_event_p;
-        if (event == 0) {
-            // We've been told to return
-            info.d_processedZero = true;
-            return;  // RETURN
-        }
```
678098, line 968: Removing this check should slightly improve performance of MQTP

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -1161,20 +1170,26 @@ inline void MultiQueueThreadPool<TYPE>::stop()
     for (size_t i = 0; i < d_queues.size(); ++i) {
         QueueInfo& info = d_queues[i];
 
-        QueueItem item = {false  // isMonitorEvent
-                          ,
-                          0};  // pointer to event
-        info.d_queue_p->pushBack(item);
+        // The following operations (1) and (2) can be removed from the code
```
dorjesinpo: let's remove (1) and (2) and just leave a comment

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -1161,20 +1170,26 @@ inline void MultiQueueThreadPool<TYPE>::stop()
...
+        // It is possible that something is enqueued to the queue between the
+        // last monitor event and `disablePushBack()` call, this is expected.
         info.d_queue_p->disablePushBack();
 
-        // TODO: Change 'd_processedZero' to semaphore instead to avoid busy
-        //       waiting?
-        while (!info.d_processedZero) {
-            bslmt::ThreadUtil::yield();
-        }
+        info.d_finished.wait();
```
dorjesinpo: theoretically, it is impossible not to `post` but maybe use `timedWait`?

678098: Changed the type `bslmt::Semaphore` -> `bslmt::TimedSemaphore` and added a timeout (300) for this operation

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -1158,23 +1172,46 @@ inline void MultiQueueThreadPool<TYPE>::stop()
...
 
-        // TODO: Change 'd_processedZero' to semaphore instead to avoid busy
-        //       waiting?
-        while (!info.d_processedZero) {
-            bslmt::ThreadUtil::yield();
+    // 2. Verify that queues have stopped
+    const bsls::TimeInterval timeout =
+        bsls::SystemTime::nowMonotonicClock().addSeconds(
+            k_MAX_WAIT_SECONDS_AT_SHUTDOWN);
+    for (size_t i = 0; i < d_queues.size(); ++i) {
```
678098: @dorjesinpo note that I have divided one loop in two loops.
First, we enqueue monitor events and disable push back for all queues.
Second, we wait for all queues to finish using the shared timeout.

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -1158,23 +1172,46 @@ inline void MultiQueueThreadPool<TYPE>::stop()
...
+        QueueInfo& info = d_queues[i];
+
+        const int rc = info.d_finished.timedWait(timeout);
+        if (0 != rc) {
+            BALL_LOG_ERROR << "#MQTP_STOP_FAILURE MQTP failed to stop in "
+                           << k_MAX_WAIT_SECONDS_AT_SHUTDOWN
+                           << " seconds while shutting down the queue (" << i
+                           << ", " << info.d_name << ", " << info.d_queue_p
+                           << "), rc:  " << rc << ".";
```
678098: I also think that it's better to just crash in this scenario, not trying to do anything else gracefully. It might be dangerous to continue.
We either have a controlled crash here or some undefined operations and a crash in a random place later.

dorjesinpo: I agree, `exit` sounds like a better option rather than to continue

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
678098: Upd: reverted this change, it causes many tests to fail and requires investigation that is out of scope for this PR

678098: https://github.com/bloomberg/blazingmq/issues/1028

PR #1025, opened 2025-11-19
[WIP]Perf[BMQ]: fast `memcpy` properties from continuous buffers

678098:
We cannot not directly cast/assign a pointer even for simple types because alignment in a `Blob` is not guaranteed. Have to call `memcpy` with hopes that it will be optimized
TODO: check asm outputs

### DIFF
--- a/src/groups/bmq/bmqu/bmqu_blob.h
+++ b/src/groups/bmq/bmqu/bmqu_blob.h
@@ -33,6 +33,7 @@
 
 // BDE
 #include <bdlbb_blob.h>
+#include <bsl_cstring.h>  // for bsl::memcpy
 #include <bsl_iostream.h>
 
 namespace BloombergLP {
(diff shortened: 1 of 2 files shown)


src/groups/bmq/bmqu/bmqu_blob.h
```diff
@@ -553,6 +559,29 @@ inline const BlobPosition& BlobSection::end() const
...
+    if (isDataContinuous()) {
+        bsl::memcpy(buf,
+                    blob.buffer(d_start.buffer()).data() + d_start.byte(),
+                    numBytes);
+    }
+    else {
+        // Rare path
+        bmqu::BlobUtil::readNBytes(buf, blob, d_start, numBytes);
+    }
+}
```
678098, line 583:
```
int BlobUtil::readUpToNBytes(char*               buf,
                             const bdlbb::Blob&  blob,
                             const BlobPosition& start,
                             int                 length)
{
    if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(!isValidPos(blob, start))) {
        return -1;  // RETURN
    }

    if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(length == 0)) {
        return 0;  // RETURN
    }

    char*              outPosition = buf;
    const int          numBuffers  = blob.numDataBuffers();
    bmqu::BlobPosition cursor(start);

    while (length != 0 && (cursor.buffer() != numBuffers)) {
        const bdlbb::BlobBuffer& buffer   = blob.buffer(cursor.buffer());
        const int                buffSize = bufferSize(blob, cursor.buffer());

        const int toCopy = bsl::min(buffSize - cursor.byte(), length);
        bsl::memcpy(outPosition, buffer.data() + cursor.byte(), toCopy);

        outPosition += toCopy;
        length -= toCopy;

        cursor.setBuffer(cursor.buffer() + 1);
        cursor.setByte(0);
    }

    return static_cast<int>(outPosition - buf);
}
```
Short `memcpy` path should skip all these other operations

678098:
Requires more deep investigation

PR #1031, opened 2025-11-21
IT[test_client_timeout]: Ensure queues are assigned

pniedzielski:
The `test_client_timeout.py` integration test creates producers with artificially short timeouts in order to test the behavior of clients on open, reopen, and close timeouts.  However, because bmqtool only supports a single timeout flag, the short timeouts can cause the initial open queue operation to fail on slow test runners.  This patch makes these test more robust, by:

  1) introducing a client with a long timeout to each test, which connects and opens the necessary queues, assigning them on the primary; and

  2) suspending the primary in each test as well as all other brokers in the cluster, to ensure that the queue already being assigned does not interfere with the short timeouts on the operations we want them for.

If we need to, we can add separate flags in bmqtool for each timeout exposed in `SessionOptions`.  This is a cheaper solution for just this test.


dorjesinpo (approved):
thank you

PR #1035, opened 2025-11-24
Test[mqbc]: reproduce CSL advisory get stuck

waldgange:
This test demonstrates how we can get a CSL advisory stuck forever so that only restart of a node can help. We should either keep it in mind or improve handling of the admin command that sets elector quorum.
Suggestion: after every quorum change, review [the uncommitted csl advisories](https://github.com/bloomberg/blazingmq/blob/eba490bf1da17c9a1c3db84ace9bb21e1835b67f/src/groups/mqb/mqbc/mqbc_incoreclusterstateledger.h#L202) to see if some of them already have more ACKs than the new quorum.
We have similar mechanism for unreceipted records when we handle ReplicationFactor change:
https://github.com/bloomberg/blazingmq/blob/eba490bf1da17c9a1c3db84ace9bb21e1835b67f/src/groups/mqb/mqbs/mqbs_filestore.cpp#L7283-L7285

### DIFF
--- a/src/integration-tests/test_cluster_node_shutdown.py
+++ b/src/integration-tests/test_cluster_node_shutdown.py
@@ -20,6 +20,7 @@ go to the relevant section in the README.md, in this directory.
 """
 
 import queue
+import re
 from time import sleep
 from typing import List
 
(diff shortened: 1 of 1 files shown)


678098 (approved):
As far as I understood, the new IT is successful on incorrect behavior.
Once the quorum thing is fixed, this IT should also change, right?
I am fine with merging this now and addressing the issue in the next PR
Thanks!

PR #396, opened 2024-08-09
[WIP]Performance[mqbi::DispatcherEvent]: replace multiple inheritance with variant

678098:

PR #1037, opened 2025-11-24
Refactor[bmqc::MQTP]: `ObjectPool` -> `SharedObjectPool`

678098:
- Use `SharedObjectPool` for `DispatcherEvent` allocation. This allows to return events to the owner pool automatically, without manual reference counting. This is an essential step before using per-thread object pools.
- Get rid of `MultiQueueThreadPoolEvent` (and "unmanaged" events). Its use for reference counting is not needed anymore when we have shared pointers to events.
- Check for finalization callback in `mqbi::DispatcherEvent::reset`. This allows to simplify `mqbi::Dispatcher`. The finalization callback feature can be easily implemented by the concrete event type, so it is not necessary to make `mqbi::Dispatcher` more complicated to allow this.
- Always get events by shared pointer and return with `move`.

### DIFF
--- a/src/groups/mqb/mqba/mqba_dispatcher.cpp
+++ b/src/groups/mqb/mqba/mqba_dispatcher.cpp
@@ -390,7 +390,7 @@ void Dispatcher::queueEventCb(mqbi::DispatcherClientType::Enum type,
 {
     if (event) {
         BALL_LOG_TRACE << "Dispatching Event to queue " << processorId
-                       << " of " << type << " dispatcher: " << event;
+                       << " of " << type << " dispatcher: " << *event;
         if (event->type() == mqbi::DispatcherEventType::e_DISPATCHER) {
             const mqbi::DispatcherDispatcherEvent* realEvent =
                 event->asDispatcherEvent();
--- a/src/groups/mqb/mqbi/mqbi_dispatcher.h
+++ b/src/groups/mqb/mqbi/mqbi_dispatcher.h
@@ -320,7 +320,7 @@ class Dispatcher {
     // TYPES
 
     typedef bsl::shared_ptr<mqbi::DispatcherEvent> DispatcherEventSp;
-    typedef bslmf::MovableRef<DispatcherEventSp>   DispatcherEventRef;
+    typedef bslmf::MovableRef<DispatcherEventSp>   DispatcherEventRvRef;
 
     /// Type representing a handle to a processor in the dispatcher.
     typedef int ProcessorHandle;
@@ -382,14 +382,14 @@ class Dispatcher {
     /// Dispatch the specified `event` to the specified `destination`.  The
     /// behavior is undefined unless `event` was obtained by a call to
     /// `getEvent` with a type matching the one of `destination`.
-    virtual void dispatchEvent(DispatcherEventRef event,
-                               DispatcherClient*  destination) = 0;
+    virtual void dispatchEvent(DispatcherEventRvRef event,
+                               DispatcherClient*    destination) = 0;
 
     /// Dispatch the specified `event` to the processor in charge of clients
     /// of the specified `type` and associated with the specified `handle`.
     /// The behavior is undefined unless `event` was obtained by a call to
     /// `getEvent` with a matching `type`..
-    virtual void dispatchEvent(DispatcherEventRef         event,
+    virtual void dispatchEvent(DispatcherEventRvRef       event,
                                DispatcherClientType::Enum type,
                                ProcessorHandle            handle) = 0;
 
@@ -1623,7 +1623,7 @@ inline void DispatcherEvent::reset()
         !d_finalizeCallback.empty()) {
         // We only set finalizeCallback on e_DISPATCHER events
 
-        // TODO: make a special event type that handles this case
+        // TODO(678098): make a special event type that handles this case
         d_finalizeCallback();
     }
 
--- a/src/groups/mqb/mqba/mqba_dispatcher.h
+++ b/src/groups/mqb/mqba/mqba_dispatcher.h
@@ -414,15 +414,15 @@ class Dispatcher BSLS_CPP11_FINAL : public mqbi::Dispatcher {
 
     /// Dispatch the specified `event` to the specified `destination`.
     void
-    dispatchEvent(mqbi::Dispatcher::DispatcherEventRef event,
+    dispatchEvent(mqbi::Dispatcher::DispatcherEventRvRef event,
                   mqbi::DispatcherClient* destination) BSLS_KEYWORD_OVERRIDE;
 
     /// Dispatch the specified `event` to the queue associated with the
     /// specified `type` and `handle`.  The behavior is undefined unless the
     /// `event` was obtained by a call to `getEvent`.
-    void dispatchEvent(mqbi::Dispatcher::DispatcherEventRef event,
-                       mqbi::DispatcherClientType::Enum     type,
-                       mqbi::Dispatcher::ProcessorHandle    handle)
+    void dispatchEvent(mqbi::Dispatcher::DispatcherEventRvRef event,
+                       mqbi::DispatcherClientType::Enum       type,
+                       mqbi::Dispatcher::ProcessorHandle      handle)
         BSLS_KEYWORD_OVERRIDE;
 
     /// Execute the specified `functor` in the processors in charge of
@@ -536,8 +536,8 @@ Dispatcher::getEvent(const mqbi::DispatcherClient* client)
 }
 
 inline void
-Dispatcher::dispatchEvent(mqbi::Dispatcher::DispatcherEventRef event,
-                          mqbi::DispatcherClient*              destination)
+Dispatcher::dispatchEvent(mqbi::Dispatcher::DispatcherEventRvRef event,
+                          mqbi::DispatcherClient*                destination)
 {
     BALL_LOG_TRACE << "Enqueuing Event to '" << destination->description()
                    << "': " << *event;
@@ -550,9 +550,9 @@ Dispatcher::dispatchEvent(mqbi::Dispatcher::DispatcherEventRef event,
 }
 
 inline void
-Dispatcher::dispatchEvent(mqbi::Dispatcher::DispatcherEventRef event,
-                          mqbi::DispatcherClientType::Enum     type,
-                          mqbi::Dispatcher::ProcessorHandle    handle)
+Dispatcher::dispatchEvent(mqbi::Dispatcher::DispatcherEventRvRef event,
+                          mqbi::DispatcherClientType::Enum       type,
+                          mqbi::Dispatcher::ProcessorHandle      handle)
 {
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(handle != mqbi::Dispatcher::k_INVALID_PROCESSOR_HANDLE);
--- a/src/groups/mqb/mqbmock/mqbmock_dispatcher.cpp
+++ b/src/groups/mqb/mqbmock/mqbmock_dispatcher.cpp
@@ -80,14 +80,14 @@ Dispatcher::getEvent(BSLA_UNUSED mqbi::DispatcherClientType::Enum type)
     return mqbi::Dispatcher::DispatcherEventSp();
 }
 
-void Dispatcher::dispatchEvent(mqbi::Dispatcher::DispatcherEventRef event,
+void Dispatcher::dispatchEvent(mqbi::Dispatcher::DispatcherEventRvRef event,
                                mqbi::DispatcherClient* destination)
 {
     destination->onDispatcherEvent(*event);
 }
 
 void Dispatcher::dispatchEvent(
-    BSLA_UNUSED mqbi::Dispatcher::DispatcherEventRef event,
+    BSLA_UNUSED mqbi::Dispatcher::DispatcherEventRvRef event,
     BSLA_UNUSED mqbi::DispatcherClientType::Enum type,
     BSLA_UNUSED mqbi::Dispatcher::ProcessorHandle handle)
 {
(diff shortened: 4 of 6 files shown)


src/groups/mqb/mqbi/mqbi_dispatcher.h
```diff
@@ -1616,6 +1619,14 @@ DispatcherEvent::setState(const bsl::shared_ptr<bmqu::AtomicState>& state)
 
 inline void DispatcherEvent::reset()
 {
+    if (d_type == mqbi::DispatcherEventType::e_DISPATCHER &&
+        !d_finalizeCallback.empty()) {
+        // We only set finalizeCallback on e_DISPATCHER events
+
+        // TODO: make a special event type that handles this case
+        d_finalizeCallback();
+    }
```
678098, line 1628: We call `reset()` when we release the last reference to `event` and just before we return it to the shared object pool.
This automatically means that the event was already processed by all queues it was enqueued to.

src/groups/mqb/mqba/mqba_dispatcher.cpp
```diff
@@ -412,39 +412,21 @@ void Dispatcher::queueEventCb(mqbi::DispatcherClientType::Enum type,
...
-        if (event->object().type() ==
-            mqbi::DispatcherEventType::e_DISPATCHER) {
-            const mqbi::DispatcherDispatcherEvent* realEvent =
-                event->object().asDispatcherEvent();
-
-            if (!realEvent->finalizeCallback().empty()) {
-                BALL_LOG_TRACE << "Calling finalizeCallback on queue "
-                               << processorId << " of " << type
-                               << " dispatcher: " << event->object();
-                realEvent->finalizeCallback()();
```
678098, line 443: This complexity moved to `mqbi::Dispatcher::reset`

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -935,33 +783,22 @@ inline void MultiQueueThreadPool<TYPE>::processMonitorEvents()
     }
 }
 
-template <typename TYPE>
-inline bool MultiQueueThreadPool<TYPE>::unrefEvent(Event* event, bool release)
-{
-    const bool isLastRef = (--event->d_refCount == 0);
-    if (isLastRef && release) {
-        d_pool.releaseObject(event);
```
678098, line 943: Manual ref counting replaced by shared pointer ref counting mechanism

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -993,27 +830,13 @@ inline void MultiQueueThreadPool<TYPE>::processQueue(int queue)
...
-                             Config::BMQC_FINALIZE_ALL);
-            finalize      = finalize || ((d_config.d_finalizeType ==
-                                     Config::BMQC_FINALIZE_MULTI_QUEUE) &&
-                                    event->d_enqueuedOnMultipleQueues);
-
-            if (finalize) {
-                event->d_type = Event::BMQC_FINALIZE_EVENT;
-                d_config.d_eventCallbackFn(queue,
-                                           info.d_context_p.get(),
-                                           event);
```
678098, line 1008: Handled by `mqbi::DispatcherEvent::reset()`

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -993,27 +830,13 @@ inline void MultiQueueThreadPool<TYPE>::processQueue(int queue)
         }
 
         d_config.d_eventCallbackFn(queue, info.d_context_p.get(), event);
-
-        if (unrefEvent(event, false)) {
-            bool finalize = (d_config.d_finalizeType ==
-                             Config::BMQC_FINALIZE_ALL);
-            finalize      = finalize || ((d_config.d_finalizeType ==
-                                     Config::BMQC_FINALIZE_MULTI_QUEUE) &&
-                                    event->d_enqueuedOnMultipleQueues);
```
678098: TODO: want to remove these settings fully `Config::BMQC_FINALIZE_MULTI_QUEUE`

src/groups/mqb/mqba/mqba_dispatcher.cpp
```diff
@@ -326,7 +328,6 @@ int Dispatcher::startContext(bsl::ostream&                    errorDescription,
 
     processorPoolConfig.setName(mqbi::DispatcherClientType::toAscii(type))
         .setEventScheduler(d_scheduler_p)
-        .setFinalizeEvents(ProcessorPool::Config::BMQC_FINALIZE_MULTI_QUEUE)
```
678098, line 329: The broker fundamentally will not work if we ever change this setting, removed it for simplicity. The code now assumes that we always need to call finalization callback if conditions are met (done in `mqbi::Dispatcher::reset()`)

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -354,7 +265,8 @@ class MultiQueueThreadPoolConfig {
     /// Callback invoked when processing the specified `event` popped from
     /// the queue having the specified `queueId` and created with the
     /// specified `queueContext`.
-    typedef bsl::function<void(int queueId, void* queueContext, Event* event)>
+    typedef bsl::function<
+        void(int queueId, void* queueContext, const EventSp& event)>
```
678098, line 270: This is a design choice to pass events to client function like `const sp&` and enqueue them only like `bslmf::MovableRef`. This way, it is impossible to re-enqueue the same event to the processing queue and cause infinite processing. Arg type `const sp&` is enough for all our use cases

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -317,30 +241,17 @@ class MultiQueueThreadPool_QueueCreatorRet {
...
-                                   // on multiple queues
-        ,
-        BMQC_FINALIZE_ALL  // Finalize all events
-    };
-
   private:
     // PRIVATE TYPES
-    typedef MultiQueueThreadPoolEvent<TYPE> Event;
+    typedef TYPE                   Event;
+    typedef bsl::shared_ptr<Event> EventSp;
```
hallfox, line 248: Are we going with `Sp` or `SP` for the abbreviation of `shared_ptr`? Just asking so we can be consistent. Seems like there's a lot more instances of `Sp`.

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -659,37 +556,18 @@ class MultiQueueThreadPool BSLS_KEYWORD_FINAL {
...
     /// that if any queue is full, this will block.
-    int enqueueEventOnAllQueues(Event* event);
-
-    /// Enqueue the specified `event` on all queues and return `0` if this
-    /// `MultiQueueThreadPool` is started, and return `-1` otherwise'.  Note
-    /// that if any queue is full, this will block.  WARNING: The behavior
-    /// is undefined unless the specified `userEvent` is a member of an
-    /// event obtained from a call to `getUnmanagedEvent`.
-    int enqueueEventOnAllQueues(TYPE* userEvent);
+    int enqueueEventOnAllQueues(bslmf::MovableRef<EventSp> event);
```
hallfox, line 566: Big fan of taking these raw pointers and turning them into rvalue references 👍🏻.

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -1250,60 +1031,40 @@ inline int MultiQueueThreadPool<TYPE>::setMonitorAlarmTimeout(
...
-inline int MultiQueueThreadPool<TYPE>::enqueueEvent(TYPE* userEvent,
-                                                    int   queueId)
-{
-    // The userEvent was 'extracted' from an event, retrieve the event back by
-    // pointing to the correct address in memory.
-    Event* event = reinterpret_cast<Event*>(
-        (reinterpret_cast<char*>(userEvent) - d_eventOffset));
-
-    return enqueueEvent(event, queueId);
-}
```
hallfox, line 1282: I'm glad we don't need this anymore.

src/groups/mqb/mqba/mqba_dispatcher.cpp
```diff
@@ -385,16 +386,14 @@ Dispatcher::queueCreator(mqbi::DispatcherClientType::Enum             type,
...
+                              const ProcessorPool::EventSp&    event)
 {
-    switch (event->type()) {
-    case ProcessorPool::Event::BMQC_USER: {
+    if (event) {
         BALL_LOG_TRACE << "Dispatching Event to queue " << processorId
-                       << " of " << type << " dispatcher: " << event->object();
-        if (event->object().type() ==
-            mqbi::DispatcherEventType::e_DISPATCHER) {
+                       << " of " << type << " dispatcher: " << event;
```
hallfox:
```suggestion
                       << " of " << type << " dispatcher: " << *event;
```

src/groups/mqb/mqba/mqba_dispatcher.cpp
```diff
@@ -412,39 +411,21 @@ void Dispatcher::queueEventCb(mqbi::DispatcherClientType::Enum type,
...
-
-            if (!realEvent->finalizeCallback().empty()) {
-                BALL_LOG_TRACE << "Calling finalizeCallback on queue "
-                               << processorId << " of " << type
-                               << " dispatcher: " << event->object();
-                realEvent->finalizeCallback()();
-            }
-        }
-    } break;
-    default: BSLA_UNREACHABLE;
```
hallfox, line 447: Do we still want to keep this case?

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -993,74 +794,9 @@ inline void MultiQueueThreadPool<TYPE>::processQueue(int queue)
...
-        /// No error
-        rc_SUCCESS = 0,
-
-        /// An error was encountered while pushing back (queue is disabled)
-        rc_PUSH_BACK_ERROR = -1
-    };
-
-    // Determine start and end queues for which to enqueue
-    int startQueue = queue;
-    int endQueue   = queue + 1;
```
678098, line 1033: Don't want to pay for these operations when we want to enqueue to just 1 queue, simplified

src/groups/mqb/mqbi/mqbi_dispatcher.h
```diff
@@ -319,6 +319,9 @@ class Dispatcher {
   public:
     // TYPES
 
+    typedef bsl::shared_ptr<mqbi::DispatcherEvent> DispatcherEventSp;
+    typedef bslmf::MovableRef<DispatcherEventSp>   DispatcherEventRef;
```
hallfox: `DispatcherEventRef` made me think the type was `DispatcherEventSp&` above, perhaps we could use an abbreviation like `RvRef` (for "r-value reference")?

src/groups/mqb/mqbi/mqbi_dispatcher.h
```diff
@@ -1616,6 +1619,14 @@ DispatcherEvent::setState(const bsl::shared_ptr<bmqu::AtomicState>& state)
 
 inline void DispatcherEvent::reset()
 {
+    if (d_type == mqbi::DispatcherEventType::e_DISPATCHER &&
+        !d_finalizeCallback.empty()) {
+        // We only set finalizeCallback on e_DISPATCHER events
+
+        // TODO: make a special event type that handles this case
```
hallfox: Is this TODO to be handled in this PR?

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
hallfox: Can we pass them as `const Event&` then? I think it would totally still be possible to do this:

```c++
void callback(int, void* context, const EventSp& event) {
  EventSp event2 = event;
  context->enqueueEvent(bsl::move(event2));
}
```

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
678098: I think I've seen `Sp` more often. Let's keep it?

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
678098: Currently pointer type serves 2 purposes: it is either a user event or a queue empty event. Pointer types are good to signal this, with empty (NULL) pointer meaning empty queue event. `const Event&` doesn't have a special value for this case.
There are some alternatives. It is possible to provide a separate "empty queue callback" that doesn't accept an event. It is also possible to make events more complex.
For this PR, I would rather keep the existing code, we can do the next refactor later.

src/groups/mqb/mqba/mqba_dispatcher.cpp
678098: We don't have switch/case anymore.
Instead, we check `event` shared pointer:
```
if (event) {
    // user event, can process
} else {
    // empty `event` means queue is empty
}
```

src/groups/mqb/mqbi/mqbi_dispatcher.h
678098: This one should be done once we refactor the inheritance of dispatcher event classes, similar to this PR https://github.com/bloomberg/blazingmq/pull/396

We will have an abstract `DispatcherEvent` and multiple concrete implementations. Each of these implementations will only have fields needed for it and will have a `reset()` that cleans exactly what is needed. This way, we can keep execute on `reset()` only for `DispatcherDispatcherEvent`

src/groups/mqb/mqba/mqba_dispatcher.cpp
678098: Nice catch!

src/groups/mqb/mqbi/mqbi_dispatcher.h
hallfox:
```suggestion
        // TODO(678098): make a special event type that handles this case
```

Marking who to refer to when we get to following up on this

hallfox (approved):

PR #1042, opened 2025-12-04
Fix[MQB]: downstream clears app too early

dorjesinpo:
In the following sequence, proxy clears its state too early, while primary still relies on its state.
Happens when closing last consumer and opening new one are interleaved.

```
proxy                                   primary

-> open (first handle)                  open (readCount = 1)
open <-
(readCount = 1)

-> configure                            configure
configure <-

post                                    post

tearDown (first handle)
-> deconfigure                          deconfigure 

-> open (second handle)                 open (readCount = 2)
deconfigure <-                

-> close (first handle)                 close (readCount = 1)  
                                        decision is made on App readCount. No RDL!

open <- (second handle)

-> configure                            configure (empty RDL)

close <- (first handle) 
clear the state!
decision is made on handle readCount (0)

configure <-
```

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
@@ -1146,15 +1146,17 @@ void RootQueueEngine::releaseHandle(
 
             BSLS_ASSERT_SAFE(itApp != d_apps.end());
 
-            AppState* app(itApp->second.get());
+            AppStateSp app = itApp->second;
 
             BSLS_ASSERT_SAFE(itApp->first == app->appId());
+
+            bool isConfigured = app->find(handle);
 
             if (result.hasNoHandleStreamConsumers()) {
                 // No re-delivery attempts until entire handle stops consuming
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1146,15 +1146,17 @@ void RootQueueEngine::releaseHandle(
 
             BSLS_ASSERT_SAFE(itApp != d_apps.end());
 
-            AppState* app(itApp->second.get());
+            AppStateSp app = itApp->second;
```
678098, line 1149: Do we have to make a copy of a `shared_ptr`?
Depending on its usage next, can be either `AppStateSp&` or `const AppStateSp&`

dorjesinpo: Yes. It is better to make sure the object is not destructed.  Note the line `d_apps.erase(itApp);` and the use of the `app` after.
(This is not in the performance critical path - once per QueueClose)

PR #1046, opened 2025-12-11
Fix[MQB]: stop catchup before reaching common iterator

dorjesinpo:
When `RelayQueueEngine` calls `catchUp` for a single App, the PushStream may have a gap resulting in the App first element being behind common iterator.
A solution is to introduce a seqeunce number (only for PushStream messages)
For `RootQueueEngine`, `catchUp` still can rely on comparing current guid against the common iterator.
Since both engines use the same `mqbi::StorageIterator` interface, the duality is expressed as `bdlb::Variant<bsls::Types::Uint64, bmqt::MessageGUID>`

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_pushstream.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_pushstream.cpp
@@ -81,8 +81,6 @@ bool PushStreamIterator::loadMessageAndAttributes() const
 
 const PushStream::Message& PushStreamIterator::message() const
 {
-    BSLS_ASSERT_SAFE(d_iterator != d_owner_p->d_stream.end());
-
     return d_iterator->second;
 }
 
--- a/src/groups/mqb/mqbblp/mqbblp_pushstream.h
+++ b/src/groups/mqb/mqbblp/mqbblp_pushstream.h
@@ -131,10 +131,10 @@ class PushStream {
     };
 
     struct Message {
-        Elements           d_appMessages;
-        bsls::Types::Int64 d_sequenceNumber;
+        Elements            d_appMessages;
+        bsls::Types::Uint64 d_sequenceNumber;
 
-        explicit Message(bsls::Types::Int64 sequenceNumber);
+        explicit Message(bsls::Types::Uint64 sequenceNumber);
 
         /// Return number of Elements in the list
         unsigned int numElements() const;
@@ -221,7 +221,7 @@ class PushStream {
 
     bsl::shared_ptr<bdlma::ConcurrentPool> d_pushElementsPool_sp;
 
-    bsls::Types::Int64 d_nextSequenceNumber;
+    bsls::Types::Uint64 d_nextSequenceNumber;
 
     // CREATORS
     /// @brief Construct this object.
@@ -269,7 +269,7 @@ class PushStream {
                  const Apps::iterator& iteratorApp);
 
     /// Generate and return unique, monotonically increasing integer
-    bsls::Types::Int64 nextSequenceNumber();
+    bsls::Types::Uint64 nextSequenceNumber();
 };
 
 // ========================
@@ -641,7 +641,7 @@ inline unsigned int PushStream::Elements::numElements() const
     return d_numElements;
 }
 
-inline PushStream::Message::Message(bsls::Types::Int64 sequenceNumber)
+inline PushStream::Message::Message(bsls::Types::Uint64 sequenceNumber)
 : d_appMessages()
 , d_sequenceNumber(sequenceNumber)
 {
@@ -696,7 +696,7 @@ inline const PushStream::Element* PushStream::App::last() const
 // struct PushStream
 // -----------------
 
-inline bsls::Types::Int64 PushStream::nextSequenceNumber()
+inline bsls::Types::Uint64 PushStream::nextSequenceNumber()
 {
     return ++d_nextSequenceNumber;
 }
--- a/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
@@ -178,21 +178,30 @@ inline bsl::ostream& operator<<(bsl::ostream& stream, const Event& event)
 
 struct VirtualIterator : mqbblp::QueueEngineUtil_AppState::VirtualIterator {
     mqbblp::PushStreamIterator* d_start_p;
-    mqbblp::PushStreamIterator* d_stop_p;
+    const bsls::Types::Uint64   d_stop;
     bool                        d_doAdvance;
 
     VirtualIterator(mqbblp::PushStreamIterator* start,
                     mqbblp::PushStreamIterator* stop)
     : d_start_p(start)
-    , d_stop_p(stop)
+    , d_stop((stop && !stop->atEnd()) ? stop->sequenceNumber() : 0)
     , d_doAdvance(false)
     {
         // PRECONDITIONS
         BSLS_ASSERT_SAFE(d_start_p);
-        BSLS_ASSERT_SAFE(d_stop_p);
     }
+    ~VirtualIterator() BSLS_KEYWORD_OVERRIDE;
     const mqbi::StorageIterator* next() BSLS_KEYWORD_OVERRIDE;
 };
+
+VirtualIterator::~VirtualIterator()
+{
+    // `start` might keep a shared pointer to a memory mapped file area, and
+    // this prevents file set from closing possibly for a very long time.
+    // Make sure to invalidate any cached data within this iterator after use.
+    // TODO: refactor iterators to remove cached data.
+    d_start_p->clearCache();
+}
 
 const mqbi::StorageIterator* VirtualIterator::next()
 {
@@ -207,8 +216,8 @@ const mqbi::StorageIterator* VirtualIterator::next()
         return 0;
     }
 
-    if (!d_stop_p->atEnd()) {
-        if (d_start_p->sequenceNumber() >= d_stop_p->sequenceNumber()) {
+    if (d_stop) {
+        if (d_start_p->sequenceNumber() >= d_stop) {
             return 0;
         }
     }
--- a/src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
@@ -78,18 +78,17 @@ const bsls::Types::Int64 k_NS_PER_MESSAGE =
 // Time interval between messages logged with throttling.
 
 struct VirtualIterator : mqbblp::QueueEngineUtil_AppState::VirtualIterator {
-    mqbi::StorageIterator* d_start_p;
-    mqbi::StorageIterator* d_stop_p;
-    bool                   d_doAdvance;
+    mqbi::StorageIterator*  d_start_p;
+    const bmqt::MessageGUID d_stop;
+    bool                    d_doAdvance;
 
     VirtualIterator(mqbi::StorageIterator* start, mqbi::StorageIterator* stop)
     : d_start_p(start)
-    , d_stop_p(stop)
+    , d_stop((stop && !stop->atEnd()) ? stop->guid() : bmqt::MessageGUID())
     , d_doAdvance(false)
     {
         // PRECONDITIONS
         BSLS_ASSERT_SAFE(d_start_p);
-        BSLS_ASSERT_SAFE(d_stop_p);
     }
     ~VirtualIterator() BSLS_KEYWORD_OVERRIDE;
     const mqbi::StorageIterator* next() BSLS_KEYWORD_OVERRIDE;
@@ -108,8 +107,8 @@ const mqbi::StorageIterator* VirtualIterator::next()
         return 0;
     }
 
-    if (!d_stop_p->atEnd()) {
-        if (d_start_p->guid() == d_stop_p->guid()) {
+    if (!d_stop.isUnset()) {
+        if (d_start_p->guid() == d_stop) {
             return 0;
         }
     }


src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -901,20 +901,20 @@ QueueEngineUtil_AppState::~QueueEngineUtil_AppState()
...
 
-size_t QueueEngineUtil_AppState::catchUp(bsls::TimeInterval*          delay,
-                                         mqbi::StorageIterator*       reader,
-                                         mqbi::StorageIterator*       start,
-                                         const mqbi::StorageIterator* end)
+size_t QueueEngineUtil_AppState::catchUp(
+    bsls::TimeInterval*                                          delay,
+    mqbi::StorageIterator*                                       reader,
+    mqbi::StorageIterator*                                       start,
+    const bdlb::Variant<bsls::Types::Uint64, bmqt::MessageGUID>& stop)
```
678098: As discussed with @dorjesinpo, I will try to find a way to provide `stop` without building a `Variant` object. Template arg might work well, however, there might be cleaner ways to refactor the code

src/groups/mqb/mqbblp/mqbblp_pushstream.cpp
```diff
@@ -245,25 +253,43 @@ bool PushStreamIterator::hasReceipt() const
...
+    const bdlb::Variant<bsls::Types::Uint64, bmqt::MessageGUID>& stop) const
+{
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(!atEnd());
+
+    BSLS_ASSERT_SAFE(stop.is<bsls::Types::Uint64>());
+
+    bsls::Types::Uint64 value = stop.the<bsls::Types::Uint64>();
+
+    return value && sequenceNumber() >= value;
```
678098:
```suggestion
    return value && value <= sequenceNumber();
```
Would prefer left-to-right comparisons

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.t.cpp
```diff
@@ -1906,6 +1906,125 @@ static void test21_deliverOutOfOrder()
...
+//  It starts from 5 but since 5 is past the common iterator, there is no push.
+//  Then trigger 'processAppRedelivery' for "b".
+//  It delivers 3 and stops at 4.  It then attempts to deliver all Apps which
+//  results in "b" receiving 4 and then both "a" and "b" receiving 5 and 6.
+//
+// Testing:
+//   'processAppRedelivery'.
+// ------------------------------------------------------------------------
+{
+    bmqtst::TestHelper::printTestName("ROUND-ROBIN AND REDELIVERY");
```
678098: The test name is not relevant, it looks like a copy-paste problem

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -627,6 +641,30 @@ inline unsigned int PushStream::Elements::numElements() const
...
+}
+
+inline bsls::Types::Uint64 PushStream::Message::sequenceNumber() const
+{
+    return d_sequenceNumber;
+}
+
+inline bsls::Types::Int64 PushStream::Message::nextSequenceNumber()
+{
+    static bsls::Types::Int64 s_value = 0;
```
678098: 1. This is shared across all apps/queues/etc. Note that this type is not atomic, and it is read/modified from different threads
2. signed64 type, Variant has unsigned64.
3. For safe multithreaded initialization, has to do BSLMT_ONCE_DO

https://github.com/bloomberg/blazingmq/blob/9b7ddca44575d8d2d01bac64ec51303ab2c891ee/src/groups/bmq/bmqt/bmqt_subscription.cpp#L45-L57

678098 (changes requested):
Counter thing introduces a multi-threading bug

src/groups/mqb/mqbblp/mqbblp_pushstream.cpp
```diff
@@ -41,6 +41,7 @@ PushStream::PushStream(
           : bsl::allocate_shared<bdlma::ConcurrentPool>(allocator,
                                                         sizeof(Element),
                                                         allocator))
+, s_nextSequenceNumber(0)
```
678098:
```suggestion
, d_nextSequenceNumber(0)
```

src/groups/mqb/mqbblp/mqbblp_pushstream.cpp
```diff
@@ -121,17 +133,14 @@ void PushStreamIterator::removeCurrentElement()
     d_owner_p->remove(del, false);
     // cannot erase the GUID because of the d_iterator
 
-    if (d_iterator->second.numElements() == 0) {
+    if (message().numElements() == 0) {
```
678098, line 134: `message()` accessor now checks that `!atEnd()`, the same check happens at line 124. It means that the same thing is being checked twice during `removeCurrentElement` call

src/groups/mqb/mqbblp/mqbblp_pushstream.cpp
```diff
@@ -147,7 +156,7 @@ bool PushStreamIterator::advance()
 
     clearCache();
 
-    if (d_iterator->second.numElements() == 0) {
+    if (message().numElements() == 0) {
```
678098, line 157: `!atEnd` is checked twice now during `advance` call

src/groups/mqb/mqbblp/mqbblp_pushstream.cpp
```diff
@@ -191,7 +200,7 @@ PushStream::Element* PushStreamIterator::element(unsigned int appOrdinal) const
 
     if (d_currentOrdinal > appOrdinal) {
         d_currentOrdinal = 0;
-        d_currentElement = d_iterator->second.front();
+        d_currentElement = message().d_appMessages.front();
```
678098, line 201: `!atEnd` is checked twice now in this function

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -209,6 +221,8 @@ class PushStream {
 
     bsl::shared_ptr<bdlma::ConcurrentPool> d_pushElementsPool_sp;
 
+    bsls::Types::Int64 s_nextSequenceNumber;
```
678098:
```suggestion
    bsls::Types::Int64 d_nextSequenceNumber;
```

src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -941,25 +925,20 @@ size_t QueueEngineUtil_AppState::catchUp(bsls::TimeInterval*          delay,
...
-        if (!end->atEnd()) {
-            if (start->guid() == end->guid()) {
-                // Deliver the rest by 'QueueEngineUtil_AppsDeliveryContext'
-                break;
-            }
-        }
 
+    size_t                       numMessages = 0;
+    const mqbi::StorageIterator* current     = 0;
+    while (BSLS_PERFORMANCEHINT_PREDICT_LIKELY(current = start->next())) {
```
678098: I am not actually sure if we can use performance hints at all in this context. It is a normal situation when where is no `next` element.
Also, it is normal if we reach the state `result == Routers::e_NO_CAPACITY_ALL`.

src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -1122,6 +1095,13 @@ QueueEngineUtil_AppState::processDeliveryLists(bsls::TimeInterval*    delay,
         // The only excuse for stopping the iteration is poisonous message
         numMessages += processDeliveryList(delay, reader, d_putAsideList);
     }
+
+    // `reader` might keep a shared pointer to a memory mapped file area, and
+    // this prevents file set from closing possibly for a very long time.
+    // Make sure to invalidate any cached data within this iterator after use.
+    // TODO: refactor iterators to remove cached data.
+    reader->clearCache();
```
678098, line 1115: Why cache clearing is moved here? Don't we need to clear cache in the end of `catchUp` anymore?

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -176,6 +176,44 @@ inline bsl::ostream& operator<<(bsl::ostream& stream, const Event& event)
     return stream;
 }
 
+struct VirtualIterator : mqbblp::QueueEngineUtil_AppState::VirtualIterator {
+    mqbblp::PushStreamIterator* d_start;
+    mqbblp::PushStreamIterator* d_stop;
```
```suggestion
    mqbblp::PushStreamIterator* d_start_p;
    mqbblp::PushStreamIterator* d_stop_p;
```

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -671,38 +709,34 @@ void RelayQueueEngine::processAppRedelivery(unsigned int upstreamSubQueueId,
...
-                 start,
-                 d_storageIter_mp.get());
+        // Always start with the first element in the App and always stop at
+        // the d_storageIter_mp (the start of all Apps processing).
+
+        start.load(new (*d_allocator_p)
+                       VirtualPushStreamIterator(upstreamSubQueueId,
+                                                 storage(),
+                                                 &d_pushStream),
+                   d_allocator_p);
```
678098: I think we don't actually need to allocate it in heap

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -77,6 +77,53 @@ const bsls::Types::Int64 k_NS_PER_MESSAGE =
     bdlt::TimeUnitRatio::k_NANOSECONDS_PER_MINUTE / k_MAX_INSTANT_MESSAGES;
 // Time interval between messages logged with throttling.
 
+struct VirtualIterator : mqbblp::QueueEngineUtil_AppState::VirtualIterator {
+    mqbi::StorageIterator* d_start;
+    mqbi::StorageIterator* d_stop;
```
```suggestion
    mqbi::StorageIterator* d_start_p;
    mqbi::StorageIterator* d_stop_p;
```

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -176,6 +176,44 @@ inline bsl::ostream& operator<<(bsl::ostream& stream, const Event& event)
...
+    mqbblp::PushStreamIterator* d_stop;
+    bool                        d_doAdvance;
+
+    VirtualIterator(mqbblp::PushStreamIterator* start,
+                    mqbblp::PushStreamIterator* stop)
+    : d_start(start)
+    , d_stop(stop)
+    , d_doAdvance(false)
+    {
+        // NOTHING
```
```suggestion
        // PRECONDITIONS
        BSLS_ASSERT_SAFE(d_start_p);
        BSLS_ASSERT_SAFE(d_end_p);
```

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -77,6 +77,53 @@ const bsls::Types::Int64 k_NS_PER_MESSAGE =
...
+    bool                   d_doAdvance;
+
+    VirtualIterator(mqbi::StorageIterator* start, mqbi::StorageIterator* stop)
+    : d_start(start)
+    , d_stop(stop)
+    , d_doAdvance(false)
+    {
+        // NOTHING
```
```suggestion
        // PRECONDITIONS
        BSLS_ASSERT_SAFE(d_start_p);
        BSLS_ASSERT_SAFE(d_end_p);
```

src/groups/mqb/mqbblp/mqbblp_pushstream.cpp
dorjesinpo: Correct.  We can remove the assert in `message()` if we want to.  Everywhere it is called, there is the same (duplicate) check prior to the call.  But, it looks safer to keep the assert.  If we want optimization, we can do something about asserts maybe.

src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
dorjesinpo: We do, in `VirtualIterator::~VirtualIterator()`

src/groups/mqb/mqbblp/mqbblp_pushstream.h
```diff
@@ -130,6 +130,18 @@ class PushStream {
...
+    struct Message {
+        Elements           d_appMessages;
+        bsls::Types::Int64 d_sequenceNumber;
+
+        explicit Message(bsls::Types::Int64 sequenceNumber);
+
+        /// Return number of Elements in the list
+        unsigned int numElements() const;
+
+        bsls::Types::Uint64 sequenceNumber() const;
```
678098, line 142: Is there a reason to keep data signed and accessor unsigned? Let's keep both unsigned?

src/groups/mqb/mqbblp/mqbblp_pushstream.cpp
678098: `const PushStream::Message& message() const` is a protected API, we fully control how to use it from within this class. All its usages are visible and easy to review.
I think it's worth to add checks where we use the class from external code  but it's not worth to check any action in internal API, if we have trust in object's state due to the checks that were already performed.

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -176,6 +176,46 @@ inline bsl::ostream& operator<<(bsl::ostream& stream, const Event& event)
...
+    }
+    else {
+        d_doAdvance = true;
+    }
+
+    if (d_start_p->atEnd()) {
+        return 0;
+    }
+
+    if (!d_stop_p->atEnd()) {
```
678098: We only advance `d_start_p`. This means that the result of `d_stop_p->atEnd()` never changes unless we pass a pointer to the same `mqbi::StorageIterator`:

```
VirtualIterator iter(start_p, start_p);
```

In this case we will `return 0` on line 206 and never reach the check `if (!d_stop_p->atEnd()) {`


It is possible to compute `d_stop_p->sequenceNumber()` just once in constructor:

```
    const bsls::Types::Uint64 d_stopSequenceNumber;

    VirtualIterator(mqbblp::PushStreamIterator* start,
                    mqbblp::PushStreamIterator* stop)
    : d_start_p(start)
    , d_stopSequenceNumber((stop && !stop->atEnd()) ? stop->sequenceNumber() : bsl::numeric_limits<bsls::Types::Uint64>::max())
    , d_doAdvance(false)
    {
        // PRECONDITIONS
        BSLS_ASSERT_SAFE(d_start_p);
    }
```

And simplify the condition:

```
if (d_stopSequenceNumber <= d_start_p->sequenceNumber()) {
    return 0;
}
```

Note that when we call `next()` multiple times, we always check 2 if conditions, after this change we will only check 1.

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -77,6 +77,55 @@ const bsls::Types::Int64 k_NS_PER_MESSAGE =
...
+    else {
+        d_doAdvance = true;
+    }
+
+    if (d_start_p->atEnd()) {
+        return 0;
+    }
+
+    if (!d_stop_p->atEnd()) {
+        if (d_start_p->guid() == d_stop_p->guid()) {
```
678098: Same here. When we call `next()` multiple times we always check these 2 conditions. We can just cache `MessageGuid` and check one:

```
if (d_start_p->guid() == d_stopGuid) {
    return 0;
}
```

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -77,6 +77,55 @@ const bsls::Types::Int64 k_NS_PER_MESSAGE =
...
+    return d_start_p;
+}
+
+VirtualIterator::~VirtualIterator()
+{
+    // `start` might keep a shared pointer to a memory mapped file area, and
+    // this prevents file set from closing possibly for a very long time.
+    // Make sure to invalidate any cached data within this iterator after use.
+    // TODO: refactor iterators to remove cached data.
+    d_start_p->clearCache();
```
678098, line 125: Don't we need to do the same for `RelayQueueEngine::VirtualIterator`?

PR #833, opened 2025-07-15
Performance[MQB]: Inline Put, Push, Ack, Confirm

dorjesinpo:
duplicate of https://github.com/bloomberg/blazingmq/pull/177

PR #1049, opened 2025-12-12
Refactor[MQB]: detach stats from cluster thread

dorjesinpo:
This is part of [https://github.com/bloomberg/blazingmq/pull/833](https://github.com/bloomberg/blazingmq/pull/833), separated to ease the review.

The main idea is to make stats updates possible bypassing cluster thread.


src/groups/mqb/mqbstat/mqbstat_clusterstats.h
```diff
@@ -393,115 +393,6 @@ class PartitionStats {
...
-    /// `clusterNodesStatContext`, using the specified `allocator`.
-    void initialize(const bmqt::Uri&    uri,
-                    bmqst::StatContext* clusterNodesStatContext,
-                    bslma::Allocator*   allocator);
-
-    /// Update statistics for the event of the specified `type` and with the
-    /// specified `value` (depending on the `type`, `value` can represent
-    /// the number of bytes, a counter, ...
-    template <EventType::Enum type>
-    void onEvent(bsls::Types::Int64 value);
```
678098, line 499: The templated version was introduced to avoid passing the enum value we know at compile time

PR #1052, opened 2025-12-16
Fix[mqb]: isFinal race when dropping handle

dorjesinpo:
1.
When dropping `QueueHandle`, `mqbblp::Queue` sends deconfigure request for each subStream followed by close request.
After [https://github.com/bloomberg/blazingmq/pull/1002](https://github.com/bloomberg/blazingmq/pull/1002), queue needs to wait for deconfigure response before sending close request

2.  
Close requests have `isFinal` parameter which enforces zeroing out all counters even if Close request counters do not match existing one.  (we do not use `isFinal` inside clusters, there is some use in the SDK).

3.  
Producers do not need (de)configuration, and attempt to deconfigure results in an immediate response.

There is a race - if a producer is the last subStream, its close request has `isFinal` as `true`.  But, since it is short circuited, the producer close could get processed before the previous consumer close.   After `RelayQueueEngine` sees `isFinal`, it ignores subsequent consumer close resulting in failure to redeliver all pending data for this consumer.

Multiple runs of `test_dead_replica` can expose this race.

Surprisingly, unconditionally using `false` as `isFinal` did not pass fuzz testing.  Switching to plan B, using a counter for the decision on `isFinal` value.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_queue.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_queue.cpp
@@ -60,12 +60,38 @@ namespace mqbblp {
 
 namespace {
 
-void onHandleDeconfigured(const bmqp_ctrlmsg::Status&,
-                          const bmqp_ctrlmsg::StreamParameters&,
-                          const bsl::function<void(void)>& cb)
+struct Counter {
+    bsls::AtomicUint d_count;
+
+    explicit Counter(unsigned count)
+    : d_count(count)
+    {
+        // NOTHING
+    }
+    unsigned decrement()
+    {
+        BSLS_ASSERT_SAFE(d_count.load());
+        return d_count.subtract(1);
+    }
+};
+
+void onHandleDeconfigured(
+    const bmqp_ctrlmsg::Status&,
+    const bmqp_ctrlmsg::StreamParameters&,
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqbblp/mqbblp_queue.cpp
```diff
@@ -60,12 +60,38 @@ namespace mqbblp {
 
 namespace {
 
-void onHandleDeconfigured(const bmqp_ctrlmsg::Status&,
-                          const bmqp_ctrlmsg::StreamParameters&,
-                          const bsl::function<void(void)>& cb)
+struct Counter {
+    bsls::AtomicUint d_count;
```
678098, line 64: What about using `bsl::shared_ptr<bsls::AtomicUint>` directly?
I think that defining a wrapper class increases complexity when it's not really needed.

dorjesinpo: > What about using `bsl::shared_ptr<bsls::AtomicUint>` directly? I think that defining a wrapper class increases complexity when it's not really needed.

No strong opinion.  I though, separate type makes the purpose clearer which justifiers few extra lines of code.  Of course, `bsl::shared_ptr<bsls::AtomicUint>` works.

PR #1054, opened 2025-12-17
ci: Teach CI to build C++03 deps

pniedzielski:
This pull request teaches our CI to build bmqbrkr and bmqtool in C++03, which is a supported configuration for them.

Some notes about this PR:
  - The `configure` script from the version of ntf-core we were targeting will overwrite whatever UFID we pass it on Linux to compile with C++20.  https://github.com/bloomberg/ntf-core/pull/353 changes this behavior and is necessary for this to work.  So, we need to bump the version of ntf we build against.
  - There are two commits that should not be merged in.  They're titled `chore[DO NOT MERGE]`.
    - [x] ~The first temporarily uses a tag in my personal fork of ntf-core that includes the changes in https://github.com/bloomberg/ntf-core/pull/353.   Once that PR is merged, we should remove this commit.~ **This has been merged.  In its place, there's a commit that bumps the version of NTF we use.**
    - [x] ~The second is a line of code that is a syntax error in C++03 but not in C++11 or later.  This commit just serves to prove that we are building BlazingMQ in C++03, and causes only the C++03 job to fail.~. **This has been removed.  I verified that only the C++03 job failed, and that the failure was due to this line of code.**
  - There's one real C++03 fix in this PR as well, a conversion to `shared_ptr` that seems to work fine on Solaris but not Linux+GCC.

### DIFF
--- a/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
@@ -1717,7 +1717,7 @@ TCPSessionFactory::PortManager::addChannelContext(bmqst::StatContext* parent,
         bmqst::StatContextConfiguration portConfig(
             static_cast<bsls::Types::Int64>(port),
             &localAllocator);
-        bsl::shared_ptr<bmqst::StatContext> portStatContext =
+        bslma::ManagedPtr<bmqst::StatContext> portStatContext =
             parent->addSubcontext(
                 portConfig.storeExpiredSubcontextValues(true));
         channelStatContext      = portStatContext->addSubcontext(statConfig);


pniedzielski:
> ```
> Run [[ false ]] && CPP_VERSION_TAG="cpp03-" || CPP_VERSION_TAG=""
>   [[ false ]] && CPP_VERSION_TAG="cpp03-" || CPP_VERSION_TAG=""
>   BUILD_DEPS_SHA=$(cat docker/build_deps.sh | shasum)
>   echo "cache_key=deps-$CPP_VERSION_TAG$BUILD_DEPS_SHA" >> $GITHUB_OUTPUT
> 0s
> Run [[ false ]] && CPP_VERSION_TAG="cpp03-" || CPP_VERSION_TAG=""
>   [[ false ]] && CPP_VERSION_TAG="cpp03-" || CPP_VERSION_TAG=""
>   BUILD_DEPS_SHA=$(cat docker/build_deps.sh | shasum)
>   echo "cache_key=deps-$CPP_VERSION_TAG$BUILD_DEPS_SHA" >> $GITHUB_OUTPUT
>   shell: /usr/bin/bash -e {0}
> 1s
> Run actions/cache/restore@v4
> Cache hit for: deps-cpp03-1ef1fd1307a7392d2465bae7f687eef2aead9067  -
> Lookup only - skipping download
> Cache found and can be restored from key: deps-cpp03-1ef1fd1307a7392d2465bae7f687eef2aead9067  -
> 1s
> Run actions/cache/restore@v4
> Cache hit for: deps-cpp03-1ef1fd1307a7392d2465bae7f687eef2aead9067  -
> Lookup only - skipping download
> Cache found and can be restored from key: deps-cpp03-1ef1fd1307a7392d2465bae7f687eef2aead9067  -
> Run actions/cache/restore@v4
> Cache hit for: deps-cpp03-1ef1fd1307a7392d2465bae7f687eef2aead9067  -
> Lookup only - skipping download
> Cache found and can be restored from key: deps-cpp03-1ef1fd1307a7392d2465bae7f687eef2aead9067  -
> ```
> 
> PR is using c++03 deps cache entry even in non c++03 build, let me see what's up.

Extra `[[ ... ]]` around `false`

.github/workflows/dependencies.yaml
```diff
@@ -27,7 +33,9 @@ jobs:
       - name: Generate dependencies cache key
         id: get-cache-key
         run: |
-          echo "cache_key=deps-`date +%y-%m-%d`-`cat docker/build_deps.sh | shasum`" >> $GITHUB_OUTPUT
+          ${{ inputs.cpp03 }} && CPP_VERSION_TAG="cpp03-" || CPP_VERSION_TAG=""
+          BUILD_DEPS_SHA=$(cat docker/build_deps.sh | shasum)
+          echo "cache_key=deps-$CPP_VERSION_TAG$BUILD_DEPS_SHA" >> $GITHUB_OUTPUT
```
678098: I think this cache key is one per `build` workflow run

.github/workflows/build-ubuntu.yaml
```diff
@@ -105,11 +113,29 @@ jobs:
       - name: Install cached non packaged dependencies
         if: steps.build-cache-restore-step.outputs.cache-hit != 'true' # Variable type is string, thus using quotes
         working-directory: deps
-        run: ../docker/build_deps.sh
+        run: |
+          ${{ inputs.cpp03 }} ]] && ../docker/build_deps.sh --cpp03 || ../docker/build_deps.sh
```
678098:
```suggestion
          [[ ${{ inputs.cpp03 }} ]] && ../docker/build_deps.sh --cpp03 || ../docker/build_deps.sh
```
Don't understand how it worked before

.github/workflows/build-ubuntu.yaml
```diff
@@ -32,6 +32,11 @@ on:
         required: false
         type: string
         default: "build-ubuntu"
+      cpp03:
```
678098: What about adding a `string` property `cxx_standard` instead and explicitly pass `cpp03` or `cpp17` (or what version do we build here)?
This allows to normalize the code and remove pattern `flag && A || B` like `[[ ${{ inputs.cpp03 }} ]] && CPP_VERSION_TAG="cpp03-" || CPP_VERSION_TAG=""`

.github/workflows/build-ubuntu.yaml
```diff
@@ -105,11 +113,29 @@ jobs:
...
+        if: inputs.cpp03 == true && steps.build-cache-restore-step.outputs.cache-hit != 'true' # Variable type is string, thus using quotes
+        env:
+          PKG_CONFIG_PATH: /usr/lib/x86_64-linux-gnu/pkgconfig:/opt/bb/lib64/pkgconfig
+        run: |
+          cmake -S . -B build/blazingmq -G Ninja \
+            -DCMAKE_TOOLCHAIN_FILE=${{ github.workspace }}/deps/srcs/bde-tools/BdeBuildSystem/toolchains/linux/gcc-default.cmake \
+            -DCMAKE_BUILD_TYPE=Debug \
+            -DBDE_BUILD_TARGET_SAFE=ON \
+            -DBDE_BUILD_TARGET_64=ON \
+            -DBDE_BUILD_TARGET_CPP03=ON \
```
678098: We don't provide `configure` script with `ufid` but can support something similar and simple like standard configuration if we update `CMakeLists.txt`

.github/workflows/dependencies.yaml
```diff
@@ -2,6 +2,12 @@ name: Dependencies
 
 on:
   workflow_call:
+    inputs:
+      cpp03:
```
678098: Can also have `cxx_standard`

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -1599,7 +1599,7 @@ TCPSessionFactory::PortManager::addChannelContext(bmqst::StatContext* parent,
         bmqst::StatContextConfiguration portConfig(
             static_cast<bsls::Types::Int64>(port),
             &localAllocator);
-        bsl::shared_ptr<bmqst::StatContext> portStatContext =
+        bslma::ManagedPtr<bmqst::StatContext> portStatContext =
```
678098, line 1720: Is this change required for cpp03?
Changing `shared_ptr` to `ManagedPtr` looks like a step back in this case.

Somehow we are able to build a `shared_ptr` later when we construct a `PortContext`

```
        struct PortContext {
            bsl::shared_ptr<bmqst::StatContext> d_portContext;
            bsl::size_t                         d_numChannels;
        };
```

.github/workflows/build-ubuntu.yaml
pniedzielski: Hm not sure how it worked, but crucially we don't want `[[ ... ]]`.  `inputs.cpp03` expands to `true` or `false`, and *both* `[[ true ]]; echo $?` and `[[ false ]]; echo $?` evaluate to `0`.

.github/workflows/build-ubuntu.yaml
pniedzielski: Good idea, I'll go this route.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
pniedzielski: It is required for C++03 unfortunately.  GCC is stricter about this than the Solaris compiler.

Here's [the relevant failure](https://github.com/bloomberg/blazingmq/actions/runs/22238399421/job/64335734890?pr=1054), truncated down a lot:
```
/home/runner/work/blazingmq/blazingmq/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp: In member function ‘BloombergLP::bslma::ManagedPtr<BloombergLP::bmqst::StatContext> BloombergLP::mqbnet::TCPSessionFactory::PortManager::addChannelContext(BloombergLP::bmqst::StatContext*, const bsl::string&, uint16_t)’:
Error: /home/runner/work/blazingmq/blazingmq/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp:1722:62: error: no matching function for call to ‘BloombergLP::bslma::ManagedPtr<BloombergLP::bmqst::StatContext>::ManagedPtr(BloombergLP::bslma::ManagedPtr<BloombergLP::bmqst::StatContext>)’
 1722 |                 portConfig.storeExpiredSubcontextValues(true));
      |                                                              ^
...
/opt/bb/include/bslma_managedptr_cpp03.h:1837:1: note:   conversion of argument 1 would be ill-formed:
Error: /home/runner/work/blazingmq/blazingmq/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp:1721:34: error: cannot bind non-const lvalue reference of type ‘BloombergLP::bslma::ManagedPtr<BloombergLP::bmqst::StatContext>&’ to an rvalue of type ‘BloombergLP::bslma::ManagedPtr<BloombergLP::bmqst::StatContext>’
 1721 |             parent->addSubcontext(
      |             ~~~~~~~~~~~~~~~~~~~~~^
 1722 |                 portConfig.storeExpiredSubcontextValues(true));
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

So we're returning a rvalue `bslma::ManagedPtr<bmqst::StatContext>` from `addSubcontext` and trying to use that to construct a `bsl::shared_ptr<bmqst::StatContext>` using this constructor:

```cpp
    template <class CONVERTIBLE_TYPE
              BSLSTL_SHAREDPTR_DECLARE_IF_CONVERTIBLE>
    shared_ptr(
         BloombergLP::bslma::ManagedPtr<CONVERTIBLE_TYPE>  managedPtr,
         BloombergLP::bslma::Allocator                    *basicAllocator = 0);
                                                                    // IMPLICIT
```

Calling this constructor needs to invoke `bslma::ManagedPtr<bmqst::StatContext>`'s copy constructor (which doesn't exist).  Compiler tries to use the constructor that takes a non-`const` `bslma::ManagedPtr&`, but we can't convert an rvalue to a non-`const` lvalue ref.

After this patch, you're right, we do construct a `shared_ptr` later from the same `ManagedPtr`, but crucially then we're constructing the `shared_ptr` out of an lvalue `ManagedPtr`, to which we can take a non-`const` `ManagedPtr&`.

With move semantics in C++11 this works just fine of course, and somehow Solaris compiler does not care.  But I think GCC is right here.

pniedzielski: I'll add the above to the commit message for future `git log`ers.

docker/build_deps.sh
```diff
@@ -43,13 +75,13 @@ fetch_git() {
 fetch_deps() {
     fetch_git bloomberg bde-tools 4.28.0.0
     fetch_git bloomberg bde 4.28.0.0
-    fetch_git bloomberg ntf-core 2.4.2
+    fetch_git pniedzielski ntf-core tmp-tag
```
678098:
```suggestion
    fetch_git bloomberg ntf-core 2.6.10
```

678098 (commented):
Have only one comment

docker/build_deps.sh
pniedzielski: Oh shoot, thanks for catching that

pniedzielski:
@678098 back to you one more time

678098:
@pniedzielski thank you

PR #1061, opened 2025-12-24
Perf[MQB]: optimize `inDispatcherThread`

678098:
Commit 1:

1. Replace `bslmt::ThreadUtil::Handle` usage with `bslmt::ThreadUtil::Id`. According to docs, `Handle` doesn't allow usage from a thread different from the one it corresponds to. `Id` allows this.
2. Keep a copy of the assigned dispatcher thread id in `mqbi::DispatcherClientData`. This allows to simplify lookups for `inDispatcherThread` checks.
3. Move `mqbi::Dispatcher::inDispatcherThread(int)` -> `mqbi::DispatcherClient::inDispatcherThread(void)`. The previous declaration was abstract and it was not possible to inline `inDispatcherThread`, now it is possible to inline it.
4. Simplify `inDispatcherThread` calls. Before: `BSLS_ASSERT_SAFE(dispatcher()->inDispatcherThread(this));`, after: `BSLS_ASSERT_SAFE(inDispatcherThread());`

Commit 2:

1. Move `d_threadId`: `mqbi::DispatcherClientData::d_threadId` -> `mqbi::DispatcherClient::d_threadId`. This allows to simplify calls in `inDispatcherThread()` and make it even faster.

# Benchmarks

## Darwin M2 - RelWithDebInfo

**Before**: 4.1 ns per iteration `BSLS_ASSERT_OPT(dispatcher_p->inDispatcherThread(client_p));`
**After (commit 1)**: 2.7 ns per iteration `BSLS_ASSERT_OPT(client_p->inDispatcherThread());`
**After (commit 2)**: 2.0 ns per iteration `BSLS_ASSERT_OPT(client_p->inDispatcherThread());`

## Linux amd64 - RelWithDebInfo

**Before**: 4.6 ns per iteration `BSLS_ASSERT_OPT(dispatcher_p->inDispatcherThread(client_p));`
**After (commit 1)**: 3.4 ns per iteration `BSLS_ASSERT_OPT(client_p->inDispatcherThread());`
**After (commit 2)**: 0.6 ns per iteration `BSLS_ASSERT_OPT(client_p->inDispatcherThread());`

Note that the speedup on M2 Darwin is 2x, on amd64 Linux is 7.5x.

### DIFF
--- a/src/groups/mqb/mqbi/mqbi_dispatcher.h
+++ b/src/groups/mqb/mqbi/mqbi_dispatcher.h
@@ -1124,9 +1124,6 @@ class DispatcherClientData {
     /// The dispatcher associated with the client.
     Dispatcher* d_dispatcher_p;
 
-    /// The id of the thread this dispatcher client is assigned to.
-    bslmt::ThreadUtil::Id d_threadId;
-
     /// The flag indicating whether the dispatcher have added the corresponding
     /// client to its internal flush list -- this is a Dispatcher internal
     /// member that should only be manipulated by the dispatcher, and not the
@@ -1149,8 +1146,6 @@ class DispatcherClientData {
     /// reference offering modifiable access to this object.
     DispatcherClientData& setDispatcher(Dispatcher* value);
 
-    DispatcherClientData& setThreadId(bslmt::ThreadUtil::Id threadId);
-
     /// Return a pointer to the dispatcher associated with this object; or
     /// null is this client is not (yet) registered to a dispatcher.
     Dispatcher* dispatcher();
@@ -1159,7 +1154,6 @@ class DispatcherClientData {
     DispatcherClientType::Enum  clientType() const;
     Dispatcher::ProcessorHandle processorHandle() const;
     bool                        addedToFlushList() const;
-    bslmt::ThreadUtil::Id       threadId() const;
 
     /// Return the value of the corresponding member.
     const Dispatcher* dispatcher() const;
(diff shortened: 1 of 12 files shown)


src/groups/mqb/mqbi/mqbi_dispatcher.h
```diff
@@ -1231,6 +1224,17 @@ class DispatcherClient {
...
+
+    /// Return whether the current thread is the thread this client is
+    /// associated with in dispatcher.
+    inline bool inDispatcherThread() const
+    {
+        // In most cases the following condition should short-circuit on
+        // the first operand:
+        return (dispatcherClientData().threadId() ==
+                bslmt::ThreadUtil::selfId()) ||
+               (dispatcherClientData().threadId() == 0);
```
678098: Just to be sure I verified in my benchmark that `inDispatcherThread()` takes the same time with or without the second part of the expression when the first part is true (short-circuit).

678098: Also `dispatcherClientData()` is a virtual function and makes `inDispatcherThread()` slower than it can be. If we just keep `threadId` as a field in `mqbi::DispatcherClient`, we can address it directly without virtual function call and save time on this check.

678098: > If we just keep `threadId` as a field in `mqbi::DispatcherClient`, we can address it directly without virtual function call and save time on this check.

I applied this in the second commit and updated the benchmarks

Note that amd64 Linux time improved much more:
```
TEST /blazingmq/src/groups/mqb/mqba/mqba_dispatcher.t.cpp CASE -1
inDispatcherThread():
       total: 6.11 ms (10000000 iterations)
    per call: 0 ns
```

PR #1062, opened 2025-12-24
Feat[mqbcfg,mqbnet]: enhance TcpInterfaceListener config with `address` field

alexander-e1off:
Today, the broker listens on all interfaces (on the specified port).
The broker's listeners config  is extended to allow an ipv4 interface to be specified.

For example, it is reasonable to want to configure a local broker/proxy to only listen on localhost and not accept remote connections. Or, a listener may be configured on the machines private IP, not a public IP.

Example of listeners config with new `address` field:
```
"listeners": [
  {
     "name": "myListener1",
     "address": "hostName or IP address, 0.0.0.0 by default",
     "port": 30114
  }
]
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,7 +5,6 @@ on:
     branches:
       - main
       - 'integration/**'
-      - enhance-listener-config
   pull_request:
     branches:
       - main


src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -1400,7 +1400,7 @@ int TCPSessionFactory::listen(const mqbcfg::TcpInterfaceListener& listener,
 
     bdlma::LocalSequentialAllocator<64> localAlloc(d_allocator_p);
     bmqu::MemOutStream                  endpoint(&localAlloc);
-    endpoint << ":" << port;  // Empty hostname, listen from all interfaces
+    endpoint << listener.address() << ":" << listener.port();
```
alexander-e1off, line 1403: TBD: `d_listenContexts` and `d_listeningHandles` maps use `port` as a key (in lines 1385-1387 above). Do we need to support multiple addresses for one port?

678098: I think we should ideally, but we can do it later I think.
The use case is when we have several network interfaces and want to allow connections to a specific port only from a subset of interfaces, and want to forbid it from any other interface.

PR #1064, opened 2025-12-29
Misc mqbc::StorageMgr: Relax assert when process live data

kaikulimu:
[Internal Ticket 182146228]

`StorageUtil::validateStorageEvent` will alarm if source is not primary. That is sufficient. Asserting and crashing the broker is overkill.


pniedzielski (approved):
`StorageUtil::validateStorageEvent` will alarm only when `skipAlarm` is false.  From the looks of it, it's true when `partitionHealthState(partitionId) == PartitionFSM::State::e_UNKNOWN` in `do_processLiveData`.  Okay, very slight behavior change here, but no effect other than noted in PR.

PR #1065, opened 2026-01-02
UT[bmqp]: Add test coverage for PutEventBuilder compression ratio accessor

AneesPatel:
*Issue number of the reported bug or feature request: N/A (proactive test coverage)*

**Describe your changes**

Adds unit test coverage for the `lastPackedMesageCompressionRatio()` accessor in `bmqp::PutEventBuilder` (Case 8 in `bmqp_puteventbuilder.t.cpp`).

This accessor is part of the public API but previously lacked test coverage. Ensuring this metric is accurate is critical for monitoring compression efficiency in production.

**Testing performed**

Added `test8_compressionRatioAccessor` which verifies:
1. Initial value is -1.0 before any message is packed
2. Value is 1.0 when compression is disabled
3. Value returns to -1.0 after `reset()`
4. Value > 1.0 when compression successfully reduces size
5. Value is preserved after a failed `packMessage()` call
6. Builder recovery after failure

**Built and ran locally: `./bmqp_puteventbuilder.t 8`; where all assertions passed.**

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_puteventbuilder.t.cpp
+++ b/src/groups/bmq/bmqp/bmqp_puteventbuilder.t.cpp
@@ -2614,23 +2614,29 @@ static void test8_compressionRatioAccessor()
 // Concerns:
 //   1. 'lastPackedMesageCompressionRatio' returns -1.0 initially and after
 //      'reset()'.
-//   2. Returns 1.0 when compression is disabled or payload is small.
+//   2. Returns 1.0 when compression is disabled.
 //   3. Returns > 1.0 when compression is successful.
-//   4. Preserves previous ratio after a failed 'packMessage' call.
 //
 // Plan:
 //   1. Verify initial state (-1.0).
 //   2. Pack uncompressed message, verify 1.0.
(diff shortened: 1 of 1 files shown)


src/groups/bmq/bmqp/bmqp_puteventbuilder.t.cpp
```diff
@@ -2606,6 +2606,109 @@ static void testN1_decodeFromFile()
...
+
+    obj.startMessage();
+    obj.setMessagePayload(&bigBlob);
+    obj.setMessageGUID(bmqp::MessageGUIDGenerator::testGUID());
+
+    BMQTST_ASSERT_EQ(obj.packMessage(k_QID),
+                     bmqt::EventBuilderResult::e_PAYLOAD_TOO_BIG);
+
+    // Ratio should remain from the last successful pack
+    BMQTST_ASSERT_EQ(obj.lastPackedMesageCompressionRatio(), ratio);
```
678098: `ratio` is a const, the value is the same with line 2683

678098:
Thank you for contribution @AneesPatel!

PR #1070, opened 2026-01-07
Fix[mqb]: queue/client broker stat config order

chrisbeard:
**Describe your changes**
In b536f50e9d43abacb499ee9a927ca84f6eabfa33, I made the enum values consistent which fixed one bug. Namely, these stats now printed correctly in the broker stat log file.

Stat consumer plugins consuming these stats via the getValue() interface worked correctly, no issues there.

However, the positional collection of stat names and the defined enums remained mismatched. A hypothetical stat consumer reading values by name rather than by enum would see the values swapped. This commit fixes this mismatch.


678098 (approved):
Looks good, the IT failure is unrelated

PR #1075, opened 2026-01-12
Feat[mqbc]: fix uncommited advisory gets stuck forever

waldgange:
Fix: After every quorum change, review [the uncommitted csl advisories](https://github.com/bloomberg/blazingmq/blob/eba490bf1da17c9a1c3db84ace9bb21e1835b67f/src/groups/mqb/mqbc/mqbc_incoreclusterstateledger.h#L202) to see if some of them already have more ACKs than the new quorum. It fixes the bug that leads to openQueue requests get stuck forever.
Updated UT and IT

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_incoreclusterstateledger.cpp
+++ b/src/groups/mqb/mqbc/mqbc_incoreclusterstateledger.cpp
@@ -462,17 +462,15 @@ int IncoreClusterStateLedger::applyRecordInternalImpl(
         ,
         rc_CREATE_ACK_FAILURE = -3  // Fail to create leader advisory ack
         ,
-        rc_CREATE_COMMIT_FAILURE = -4  // Fail to create leader advisory
-                                       // commit
+        rc_COMMIT_FAILURE = -4  // Fail to create leader advisory
+                                // commit
         ,
         rc_APPLY_ACK_FAILURE = -5  // Fail to apply leader advisory ack
         ,
-        rc_APPLY_COMMIT_FAILURE = -6  // Fail to apply leader advisory commit
-        ,
-        rc_SEND_ACK_FAILURE = -7  // Fail to send leader advisory ack
+        rc_SEND_ACK_FAILURE = -6  // Fail to send leader advisory ack
                                   // back to leader
         ,
-        rc_ADVISORY_NOT_FOUND = -8  // Advisory not found
+        rc_ADVISORY_NOT_FOUND = -7  // Advisory not found
     };
 
     int rc = rc_UNKNOWN;


src/groups/mqb/mqbc/mqbc_incoreclusterstateledger.cpp
```diff
@@ -350,6 +350,28 @@ int IncoreClusterStateLedger::onLogRolloverCb(const mqbu::StorageKey& oldLogId,
     return rc_SUCCESS;
 }
 
+void IncoreClusterStateLedger::onQuorumChangeCb(const unsigned int ackQuorum)
```
678098:
```suggestion
void IncoreClusterStateLedger::onQuorumChangeCb(unsigned int ackQuorum)
```
Might be simpler

src/groups/mqb/mqbc/mqbc_incoreclusterstateledger.cpp
```diff
@@ -780,8 +767,70 @@ int IncoreClusterStateLedger::applyRecordInternal(
                                    recordType);
 }
 
+int IncoreClusterStateLedger::applyCommit(
+    const bmqp_ctrlmsg::LeaderMessageSequence& sequenceNumber,
+    const unsigned int                         ackQuorum)
```
678098:
```suggestion
    unsigned int                               ackQuorum)
```
Might be simpler

src/groups/mqb/mqbcfg/mqbcfg_clusterquorummanager.h
```diff
@@ -100,6 +120,18 @@ inline void ClusterQuorumManager::setQuorum(unsigned int quorum,
...
+        d_callback(quorum);
+    }
+}
+
+inline void
+ClusterQuorumManager::setCallback(const UpdateQuorumCallback& callback)
+{
+    bslmt::LockGuard<bslmt::Mutex> lock(&d_mutex);
+    d_callback = callback;
 }
```
678098: We don't call `setCallback` and `setQuorum` often, right?
In this case the implementations can be moved to `cpp` file without harming performance.
This allows to move `#include <bslmt_lockguard.h>` to `cpp` file and include less.

src/groups/mqb/mqbcfg/mqbcfg_clusterquorummanager.h
```diff
@@ -100,6 +120,18 @@ inline void ClusterQuorumManager::setQuorum(unsigned int quorum,
     // elected.
 
     d_quorum.store(quorum);
+
+    bslmt::LockGuard<bslmt::Mutex> lock(&d_mutex);
+    if (d_callback != 0) {
+        d_callback(quorum);
+    }
```
678098: I am thinking if `setQuorum` might be called from different threads at the same time. If this is the case, maybe we should also update `d_quorum` under the mutex, to always have a consistent state of `ClusterQuorumManager`?

Example of 2 thread access:
```
T1: start ClusterQuorumManager::setQuorum
T1: d_quorum.store(5)
T2: start ClusterQuorumManager::setQuorum
T2: d_quorum.store(3)
T1: lock d_mutex
T1: execute d_callback(5), at the same time d_quorum contains 3 from T2
```
In the example d_callback is executed when d_quorum is different, so `ClusterQuorumManager` is in inconsistent state

src/groups/mqb/mqbcfg/mqbcfg_clusterquorummanager.h
```diff
@@ -44,10 +47,22 @@ namespace mqbcfg {
 
 /// This class provides a mechanism to manage the quorum for a cluster.
 class ClusterQuorumManager {
+  public:
+    // TYPES
+
+    /// Invoked on every quorum update
+    typedef bsl::function<void(const unsigned int)> UpdateQuorumCallback;
```
678098:
```suggestion
    typedef bsl::function<void(unsigned int)> UpdateQuorumCallback;
```
Might be simpler

src/groups/mqb/mqbcfg/mqbcfg_clusterquorummanager.h
waldgange: I think it's fine as long as we don't access `QuorumManager` from the callback. In the example you described, let's assume the number of nodes is 4. In this case `d_callback(5)` will do nothing and `d_callback(3)` will trigger processing of the uncommitted advisories. And the order of execution doesn't matter. I left `d_quorum` atomic and outside the lock because I didn't want to affect its accessability for readers.
P.S. Now we call `setQuorum` only from the cluster dispatcher thread. And we call `setCallback` before this thread gets started. So technically there is no concurrent access to the callback. But my idea was to not rely on this assumption and keep `QuorumManager` thread safe

waldgange: Alternatively I can make `d_quorum` non-atomic and move it under the mutex. In this case I believe we should use shared mutex, because this variable is modified very rarely but accessed really often

src/groups/mqb/mqbc/mqbc_incoreclusterstateledger.cpp
```diff
@@ -688,42 +710,7 @@ int IncoreClusterStateLedger::applyRecordInternalImpl(
...
-
-            rc = applyRecordInternal(*commitRecord,
-                                     0,
-                                     commitMessage,
-                                     commitAdvisory.sequenceNumber(),
-                                     ClusterStateRecordType::e_COMMIT);
-            if (rc != 0) {
-                return 10 * rc + rc_APPLY_COMMIT_FAILURE;  // RETURN
-            }
+            return applyCommit(ack.sequenceNumberAcked(), ackQuorum);
```
678098:
```suggestion
            rc = applyCommit(ack.sequenceNumberAcked(), ackQuorum);
            if (0 != rc) {
                return 10 * rc + rc_COMMIT_FAILURE;
            }
```
And `return rc_SUCCESS` at the end of this routine on success

src/groups/mqb/mqbc/mqbc_incoreclusterstateledger.cpp
```diff
@@ -780,8 +767,70 @@ int IncoreClusterStateLedger::applyRecordInternal(
...
+{
+    enum RcEnum {
+        // Value for the various RC error categories
+        rc_SUCCESS = 0  // Success
+        ,
+        rc_UNKNOWN = -1  // Unknown result
+        ,
+        rc_CREATE_COMMIT_FAILURE = -4  // Fail to create leader advisory
+        ,
+        rc_APPLY_COMMIT_FAILURE = -6  // Fail to apply leader advisory commit
```
678098:
```suggestion
        rc_CREATE_COMMIT_FAILURE = -1  // Fail to create leader advisory
        ,
        rc_APPLY_COMMIT_FAILURE = -2  // Fail to apply leader advisory commit
```
The numeration can be independent from the caller routine, since we have `10 * rc` anyway

src/groups/mqb/mqbc/mqbc_incoreclusterstateledger.cpp
```diff
@@ -780,8 +767,70 @@ int IncoreClusterStateLedger::applyRecordInternal(
...
+
+    BALL_LOG_INFO << description() << " Quorum of " << ackQuorum
+                  << " acks is achieved for advisory of seqNum "
+                  << sequenceNumber
+                  << ", creating and applying commit advisory: "
+                  << commitMessage << ".";
+
+    int                          rc           = rc_UNKNOWN;
+    bsl::shared_ptr<bdlbb::Blob> commitRecord = d_blobSpPool_p->getObject();
+    rc = ClusterStateLedgerUtil::appendRecord(
```
678098:
```suggestion
    bsl::shared_ptr<bdlbb::Blob> commitRecord = d_blobSpPool_p->getObject();
    int rc = ClusterStateLedgerUtil::appendRecord(
```
To get rid of `rc_UNKNOWN`

src/groups/mqb/mqbc/mqbc_incoreclusterstateledger.h
```diff
@@ -257,10 +261,25 @@ class IncoreClusterStateLedger BSLS_KEYWORD_FINAL : public ClusterStateLedger {
         const bmqp_ctrlmsg::LeaderMessageSequence& sequenceNumber,
         ClusterStateRecordType::Enum               recordType);
 
+    /// Internal helper method to apply commit for the advisory with the
+    /// specified `sequenceNumber` as the specified `ackQuorum` is reached.
+    int applyCommit(const bmqp_ctrlmsg::LeaderMessageSequence& sequenceNumber,
+                    const unsigned int                         ackQuorum);
```
678098:
```suggestion
                    unsigned int                               ackQuorum);
```

PR #1077, opened 2026-01-13
Feat[mqb]: Throttling subscriptions overhead

dorjesinpo:
Measure the rate of subscription evaluations using leaky bucket.
Measure the dispatcher queue depth
Limit the rate of subscription evaluations when the depth reaches high threshold

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
@@ -1441,7 +1441,7 @@ void RootQueueEngine::afterNewMessage()
 
     bsls::Types::Int64 numMessages = queue->storage()->numMessages(
         mqbu::StorageKey::k_NULL_KEY);
-    unsigned int batch = k_FLOW_CONTROL_BATCH;
+    unsigned int batch = 0;
 
     while (d_appsDeliveryContext.reset(storageIt)) {
         // Assume, all Apps need to deliver (some may be at capacity)
@@ -1490,10 +1490,8 @@ void RootQueueEngine::afterNewMessage()
             storageIt = 0;
         }
         else if (result > mqbu::FlowController::Watermark::e_ZERO) {
-            BSLS_ASSERT_SAFE(batch > 0);
-
-            if (--batch == 0) {
-                batch = k_FLOW_CONTROL_BATCH;
+            if (++batch >= k_FLOW_CONTROL_BATCH) {
+                batch = 0;
 
                 now = bsls::SystemTime::now(
                     bsls::SystemClockType::e_MONOTONIC);
@@ -2076,7 +2074,6 @@ void RootQueueEngine::updateFlowControl(bsls::Types::Int64 nowMs)
 
     // Provide time to the bucket algorithm.
     mqbi::Dispatcher* dispatcher = queue->dispatcher();
-    BSLS_ASSERT_SAFE(dispatcher);
 
     bsls::Types::Int64 numEvents = dispatcher->numProcessorEvents(queue);
 
--- a/src/groups/mqb/mqbu/mqbu_flowcontroller.h
+++ b/src/groups/mqb/mqbu/mqbu_flowcontroller.h
@@ -28,9 +28,6 @@
 ///-------------
 // This object is *not* thread safe.
 //
-/// Usage
-///-----
-//
 
 // MQB
 #include <mqbscm_version.h>
@@ -53,9 +50,10 @@ class FlowController {
 
     struct Policy {
         enum Enum {
-            e_NONE = 0  // Just calculate the moving (5 min) average
-            ,
-            e_LIMIT = 1  // Enforce rate limit
+            /// Just calculate the moving (5 min) average
+            e_NONE = 0,
+            /// Enforce rate limit
+            e_LIMIT = 1
         };
     };
 
@@ -93,13 +91,14 @@ class FlowController {
 
     struct Watermark {
         enum Enum {
-            e_ZERO = 0  // No data
-            ,
-            e_LOW = 1  // No overload
-            ,
-            e_HIGH = 2  // Resource(s) is(are) at the first high wm
-            ,
-            e_STRICT = 3  // Drop the data
+            /// No data
+            e_ZERO = 0,
+            /// No overload
+            e_LOW = 1,
+            /// Resource(s) is(are) at the first high wm
+            e_HIGH = 2,
+            /// Drop the data
+            e_STRICT = 3
         };
     };
 


src/groups/mqb/mqbblp/mqbblp_queueengineutil.h
```diff
@@ -31,6 +31,7 @@
 #include <mqbconfm_messages.h>
 #include <mqbi_queue.h>
 #include <mqbi_storage.h>
+#include <mqbu_flowcontroller.h>
```
678098: This include is not used in this header (so it's unrelated), can be removed

```suggestion
```

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1373,9 +1405,50 @@ void RootQueueEngine::afterNewMessage()
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(d_queueState_p->queue()->inDispatcherThread());
 
+    mqbi::Queue* queue = d_queueState_p->queue();
+    BSLS_ASSERT_SAFE(queue);
+
+    mqbi::Dispatcher* dispatcher = queue->dispatcher();
+    BSLS_ASSERT_SAFE(dispatcher);
```
```suggestion
```
Not used in this scope

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1373,9 +1405,50 @@ void RootQueueEngine::afterNewMessage()
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(d_queueState_p->queue()->inDispatcherThread());
 
+    mqbi::Queue* queue = d_queueState_p->queue();
+    BSLS_ASSERT_SAFE(queue);
```
678098, line 1405: Line 1406: `d_queueState_p->queue()->inDispatcherThread()`, queue ptr is dereferenced already before, so if it is zero, it is UB or more typically a segfault.

Either the check should be placed before checking `inDispatcherThread` or it should be removed (can we even have a zero ptr `queueState->queue()`?)

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1373,9 +1405,50 @@ void RootQueueEngine::afterNewMessage()
...
+    BSLS_ASSERT_SAFE(dispatcher);
+
+    mqbi::StorageIterator* storageIt = d_storageIter_mp.get();
+    bsls::TimeInterval     now       = bsls::SystemTime::now(
+        bsls::SystemClockType::e_MONOTONIC);
+    updateFlowControl(now.totalMilliseconds());
+
+    if (d_flowController.isFull()) {
+        BMQ_LOGTHROTTLE_INFO << "Local queue: " << d_queueState_p->uri()
+                             << " throttling delivery; the timer was "
```
```suggestion
                             << " throttling delivery; the timer is "
```
I would prefer present time in this log. `was` means that it was before (something), this raises a question "okay, it was before, but what about now?"

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1390,27 +1463,62 @@ void RootQueueEngine::afterNewMessage()
...
                 ->onEvent<mqbstat::QueueStatsDomain::EventType::e_QUEUE_TIME>(
                     d_appsDeliveryContext.timeDelta());
         }
         d_appsDeliveryContext.deliverMessage();
+
+        mqbu::FlowController::Watermark result = d_flowController.add(numHits);
+
+        if (result == mqbu::FlowController::e_Strict) {
+            BMQ_LOGTHROTTLE_INFO << "Local queue: " << d_queueState_p->uri()
+                                 << " throttling delivery; the overhead was "
```
```suggestion
                                 << " throttling delivery; the overhead is "
```

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1956,6 +2064,52 @@ void RootQueueEngine::logAlarmCb(
...
+
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(d_queueState_p->queue()->inDispatcherThread());
+
+    if (d_flowController.isIdle()) {
+        return;
+    }
+
+    mqbi::Queue* queue = d_queueState_p->queue();
+    BSLS_ASSERT_SAFE(queue);
```
678098: Same, `queue` is already dereferenced on line 2072

src/groups/mqb/mqbmock/mqbmock_dispatcher.cpp
```diff
@@ -209,6 +209,14 @@ Dispatcher::executor(BSLA_UNUSED const mqbi::DispatcherClient* client) const
     return bmqex::Executor();
 }
 
+bsls::Types::Int64
+Dispatcher::numProcessorEvents(const mqbi::DispatcherClient* client) const
```
678098:
```suggestion
Dispatcher::numProcessorEvents(BSLA_MAYBE_UNUSED const mqbi::DispatcherClient* client) const
```

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,335 @@
...
+// limitations under the License.
+
+// mqbu_flowcontroller.h                                              -*-C++-*-
+#ifndef INCLUDED_MQBU_FLOWCONTROLLER
+#define INCLUDED_MQBU_FLOWCONTROLLER
+
+//@PURPOSE: Provide a mechanism to limit data rate
+//
+//@CLASSES:
+//  mqbu::FlowController: mechanism to monitor/ manage a rate of actions.
```
678098:
```suggestion
//  mqbu::FlowController: mechanism to monitor/manage a rate of actions.
```

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,335 @@
...
+//
+//@DESCRIPTION:
+//
+/// Thread-safety
+///-------------
+// This object is *thread* *safe*.
```
678098:
```suggestion
// This object is *not* thread safe.
```

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,335 @@
...
+
+// MQB
+#include <mqbscm_version.h>
+
+// BDE
+#include <bdlb_print.h>
+#include <bdlt_timeunitratio.h>
+#include <bsl_ostream.h>
+#include <bslma_allocator.h>
+#include <bsls_atomic.h>
```
678098:
```suggestion
```
Not used

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,335 @@
...
+class FlowController {
+    // Mechanism to apply 'leaky bucket' logic to incoming traffic.
+
+  public:
+    // PUBLIC DATA
+
+    enum Policy {
+        e_None = 0  // Just calculate the moving (5 min) average
+        ,
+        e_Limit = 1  // Enforce rate limit
```
678098:
```suggestion
        /// Just calculate the moving (5 min) average
        e_NONE = 0,
        
        /// Enforce rate limit
        e_LIMIT = 1
```

Enums are not consistent with other codebase, also, `,` on its separate line looks odd, don't know why we keep this style

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,335 @@
...
+
+namespace mqbu {
+
+class FlowController {
+    // Mechanism to apply 'leaky bucket' logic to incoming traffic.
+
+  public:
+    // PUBLIC DATA
+
+    enum Policy {
```
678098: Typically we do

```
struct Policy {
    enum Enum {
        // ...
    }
};
```

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,335 @@
...
+        e_None = 0  // Just calculate the moving (5 min) average
+        ,
+        e_Limit = 1  // Enforce rate limit
+    };
+
+    struct Config {
```
678098:
```suggestion
    class Config {
```

This is not a simple data type, it uses class features fully

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,335 @@
...
+        /// Leaky bucket drain rate.
+        bsls::Types::Int64 d_ratePerMs;
+
+        /// Leaky bucket size.
+        bsls::Types::Int64 d_burst;
+
+      public:
+        // CREATORS
+        Config();
+        Config(Policy policy, int rate, int burst);
```
678098:
```suggestion
        explicit Config(Policy policy, int rate, int burst);
```

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,335 @@
...
+
+      private:
+        // PRIVATE DATA
+        Policy d_policy;
+
+        /// Leaky bucket drain rate.
+        bsls::Types::Int64 d_ratePerMs;
+
+        /// Leaky bucket size.
+        bsls::Types::Int64 d_burst;
```
678098:
```suggestion
        int d_ratePerMs;

        /// Leaky bucket size.
        int d_burst;
```

The fields are `int64`, the constructor args are `int`.
Should be one way in both cases.
Alternatively `int64` in constructor

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,335 @@
...
+        // PUBLIC ACCESSORS
+        Policy             policy() const;
+        bsls::Types::Int64 ratePerMs() const;
+        bsls::Types::Int64 burst() const;
+
+        bsl::ostream&
+        print(bsl::ostream& stream, int level, int spacesPerLevel) const;
+    };
+
+    enum Watermark {
```
678098:
```
struct Watermark {
    enum Enum {

    }
};
```
To be fair, I don't remember why we do this everywhere. Maybe because we implement `toAscii`, `fromAscii` closer to enum?

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,335 @@
...
+        e_Zero = 0  // No data
+        ,
+        e_Low = 1  // No overload
+        ,
+        e_High = 2  // Resource(s) is(are) at the first high wm
+        ,
+        e_Strict = 3  // Drop the data
```
678098:
```suggestion
        e_ZERO = 0  // No data
        ,
        e_LOW = 1  // No overload
        ,
        e_HIGH = 2  // Resource(s) is(are) at the first high wm
        ,
        e_STRICT = 3  // Drop the data
```

Enum names should be uppercase for consistency

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,335 @@
...
+    /// Number of `update` calls for the window calculating average watermark
+    /// and max burst.  Once the number is greater than the window size, save
+    /// the current value as `previous` and then reset it.  The result is then
+    /// a function (avg or max) of two values: the previous and the current.
+    int                d_currentHits;
+    bsls::Types::Int64 d_currentAverageWatermark;
+    bsls::Types::Int64 d_previousAverageWatermark;
+
+    int d_currentMaxBurst;
+    int d_previousMaxBurst;
```
678098: More optimized memory layout. It only has 3 unused padding bytes in the end of the class:
```suggestion
    bsls::Types::Int64 d_currentAverageWatermark;
    bsls::Types::Int64 d_previousAverageWatermark;
    int                d_currentHits;

    int d_currentMaxBurst;
    int d_previousMaxBurst;
    
    bool d_isHistoryFull;
```

The original version had 7 bytes padding after `bool d_isHistoryFull` and another 4 bytes after `d_currentHits`: 11 extra unused bytes total.

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,335 @@
...
+    /// Return max burst in the window.
+    int maxBurst() const;
+
+    /// Return `true` if there is no recorded activity and the policy does not
+    /// limit.  Return `false` otherwise.
+    bool isIdle() const;
+
+    /// Return `true` if the bucket is full and the policy does limit.
+    /// Return `false` otherwise.
+    inline bool isFull() const;
```
678098:
```suggestion
    bool isFull() const;
```

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,335 @@
...
+    print(bsl::ostream& stream, int level, int spacesPerLevel) const;
+};
+
+// ============================================================================
+//                             INLINE DEFINITIONS
+// ============================================================================
+
+// --------------------------
+// class FlowController::Rate
+// --------------------------
```
678098:
```suggestion
// ----------------------------
// class FlowController::Config
// ----------------------------
```

src/groups/mqb/mqbu/mqbu_flowcontroller.cpp
```diff
@@ -0,0 +1,233 @@
...
+, d_previousAverageWatermark(0)
+, d_currentMaxBurst(0)
+, d_previousMaxBurst(0)
+{
+    for (int i = 0; i < k_HISTORY_SIZE; i++) {
+        d_history[i] = 0;
+    }
+}
+
+FlowController::FlowController::~FlowController()
```
678098:
```suggestion
FlowController::~FlowController()
```

src/groups/mqb/mqbu/mqbu_flowcontroller.cpp
```diff
@@ -0,0 +1,233 @@
...
+        if (config().policy() == e_Limit) {
+            result = e_Strict;
+            // Possibly could Undo by d_count.subtract(howMany);
+        }
+        else {
+            result = e_High;
+        }
+    }
+
+    return result;
```
678098: Possible to get rid of a tmp variable for simplicity
```suggestion
    if (d_count > config().burst()) {
        // Over the bucket
        // Need to know what is the current policy and the strict threshold

        if (config().policy() == e_LIMIT) {
            return e_STRICT;
            // Possibly could Undo by d_count.subtract(howMany);
        }
        else {
            return e_HIGH;
        }
    }

    return e_LOW;
```

Or even

```suggestion
    if (d_count > config().burst()) {
        // Over the bucket
        // Need to know what is the current policy and the strict threshold

        // Possibly could Undo by d_count.subtract(howMany); on e_STRICT
        return (config().policy() == e_LIMIT) ? e_STRICT : e_HIGH;
    }

    return e_LOW;
```

678098: By the way, I remembered another reason why we use
```
struct MyType {
    enum Enum {
        ...
    }
};
```

This way, we don't poison the current scope with free enums.
We cannot simply refer to them like `e_SOMETHING`, we always has to specify their group: `MyType::e_SOMETHING`
This makes the code more specific

src/groups/mqb/mqbu/mqbu_flowcontroller.cpp
```diff
@@ -0,0 +1,233 @@
...
+    const bsls::Types::Int64 deltaMs = ms - d_lastUpdateMs;
+
+    BSLS_ASSERT_SAFE(0 <= deltaMs);
+
+    d_lastUpdateMs = ms;
+
+    const Config& config = d_config;
+
+    if (config.policy() > e_None) {
+        bsls::Types::Int64 rate = config.ratePerMs() * deltaMs;
```
678098, line 126: Can `deltaMs` be huge? I don't see it being limited by anything, same with `ratePerMs`. Can we have a signed integer overflow (UB) here?

This code should work fine if message flow is consistent, but what if we have a long pause in messages?

src/groups/mqb/mqbu/mqbu_flowcontroller.cpp
```diff
@@ -0,0 +1,233 @@
...
+FlowController::Config FlowController::survey(Policy policy) const
+{
+    bsls::Types::Int64 sum = d_currentSecondCount + d_totalCount;
+    const int numRecords  = d_isHistoryFull ? k_HISTORY_SIZE : d_currentRecord;
+    bsls::Types::Int64 ms = d_currentSecondMs + k_RECORD_MS * numRecords;
+
+    // Calculate moving average within approximately 5 min
+    // Use 'd_currentMaxBurst' as the burst size
+
+    return Config(policy, sum * 1000 / ms, maxBurst());
```
678098, line 203: Division by 0 if we happen to call `survey` in the following conditions:
- History is not full, `d_currentRecord == 0`
- `d_currentSecondMs == 0` (might happen if we start calling `survey` fast enough)

678098: Anyway, do we need to bother to support `d_isHistoryFull`?
Even if history is not initialized fully for a few seconds, there is no harm in setting
```
bsls::Types::Int64 ms = d_currentSecondMs + k_RECORD_MS * k_HISTORY_SIZE;
```

src/groups/mqb/mqbu/mqbu_flowcontroller.cpp
dorjesinpo: As I understand, `deltaMs` is always positive and it cannot overflow.
The code should work fine, but the loop below should be limited to the history size.

src/groups/mqb/mqbu/mqbu_flowcontroller.cpp
dorjesinpo: > Anyway, do we need to bother to support `d_isHistoryFull`? Even if history is not initialized fully for a few seconds, there is no harm in setting
> 
> ```
> bsls::Types::Int64 ms = d_currentSecondMs + k_RECORD_MS * k_HISTORY_SIZE;
> ```

Prefer to avoid that, the logic will start with incorrect rate (too low) if `survey` gets called before 5 sec pass.

dorjesinpo: > Division by 0 if we happen to call `survey` in the following conditions:
> 
> * History is not full, `d_currentRecord == 0`
> * `d_currentSecondMs == 0` (might happen if we start calling `survey` fast enough)

In other words, if `survey` gets called before `update` gets called with non-zero delta.  Adding this to UT.

src/groups/mqb/mqbu/mqbu_flowcontroller.h
dorjesinpo: I see this style in multiple places

678098: https://bloomberg.github.io/bde/knowledge_base/coding_standards.html#types-section-rules-specific-to-enumerations

BDE code style

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -125,6 +137,16 @@ VirtualIterator::~VirtualIterator()
     d_start_p->clearCache();
 }
 
+void executeInQueueDispatcher(mqbi::Queue*                 queue,
+                              const bsl::function<void()>& fn)
+{
+    // executed by the *SCHEDULER DISPATCHER* thread
```
```suggestion
    // executed by the *SCHEDULER* thread
```

Or why don't we allow ANY thread, it is thread-safe anyway

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
678098: No change here?

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1956,6 +2064,54 @@ void RootQueueEngine::logAlarmCb(
...
 
+void RootQueueEngine::updateFlowControl(bsls::Types::Int64 nowMs)
+{
+    // executed by the *QUEUE DISPATCHER* thread
+
+    // PRECONDITIONS
+    mqbi::Queue* queue = d_queueState_p->queue();
+    BSLS_ASSERT_SAFE(queue);
+
+    BSLS_ASSERT_SAFE(d_queueState_p->queue()->inDispatcherThread());
```
```suggestion
    BSLS_ASSERT_SAFE(queue->inDispatcherThread());
```

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1956,6 +2064,54 @@ void RootQueueEngine::logAlarmCb(
...
+    mqbi::Dispatcher* dispatcher = queue->dispatcher();
+    BSLS_ASSERT_SAFE(dispatcher);
+
+    bsls::Types::Int64 numEvents = dispatcher->numProcessorEvents(queue);
+
+    d_flowController.update(nowMs, numEvents);
+
+    const mqbu::FlowController::Config before = d_flowController.config();
+    d_flowController.checkWatermark(k_FLOW_CONTROL_LOW_WATERMARK,
+                                    k_FLOW_CONTROL_MAX_RATE_LIMIT);
```
678098, line 2084: If these values never change during `FlowController` lifetime, it is better to make them `FlowController`'s constructor args.

We don't have to provide the same constants on every `checkWatermark` call
```
FlowController controller(k_FLOW_CONTROL_LOW_WATERMARK,
                          k_FLOW_CONTROL_MAX_RATE_LIMIT);
controller.checkWatermark();
```

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1956,6 +2064,54 @@ void RootQueueEngine::logAlarmCb(
...
+
+    if (before != d_flowController.config()) {
```
678098, line 2086: I think we don't have to expose `config` to another component. It is an implementation detail of `FlowController`.

Why don't we just
```
if (d_flowController.checkWatermark()) {
```

Or with a better name
```
if (d_flowController.hasThrottlingChanged()) {
```

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1371,11 +1403,51 @@ void RootQueueEngine::afterNewMessage()
...
+
+            d_scheduler_p->scheduleEvent(
+                &d_flowControlEventHandle,
+                now + timeout,
+                bdlf::BindUtil::bindS(d_allocator_p,
+                                      executeInQueueDispatcher,
+                                      queue,
+                                      d_flowControlTimerCb));
+
+            BSLS_ASSERT_SAFE(d_flowControlEventHandle);
```
```suggestion
```
I think this check is excessive, because `scheduleEvent` always sets event handle if it is provided. It is also `void` - no return code, that hints that there is nothing to check

It is similar to checking a vector if it is really empty after calling `clear()`:
```
vec.clear();
BSLS_ASSERT_SAFE(vec.empty());
```

Also we don't check for this in other places where we `scheduleEvent`

src/groups/mqb/mqbu/mqbu_flowcontroller.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    // Accumulate count per each min within approximately 5 min
+    bsls::Types::Int64 elapsedSinceLastSecondMs = d_currentSecondMs + deltaMs;
+
+    if (elapsedSinceLastSecondMs > k_RECORD_MS * k_HISTORY_SIZE) {
+        // it has been more than 5 sec since the last update.
+        // reset the entire history.
+        elapsedSinceLastSecondMs = k_RECORD_MS * k_HISTORY_SIZE + 1;
+    }
+    while (elapsedSinceLastSecondMs > k_RECORD_MS) {
```
678098, line 149: What concerns me is that this loop is unnecessary unbounded.
10000 queues, 10 hours without updates: 3,600 * 10,000 * 10,000 = 360kkk iterations
This is for k_RECORDS_MS = 1000, reducing it to 100, 10 or 1 makes things even worse. Can we ensure by design this thing cannot take so much time?

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
dorjesinpo: We do want to log config.  It is not an implementation detail

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
dorjesinpo: Yes.  
```
    // PRECONDITIONS
    mqbi::Queue* queue = d_queueState_p->queue();
    BSLS_ASSERT_SAFE(queue);

    BSLS_ASSERT_SAFE(d_queueState_p->queue()->inDispatcherThread());
```

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
dorjesinpo: I am not sure, we won't want to change them later, so let's keep them as arguments for now

src/groups/mqb/mqbu/mqbu_flowcontroller.cpp
dorjesinpo: That is why we reset the `elapsedSinceLastSecondMs` on the line 147?

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1371,11 +1401,49 @@ void RootQueueEngine::afterNewMessage()
...
+            // Our Schedule is of the type e_MONOTONIC,
+            // see m_bmqbrkr::Task::Task.
+
+            d_scheduler_p->scheduleEvent(
+                &d_flowControlEventHandle,
+                now + timeout,
+                bdlf::BindUtil::bindS(d_allocator_p,
+                                      executeInQueueDispatcher,
+                                      queue,
+                                      d_flowControlTimerCb));
```
678098, line 1433: It is possible to bind directly to `mqbi::Dispatcher::execute`, and remove intermediate `executeInQueueDispatcher`:
```
bindS(d_allocator_p,
      mqbi::Dispatcher::execute,
      dispatcher(),
      d_flowControllerCb,
      queue)
```

No strong opinion here

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1390,27 +1458,63 @@ void RootQueueEngine::afterNewMessage()
...
+            d_appsDeliveryContext.reset(0);
+
+            // break the loop and go to the cleaning at the end.
+            storageIt = 0;
+        }
+        else if (result > mqbu::FlowController::Watermark::e_ZERO) {
+            BSLS_ASSERT_SAFE(batch > 0);
+
+            if (--batch == 0) {
+                batch = k_FLOW_CONTROL_BATCH;
```
678098: Note that `batch` is unsigned. It is very easy to have unsigned overflow by decrementing because value 0 is near: we can easily reach it on most iterations. Due to this, you have added an extra check `BSLS_ASSERT_SAFE(batch > 0);`, to ensure that we never actually make a mistake and overflow to `numeric_limits<unsigned int>::max()`.

I propose another solution: iterate in positive direction, starting from 0. This way, we don't have much room to overflow _by design_, and the check can be removed:

```suggestion
            if (k_FLOW_CONTROL_BATCH <= ++batch) {
                batch = 0;
```

As a bonus, `batch` will mean the current number of iterations after the last update, not the remaining number of iterations.

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1956,6 +2060,54 @@ void RootQueueEngine::logAlarmCb(
...
+
+    BSLS_ASSERT_SAFE(queue->inDispatcherThread());
+
+    if (d_flowController.isIdle()) {
+        return;
+    }
+
+    // Provide time to the bucket algorithm.
+    mqbi::Dispatcher* dispatcher = queue->dispatcher();
+    BSLS_ASSERT_SAFE(dispatcher);
```
```suggestion
    mqbi::Dispatcher* dispatcher = queue->dispatcher();
    BSLS_ASSERT_SAFE(dispatcher);
    BSLS_ASSERT_SAFE(queue->inDispatcherThread());

    if (d_flowController.isIdle()) {
        return;
    }

    // Provide time to the bucket algorithm.
```

I still think that checking that dispatcher exists is an unnecessary use of CPU cycles. I cannot imagine a situation where we have not provided a `dispatcher` to a queue or set it to NULL in the middle of run. Anyway, the NULL check should ideally be done in `Queue` constructor once per construction.

Alternatively, the `dispatcher` NULL check in this function should be done before checking `inDispatcherThread()`

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,338 @@
...
+//
+//@DESCRIPTION:
+//
+/// Thread-safety
+///-------------
+// This object is *not* thread safe.
+//
+/// Usage
+///-----
+//
```
678098:
```suggestion
```
Can remove if it is empty

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,338 @@
...
+    // Mechanism to apply 'leaky bucket' logic to incoming traffic.
+
+  public:
+    // PUBLIC DATA
+
+    struct Policy {
+        enum Enum {
+            e_NONE = 0  // Just calculate the moving (5 min) average
+            ,
+            e_LIMIT = 1  // Enforce rate limit
```
678098: We have this fix for enums in main:
https://github.com/bloomberg/blazingmq/pull/1124

src/groups/mqb/mqbu/mqbu_flowcontroller.h
```diff
@@ -0,0 +1,338 @@
...
+
+    struct Watermark {
+        enum Enum {
+            e_ZERO = 0  // No data
+            ,
+            e_LOW = 1  // No overload
+            ,
+            e_HIGH = 2  // Resource(s) is(are) at the first high wm
+            ,
+            e_STRICT = 3  // Drop the data
```
678098: Same
https://github.com/bloomberg/blazingmq/pull/1124

src/groups/mqb/mqbu/mqbu_flowcontroller.t.cpp
```diff
@@ -0,0 +1,136 @@
...
+    mqbu::FlowController::Watermark::Enum result;
+    bsls::Types::Int64                    time = 0;
+    mqbu::FlowController::Config          config;
+
+    config = fc.survey(mqbu::FlowController::Policy::e_NONE);
+
+    // 500 at 0.5 sec
+    result = fc.add(500);
+    BMQTST_ASSERT_EQ(result, mqbu::FlowController::Watermark::e_HIGH);
+    fc.update(time += 500, 100);
```
678098, line 56: `fc.update(time += 500, 100);`
Does this build in C++03/Solaris?
Could you rebase from the latest main to run newly added C++03 checks?

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
dorjesinpo: this cannot infer template argument 't_FUNC'

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
dorjesinpo: removing `BSLS_ASSERT_SAFE(dispatcher);`

src/groups/mqb/mqbu/mqbu_flowcontroller.t.cpp
dorjesinpo: https://github.com/bloomberg/blazingmq/actions/runs/22375906495/job/64765945986?pr=1077

PR #1079, opened 2026-01-14
Fix: Clean up command line directory inconsistencies

pniedzielski:
We noticed two inconsistencies while reviewing the `bmqbrkr` main function:

  1. `--config` was parsed as an optional argument, but if it was not given on the command line, we would detect and fail after parsing.  We can just make it required.
  
  2. All environment variables that the BlazingMQ broker knows about can be overridden by command line arguments, *except* `BMQ_PREFIX`.  We can add a new command line argument to do that.

### DIFF
--- a/src/applications/bmqbrkr/bmqbrkr.m.cpp
+++ b/src/applications/bmqbrkr/bmqbrkr.m.cpp
@@ -90,7 +90,7 @@ struct TaskEnvironment {
     // otherwise)
 
     bsl::string d_bmqPrefix;
-    // BMQ_PREFIX directory path
+    // Directory path to store PID, HIST, and CTL files under.
 
     bsl::string d_configJson;
     // JSON content ouput of the generated
@@ -414,7 +414,7 @@ static int initializeTask(bsl::ostream&    errorDescription,
                              bdlf::PlaceHolders::_1,    // prefix
                              bdlf::PlaceHolders::_2));  // istream
 
-    // Save the PID of the process in the '${BMQ_PREFIX}/bmqbrkr.pid' file
+    // Save the PID of the process in the '{prefix}/bmqbrkr.pid' file
     const bsl::string pidFile = taskEnv->d_bmqPrefix + "/bmqbrkr.pid";
     bsl::ofstream     pidFd(pidFile.c_str());
     if (!pidFd) {
@@ -499,9 +499,9 @@ static void shutdownApplication(TaskEnvironment* taskEnv)
     app.mqba::Application::~Application();
 }
 
-/// Update the `bmqbrkr.hist` file (in the BMQ_PREFIX directory) using the
+/// Update the `bmqbrkr.hist` file (in the prefix directory) using the
 /// specified `taskEnv`.  This file contains information about the last `n`
-/// successfull start of the broker, in reverse time order.
+/// successful starts of the broker, in reverse time order.
 /// Each line entry has the following format:
 ///    <currentTime_UTC>|<brokerVersion>|<configVersion>|<brokerId>
 ///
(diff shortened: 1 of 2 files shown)


src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -692,8 +692,11 @@ int main(int argc, const char* argv[])
     TaskEnvironment taskEnv;
     s_taskEnv_p = &taskEnv;
 
-    const char* prefixEnvVar = bsl::getenv("BMQ_PREFIX");
-    taskEnv.d_bmqPrefix      = (prefixEnvVar != 0 ? prefixEnvVar : "./");
+    // Default prefix directory to `BMQ_PREFIX` or ./
+    if (prefixDir.empty()) {
+        prefixDir = bsl::getenv("BMQ_PREFIX");
+    }
```
chrisbeard, line 699: I guess the only real question here is should the env var win even if `--prefixDir` is provided on the cli? Unsure.

hallfox: Usually the order of precedence I expect is config file, CLI arg, environment variable in order of lowest to highest precedence (i.e. `BMQ_PREFIX` should win)

pniedzielski: I have no strong opinions either way.

With bmqtool, it looks like the CLI param `--broker` overrides `BMQ_PORT`: https://github.com/bloomberg/blazingmq/blob/112f750ba63089918f7f5d4b10a26e1129020625/src/applications/bmqtool/bmqtool.m.cpp#L173
But `BMQ_BROKER_URI` overrides the user's session options: https://github.com/bloomberg/blazingmq/blob/112f750ba63089918f7f5d4b10a26e1129020625/src/groups/bmq/bmqa/bmqa_session.cpp#L298

So it's a mess even with the only two examples we have.

pniedzielski: Did a little research, it seems that most other CLI tools have CLI arg as the highest precedence.

  - **Docker**: https://docs.docker.com/reference/cli/docker/#configuration-files "Command line options override environment variables and environment variables override properties you specify in a config.json file."
  - **Git**: Does not document this very clearly from what I can tell (https://git-scm.com/docs/git#_environment_variables).  But, running the following test confirms that the CLI argument takes higher precedence:
    ```bash
    cd ~
    GIT_DIR=~/blazingmq/.git git --git-dir=~/blazingmq-sdk-java/.git fetch
    # remote: Enumerating objects: 23, done.
    # remote: Counting objects: 100% (23/23), done.
    # remote: Compressing objects: 100% (9/9), done.
    # remote: Total 23 (delta 8), reused 13 (delta 8), pack-reused 0 (from 0)
    # Unpacking objects: 100% (23/23), 5.93 KiB | 276.00 KiB/s, done.
    # From https://github.com/bloomberg/blazingmq-sdk-java
    #    9203baf..a342729  main       -> upstream/main
    ```
  - **HashiCorp Vault**: I think this is a bit more like BlazingMQ as a network server.  https://developer.hashicorp.com/vault/docs/commands#configure-environment-variables "You can use environment variables to configure the CLI globally. Some configuration settings have a corresponding CLI flag to configure a specific command.  For example, export `VAULT_ADDR='http://localhost:8200'` sets the address of your Vault server globally, while `-address='http://someotherhost:8200'` overrides the value for a specific command."
  
I have not been able to find anything that follows the precedence config file < CLI arg < environment variable.
  
Second thing I've found is it's hard to find this documented explicitly.  So I think we should document explicitly what we choose.

hallfox: Coming back to this after two months but I have no idea what I was thinking when I wrote this originally. Most tools I use do the config file < environment var < CLI arg order. Maybe I was thinking about how one might control broker configurations from a docker image?

Either way I do think the most important thing is to document the precedence regardless of what takes precedent. I just don't agree with what I originally wrote anymore.

PR #1081, opened 2026-01-15
Fix: fix python format

emelialei88:
Fix a Python format error. This happens after upgrading to Python3.12.

### DIFF
--- a/src/integration-tests/pytest.ini
+++ b/src/integration-tests/pytest.ini
@@ -1,8 +1,8 @@
 [pytest]
 log_cli_format = %(bmqprocess)-16s %(filename)s:%(lineno)d %(message)s
 log_level = INFO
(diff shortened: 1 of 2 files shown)


678098 (dismissed):

src/integration-tests/pytest.ini
```diff
@@ -1,8 +1,8 @@
 [pytest]
 log_cli_format = %(bmqprocess)-16s %(filename)s:%(lineno)d %(message)s
 log_level = INFO
-log_format = %(bmqprocess16)s %(asctime)s.%(msecs)03d (%(thread)15d) %(levelname)-8s %(name24)s %(filename)10s:%(lineno)-03d %(message)s
-log_file_format = %(bmqprocess16)s %(asctime)s.%(msecs)03d (%(thread)15d) %(levelname)-8s %(name24)s %(filename)10s:%(lineno)-03d %(message)s
```
jll63, line 5: We should also remove the code that synthesizes bmqprocess16 etc. It is [here](https://github.com/bloomberg/blazingmq/blob/main/src/python/blazingmq/dev/it/logging.py#L42-L58).

PR #1086, opened 2026-01-27
Fix[mqb]: RelayQE accepts temporary configure failures

dorjesinpo:
For RelayQueueEngine, Configure request is a result of merging configuration for _all_ clients _and_ multiple concurrent configure requests can be in transit.  RelayQueueEngine keeps a cache of everything it has sent upstream and uses the cache when building new (merged) request.  Otherwise, one concurrent request can conflict with another (earlier) request in transit.  Now, suppose the earlier request fails and the other succeeds.  The resulting configuration in effect will include material (ids) from the failed request.  
Unless the failure is permanent, RelayQueueEngine should consider failed config requests as successful.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
@@ -225,6 +225,40 @@ const mqbi::StorageIterator* VirtualIterator::next()
     return d_start_p;
 }
 
+bool isConfigureErrorPermanent(
+    const bmqp_ctrlmsg::StatusCategory::Value& value)
+{
+    switch (value) {
+    case bmqp_ctrlmsg::StatusCategory::E_SUCCESS: {
+        return false;
+    }
+    case bmqp_ctrlmsg::StatusCategory::E_UNKNOWN: {
+        return false;
+    }
+    case bmqp_ctrlmsg::StatusCategory::E_TIMEOUT: {
+        return false;
+    }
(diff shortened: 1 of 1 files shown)


src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -225,6 +225,40 @@ const mqbi::StorageIterator* VirtualIterator::next()
     return d_start_p;
 }
 
+bool isConfigureErrorPermanent(
+    const bmqp_ctrlmsg::StatusCategory::Value& value)
```
678098, line 229:
```suggestion
    bmqp_ctrlmsg::StatusCategory::Value value)
```

Enums are backed by integer types, we don't need to pass basic types by const ref

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -225,6 +225,40 @@ const mqbi::StorageIterator* VirtualIterator::next()
...
+    }
+    case bmqp_ctrlmsg::StatusCategory::E_REFUSED: {
+        return false;
+    }
+    case bmqp_ctrlmsg::StatusCategory::E_INVALID_ARGUMENT: {
+        return true;
+    }
+    case bmqp_ctrlmsg::StatusCategory::E_NOT_READY: {
+        return false;
+    }
```
678098, line 258: This function might be ~3 times shorter if fallthrough is used

678098 (approved):
We can merge as-is, have small comments

PR #1094, opened 2026-01-30
Fix mqbc::StorageMgr: Prevent Duplicate ReplicaDataRequests to Same Node

kaikulimu:
## Problem

When the primary's Partition FSM receives multiple `REPLICA_STATE_RSPN` or `PRIMARY_STATE_RQST` events from different replicas in quick succession, it may send `ReplicaDataRequestPush` and/or `ReplicaDataRequestDrop` to the same replica multiple times.

## Root Cause:

The functions `do_replicaDataRequestPush`, `do_replicaDataRequestDrop`, and `do_startSendDataChunks` were iterating through all replicas in `nodeToSeqNumCtxMap` regardless of which replica triggered the current event. This meant:

  1. Primary receives `REPLICA_STATE_RSPN` from Replica A
    - Processes event, iterates all replicas, sends requests to A
  2. Primary receives `REPLICA_STATE_RSPN` from Replica B shortly after
    - Processes event, iterates all replicas again, potentially sends duplicate requests to A

While the `d_isRecoveryDataSent` flag provided some protection, it was intentionally reset when new sequence numbers arrived or on `PRIMARY_STATE_RQST` events, making it insufficient for preventing all duplicate sends.

## Solution

Modified the three affected functions to distinguish between broadcast and targeted behavior:
- Use event type and event source to determine the set of potential targets; do not automatically iterate the entire `nodeToSeqNumCtxMap`. For example, if event type is `REPLICA_STATE_RSPN` and event source is a replica, that replica is the only possible destination of our requests.
- Remove unused `d_isRecoveryDataSent` flag

## New Tests in [`test_fsm_partition_sync.py`](https://github.com/bloomberg/blazingmq/blob/main/src/integration-tests/test_fsm_partition_sync.py)
1. `test_sync_after_replicas_missed_various_records`: Test replicas journal file synchronization with cluster after each of them missed a different number of records.
2. `test_sync_after_replicas_missed_or_extra_records`: Test replicas journal file synchronization with cluster after some of them missed records while some had extra records.
    - This test fails near the end, but we early return and put a TODO block for now. See the code for explanation.

### DIFF
--- a/src/python/blazingmq/dev/it/process/broker.py
+++ b/src/python/blazingmq/dev/it/process/broker.py
@@ -92,17 +92,6 @@ class Broker(blazingmq.dev.it.process.bmqproc.BMQProcess):
     def __repr__(self):
         return f"Broker({self.name})"
 
-    def start(self):
-        """
-        Start the broker process after cleaning up the named pipe.
-        """
-        # Clean up the named pipe before starting to avoid conflicts from
-        # previous runs. This is necessary when restarting a broker without
-        # creating a new Broker object.
-        ctl_path = self._cwd / "bmqbrkr.ctl"
-        ctl_path.unlink(missing_ok=True)
-        super().start()
-
     @property
     def datacenter(self) -> str:
         return self.config.config.app_config.host_data_center


dorjesinpo (changes requested):
Summary of the meeting. 
1.  Introducing new state on Replica indicating that it is waiting for PrimaryStateResponse
2.  In this state, answer ReplicaStateRequest with a status preventing the Primary from starting (extra) sync
3.  Upon any PrimaryStateResponse Replica transitions to the existing HEALING and now can answer ReplicaStateRequest with a good status
4.  Primary needs to answer PrimaryStateRequest in UNKNOWN state (with a status)
5.  This all depends on the order:  Primary sends ReplicaStateRequest _before_ processing PrimaryStateRequest

PR #1098, opened 2026-02-04
Feat[MQB]: dispatcher stats

waldgange:
This PR introduces 4 new metrics collection and reporting for each [DispatcherEventType](https://github.com/bloomberg/blazingmq/blob/daf7701c5e8749aa71aac749f54ca2df449d203e/src/groups/mqb/mqbi/mqbi_dispatcher.h#L216-L244) (4*12=48 in total) :
- e_PROCESSING_TIME_MAX
- e_PROCESSING_TIME_AVG
- e_PROCESSING_TIME_SUM
- e_PROCESSED_COUNT

### DIFF
--- a/src/groups/mqb/mqbstat/mqbstat_dispatcherstats.cpp
+++ b/src/groups/mqb/mqbstat/mqbstat_dispatcherstats.cpp
@@ -205,8 +205,6 @@ DispatcherStatsUtil::initializeStatContext(int               historySize,
         .defaultHistorySize(historySize)
         .statValueAllocator(allocator)
         .storeExpiredSubcontextValues(true)
-        .value("queued_count")
-        .value("queued_time", bmqst::StatValue::e_DISCRETE)
         .value("processing_time_undefined", bmqst::StatValue::e_DISCRETE)
         .value("processing_time_dispatcher", bmqst::StatValue::e_DISCRETE)
         .value("processing_time_callback", bmqst::StatValue::e_DISCRETE)
@@ -220,7 +218,9 @@ DispatcherStatsUtil::initializeStatContext(int               historySize,
         .value("processing_time_storage", bmqst::StatValue::e_DISCRETE)
         .value("processing_time_recovery", bmqst::StatValue::e_DISCRETE)
         .value("processing_time_replication_receipt",
-               bmqst::StatValue::e_DISCRETE);
+               bmqst::StatValue::e_DISCRETE)
+        .value("queued_count")
+        .value("queued_time", bmqst::StatValue::e_DISCRETE);
 
     return bsl::shared_ptr<bmqst::StatContext>(
         new (*allocator) bmqst::StatContext(config, allocator),
--- a/src/groups/mqb/mqbstat/mqbstat_dispatcherstats.h
+++ b/src/groups/mqb/mqbstat/mqbstat_dispatcherstats.h
@@ -122,9 +122,6 @@ class DispatcherStats {
     /// dispatcher queues from the clients.
     struct DispatcherStatsIndex {
         enum Enum {
-            e_STAT_QUEUE = -2,  // Queue/Dequeue
-            e_STAT_TIME  = -1,  // Event queued time
-
             /// Processing times for each event type.
             /// Each enum value MUST be equal to the corresponding
             /// event type value in mqbi::DispatcherEventType
@@ -140,7 +137,10 @@ class DispatcherStats {
             e_STAT_PROCESSING_TIME_CLUSTER_STATE       = 9,
             e_STAT_PROCESSING_TIME_STORAGE             = 10,
             e_STAT_PROCESSING_TIME_RECOVERY            = 11,
-            e_STAT_PROCESSING_TIME_REPLICATION_RECEIPT = 12
+            e_STAT_PROCESSING_TIME_REPLICATION_RECEIPT = 12,
+            /// Other queue metrics
+            e_STAT_QUEUE = 13,  // Queue/Dequeue
+            e_STAT_TIME  = 14,  // Event queued time
         };
     };
 


src/groups/mqb/mqba/mqba_dispatcher.cpp
```diff
@@ -329,6 +355,14 @@ void Dispatcher::queueEventCb(mqbi::DispatcherClientType::Enum type,
     if (event) {
         BALL_LOG_TRACE << "Dispatching Event to queue " << processorId
                        << " of " << type << " dispatcher: " << *event;
+
+        const bsls::Types::Int64 queuedTime =
+            bmqsys::Time::highResolutionTimer() - event->enqueueTime();
+        const bsls::Types::Int64 processingTimeStart =
+            bmqsys::Time::highResolutionTimer();
```
678098: Let's reuse the current time point because the call `bmqsys::Time::highResolutionTimer()` is not free


```suggestion
        const bsls::Types::Int64 processingTimeStart =
            bmqsys::Time::highResolutionTimer();
        const bsls::Types::Int64 queuedTime =
            processingTimeStart - event->enqueueTime();
```

src/groups/mqb/mqba/mqba_dispatcher.h
```diff
@@ -45,9 +45,14 @@
 /// be executed in-place.  A call to `dispatch` from outside of the executor's
 /// associated processor thread is equivalent to a call to `post`.
 
+// BMQ
+#include <bmqst_statcontext.h>
+#include <bmqsys_time.h>
```
678098: This header is used only in this function:
```
Dispatcher::dispatchEvent(mqbi::Dispatcher::DispatcherEventRvRef event,
                          mqbi::DispatcherClientType::Enum       type,
                          mqbi::Dispatcher::ProcessorHandle      handle)
```

However, this function is an override of a virtual function, and we never actually inline it. This means that we might move both the header and `dispatchEvent` implementation to `cpp` file

After this, we can also move these new headers to `cpp` file:
```
#include <bmqst_statcontext.h>
#include <mqbstat_dispatcherstats.h>
```

src/groups/mqb/mqba/mqba_dispatcher.h
```diff
@@ -266,6 +273,12 @@ class Dispatcher BSLS_KEYWORD_FINAL : public mqbi::Dispatcher {
         bsl::vector<bsl::shared_ptr<mqbi::DispatcherEventSource> >
             d_eventSources;
 
+        /// Pointer to stat context for client
+        bslma::ManagedPtr<bmqst::StatContext> d_clientStatContext_mp;
+
+        /// Vector of stat contexts pointers, one per client's processor.
+        bsl::vector<bslma::ManagedPtr<bmqst::StatContext> > d_statContexts;
```
678098: I think this will not compile on Solaris
The solution might be to change it to `shared_ptr`s

src/groups/mqb/mqbstat/mqbstat_dispatcherstats.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+}  // close unnamed namespace
+
+// ---------------------
+// class DispatcherStats
+// ---------------------
+
+bsls::Types::Int64 DispatcherStats::getValue(const bmqst::StatContext& context,
+                                             int               snapshotId,
+                                             const Stat::Enum& stat)
```
678098:
```suggestion
                                             Stat::Enum        stat)
```

Enum values are integer types, there is no need to provide basic types as const references

src/groups/mqb/mqbstat/mqbstat_dispatcherstats.cpp
```diff
@@ -0,0 +1,244 @@
...
+        .value("processing_time_put", bmqst::StatValue::e_DISCRETE)
+        .value("processing_time_ack", bmqst::StatValue::e_DISCRETE)
+        .value("processing_time_cluster_state", bmqst::StatValue::e_DISCRETE)
+        .value("processing_time_storage", bmqst::StatValue::e_DISCRETE)
+        .value("processing_time_recovery", bmqst::StatValue::e_DISCRETE)
+        .value("processing_time_replication_receipt",
+               bmqst::StatValue::e_DISCRETE);
+
+    return bsl::shared_ptr<bmqst::StatContext>(
+        new (*allocator) bmqst::StatContext(config, allocator),
```
678098, line 226: `new (*allocator)` will probably segfault if `allocator = 0` provided to this function.
You might either add a precondition check, or use a shorter `bsl::allocate_shared`

src/groups/mqb/mqbstat/mqbstat_dispatcherstats.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+// -------------------------
+// class DispatcherStatsUtil
+// -------------------------
+
+bsl::shared_ptr<bmqst::StatContext>
+DispatcherStatsUtil::initializeStatContext(int               historySize,
+                                           bslma::Allocator* allocator)
+{
+    bdlma::LocalSequentialAllocator<2048> localAllocator(allocator);
```
678098: I think we allocate stat contexts for dispatcher/its queues so rarely that using LSA is overcomplication. Can use `allocator` just fine, and there will be less lines and one less include

src/groups/mqb/mqba/mqba_dispatcher.h
```diff
@@ -559,9 +576,19 @@ Dispatcher::dispatchEvent(mqbi::Dispatcher::DispatcherEventRvRef event,
     case mqbi::DispatcherClientType::e_SESSION:
     case mqbi::DispatcherClientType::e_QUEUE:
     case mqbi::DispatcherClientType::e_CLUSTER: {
-        d_contexts[type]->d_processorPool_mp->enqueueEvent(
+        DispatcherContext* dispatcherContext = d_contexts[type].get();
+
+        bslmf::MovableRefUtil::access(event)->setEnqueueTime(
+            bmqsys::Time::highResolutionTimer());
```
678098: `setEnqueueTime` should also be called in `execute` functions:
https://github.com/waldgange/blazingmq/blob/95c6685b726426b6009fafd91f2f13508e4d50a3/src/groups/mqb/mqba/mqba_dispatcher.h#L600-L633

src/groups/mqb/mqbstat/mqbstat_dispatcherstats.h
```diff
@@ -87,18 +139,48 @@ class DispatcherStats {
...
+            e_STAT_PROCESSING_TIME_REJECT,
+            e_STAT_PROCESSING_TIME_PUSH,
+            e_STAT_PROCESSING_TIME_PUT,
+            e_STAT_PROCESSING_TIME_ACK,
+            e_STAT_PROCESSING_TIME_CLUSTER_STATE,
+            e_STAT_PROCESSING_TIME_STORAGE,
+            e_STAT_PROCESSING_TIME_RECOVERY,
+            e_STAT_PROCESSING_TIME_REPLICATION_RECEIPT,
+            e_STAT_PROCESSING_TIME_END =
+                e_STAT_PROCESSING_TIME_REPLICATION_RECEIPT
```
678098: You might also make STAT types equal to their corresponding mqbi::DispatcherEventType. This will simplify offsets and decrease cognitive difficulty because the values will be matching.

```suggestion
            e_STAT_QUEUE                 = -2,  // Queue/Dequeue
            e_STAT_TIME                  = -1,  // Event queued time

            /// Processing times for each event type.
            /// Each enum value MUST be equal to the corresponding
            /// event type value in mqbi::DispatcherEventType
            e_STAT_PROCESSING_TIME_UNDEFINED = 0,
            e_STAT_PROCESSING_TIME_DISPATCHER = 1,
            e_STAT_PROCESSING_TIME_CALLBACK = 2,
            e_STAT_PROCESSING_TIME_CONTROL_MSG = 3,
            e_STAT_PROCESSING_TIME_CONFIRM = 4,
            e_STAT_PROCESSING_TIME_REJECT = 5,
            e_STAT_PROCESSING_TIME_PUSH = 6,
            e_STAT_PROCESSING_TIME_PUT = 7,
            e_STAT_PROCESSING_TIME_ACK = 8,
            e_STAT_PROCESSING_TIME_CLUSTER_STATE = 9,
            e_STAT_PROCESSING_TIME_STORAGE = 10,
            e_STAT_PROCESSING_TIME_RECOVERY = 11,
            e_STAT_PROCESSING_TIME_REPLICATION_RECEIPT = 12
```

Note that new random dispatcher stats are extendable in negative direction, new event type stats are extendable in positive direction.

678098: Also it will be nice to add a few compile-time validations for these enums matching mqbi ones, somewhere in unit tests or just in cpp file:
```
BSLMF_ASSERT(static_cast<int>(DispatcherStatsIndex::e_STAT_PROCESSING_TIME_UNDEFINED) == static_cast<int>(mqbi::DispatcherEventType::e_UNDEFINED));
```

waldgange: I added the compile-time checks. However I desided to get rid of the negative values in the enums. These values are used as indices to access elements of the vector inside `StatContext::adjustValue`, `StatContext::reportValue` and so on. To make it working we would have to incremect the enums every time we call the reporting functions. It's not convinient so I just moved the first two indices to the end of the list.

PR #1099, opened 2026-02-04
Feat[bmq]: Keep counted reference to blob in Event

dorjesinpo:
Detach PUT builder and events it generates by using `shared_ptr<bsl::shared_ptr<bdlbb::Blob>` instead of `bdlbb::Blob`

### DIFF
--- a/src/groups/bmq/bmqio/bmqio_channelutil.h
+++ b/src/groups/bmq/bmqio/bmqio_channelutil.h
@@ -27,7 +27,6 @@
 
 // BDE
 #include <bdlbb_blob.h>
-#include <bdlcc_sharedobjectpool.h>
 #include <bsl_string.h>
 #include <bsl_vector.h>
 


src/groups/bmq/bmqio/bmqio_channelutil.cpp
```diff
@@ -195,11 +194,12 @@ int ChannelUtil::handleRead(bsl::vector<bdlbb::Blob>* outPackets,
...
-        bdlbb::BlobUtil::append(&outPackets->back(),
-                                *inBlob,
-                                consumedBytes,
-                                packetLength);
+        reader(*inBlob, consumedBytes, packetLength);
+        //        outPackets->emplace_back();
+        //        bdlbb::BlobUtil::append(&outPackets->back(),
+        //                                *inBlob,
+        //                                consumedBytes,
+        //                                packetLength);
```
hallfox:
```suggestion
```

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1050,6 +1050,13 @@ void RootQueueEngine::configureHandle(
                                      d_queueState_p->routingContext(),
                                      d_allocator_p),
                                  d_allocator_p);
+    const QueueHandleCatalog::Visitor& visitor = bdlf::BindUtil::bind(
+        &RootQueueEngine::rebuildSelectedApp,
+        this,
+        bdlf::PlaceHolders::_1,  // handle
+        bdlf::PlaceHolders::_2,  // info
+        iter,
+        previous.get());
```
678098: This code is not used
Is this a bad merge?

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -1246,7 +1253,6 @@ void RootQueueEngine::releaseHandle(
                     // 'releaseHandle' that releases the handle.
                     BALL_LOG_ERROR
                         << "#QUEUE_IMPROPER_BEHAVIOR "
-                        << "For queue [" << d_queueState_p->uri() << "],  "
```
678098, line 1250: Should we really remove this line?
Looks not intentional

src/groups/bmq/bmqio/bmqio_channelutil.t.cpp
```diff
@@ -217,9 +227,23 @@ static void test2_handleRead_multiplePackets()
 
         const int totalLength = k_MINIMUM_PACKET_LENGTH + 12;
         appendToInputBlob(&input, totalLength, totalLength);
-        const int rc = bmqio::ChannelUtil::handleRead(&packets,
-                                                      &numNeeded,
-                                                      &input);
+
+        const bsl::function<
+            void(const bdlbb::Blob& source, int offset, int length)>& reader =
```
678098: `reader` is not used

src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp
```diff
@@ -1299,11 +1299,10 @@ void ClusterOrchestrator::processElectorEvent(const bmqp::Event&   event,
         dispatcher()->getDefaultEventSource()->getEvent();
     (*clusterEvent).setType(mqbi::DispatcherEventType::e_CALLBACK);
 
-    bmqp::Event clonedEvent = event.clone(d_allocator_p);
     clusterEvent->callback()
         .createInplace<ClusterOrchestrator::OnElectorEventFunctor>(
             this,
-            bslmf::MovableRefUtil::move(clonedEvent),
+            bslmf::MovableRefUtil::move(event),
```
678098: How does this method work now?
Note that the arg is `const bmqp::Event&   event` and we cannot perfectly move it now, we have to create an implicit copy

src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
```diff
@@ -4405,7 +4404,7 @@ void BrokerSession::transferAckEvent(bmqp::AckEventBuilder*  ackBuilder,
     // Our event is full at this point so send this ack event to the user and
     // reset the builder to append the ack that was rejected.
 
-    bmqp::Event event(ackBuilder->blob().get(), d_allocator_p, true);
+    bmqp::Event event(ackBuilder->blob(), d_allocator_p);
     // clone = true
```
678098:
```suggestion
```
The comment is outdated

src/groups/mqb/mqbblp/mqbblp_clusterorchestrator.cpp
dorjesinpo: Good point.  The `event` comes from `TCPSessionFactory::read` and has `bsl::shared_ptr<const bdlbb::Blob>`.  So, copy.

src/groups/bmq/bmqimp/bmqimp_brokersession.h
```diff
@@ -918,7 +918,8 @@ class BrokerSession BSLS_CPP11_FINAL {
     /// Send the specified `blob`, representing a `Confirm` packet over the
     /// channel and containing the specified `msgCount` message confirmation
     /// in it (count used for statistics).
-    void sendConfirm(const bdlbb::Blob& blob, const int msgCount);
+    void sendConfirm(const bsl::shared_ptr<const bdlbb::Blob>& blob,
+                     const int                                 msgCount);
```
678098, line 922:
```suggestion
    void sendConfirm(const bsl::shared_ptr<const bdlbb::Blob>& blob_sp,
                     int                                       msgCount);
```

src/groups/bmq/bmqimp/bmqimp_brokersession.h
```diff
@@ -1460,7 +1461,7 @@ class BrokerSession BSLS_CPP11_FINAL {
     bmqt::GenericResult::Enum writeOrBuffer(const bdlbb::Blob& eventBlob,
                                             bsls::Types::Int64 highWaterMark);
 
-    bool acceptUserEvent(const bdlbb::Blob& eventBlob);
+    bool acceptUserEvent(const bsl::shared_ptr<const bdlbb::Blob>& eventBlob);
```
678098:
```suggestion
    bool acceptUserEvent(const bsl::shared_ptr<const bdlbb::Blob>& blob_sp);
```

src/groups/bmq/bmqimp/bmqimp_brokersession.h
```diff
@@ -1603,12 +1605,12 @@ class BrokerSession BSLS_CPP11_FINAL {
                         bsls::TimeInterval            timeout,
                         const EventCallback& eventCallback = EventCallback());
 
-    int post(const bdlbb::Blob& eventBlob);
+    int post(const bsl::shared_ptr<const bdlbb::Blob>& eventBlob);
```
678098:
```suggestion
    int post(const bsl::shared_ptr<const bdlbb::Blob>& blob_sp);
```

src/groups/bmq/bmqimp/bmqimp_brokersession.h
```diff
@@ -1603,12 +1605,12 @@ class BrokerSession BSLS_CPP11_FINAL {
...
 
     int confirmMessage(const bsl::shared_ptr<bmqimp::Queue>& queue,
                        const bmqt::MessageGUID&              messageId);
 
-    int confirmMessages(const bdlbb::Blob& blob);
+    int confirmMessages(const bsl::shared_ptr<const bdlbb::Blob>& blob);
```
678098:
```suggestion
    int confirmMessages(const bsl::shared_ptr<const bdlbb::Blob>& blob_sp);
```

src/groups/bmq/bmqa/bmqa_confirmeventbuilder.h
```diff
@@ -212,6 +212,8 @@ class ConfirmEventBuilder {
     /// `length == 0`.  Behavior is undefined unless this instance was
     /// obtained using `bmqa::Session::loadConfirmEventBuilder`.
     const bdlbb::Blob& blob() const;
+
+    const bsl::shared_ptr<bdlbb::Blob>& sharedBlob() const;
```
678098, line 216: Can we remove `const bdlbb::Blob& blob() const;` and have only `const bsl::shared_ptr<bdlbb::Blob>& blob() const;` instead?
This will make API consistent with `PutEventBuilder` and other builders that return only shared pointer on `blob()` call

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -432,19 +435,16 @@ void TCPSessionFactory::readCallback(const bmqio::Status& status,
...
-
-    bsl::vector<bdlbb::Blob> readBlobs(&lsa);
-    readBlobs.reserve(32);
+    const int rc = bmqio::ChannelUtil::handleRead(
+        bdlf::BindUtil::bind(&TCPSessionFactory::read,
+                             this,
+                             channelInfo,
+                             bdlf::PlaceHolders::_1,
+                             bdlf::PlaceHolders::_2,
+                             bdlf::PlaceHolders::_3),
```
678098: Getting rid of local `bsl::vector<bdlbb::Blob> readBlobs(&lsa);` is good, however, using `Bind` here on hot path is super expensive.
First, its construction and cast to `bsl::function` is rather expensive.
Second, each invokation of this binded function is done with multiple function calls, which produces 6-7 stack frames. These function calls are not free, and now we do this for every complete event.

At least replacing this `Bind` with functor object will be much better in terms of performance. Still not sure if it is enough.

src/groups/bmq/bmqio/bmqio_channelutil.cpp
```diff
@@ -195,11 +194,7 @@ int ChannelUtil::handleRead(bsl::vector<bdlbb::Blob>* outPackets,
 
         // We have a complete message.  Move the bytes out of 'inBlob' into a
         // standalone packet in 'outPackets'.
```
678098: This comment is outdated

src/groups/bmq/bmqio/bmqio_channelutil.cpp
```diff
@@ -195,11 +194,7 @@ int ChannelUtil::handleRead(bsl::vector<bdlbb::Blob>* outPackets,
...
-        outPackets->emplace_back();
-        bdlbb::BlobUtil::append(&outPackets->back(),
-                                *inBlob,
-                                consumedBytes,
-                                packetLength);
+        reader(*inBlob, consumedBytes, packetLength);
```
678098, line 196: Why don't we prepare a complete blob here?
Something like this (TBD where to get blob sp pool):
```suggestion
        bsl::shared_ptr<bdlbb::Blob> readBlob = d_blobSpPool.getObject();
        bdlbb::BlobUtil::append(readBlob.get(), source, offset, length);

        reader(bslmf::MovableRefUtil::move(readBlob));
```
This allows to simplify `reader`'s API and pay less for passing its args (note that we call this on every complete event).

678098: This also allows to get rid of `d_blobSpPool` usage from `TCPSessionFactory::read`, making it easier to convert it to a lightweight functor.

src/groups/bmq/bmqio/bmqio_channelutil.h
```diff
@@ -42,6 +42,11 @@ namespace bmqio {
 
 /// Utility namespace for channel manipulation.
 struct ChannelUtil {
+  public:
+    typedef const bsl::function<
```
678098, line 46:
```suggestion
    typedef bsl::function<
```

`handleRead` already expects explicit `const Reader&`

src/groups/bmq/bmqp/bmqp_event.h
```diff
@@ -76,10 +76,9 @@ class Event {
     bslma::Allocator* d_allocator_p;
     // Allocator to use by this object.
 
-    bsl::shared_ptr<const bdlbb::Blob> d_clonedBlob_sp;
+    bsl::shared_ptr<const bdlbb::Blob> d_blob_sp;
```
678098, line 79: Can we remove raw pointer field `d_blob_p`?

src/groups/bmq/bmqp/bmqp_event.h
```diff
@@ -372,42 +368,53 @@ int Event::loadSchemaEvent(TYPE* message) const
...
 
 inline Event::Event(const Event& src, bslma::Allocator* allocator)
 : d_allocator_p(allocator)
-, d_clonedBlob_sp(0, allocator)
+, d_blob_sp(0, allocator)
 , d_blob_p(0)
 {
-    d_clonedBlob_sp = src.d_clonedBlob_sp;  // src could be a clone
-    initialize(src.d_blob_p, false);
+    d_blob_sp = src.d_blob_sp;  // src could be a clone
```
678098:
```suggestion
    d_blob_sp = src.d_blob_sp;
```
No more clones
Should this logic change?

src/groups/bmq/bmqp/bmqp_event.h
```diff
@@ -418,8 +425,8 @@ inline Event& Event::operator=(const Event& rhs)
         return *this;
     }
 
-    d_clonedBlob_sp = rhs.d_clonedBlob_sp;  // src could be a clone
-    initialize(rhs.d_blob_p, false);
+    d_blob_sp = rhs.d_blob_sp;  // src could be a clone
```
678098:
```suggestion
    d_blob_sp = rhs.d_blob_sp;
```

src/groups/bmq/bmqp/bmqp_eventutil.h
```diff
@@ -84,7 +83,7 @@ struct EventUtilEventInfo {
     typedef bsl::vector<EventUtilQueueInfo> Ids;
 
     // PUBLIC DATA
-    const bdlbb::Blob d_blob;
+    const bsl::shared_ptr<bdlbb::Blob> d_blob;
```
678098:
```suggestion
    const bsl::shared_ptr<bdlbb::Blob> d_blob_sp;
```

src/groups/bmq/bmqp/bmqp_eventutil.h
```diff
@@ -101,9 +100,9 @@ struct EventUtilEventInfo {
     /// Create a `bmqp::EventUtil_EventInfo` object using the specified
     /// `blob` and `queueIds`.  All memory allocations will be done using
     /// the specified `allocator`.
-    EventUtilEventInfo(const bdlbb::Blob& blob,
-                       const Ids&         queueIds,
-                       bslma::Allocator*  allocator);
+    EventUtilEventInfo(const bsl::shared_ptr<bdlbb::Blob>& blob,
```
678098:
```suggestion
    EventUtilEventInfo(const bsl::shared_ptr<bdlbb::Blob>& blob_sp,
```

src/groups/bmq/bmqa/bmqa_confirmeventbuilder.h
dorjesinpo: Not sure!  This is public API

src/groups/bmq/bmqp/bmqp_event.h
dorjesinpo: No, because Event can be still initialized with a raw pointer (like before).  This PR did not go that far as to remove those uses.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
678098: I also run a few benchmarks.

## Call

```
bsl::function from BindUtil (4 args):
       total: 338.47 ms (100000000 iterations)
    per call: 3 ns

bsl::function from functor (4 args):
       total: 216.54 ms (100000000 iterations)
    per call: 2 ns

pure virtual interface (4 args):
       total: 172.21 ms (100000000 iterations)
    per call: 1 ns
```

## Assignment

```
assign from BindUtil (4 args):
       total: 1.48 s (10000000 iterations)
    per call: 148 ns

assign from functor (4 args):
       total: 711.95 ms (10000000 iterations)
    per call: 71 ns
```

Functor behaves better but still is expensive when used with intermediate `bsl::function`.
We either have to cache it for reuse or use stack-constructed `Callable` that is called from the caller function (without `bsl::function` use).
I can do it in the future if you don't mind.

src/groups/bmq/bmqio/bmqio_channelutil.h
```diff
@@ -27,6 +27,7 @@
 
 // BDE
 #include <bdlbb_blob.h>
+#include <bdlcc_sharedobjectpool.h>
```
678098: This include doesn't seem to be used in this header

PR #1101, opened 2026-02-04
Feat[bmq]: SessionOptions::d_channelWriteTimeout

dorjesinpo:
Introducing new session option - `channelWriteTimeoutMs` (the default is 5000).
When `0`, `Session::post` never blocks (the default blocks when channel is at high watermark)

### DIFF
--- a/src/groups/bmq/bmqt/bmqt_sessionoptions.h
+++ b/src/groups/bmq/bmqt/bmqt_sessionoptions.h
@@ -240,7 +240,7 @@ class SessionOptions {
     /// When the session channel is at high watermark, block `post` calls for
     /// up to this value until the channel returns to low watermark and drains
     /// buffered data.
-    bsls::TimeInterval d_channelWriteTimeoutMs;
+    bsls::TimeInterval d_channelWriteTimeout;
 
   public:
     // TRAITS
@@ -573,7 +573,7 @@ SessionOptions::setUserAgentPrefix(bsl::string_view value)
 inline SessionOptions&
 SessionOptions::setChannelWriteTimeout(const bsls::TimeInterval& value)
 {
-    d_channelWriteTimeoutMs = value;
+    d_channelWriteTimeout = value;
 
     return *this;
 }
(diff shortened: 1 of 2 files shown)


src/groups/bmq/bmqt/bmqt_sessionoptions.h
```diff
@@ -327,6 +334,10 @@ class SessionOptions {
     /// characters and `value.size() < 128`.
     SessionOptions& setUserAgentPrefix(bsl::string_view value);
 
+    /// Set the timeout in ms to block `post` when at high watermark.
+    /// Zero means no blocking.
+    SessionOptions& setChannelWriteTimeoutMs(unsigned int ms);
```
678098: The arg type is not consistent with types in other timeout setters in this source:
```
    /// Set the timeout for connecting to the broker to the specified `value`.
    SessionOptions& setConnectTimeout(const bsls::TimeInterval& value);

    /// Set the timeout for disconnecting from the broker to the specified
    /// `value`.
    SessionOptions& setDisconnectTimeout(const bsls::TimeInterval& value);

    /// Set the timeout for opening a queue to the specified `value`.
    SessionOptions& setOpenQueueTimeout(const bsls::TimeInterval& value);
```

dorjesinpo: True.  I would prefer to avoid unnecessary dependency on bde but for the sake of consistency, let's change it

678098: I agree that extra dependencies are worth avoiding, however, we already have it here.
And this generic time type is more flexible

src/groups/bmq/bmqt/bmqt_sessionoptions.h
```diff
@@ -240,7 +240,7 @@ class SessionOptions {
     /// When the session channel is at high watermark, block `post` calls for
     /// up to this value until the channel returns to low watermark and drains
     /// buffered data.
-    unsigned int d_channelWriteTimeoutMs;
+    bsls::TimeInterval d_channelWriteTimeoutMs;
```
pniedzielski: Should this guy still be `d_channelWriteTimeoutMs` with the `Ms` suffix still?

pniedzielski (approved):

PR #1106, opened 2026-02-12
Fix[mqbs::FileStore]: SAN issue on destruction

678098:
The data race happens on broker shutdown.

We stop all bmqbrkr components in order. First we destruct `mqbc::StorageManager`, after this we stop `Dispatcher`. When we destruct `mqbc::StorageManager`, we also destruct all `FileStore`s owned by it. We do it from the main thread, and at the same time `Dispatcher` might enqueue an event to `FileStore` being destructed through flush list.

The `mqbblp::StorageManager` doesn't have this issue

The safe solution is to disable `flushClients` on `shutdown`.

### DIFF
--- a/docker/sanitizers/build_sanitizer.sh
+++ b/docker/sanitizers/build_sanitizer.sh
@@ -84,6 +84,14 @@ ln -sf /usr/bin/llvm-symbolizer-${LLVM_VERSION} /usr/bin/llvm-symbolizer
 # Set some initial constants
 PARALLELISM=8
 
+# Helper function to print disk usage
+print_disk_usage() {
+    local component="${1:-unknown}"
+    echo "=== Disk usage after building ${component} ==="
+    df -h /
+    echo "================================================"
+}
+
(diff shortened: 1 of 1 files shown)


src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -523,6 +523,11 @@ void Application::stop()
 
     // STOP everything.
 
+    // Dispatcher is running while some other components are being stopped
+    // and destructed, make sure that Dispatcher will not call `flush` for
+    // any DispatcherClient that might be in the middle of its destruction.
+    d_dispatcher_mp->disableFlushClients();
```
678098, line 529: This is not fully thread-safe with `AtomicBool`, because `flushClients` might start executing just before disabling flush and might take a long time.

The safer solution is to use `AtomicGate`, it requires this PR to be merged:
https://github.com/bloomberg/blazingmq/pull/1109

PR #1107, opened 2026-02-12
Cleanup[mqba::Dispatcher]: get rid of pointless `vector::reserve`

678098:


src/groups/mqb/mqba/mqba_dispatcher.cpp
```diff
@@ -173,39 +173,6 @@ Dispatcher::DispatcherContext::DispatcherContext(
...
-    // executed by the *DISPATCHER* thread
-
-    // Resize the 'd_flushList' vector for that specified 'processorId', if
-    // needed, to ensure it has enough space to hold all clients associated to
-    // that processorId.
-    DispatcherContext& context = *(d_owner_p->d_contexts[d_type]);
-
-    int count = context.d_loadBalancer.clientsCountForProcessor(d_processorId);
-    if (static_cast<int>(context.d_flushList[d_processorId].capacity()) <
-        count) {
```
678098, line 204: This condition is super strange, it duplicates the logics in `vector::reserve`

src/groups/mqb/mqba/mqba_dispatcher.cpp
```diff
@@ -173,39 +173,6 @@ Dispatcher::DispatcherContext::DispatcherContext(
...
-void Dispatcher::OnNewClientFunctor::operator()() const
-{
-    // executed by the *DISPATCHER* thread
-
-    // Resize the 'd_flushList' vector for that specified 'processorId', if
-    // needed, to ensure it has enough space to hold all clients associated to
-    // that processorId.
-    DispatcherContext& context = *(d_owner_p->d_contexts[d_type]);
-
-    int count = context.d_loadBalancer.clientsCountForProcessor(d_processorId);
```
678098, line 202: We don't save much from reserving this, we spend much more enqueueing lots of events, and in the worst case we do `reserve` on every +1 dispatcher client

Just let the vector resize itself on flush list updates, it will allocate enough memory

678098: Also note that `clientsCountForProcessor` locks a mutex

PR #1109, opened 2026-02-12
Refactor[BMQ,MQB]: move `AtomicGate`/`GateKeeper` to `bmqu`

678098:
- `AtomicGate`/`GateKeeper` do not belong to cluster node session, they might be a separate component.
- Introduce UTs for these classes.

### DIFF
--- a/src/groups/bmq/bmqu/bmqu_atomicgate.cpp
+++ b/src/groups/bmq/bmqu/bmqu_atomicgate.cpp
@@ -1,0 +1,48 @@
+// Copyright 2026 Bloomberg Finance L.P.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// bmqu_atomicgate.cpp                                                -*-C++-*-
+#include <bmqu_atomicgate.h>
+
+#include <bmqscm_version.h>
+
+// BDE
+#include <bslmt_threadutil.h>
+
+namespace BloombergLP {
+namespace bmqu {
+
+// ----------------
+// class AtomicGate
+// ----------------
+
+// MANIPULATORS
+void AtomicGate::closeAndDrain()
+{
+    int result = d_value.add(e_CLOSE);
+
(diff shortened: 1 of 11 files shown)


src/groups/bmq/bmqu/bmqu_atomicgate.h
```diff
@@ -0,0 +1,259 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// bmqu_atomicgate.h                                                  -*-C++-*-
+#ifndef INCLUDED_BMQU_ATOMICGATE
+#define INCLUDED_BMQU_ATOMICGATE
+
+//@PURPOSE: Provide a thread-safe binary state (gate) mechanism.
```
pniedzielski: For new components, let's use Doxygen style:
https://github.com/bloomberg/blazingmq/blob/5dc1e54deb9b26885f2fdf22a90933a80930bf64/src/groups/bmq/bmqt/bmqt_compressionalgorithmtype.h#L20

so `///`, `@PURPOSE` -> `@brief`, and the descriptions from `@CLASSES` become `@brief` on each of the classes' Doxygen comments.

src/groups/bmq/bmqu/bmqu_atomicgate.h
```diff
@@ -0,0 +1,259 @@
...
+    class Status {
+      private:
+        // DATA
+        AtomicGate& d_gate;
+        bool        d_isOpen;
+
+      private:
+        // NOT IMPLEMENTED
+        Status(const Status& other) BSLS_KEYWORD_DELETED;
+        Status& operator=(const Status& other) BSLS_KEYWORD_DELETED;
```
pniedzielski, line 114: Good catch with this assignment operator needing to be deleted too.

src/groups/bmq/bmqu/bmqu_atomicgate.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+namespace BloombergLP {
+namespace bmqu {
+
+// ----------------
+// class AtomicGate
+// ----------------
+
+// MANIPULATORS
+void AtomicGate::closeAndDrain()
```
pniedzielski, line 32: This manipulator was changed from inline to non-inline.  Is there a particular benefit to making sure this is not inline?

pniedzielski (changes requested):

src/groups/bmq/bmqu/bmqu_atomicgate.cpp
678098: There are 2 main benefits.
1. We don't include `bslmt_threadutil` transitively from the header.
2. As you can see, there is a `while` loop in the implementation. In the worst case this loop might spin forever, and it is better to have `closeAndDrain` visible in stack trace than to have this inlined somewhere.

`closeAndDrain` is a slow operation anyway, and we typically call it when we want to shutdown a resource, so we don't lose much if this is not inlined.

PR #1112, opened 2026-02-12
Cleanup[bmqc::MultiQueueThreadPool]: remove unused queue context

678098:
`MultiQueueThreadPool_QueueCreatorRet` is never populated and never used in `mqba::Dispatcher`. It is also marked as `BSLA_UNUSED`. We don't want to provide it on any dispatcher queue event, and the interface can be simplified.

### DIFF
--- a/src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
+++ b/src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
@@ -23,7 +23,6 @@
 //@CLASSES:
 //  bmqc::MultiQueueThreadPool: Queues processed by a thread pool
 //  bmqc::MultiQueueThreadPoolConfig: Configuration for a MQTP
-//  bmqc::MultiQueueThreadPool_QueueCreatorRet: MQTP queueCreator args
 //
 //@DESCRIPTION: This component defines a mechanism,
 // 'bmqc::MultiQueueThreadPool', which encapsulates the common pattern of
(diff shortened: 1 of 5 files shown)


src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -698,9 +628,7 @@ inline void MultiQueueThreadPool<TYPE>::processQueue(int queue)
         const int popRet = info.d_queue_p->tryPopFront(&event);
         if (popRet != 0) {
             // Queue is empty
-            d_config.d_eventCallbackFn(queue,
-                                       info.d_context_p.get(),
-                                       d_queueEmptyEvent_sp);
+            d_config.d_eventCallbackFn(queue, d_queueEmptyEvent_sp);
```
678098, line 631: This will not only be faster, it will also simplify our stack traces

src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.t.cpp
```diff
@@ -79,32 +79,35 @@ struct TestItem {
...
 {
     if (event) {
         // Non-empty event pointer means user event
-        bsl::vector<int>* vec = reinterpret_cast<bsl::vector<int>*>(context);
-        vec->push_back(event->value());
+
+        bsl::map<int, bsl::vector<int> >::iterator iter =
+            queueContextMap->find(queueId);
+        BSLS_ASSERT_OPT(iter != queueContextMap->end());
+        iter->second.push_back(event->value());
```
678098, line 105: This change suggests that the same behaviour can be achieved with how we `Bind`, not necessary to require any context in interface

PR #1114, opened 2026-02-13
Feat[bmq]: Add channel watermark session events

dorjesinpo:
Two new `SessionEventType`:

`e_CHANNEL_LOW_WATERMARK`
`e_CHANNEL_HIGH_WATERMARK`


pniedzielski (changes requested):
See above^

PR #1118, opened 2026-02-14
Fix[BMQ]: use const thread safe `messageRef`

678098:
From BDE docs: https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/classball_1_1RecordAttributes.html#abc829c59980832f27efdd84eccff2b6d
**message()** Return the message attribute of this record attributes object. Note that this method will return a truncated message if it contains embedded null ('\0') characters; see `messageRef` for an alternative to this method. **Warning**: **This method is not const thread-safe**, and cannot be safely called concurrently. It may modify the stream buffer returned by `messageStreamBuf`.

Using non const thread safe `message()` in `bmqtsk::AlarmLog` causes sanitizer failures in CI:

east1            00:40:17.136 (140564709873344) INFO     blazingmq.tsk.bmqbrkr.stderr bmqproc.py:130     #0 std::__1::basic_streambuf<char, std::__1::char_traits<char>>::sputc[abi:ne180100](char) /blazingmq/deps/srcs/llvm-project/runtimes/cmake.bld/include/c++/v1/streambuf:214:13 (bmqbrkr.tsk+0x22c53be) (BuildId: 926df94d4a58580e1c6f8f91423c4fe95d3f9efe)

east1            00:40:17.136 (140564709873344) INFO     blazingmq.tsk.bmqbrkr.stderr bmqproc.py:130     #1 BloombergLP::ball::RecordAttributes::message() const /blazingmq/deps/srcs/bde/groups/bal/ball/ball_recordattributes.cpp:111:19 (bmqbrkr.tsk+0x3ad0ea8) (BuildId: 926df94d4a58580e1c6f8f91423c4fe95d3f9efe)

east1            00:40:17.136 (140564709873344) INFO     blazingmq.tsk.bmqbrkr.stderr bmqproc.py:130     #2 BloombergLP::bmqtsk::AlarmLog::publish(BloombergLP::ball::Record const&, BloombergLP::ball::Context const&) /blazingmq/src/groups/bmq/bmqtsk/bmqtsk_alarmlog.cpp:108:55 (bmqbrkr.tsk+0x387f0f7) (BuildId: 926df94d4a58580e1c6f8f91423c4fe95d3f9efe)

### DIFF
--- a/src/groups/bmq/bmqtst/bmqtst_scopedlogobserver.cpp
+++ b/src/groups/bmq/bmqtst/bmqtst_scopedlogobserver.cpp
@@ -151,7 +151,7 @@ bool ScopedLogObserverUtil::recordMessageMatch(const ball::Record& record,
     BSLS_ASSERT(rc == 0);
     BSLS_ASSERT(regex.isPrepared() == true);
 
-    const bslstl::StringRef& msg = record.fixedFields().message();
+    const bslstl::StringRef msg = record.fixedFields().messageRef();
 
     rc = regex.match(msg.data(), msg.length());
 


pniedzielski (approved):
Nice.  `message()` is a strange function, an accessor that mutates.

PR #1120, opened 2026-02-17
Fix: Mac OSX warnings

pniedzielski:
Please see each commit individually.  Every unhappy family is unhappy in its own way; so too must every fixed warning be fixed in its own way.

### DIFF
--- a/src/groups/bmq/bmqt/bmqt_sessioneventtype.cpp
+++ b/src/groups/bmq/bmqt/bmqt_sessioneventtype.cpp
@@ -70,6 +70,8 @@ const char* SessionEventType::toAscii(SessionEventType::Enum value)
         BMQT_CASE(ERROR)
         BMQT_CASE(TIMEOUT)
         BMQT_CASE(CANCELED)
+        BMQT_CASE(CHANNEL_LOW_WATERMARK)
+        BMQT_CASE(CHANNEL_HIGH_WATERMARK)
     default: return "(* UNKNOWN *)";
     }
 
@@ -106,6 +108,8 @@ bool SessionEventType::fromAscii(SessionEventType::Enum*  out,
     BMQT_CHECKVALUE(ERROR)
     BMQT_CHECKVALUE(TIMEOUT)
     BMQT_CHECKVALUE(CANCELED)
+    BMQT_CHECKVALUE(CHANNEL_LOW_WATERMARK)
+    BMQT_CHECKVALUE(CHANNEL_HIGH_WATERMARK)
 
     // Invalid string
     return false;
--- a/src/groups/bmq/bmqt/bmqt_sessioneventtype.t.cpp
+++ b/src/groups/bmq/bmqt/bmqt_sessioneventtype.t.cpp
@@ -99,6 +99,10 @@ static void test2_printTest()
         {bmqt::SessionEventType::e_ERROR, "ERROR"},
         {bmqt::SessionEventType::e_TIMEOUT, "TIMEOUT"},
         {bmqt::SessionEventType::e_CANCELED, "CANCELED"},
+        {bmqt::SessionEventType::e_CHANNEL_LOW_WATERMARK,
+         "CHANNEL_LOW_WATERMARK"},
+        {bmqt::SessionEventType::e_CHANNEL_HIGH_WATERMARK,
+         "CHANNEL_HIGH_WATERMARK"},
         {static_cast<bmqt::SessionEventType::Enum>(-1234), "(* UNKNOWN *)"},
     };
 


src/groups/bmq/bmqp/bmqp_messageguidgenerator.t.cpp
```diff
@@ -2290,7 +2290,12 @@ int main(int argc, char* argv[])
     case 3: test3_multithreadUseIP(); break;
     case 2: test2_extract(); break;
     case 1: test1_breathingTest(); break;
-    case -1: testN1_decode(); break;
+    case -1:
+        BMQTST_BENCHMARK_WITH_ARGS(testN1_decode,
```
678098: I think we should remove google-benchmark variant of this function, because this is not actually a benchmark, it is a strangely placed debug utility.
Basically it might be used by calling `./bmqp_messageguidgenerator.t -1` and it asks for `MessageGUID` to extract different component from it, including timestamp etc

678098: Might put a comment here

src/groups/bmq/bmqp/bmqp_pusheventbuilder.t.cpp
```diff
@@ -223,8 +223,8 @@ appendMessage(size_t                    iteration,
 #endif
 
     bdlbb::Blob payload(bufferFactory, allocator);
-    const int   blobSize = generateRandomInteger(0, 1024);
 #ifdef BMQ_ENABLE_MSG_GROUPID
```
678098, line 226: We might remove `BMQ_ENABLE_MSG_GROUPID` and all code blocks enabled by it completely by the way.
We do not support this thing anymore.

src/groups/mqb/mqbblp/mqbblp_recoverymanager.cpp
```diff
@@ -1577,12 +1577,8 @@ int RecoveryManager::sendFile(RequestContext*                   context,
         break;  // BREAK
 
     case bmqp::RecoveryFileChunkType::e_UNDEFINED:
```
678098: Would be nice to put `BSLA_FALLTHROUGH` here then

src/groups/mqb/mqbnet/mqbnet_dummysession.cpp
```diff
@@ -55,7 +55,7 @@ void DummySession::tearDown(BSLA_UNUSED const bsl::shared_ptr<void>& handle,
     // NOTHING
 }
 
-void DummySession::initiateShutdown(const ShutdownCb& callback)
+void DummySession::initiateShutdown(BSLA_UNUSED const ShutdownCb& callback)
```
678098, line 58:
```suggestion
void DummySession::initiateShutdown(BSLA_MAYBE_UNUSED const ShutdownCb& callback)
```

The variant without `maybe` is deprecated

src/groups/mqb/mqbu/mqbu_messageguidutil.t.cpp
```diff
@@ -1337,7 +1337,12 @@ int main(int argc, char* argv[])
     case 3: test3_print(); break;
     case 2: test2_multithread(); break;
     case 1: test1_breathingTest(); break;
-    case -1: testN1_decode(); break;
+    case -1:
+        BMQTST_BENCHMARK_WITH_ARGS(testN1_decode,
```
678098: The same situation with `bmqp_messageguidutil`, better to remove google-benchmark variant
Not sure what are the differences in util's implementations

src/groups/bmq/bmqp/bmqp_pusheventbuilder.t.cpp
pniedzielski: I agree with this.  I think it's a bigger change, and it's in the protocol in a way we can't naively rip out.  Let's keep it in mind, I think it would clean up lots of the code (and make the build of the Python library easier, as well).

src/groups/mqb/mqbnet/mqbnet_dummysession.cpp
pniedzielski: While I'm at it I'll clean up the other `BSLA_UNUSED` in this file.

pniedzielski: Actually, I'll make a quick separate PR that `s/BSLA_UNUSED/BSLA_MAYBE_UNUSED/g`.  I think that will be better and easier to review?

pniedzielski: Work started in #1142, will update after this is merged.

678098: > Actually, I'll make a quick separate PR that `s/BSLA_UNUSED/BSLA_MAYBE_UNUSED/g`. I think that will be better and easier to review?

Yes, sure

src/groups/bmq/bmqp/bmqp_messageguidgenerator.t.cpp
```diff
@@ -1726,108 +1726,6 @@ static void testN10_hashCollisionsComparison()
 // Begin Benchmarking Tests
 
 #ifdef BMQTST_BENCHMARK_ENABLED
-static void testN1_decode_GoogleBenchmark(benchmark::State& state)
```
678098, line 1729: By the way, I think we can just drop google-benchmark tests entirely.
Will do it one day
https://github.com/bloomberg/blazingmq/issues/1148

678098 (approved):
Looks good

PR #1121, opened 2026-02-17
Feat[BMQ]: `bsl::string_view` in `setPropertyAsString`

dorjesinpo:
`bmqa::MessageProperties` has a set of setPropertyAs* functions, each taking the property name as a const `bsl::string&` parameter. This is fine if the caller of the function already happens to have a bsl::string object. If not, this is inconvenient and inefficient. Not only a temporary string object is created, but if the caller happens to have the string as a `bsl::string_view`, which is not implicitly-convertible to `bsl::string`, a temporary string object has to be created on the stack prior to calling the function.

A simple solution would be to replace the usage of `const bsl::string&`, with `bsl::string_view`. Event though this technically alters the function signature, we do not expect any real compatibility issue, as `bsl::string_view` is implicitly-constructible from a `bsl::string`.

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_messageproperties.h
+++ b/src/groups/bmq/bmqp/bmqp_messageproperties.h
@@ -32,7 +32,9 @@
 #include <bmqt_resultcode.h>
 
 // BDE
+#include <bdlb_transparentequalto.h>
 #include <bdlb_transparenthash.h>
+#include <bdlb_transparentless.h>
 #include <bdlb_variant.h>
 #include <bdlbb_blob.h>
 #include <bdld_datum.h>
@@ -84,7 +86,7 @@ class MessageProperties_Schema BSLS_KEYWORD_FINAL {
     typedef bsl::unordered_map<bsl::string,
                                int,
                                bdlb::TransparentHash,
-                               bsl::equal_to<> >
+                               bdlb::TransparentEqualTo>
         PropertyMap;
 
   private:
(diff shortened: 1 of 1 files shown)


src/groups/bmq/bmqp/bmqp_messageproperties.h
```diff
@@ -390,7 +395,8 @@ class MessageProperties {
     int setPropertyAsShort(const bsl::string& name, short value);
     int setPropertyAsInt32(const bsl::string& name, int value);
     int setPropertyAsInt64(const bsl::string& name, bsls::Types::Int64 value);
-    int setPropertyAsString(const bsl::string& name, const bsl::string& value);
+    int setPropertyAsString(const bsl::string&      name,
+                            const bsl::string_view& value);
```
678098:
```suggestion
                            bsl::string_view        value);
```

src/groups/bmq/bmqp/bmqp_messageproperties.h
```diff
@@ -875,8 +897,9 @@ inline int MessageProperties::setPropertyAsInt64(const bsl::string& name,
     return setProperty(name, value);
 }
 
-inline int MessageProperties::setPropertyAsString(const bsl::string& name,
-                                                  const bsl::string& value)
+inline int
+MessageProperties::setPropertyAsString(const bsl::string&      name,
+                                       const bsl::string_view& value)
```
678098:
```suggestion
                                       bsl::string_view        value)
```

src/groups/bmq/bmqp/bmqp_messageproperties.h
```diff
@@ -731,6 +744,15 @@ MessageProperties::getPropertyValueSize(const bsl::string& value) const
     return static_cast<int>(value.length());
 }
 
+template <>
+inline int
+MessageProperties::getPropertyValueSize(const bsl::string_view& value) const
```
678098, line 753:
```suggestion
MessageProperties::getPropertyValueSize(bsl::string_view value) const
```

pniedzielski (approved):

PR #1129, opened 2026-02-19
Refactor: move code to `bmqtst_table` component

678098:
`Table` is a light-weight component that might be used in other tests and benchmarks, it is good to move it to its own component.
Other changes:
- Add allocator to `Table` constructor and use it.
- Use `string_view` in interfaces.
- Doxygen-style comments.
- Usage example.

### DIFF
--- a/src/groups/bmq/bmqtst/bmqtst_table.h
+++ b/src/groups/bmq/bmqtst/bmqtst_table.h
@@ -1,0 +1,190 @@
+// Copyright 2026 Bloomberg Finance L.P.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


src/groups/bmq/bmqtst/bmqtst_table.h
```diff
@@ -0,0 +1,190 @@
...
+#include <bsls_types.h>
+
+namespace BloombergLP {
+namespace bmqtst {
+
+// ===========
+// class Table
+// ===========
+
+/// @class Table
```
pniedzielski, line 71: I think Doxygen will automatically pick up that this is documenting the `class Table` right below it.  But I don't have a strong preference whether we should have this line or not.

678098: Removed it in the next PR
https://github.com/bloomberg/blazingmq/pull/1130

PR #1130, opened 2026-02-19
IT: fix reruns hook

678098:
- Reruns hook causes IT errors when reruns are disabled
- Also fix docs from the previous PR

Should fix CI errors like
`ERROR test_admin_res_log.py::test_adminsession_res_log_purge[multi_node_fsm-strong_consistency] - KeyError: <_pytest.stash.StashKey object at 0x7f75c1cd0400>`

### DIFF
--- a/src/python/blazingmq/dev/it/testhooks.py
+++ b/src/python/blazingmq/dev/it/testhooks.py
@@ -29,12 +29,14 @@ def is_test_reported_failed(request):
     To decide if test failed
     """
 
-    report = request.node.stash[PHASE_REPORT_KEY]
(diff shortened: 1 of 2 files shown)


src/python/blazingmq/dev/it/testhooks.py
```diff
@@ -29,12 +29,14 @@ def is_test_reported_failed(request):
     To decide if test failed
     """
 
-    report = request.node.stash[PHASE_REPORT_KEY]
```
678098, line 32: This doesn't exist when reruns are disabled

PR #1132, opened 2026-02-20
IT: fix slowness due to incorrect regex checks

678098:
Some commands that we check in ITs have stars `*` or possibly other symbols conflicting with regex syntax. Due to this, regex checks were always failing in some cases, leading to long wait times in integration tests.

Also:
- Better docs and logs
- Always return int `rc`
- Check for int `rc`, not for `None`
- Add type hints and enforce keyword-only args

Should fix many of these slow 200s+ test cases

Screenshot 2026-02-20 at 13 34 05

### DIFF
--- a/src/python/blazingmq/dev/it/process/broker.py
+++ b/src/python/blazingmq/dev/it/process/broker.py
@@ -26,6 +26,7 @@ with a broker: sending commands, waiting until a leader is elected, etc.
 import itertools
 import os
 from pathlib import Path
+import re
 import signal
 
 from typing import Optional, TypeVar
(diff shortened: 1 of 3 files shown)


src/python/blazingmq/dev/it/process/broker.py
```diff
@@ -42,7 +43,7 @@
 from blazingmq.dev.it.util import internal_use, ListContextManager, Queue
 
 BLOCK_TIMEOUT = 60
-START_TIMEOUT = 120
+START_TIMEOUT = 20
```
678098, line 46: Also why should we wait 2 minutes for the broker to start?
It is not reasonable.

PR #1135, opened 2026-02-20
Fix[mqba::DomainManager]: clear cache before doing anything with DOMAIN RECONFIGURE

678098:
We need to clear cache **before** calling `locateOrCreateDomain`, because it might be using cached incorrect configuration.

This causes failures in IT `test_invalid_configuration` after I fixed regex timeouts.
In that IT, we first start brokers with incorrect configuration, try to open a queue (and fail), and try to reconfigure with good configuration.

The problem is that bad domain still exists in cache.

### DIFF
--- a/src/groups/mqb/mqba/mqba_domainmanager.cpp
+++ b/src/groups/mqb/mqba/mqba_domainmanager.cpp
@@ -677,6 +677,11 @@ int DomainManager::processCommand(mqbcmd::DomainsResult*        result,
     else if (command.isReconfigureValue()) {
         const bsl::string& name = command.reconfigure().domain();
 
+        // Clear the cache before attempting to locate or create the domain
+        // so that fresh configuration is fetched from disk, even if the
+        // domain doesn't exist yet due to previous configuration failures.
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqba/mqba_domainmanager.cpp
```diff
@@ -677,6 +677,11 @@ int DomainManager::processCommand(mqbcmd::DomainsResult*        result,
...
         const bsl::string& name = command.reconfigure().domain();
 
+        // Clear the cache before attempting to locate or create the domain
+        // so that fresh configuration is fetched from disk, even if the
+        // domain doesn't exist yet due to previous configuration failures.
+        d_configProvider_p->clearCache(name);
+
         DomainSp domainSp;
 
         if (0 != locateOrCreateDomain(&domainSp, name)) {
```
678098, line 687: `locateOrCreateDomain` might attempt to create a domain (if it doesn't exist) using the cached configuration. The cached configuration might be incorrect and this causes the entire reconfiguration to fail.

PR #1150, opened 2026-02-24
wip! Refactor: Users of `mqbblp::Queue` should not be allowed to put it in an invalid state

pniedzielski:


src/groups/mqb/mqbblp/mqbblp_queue.cpp
```diff
@@ -137,14 +204,11 @@ void Queue::configureDispatched(bool isReconfigure)
 
     bmqu::MemOutStream throwaway(d_allocator_p);
 
-    if (d_localQueue_mp) {
-        d_localQueue_mp->configure(throwaway, isReconfigure);
-    }
-    else if (d_remoteQueue_mp) {
-        d_remoteQueue_mp->configure(throwaway, isReconfigure);
+    if (d_impl.is<LocalQueueMp>()) {
+        d_impl.the<LocalQueueMp>()->configure(throwaway, isReconfigure);
```
678098, line 200: I like this `is<>`, `the<>` syntax.
Will probably use the same in `DispatcherEvent` refactor PR (I used `castTo<>` there that is less elegant)

PR #1163, opened 2026-02-26
Fix[mqbc::StorageMgr]: Determine data destinations using historic seq nums

kaikulimu:
Several improvements to the Partition FSM (omitting the lesser changes):
1. [8b257db](https://github.com/bloomberg/blazingmq/pull/1163/commits/8b257db6661f33a7d1a437ebb969059b9fa900e4) `mqbc::RecoveryMgr: Store historic highest sequence number per leaseId`:
If primary has histroic seq nums (1,18), (2,9) and replica send (1,24), primary would be able to detect a data gap from (1,19) to (1,24). Previously, this wasn't possible.
2. [4ea44e5](https://github.com/bloomberg/blazingmq/pull/1163/commits/4ea44e59c24f9a9086a7c7cbb7bb8da614d19121) `mqbc::StorageMgr: Determine data destinations before push or drop`:
Based on historic seq nums and other known information, determine for each replica whether to send it `ReplicaDataRequestPush` or `ReplicaDataRequestDrop`.
3. [09407bf](https://github.com/bloomberg/blazingmq/pull/1163/commits/09407bf6bf6abf7c79fa33c801ab16d61819c59a) `mqbc::RecoveryMgr: Alarm or terminate upon failure to send data chunks`:
More fail-fast and less silent failures.


src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
```diff
@@ -1690,6 +1790,99 @@ int RecoveryManager::recoverSeqNum(
     return rc_SUCCESS;
 }
 
+int RecoveryManager::initHistoricHighestSeqNums(int partitionId)
+{
+    // executed by the *QUEUE DISPATCHER* thread associated with 'partitionId'
+
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(partitionId >= 0 &&
```
dorjesinpo: We have at least 10 occurrences of 
```
partitionId >= 0 &&
        partitionId <
            clusterData->clusterConfig().partitionConfig().numPartitions()
```
Could be an inline method

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
```diff
@@ -1690,6 +1790,99 @@ int RecoveryManager::recoverSeqNum(
...
+                     partitionId <
+                         d_clusterConfig.partitionConfig().numPartitions());
+
+    enum rcEnum { rc_SUCCESS = 0, rc_JOURNAL_ITERATOR_FAILURE = -1 };
+
+    RecoveryContext& recoveryCtx = d_recoveryContextVec[partitionId];
+
+    // Return early if already initialized for this partition
+    if (recoveryCtx.d_historicHighestSeqNumsInitialized) {
```
dorjesinpo: Is there a point when `d_historicHighestSeqNumsInitialized` needs reset?

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
```diff
@@ -1690,6 +1790,99 @@ int RecoveryManager::recoverSeqNum(
...
+            return rc * 10 + rc_JOURNAL_ITERATOR_FAILURE;  // RETURN
+        }
+
+        unsigned int currPrimaryLeaseId =
+            journalIt.recordHeader().primaryLeaseId();
+        bsls::Types::Uint64 currSeqNum =
+            journalIt.recordHeader().sequenceNumber();
+
+        while (1 == (rc = journalIt.nextRecord())) {
+            const mqbs::RecordHeader& recordHeader = journalIt.recordHeader();
```
dorjesinpo: duplicate block, possibly can be collapsed

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
```diff
@@ -1690,6 +1790,99 @@ int RecoveryManager::recoverSeqNum(
...
+        else if (rc != 1) {
+            return rc * 10 + rc_JOURNAL_ITERATOR_FAILURE;  // RETURN
+        }
+
+        unsigned int currPrimaryLeaseId =
+            journalIt.recordHeader().primaryLeaseId();
+        bsls::Types::Uint64 currSeqNum =
+            journalIt.recordHeader().sequenceNumber();
+
+        while (1 == (rc = journalIt.nextRecord())) {
```
dorjesinpo: Do we really need to iterate journal one more time?  Is there a previous iteration where this can be done?

src/groups/mqb/mqbc/mqbc_recoverymanager.h
```diff
@@ -206,6 +211,19 @@ class RecoveryManager {
...
             d_firstSyncPointAfterRolloverSeqNum;
 
+        /// Historic map of primaryLeaseId -> highest sequence number observed
+        /// for that primary.  Note that this map will *not* contain the
+        /// highest sequence number observed for the current primary.
+        ///
+        /// TODO During ReplicaDataResponsePull, we bump up our primaryLeaseId,
+        /// we need to update this map.  We will fix this in the future PR
+        /// which fixes receive data chunks logic.
+        LeaseIdToSeqNumMap d_historicHighestSeqNums;
```
dorjesinpo: the prefix `historic` (everywhere) seems redundant (making names unnecessary long).  Historic or not, this is Highest SeqNum.

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
dorjesinpo: Why do we need the callback `doneDataChunksCb`?
`processSendDataChunks` seems to be synchronous

src/groups/mqb/mqbc/mqbc_recoverymanager.h
```diff
@@ -240,17 +253,21 @@ class RecoveryManager {
...
         PartitionDoneSendDataChunksCb;
 
     /// Vector per partition of `RecoveryContext`.
     typedef bsl::vector<RecoveryContext> RecoveryContextVec;
 
     // This callback is only used when the self node is a replica.
     bsl::function<
-        void(int partitionId, mqbnet::ClusterNode* destination, int status)>
+        void(int partitionId, mqbnet::ClusterNode* destination, int* status)>
         PartitionDoneRcvDataChunksCb;
```
dorjesinpo: seems unused

src/groups/mqb/mqbc/mqbc_storagemanager.h
```diff
@@ -338,6 +368,15 @@ class StorageManager BSLS_KEYWORD_FINAL
...
     NodeToSeqNumCtxMapPartitionVec d_nodeToSeqNumCtxMapVec;
 
+    /// Vector indexed by partitionId, of temporary data push/drop destinations
+    /// for the immediate next sends, as per the logic of the Partition FSMs.
+    /// We need to recalculate the destinations before each send.
+    ///
+    /// THREAD: Except during the ctor, the i-th index of this data member
+    ///         **must** be accessed in the associated Queue dispatcher thread
+    ///         for the i-th partitionId.
+    bsl::vector<DataDestinations> d_tempDataDestinations;
```
dorjesinpo: May be possible to avoid have temp data member.  It is always the same sequence
`determineDataDestinations_replicaDataRequestPush_replicaDataRequestDrop`
So, DataDestinations can be on stack.
(Does not mean everything is in one function)

dorjesinpo (changes requested):

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
kaikulimu: For primary, no.

But for replica who transitions as new primary, yes. The cleanest way would be that when a node detects a jump in lease id, it records the previous (lease id, seqNum) into `d_historicHighestSeqNums`, meaning that `d_historicHighestSeqNums` will be a running tally of latest seqNums instead of a one-time initialization construct. Will need a bit of refactoring.

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
kaikulimu: Did a bit of simplifying

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
kaikulimu: There is an iteration during `do_openRecoveryFileSet -> RecoveryManager::openRecoveryFileSet` earlier, but it is a backward iteration to find the last valid journal/data/qlist record positions, then breaks early once all three are found.

Later on, there is `do_attemptOpenStorage -> FileStore::open -> FileStore::openInRecoveryMode -> FileStore::recoverMessages` which has two reverse iterations through the entire journal file. I will merge `initHighestSeqNums` with `recoverMessages`, which means changing the find highest seq nums logic to also use backward iteration. The logic is actually slightly simpler with backward iteration.

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
kaikulimu: This is probably remnants of previous code transformations. And true, we don't need it to be a callback anymore, so I will simplify the code.

src/groups/mqb/mqbc/mqbc_recoverymanager.h
kaikulimu: Removed

src/groups/mqb/mqbc/mqbc_storagemanager.h
kaikulimu: Without making them one function, `determineDataDestinations` can pass on the DataDestinations via modifying the PFSM event data. However, it makes event data mutable during the action, which seems to obfuscate the code and cause more harm than good. What do you think?

kaikulimu:
Added [handling for a corner case](https://github.com/kaikulimu/blazingmq/blob/9f44938ab25b324db78f78a0a74176cd6a4635d0/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L2067-L2085) when checking replica sequence number against highest sequence numbers. This change is in the commit titled `mqbc::StorageMgr: Drop replica storage with unknown old lease ID`.

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
kaikulimu: Implemented

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
kaikulimu: I moved `d_highestSeqNums` under `FileStore`, who initializes it during `recoverMessages`. And when bumping lease id, it also updates the map. Therefore, it is now being kept up-to-date, and there is no need for the `d_highestSeqNumsInitialized` variable.

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
```diff
@@ -527,12 +533,21 @@ int RecoveryManager::processSendDataChunks(
...
+        d_qListAware ? mappedQlistFd.get() : 0);
     if (rc != 0) {
+        BMQTSK_ALARMLOG_ALARM("FILE_IO")
+            << d_clusterData.identity().description() << " Partition ["
+            << partitionId << "]: "
+            << "While sending data chunks, failed to load file descriptors, "
+               "rc: "
+            << rc << BMQTSK_ALARMLOG_END;
+
+        mqbu::ExitUtil::terminate(mqbu::ExitCode::e_RECOVERY_FAILURE);  // EXIT
```
dorjesinpo, line 547: Not clear why some errors result in `mqbu::ExitUtil::terminate` and other - in return code.

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
```diff
@@ -527,12 +533,21 @@ int RecoveryManager::processSendDataChunks(
...
+
```
dorjesinpo: Do we need `return` after `mqbu::ExitUtil::terminate`?

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -617,6 +581,34 @@ void StorageManager::processReplicaDataRequestPull(
...
 
+    mqbnet::ClusterNode* const selfNode =
+        d_clusterData_p->membership().selfNode();
+    const bmqp_ctrlmsg::PartitionSequenceNumber selfSeqNum =
+        d_nodeToSeqNumCtxMapVec.at(partitionId).at(selfNode).d_seqNum;
+    if (replicaDataRequest.endSequenceNumber() != selfSeqNum) {
+        bmqp_ctrlmsg::ControlMessage controlMsg;
+        controlMsg.rId() = message.rId();
+
+        bmqp_ctrlmsg::Status& status = controlMsg.choice().makeStatus();
```
dorjesinpo, line 1067: For the `if (d_cluster_p->isStopping())` above do we need to respond as well?

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -1383,9 +1375,16 @@ void StorageManager::do_storeSelfSeq(const EventWithData& event)
...
-    NodeSeqNumContext&   nodeSeqNumCtx =
-        d_nodeToSeqNumCtxMapVec[partitionId][selfNode];
+    NodeSeqNumContext&   nodeSeqNumCtx = d_nodeToSeqNumCtxMapVec.at(
+        partitionId)[selfNode];
     if (dataRange.second > bmqp_ctrlmsg::PartitionSequenceNumber()) {
+        // NOTE: `dataRange` is set *if and only if* the input event is
+        //       `e_REPLICA_DATA_RSPN_PULL`.  In that case, we need to update
+        //       self sequence number based on the received data range, which
+        //       is more up-to-date than what is persisted in storage or what
+        //       is reported by primary/replica state response.
```
dorjesinpo: Judging by this comment there are three:
1. sequence number based on the received REPLICA_DATA_RSPN_PULL data range
2. sequence number persisted in storage
3. sequence number reported by primary/replica state response

Does 3. mean another node?  For the same node 1 and 3 must be the same, is that correct?  Or, can a node report one sequence number in state response and another in REPLICA_DATA_RSPN_PULL data range?

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -1929,22 +1939,29 @@ void StorageManager::do_replicaDataRequestPush(const EventWithData& event)
...
+    BSLS_ASSERT_SAFE(d_partitionFSMVec[partitionId]->isSelfPrimary());
 
     if (d_recoveryManager_mp->expectedDataChunks(partitionId)) {
         BALL_LOG_INFO
             << d_clusterData_p->identity().description() << " Partition ["
             << partitionId
-            << "]: " << "Not sending ReplicaDataRequestPush to replicas yet "
-            << "because self primary is still expecting recovery data "
-            << "chunks from the up-to-date replica.";
+            << "]: " << "Not determining data destinations yet because self "
```
dorjesinpo: minor, it is easier to understand what does "sending ReplicaDataRequestPush to replicas" mean.  It is less clear what does "determining data destinations" mean

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -1984,59 +2002,169 @@ void StorageManager::do_replicaDataRequestPush(const EventWithData& event)
...
-                    bmqp_ctrlmsg::PartitionSequenceNumber() &&
-                cit->second.d_firstSyncPointAfterRolloverSeqNum !=
-                    selfFirstSyncAfterRolloverSeqNum) {
-                continue;  // CONTINUE
-            }
-
-            if (cit->second.d_seqNum <= selfSeqNum) {
-                destinationReplicas.emplace_back(source);
-            }
+            nodesToCheckBegin = cit;
```
dorjesinpo: This looks like we process only the last one in `eventDataVec`?

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -1984,59 +2002,169 @@ void StorageManager::do_replicaDataRequestPush(const EventWithData& event)
...
+                << d_clusterData_p->identity().description() << " Partition ["
+                << partitionId << "]: " << "Replica "
+                << cit->first->nodeDescription()
+                << " has partition sequence number " << nodeSeqNum
+                << ", while the highest sequence number for the same primary "
+                << "lease ID is " << highestSeqNum
+                << ", implying that it has extra irreconcilable records.  We "
+                << "need to send ReplicaDataRequestDrop to this replica.";
+
+            dataDropDestinations.emplace_back(cit);
```
dorjesinpo: just to clarify, this happens after primary fully recovers the storage including pulling if necessary?

src/groups/mqb/mqbc/mqbc_storagemanager.h
dorjesinpo: I did not mean to pass it in the event.
If this is always `determineDataDestinations_replicaDataRequestPush_replicaDataRequestDrop` sequence (is it?), then it could be one virtual API (`do_xxx`) instead of sequence of 3 keeping `tempDataDestinations` on stack and `determineDataDestinations`, `replicaDataRequestPush`, `replicaDataRequestDrop` can be implementation methods with `tempDataDestinations` as a parameter.

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -1566,6 +1575,16 @@ int FileStore::recoverMessages(QueueKeyInfoMap*     queueKeyInfoMap,
                 return rc_INVALID_SEQ_NUMBER;  // RETURN
             }
         }
+        else {
+            // Track highest sequence number per primary lease id.  Since we
+            // are iterating backward, the first record seen for a given lease
+            // id has the highest sequence number.
+            BSLS_ASSERT_SAFE(
```
dorjesinpo, line 1581: Can we encounter two different `primaryLeaseId` while enumerating?

dorjesinpo (commented):
some questions & suggestions

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
kaikulimu: Failure to access our own storage files is non-transient, so we will terminate the broker. Other failures, such as possible data gap between self and replica, or failure to write to network, can be transient, so we allow the broker to continue.

kaikulimu: Will write short comments before we call `mqbu::ExitUtil::terminate`.

src/groups/mqb/mqbc/mqbc_recoverymanager.cpp
kaikulimu: No, will update

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: It is more informative to respond with failure if we are stopping. Will add the logic.

kaikulimu: Also added failure response when stopping in `processReplicaDataRequestPush` and `processReplicaDataRequestDrop`

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: For the same node 1 and 3 must be the same.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: `do_determineDataDestinations` is always followed by `do_replicaDataRequestPush` and `do_replicaDataRequestDrop` which will print the same message anyways, so I have removed printing here.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: There's no need for special `e_REPLICA_DATA_RSPN_PULL` logic here. The normal path via `d_recoveryManager_mp->recoverSeqNum()` retrieves the correct sequence number.

For extra validation: `RecoveryManager::processReceiveDataChunks` now sets `receiveDataCtx.d_expectChunks = false` upon receiving the last data chunk, and `StorageManager::processReplicaDataResponseDispatched` checks this flag before dispatching the `PULL` response.

kaikulimu: Above set of changes are in the commit `mqbc::StorageMgr: Simplify do_storeSelfSeq; validate PULL chunks at reception`.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: Modified the logic in `do_determineDataDestinations` to fix this bug

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: Yes

src/groups/mqb/mqbs/mqbs_filestore.cpp
kaikulimu: ▎ Yes, we can encounter multiple primaryLeaseId values — the journal can span multiple
  ▎ primary terms. We iterate backward, and primaryLeaseId values are monotonically
  ▎ non-increasing (validated by the check at line 1531). Each time we see a new (lower)
  ▎ primaryLeaseId for the first time, we enter this else branch and record its highest
  ▎ sequence number. The BSLS_ASSERT_SAFE verifies we only enter this branch once per
  ▎ lease ID, which is guaranteed by the backward iteration order.
  - Mr. Claude

kaikulimu: Sorry, I didn't fully understand your question. Please clarify. But I let Claude try to answer you.

src/groups/mqb/mqbc/mqbc_storagemanager.h
kaikulimu: You're right, these three always appear together. I'll collapse them into a single `virtual do_sendDataToReplicas` that keeps `DataDestinations` on the stack and calls the three as private implementation methods. This also eliminates the `d_tempDataDestinations` data member.

kaikulimu:
@dorjesinpo Back to you for review

dorjesinpo (approved):
let's merge and work from there.  otherwise, the progress is slow

kaikulimu:
Thanks @dorjesinpo

PR #1167, opened 2026-02-27
CI: Consolidate linting workflows into one

pniedzielski:
This patch consolidates the

  - `check-mem-files`
  - `dco-check`
  - `formatting-check`
  - `license-check`
  - `pr-title-check`
  - `pylint`

workflows into a single `lint` workflow.  Previously, each workflow had only one or two jobs, which were all meant to be run in parallel on each pull request.  By including each of these jobs in a single workflow, they still run in parallel, but they do not clog up the Actions views with individual workflow runs.

In addition, this patch takes the step of running these jobs on `ubuntu-slim` runners, which are containerized runners rather than VMs, which start more quickly.  These runners are only 1 vCPU and 5 GB RAM, and jobs running on them can take a maximum of 15 minutes before they fail, which is well-suited for these lightweight checks.

There is one remaining check that could reasonably go in this workflow, the `pytest-coverage` workflow.  However, unlike all the workflows that were consolidated here, this one also runs on a cron job.  It’s not clear to me whether it should or not, but this patch does not take the step of consolidating this workflow in case we do need this.


678098 (approved):
The combined Lint view looks really good!

PR #1201, opened 2026-03-17
UT[bmqio]: fix flaky connect

678098:
There is a short time window between `listen` and `connect` when a listen handle is not ready to accept connection yet. In this case connect attempt failed return code will be returned via callback. We can safely skip this return code and retry.


src/groups/bmq/bmqio/bmqio_ntcchannelfactory.t.cpp
```diff
@@ -63,9 +63,6 @@ using namespace bmqio;
 // has been accepted only for the user to learn that it has already been
 // closed. The test that ensures that an immediately-cancelled connection
 // does not result in a CHANNEL_UP event has been adjusted to reflect this.
-//
-// 4) NTC does not report failures of individual connection retry attempts,
-// only the failure of the overall connection operation.
```
678098, line 67: This is not true, we still report individual attempt failures

kaikulimu (approved):

PR #1248, opened 2026-04-02
Fix[mqbc]: do not alarm double assignment on AppInfos mismatch

678098:
Skipping the cache when `partitionId` and `queueKey` match avoids the unnecessary unregister attempt while the normal registration flow still fixes any appInfo differences.


src/groups/mqb/mqbc/mqbc_clusterutil.cpp
```diff
@@ -1136,7 +1136,7 @@ void ClusterUtil::registerQueueInfo(ClusterState*        clusterState,
             BSLS_ASSERT_SAFE(qs);
             BSLS_ASSERT_SAFE(qs->uri() == uri);
 
-            if ((qs->equal(advisory))) {
+            if (qs->equal(advisory)) {
```
678098, line 1126: `qs->equal(advisory) == false` if `AppInfos` are different

PR #1251, opened 2026-04-02
Refactor: Trial converting a BDE style test to gtest

hallfox:
This PR takes a simple BDE test and converts it to a gtest as a trial for doing more gtest in the future (and maybe divesting away from the BDE test infrastructure we maintain).

### DIFF
--- a/docker/build_deps.sh
+++ b/docker/build_deps.sh
@@ -53,28 +53,21 @@ esac
 fetch_git() {
     local org=$1
     local repo=$2
+    local ref=$3
     mkdir -p srcs
 
     if [ -d "srcs/${repo}" ]; then
         return 0
     fi
 
-    if [ -z "${3:-}" ]; then
-        # Clone the latest 'main' branch if no specific release tag provided
-        local branch="main"
-        curl -SL "https://github.com/${org}/${repo}/archive/refs/heads/${branch}.tar.gz" | tar -xzC srcs/
-        mv "srcs/${repo}-${branch}" "srcs/${repo}"
-    else
-        local tag=$3
-        curl -SL "https://github.com/${org}/${repo}/archive/refs/tags/${tag}.tar.gz" | tar -xzC srcs/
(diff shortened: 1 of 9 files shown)


docker/build_deps.sh
```diff
@@ -53,28 +53,38 @@ esac
...
-        # Clone the latest 'main' branch if no specific release tag provided
-        local branch="main"
-        curl -SL "https://github.com/${org}/${repo}/archive/refs/heads/${branch}.tar.gz" | tar -xzC srcs/
-        mv "srcs/${repo}-${branch}" "srcs/${repo}"
-    else
-        local tag=$3
-        curl -SL "https://github.com/${org}/${repo}/archive/refs/tags/${tag}.tar.gz" | tar -xzC srcs/
-        mv "srcs/${repo}-${tag}" "srcs/${repo}"
-    fi
+    case $ref_type in
```
pniedzielski: This is fine to do.  In our [Python SDK](https://github.com/bloomberg/blazingmq-sdk-python/blob/main/bin/clone-dependencies.sh) we do `git clone --depth 1 --branch REF ...`, and despite the name, `--branch REF` there can be any ref (branch, SHA, tag).  I don't know whether `git clone` is more or less efficient than this, or if it matters.

docker/build_deps.sh
```diff
@@ -115,7 +125,34 @@ build_ntf() {
...
+        echo "Supported standard versions: cpp03, cpp11, cpp14, cpp17, cpp20, cpp23"
+        exit 1
+        ;;
+    esac
+
+    pushd srcs/googletest
+    cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+        -DCMAKE_CXX_STANDARD="$cmake_cxx_standard" \
+        -S . -B build
+    cmake --build build -j4
```
pniedzielski: NTF and BDE build with `-j8`, we could use that for consistency.

pniedzielski (changes requested):

pniedzielski (approved):

PR #1267, opened 2026-04-07
Perf: add `is_trivially_copyable` trait

678098:
Some of the data structs are safe to copy with `memcpy`

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_protocol.h
+++ b/src/groups/bmq/bmqp/bmqp_protocol.h
@@ -155,7 +155,9 @@
 
 #include <bsl_limits.h>
 #include <bsl_string.h>  // for bslstl::StringRef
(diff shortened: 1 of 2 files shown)


hallfox (commented):
The `is_trivially_copyable` changes lgtm

dorjesinpo (commented):
The initial idea was to allocate memory on demand - potentially useful with a large number of apps. 
It looks like we have a trade-off between this and slow `resize`.
Maybe, we can elaborate and still not allocate immediately but not use `resize` when there is nothing to resize?
The only use case for non empty resize is proxy when receives separate PUSH messages for the same App, same GUID.

678098:
Removed `appMessage` preallocation, kept only `is_trivially_copyable` change

hallfox (approved):

PR #1269, opened 2026-04-07
Perf[mqb]: fix `RECEIPT` flood with low nagle

678098:
# Overview

The existing implementation flushes `FileStore`'s storage event builder when the dynamic nagle count is reached. This dynamic nagle is incremented when `mqbnet::Channel` has something in its write queue (meaning, `mqbnet::Channel` is slow to send), and dynamic nagle is decremented when `mqbnet::Channel` has an empty write queue. The code assumed that slow cluster means that the `mqbnet::Channel` itself is slow, and vice versa.

The experiments show that nagle count is close to 0 in all situations: under pressure and without pressure. It happens because `mqbnet::Channel` is fast to move messages from its input queue to NTC. This also means that `FileStore` will mostly try to send only a few messages at a time, in many cases just 1 message.

Upon receiving these small messages, each replica processes it fast and sends a `RECEIPT` event. So if we have **N** nodes in a cluster, **N-1** `RECEIPT` will be sent to primary. Primary has to enqueue a dispatcher event per each `RECEIPT` from IO thread, and this results in **K x (N-1)** events per **K** `DATA` messages. `RECEIPT` optimization doesn't help here because it only happens after enqueueing these events.

This `RECEIPT` flood lowers the maximum throughput of the cluster. With 1 queue and dynamic nagle, it is possible to process at most 80k msgs/s in a strong consistency domain without NACKs. The same scenario with constant nagle (100), it is possible to process 120k msgs/s.

# Solution

Keep the default maximum nagle under pressure.
Rely on idleness to flush storage early without pressure.

# Nagle logs

**Dynamic nagle under pressure (100k msgs/s):**
**Note that a busy QUEUE dispatcher thread doesn't guarantee that mqbnet::Channel is also busy.**

Screenshot 2026-04-07 at 13 25 44

**Dynamic nagle without pressure (1k msgs/s):**

Screenshot 2026-04-07 at 13 38 12

# perf

IO thread mostly processing `RECEIPT` events:

Screenshot 2026-04-07 at 10 39 29

IO thread allocating much more callback events to handle `RECEIPT` than the number of `PUT` events.

Screenshot 2026-04-07 at 11 50 47

# Experiment 1

Strong consistency, 1 queue, 1 consumer, 1 producer.
Produce rate: 100k msgs/s.

OY axis: the number of unconfirmed messages.
OX axis: time, seconds.
Blue line: `main` (dynamic nagle)
Orange line: this PR (const nagle)

Screenshot 2026-04-07 at 14 27 58

# Experiment 2

Strong consistency, 1 queue, 1 consumer, 1 producer.
Produce rate: 120k msgs/s.
This PR.

The cluster is able to process 120k strong consistent msgs/s with good latency. This is 1.5x increase compared to `main` (that is able to process 80k msgs/s before NACKs).

Screenshot 2026-04-10 at 13 12 40

### DIFF
--- a/src/groups/mqb/mqbnet/mqbnet_cluster.h
+++ b/src/groups/mqb/mqbnet/mqbnet_cluster.h
@@ -217,7 +217,7 @@ class Cluster {
                           bmqp::EventType::Enum               type) = 0;
 
     /// Send the specified `blob` to all currently up nodes of this cluster
-    /// (exception of the current node).
+    /// (with the exception of the current node).
     virtual void broadcast(const bsl::shared_ptr<bdlbb::Blob>& blob) = 0;
 
     /// Close the channels associated to all nodes in this cluster.
--- a/src/groups/mqb/mqbnet/mqbnet_cluster.h
+++ b/src/groups/mqb/mqbnet/mqbnet_cluster.h
@@ -213,15 +213,12 @@ class Cluster {
 
     /// Write the specified `blob` of the specified `type` to all connected
     /// nodes of this cluster (with the exception of the current node).
-    /// Return the maximum number of pending items across all cluster
-    /// channels prior to broadcasting.
-    virtual int writeAll(const bsl::shared_ptr<bdlbb::Blob>& blob,
-                         bmqp::EventType::Enum               type) = 0;
+    virtual void writeAll(const bsl::shared_ptr<bdlbb::Blob>& blob,
+                          bmqp::EventType::Enum               type) = 0;
 
     /// Send the specified `blob` to all currently up nodes of this cluster
-    /// (exception of the current node).  Return the maximum number of
-    /// pending items across all cluster channels prior to broadcasting.
-    virtual int broadcast(const bsl::shared_ptr<bdlbb::Blob>& blob) = 0;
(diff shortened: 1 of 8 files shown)


src/groups/mqb/mqbnet/mqbnet_cluster.h
```diff
@@ -213,15 +213,12 @@ class Cluster {
...
-    virtual int writeAll(const bsl::shared_ptr<bdlbb::Blob>& blob,
-                         bmqp::EventType::Enum               type) = 0;
+    virtual void writeAll(const bsl::shared_ptr<bdlbb::Blob>& blob,
+                          bmqp::EventType::Enum               type) = 0;
 
     /// Send the specified `blob` to all currently up nodes of this cluster
-    /// (exception of the current node).  Return the maximum number of
-    /// pending items across all cluster channels prior to broadcasting.
-    virtual int broadcast(const bsl::shared_ptr<bdlbb::Blob>& blob) = 0;
+    /// (exception of the current node).
```
hallfox:
```suggestion
    /// (with the exception of the current node).
```

hallfox (approved):

PR #1284, opened 2026-04-09
IT: Fix and reenable flaky tests

kaikulimu:
## Summary                                                                                                                                 
  - Re-enabled three previously disabled flaky integration tests: `test_open_queue_while_cluster_blips_quorum`,
  `test_open_queue_while_cluster_blips_quorum_and_kill_all_non_leader`, and `test_force_leader_primary_divergence`.      
  - Fixed `test_force_leader_primary_divergence` by waiting for the restarted node to be ready before killing tproxies to trigger the divergence.                  
  - Split the blips-quorum test into two variants: one where the original leader wins re-election (quorum pinned via config), and one where leadership changes (strong consistency only), each with targeted recovery logic.         
  - Fixed flakiness caused by cascading leader-primary divergence: when the old primary aborts after a leadership change, the new leader can lose quorum (voter count drops below threshold), triggering yet another election and a second divergence crash — leaving too few nodes for quorum.                
  - In `mqbc::ClusterUtil`, removed explicit load uncommitted advs to merged state.

## Future Work
- Discovered that leader-primary divergence can cause a cascading leader switch. When the old leader shuts down, it could cause the new leader to lose quorum, thus a second re-election can occur. If we are unlucky, the second leader can also abort due to leader-primary divergence. Worth discussing with @chrisbeard, @dorjesinpo, and @678098 for a proper solution.

### DIFF
--- a/src/python/blazingmq/dev/configurator/configurator.py
+++ b/src/python/blazingmq/dev/configurator/configurator.py
@@ -259,6 +259,7 @@ class Configurator:
     def deploy(self, broker: Broker, site: Site) -> None:
         self.deploy_programs(broker, site)
         self.deploy_broker_config(broker, site)
+        self.deploy_clusters(broker, site)
         self.deploy_domains(broker, site)
 
     def deploy_programs(self, broker: Broker, site: Site) -> None:
@@ -318,12 +319,11 @@ class Configurator:
         if stats_dir != Path():
             site.mkdir(str(stats_dir))
 
-    def deploy_domains(self, broker: Broker, site: Site) -> None:
-        """Deploy the domains for the broker.
-        This will create the directories for the domains, and write the
-        domain definitions to JSON files in the `etc/domains` directory.
-        It will also create the clusters and proxy clusters JSON files in the
-        `etc/clusters` and `etc/proxyclusters` directories, respectively.
+    def deploy_clusters(self, broker: Broker, site: Site) -> None:
+        """Deploy the cluster configuration for the broker.
+        This will create the clusters and proxy clusters JSON files in the
+        ``etc/clusters`` and ``etc/proxyclusters`` directories, respectively,
+        and ensure that storage directories exist.
         """
 
         config_compat_mode = os.environ.get(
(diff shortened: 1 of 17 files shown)


src/python/blazingmq/dev/it/cluster_util.py
```diff
@@ -31,6 +31,22 @@
...
+    """
+    config_path = cluster.work_dir.joinpath(
+        cluster.config.nodes[node].config_dir, "clusters.json"
+    )
+    with open(config_path, "r+", encoding="utf-8") as f:
+        data = json.load(f)
+        data["myClusters"][0]["elector"]["quorum"] = quorum
+        f.seek(0)
+        json.dump(data, f, indent=4)
+        f.truncate()
```
678098: I think this should be done by using Configurator
For example, this is how domains configuration is updated in ITs:
```
    domains = {domain.name: domain for domain in cluster.configurator.domains}
    domains[
        domain_fanout
    ].definition.parameters.mode.fanout.publish_app_id_metrics = True
    cluster.deploy_domains()
```

kaikulimu: `deploy_domains` is a poor name. Will split the function into `deploy_clusters` and `deploy_domains`.

kaikulimu: Done

src/python/blazingmq/dev/configurator/configurator.py
```diff
@@ -318,12 +319,11 @@ def deploy_broker_config(self, broker: Broker, site: Site) -> None:
         if stats_dir != Path():
             site.mkdir(str(stats_dir))
 
-    def deploy_domains(self, broker: Broker, site: Site) -> None:
```
678098, line 321: There are multiple ITs where `deploy_domains` is called.
These tests rely on the previous behaviour of `deploy_domains`: it should also deploy clusters.
In this branch, clusters deploy is moved from `deploy_domains` to its own function, and ITs are not modified.
Let's wait for ITs to run, but I think this will not work.

I expect that all the places with `deploy_domains` should be modified:
```
cluster.deploy_domains()
```

=>

```
cluster.deploy_domains()
cluster.deploy_clusters()
```

678098 (commented):
Just one comment

src/python/blazingmq/dev/configurator/configurator.py
kaikulimu: I actually believe the opposite -- previous callers of `deploy_domains` only wanted to deploy domain files but accidentally deployed cluster files too. The IT run supports my hypothesis. That's why we shouldn't name functions inaccurately.

678098 (approved):

PR #1287, opened 2026-04-09
Refactor: Implement a channel factory pipeline

hallfox:
We currently manage several ChannelFactory implementations in mqbnet_tcpsessionfactory and bmqimp_application that are designed to decorate each other. The bottom of this decorator stack is the real ChannelFactory implementation, bmqio::NtcChannelFactory, and we add behaviors on top of that factory's listen() and connect() functions through classes like bmqio::ReconnectingChannelFactory, bmqio::ResolvingChannelFactory, bmqio::StatChannelFactory, bmpimp::NegotiatedChannelFactory, and bmqimp::AuthenticatedChannelFactory.

Though different components need different combinations of these decorators, their lifetimes are all managed in the same way: the channel factories are created in order from the bottom to the top of the decorator stack, then destructed in the opposite order. This is currently being managed by the class that uses these factories manually. With the introduction of a TlsChannelFactory incoming, yet another set of ChannelFactory stacks will be needed, and so we need something a little more generic.

Enter ChannelFactoryPipeline, an object whose role is to provide such a generic API for constructing and combining these deocrators in a custom order. The API is mostly driven by the addWith function, which requires a builder function to construct consecutive items in the decorator stack. Ideally in a mdoern C++ world this would just be a nice lambda, but until we get there we have a series of nasty bldf::BindUtil::bind calls. You can also directly add constructed ChannelFactory objects with the add method, but in practice this ends up not being needed all that often. Since ChannelFactoryPipeline is really just a container for several factories, it also implements the ChannelFactory interface.

### DIFF
--- a/src/groups/bmq/bmqimp/bmqimp_application.cpp
+++ b/src/groups/bmq/bmqimp/bmqimp_application.cpp
@@ -21,8 +21,13 @@
 #include <bmqex_systemexecutor.h>
 #include <bmqimp_authenticatedchannelfactory.h>
 #include <bmqimp_negotiatedchannelfactory.h>
+#include <bmqio_channelfactorypipeline.h>
 #include <bmqio_channelutil.h>
 #include <bmqio_connectoptions.h>
+#include <bmqio_reconnectingchannelfactory.h>
+#include <bmqio_resolvingchannelfactory.h>
+#include <bmqio_statchannel.h>
+#include <bmqio_statchannelfactory.h>
 #include <bmqio_status.h>
 #include <bmqio_tcpendpoint.h>
 #include <bmqma_countingallocatorutil.h>
@@ -38,6 +43,7 @@
 #include <bmqu_time.h>
 
 // BDE
+#include <ball_log.h>
 #include <bdlb_scopeexit.h>
 #include <bdlf_bind.h>
 #include <bdlf_memfn.h>
@@ -53,6 +59,7 @@
 #include <bsl_vector.h>
 #include <bslma_allocator.h>
 #include <bslma_managedptr.h>
+#include <bslmf_movableref.h>
 #include <bslmt_lockguard.h>
 #include <bslmt_mutexassert.h>
 #include <bslmt_once.h>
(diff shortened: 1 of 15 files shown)


src/groups/bmq/bmqimp/bmqimp_application.cpp
```diff
@@ -226,7 +350,15 @@ void Application::readCb(
...
-            d_authenticatedChannelFactory.processAuthenticationEvent(
+
+            // TODO(tfoxhall): This is a code smell to me, the factory should
+            // probably only be responsible for making channels. Handling
+            // request logic should probably be elsewhere.
+            AuthenticatedChannelFactory* authenticatedChannelFactory_p =
+                d_channelFactoryPipeline.get<AuthenticatedChannelFactory>();
+            BSLS_ASSERT(authenticatedChannelFactory_p);
+
+            authenticatedChannelFactory_p->processAuthenticationEvent(
```
hallfox, line 362: @pniedzielski would like your opinion on this particular case, which was not accounted for when I originally designed the pipeline. It's probably worth considering generic ways to access parts of the stack for cases like this.

src/groups/bmq/bmqio/bmqio_channelfactorypipeline.h
```diff
@@ -0,0 +1,179 @@
...
+                 bslma::ManagedPtr<OpHandle>* handle,
+                 const bmqio::ConnectOptions& options,
+                 const ResultCallback&        cb) BSLS_KEYWORD_OVERRIDE;
+
+    /// Enable this factory to start creating connections.
+    int start() BSLS_KEYWORD_OVERRIDE;
+
+    /// Stop this factory from creating connections and clean up any pending
+    /// connections.
+    void stop() BSLS_KEYWORD_OVERRIDE;
```
chrisbeard, line 151: I assume we expect `stop()` to be called before the destructor, though it's not documented. Should we call that out explicitly?

src/groups/bmq/bmqio/bmqio_channelfactorypipeline.cpp
```diff
@@ -0,0 +1,145 @@
...
+ChannelFactoryPipeline::get_allocator() const
+{
+    return d_pipeline.get_allocator();
+}
+
+const ChannelFactoryPipeline::ChannelFactorySP&
+ChannelFactoryPipeline::top() const
+{
+    return d_pipeline.back();
+}
```
chrisbeard, line 111: In practice, d_pipeline will never be empty right?

hallfox: No, though I don't know if we ever prevent the construction of an empty pipeline.

src/groups/bmq/bmqimp/bmqimp_application.cpp
hallfox: Keeping the TODO here, we agree that this behavior should probably not be on `AuthenticatedChannelFactory` but we'll fix it later.

chrisbeard (approved):

PR #1289, opened 2026-04-10
Fix[bmqp::PushMessageIterator]: validate padding byte

678098:
Similar to `PutMessageIterator`:

https://github.com/bloomberg/blazingmq/blob/04c376636ed844e6ba586ba93e6df3c6c0510f03/src/groups/bmq/bmqp/bmqp_putmessageiterator.cpp#L128-L133

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_pusheventbuilder.t.cpp
+++ b/src/groups/bmq/bmqp/bmqp_pusheventbuilder.t.cpp
@@ -221,11 +221,9 @@ appendMessage(size_t                    iteration,
 #endif
 
     bdlbb::Blob payload(bufferFactory, allocator);
-#ifdef BMQ_ENABLE_MSG_GROUPID
-    const int                        blobSize = generateRandomInteger(0, 1024);
-    const bmqp::Protocol::MsgGroupId str(blobSize, 'x', allocator);
+    const int         blobSize = generateRandomInteger(1, 1024);
+    const bsl::string str(blobSize, 'x', allocator);
     bdlbb::BlobUtil::append(&payload, str.c_str(), blobSize);
-#endif
 
     data.d_payload = payload;
 
--- a/src/groups/bmq/bmqp/bmqp_pushmessageiterator.cpp
+++ b/src/groups/bmq/bmqp/bmqp_pushmessageiterator.cpp
@@ -36,6 +36,25 @@
 namespace BloombergLP {
 namespace bmqp {
 
+namespace {
+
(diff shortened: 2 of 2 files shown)


src/groups/bmq/bmqp/bmqp_pushmessageiterator.cpp
678098, line 130: Note that it might return `-1` even in `main` and we did not check it after call

src/groups/bmq/bmqp/bmqp_pushmessageiterator.cpp
```diff
@@ -614,7 +641,16 @@ int PushMessageIterator::next()
...
+    // Validation: 'length' is an unpadded application data size, with
+    // substracted size of the padding bytes section.
+    //: o Negative 'length' is a sign that packet is malformed.
+    //: o Zero 'length' is not supported.
+    if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(length <= 0)) {
+        BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
+        return rc_INVALID_APPLICATION_DATA_SIZE;  // RETURN
+    }
 
     rc = ProtocolUtil::parse(0,  // do not separate MPs from data
```
678098, line 655: `ProtocolUtil::parse` might be called with `-1` length in main

src/groups/bmq/bmqp/bmqp_pusheventbuilder.t.cpp
```diff
@@ -222,11 +222,9 @@ appendMessage(size_t                    iteration,
 #endif
 
     bdlbb::Blob payload(bufferFactory, allocator);
-#ifdef BMQ_ENABLE_MSG_GROUPID
-    const int                        blobSize = generateRandomInteger(0, 1024);
-    const bmqp::Protocol::MsgGroupId str(blobSize, 'x', allocator);
+    const int         blobSize = generateRandomInteger(1, 1024);
+    const bsl::string str(blobSize, 'x', allocator);
```
678098, line 225: The messages generated in this UT had 0 app data length.
We do not support zero app data length in `PutMessageIterator` early on, before reaching `PushMessageIterator`.
This check:

https://github.com/bloomberg/blazingmq/blob/07d6df9a23c99847523cd61e2d1a321b4762e267/src/groups/bmq/bmqp/bmqp_putmessageiterator.cpp#L617-L625

The UT has to be updated to generate push messages with non-zero app data lengths.

dorjesinpo (approved):
thank you

Issue #1288, opened 2026-04-10
Refactor[bmqbrkr.m.cpp]: replace `Bind` with functor


678098:
`registerMTrapHandler` uses `Bind` to construct a `bsl::function` callback executed on admin commands:

https://github.com/bloomberg/blazingmq/blob/04c376636ed844e6ba586ba93e6df3c6c0510f03/src/applications/bmqbrkr/bmqbrkr.m.cpp#L408-L415

There are 2 problems with `Bind`: it is relatively slow and its invokation produces multiple intermediate stack frames that makes stack traces in debugger less readable.

It can be simplified by defining a functor object that is auto-convertible to `bsl::function`. There are multiple examples of functors being used in the codebase, for example:

https://github.com/bloomberg/blazingmq/blob/04c376636ed844e6ba586ba93e6df3c6c0510f03/src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h#L325-L338

`Bind`s intermediate stack frames in rectangle:

<img width="2542" height="869" alt="Image" src="https://github.com/user-attachments/assets/bf2319ad-7939-4192-9f55-6d00f2754c89" />

PR #1290, opened 2026-04-10
Refactor[bmqbrkr.m.cpp]: replace Bind with functor

28arnab:
Replace bdlf::BindUtil::bind usage in registerMTrapHandler with a local functor to eliminate unnecessary indirection and improve stack trace readability. This avoids intermediate Bind frames and reduces callback overhead without changing behavior.

*Issue number of the reported bug or feature request: #1288*

**Describe your changes**
Replace bdlf::BindUtil::bind usage in registerMTrapHandler with a
local functor to reduce callback overhead and improve stack trace
readability.

**Testing performed**
No functional changes were introduced. This refactor replaces
bdlf::BindUtil::bind with a functor while preserving the same
callback signature and behavior.
The project was successfully built and the broker started
without errors after the change. Existing functionality is
expected to remain unchanged.

**Additional context**
This change simplifies the callback implementation by removing
Bind usage and aligning with existing functor patterns in the
codebase.

### DIFF
--- a/src/applications/bmqbrkr/bmqbrkr.m.cpp
+++ b/src/applications/bmqbrkr/bmqbrkr.m.cpp
@@ -356,6 +356,23 @@ struct MTrapHandler {
         cmd.erase(0, 1);  // cmd starts by a space, remove it
 
         if (bdlb::String::areEqualCaseless(prefix, "EXIT")) {
+            // The functor-based callback execution (from MTrapHandler)
+            // is more efficient than the previous bind-based approach, but it
+            // creates a race condition during shutdown. When EXIT is received,
+            // any pending admin commands like domain reconfigurations may
+            // still be queued in the thread pool but not yet dequeued. If we
+            // immediately post the shutdown semaphore, the main thread will
+            // unblock and call Application::stop(), which stops thread pools
+            // before pending jobs are processed, causing commands like
+            // domainReconfigure to never execute. The following sleep ensures
+            // the thread pool has time to pick up queued jobs before we
+            // initiate shutdown. See GitHub issue #678098.
+            static const int k_SHUTDOWN_DELAY_MS = 100;  // milliseconds
+            bsl::cout << "Waiting " << k_SHUTDOWN_DELAY_MS
+                      << "ms for pending commands to be dequeued...\n"
+                      << bsl::flush;
+            usleep(k_SHUTDOWN_DELAY_MS * 1000);  // Convert ms to microseconds
+
             d_taskEnv_p->d_shutdownSemaphore.post();
             // Return now because as a consequence of the Task::shutdown()
             // call, the BALL log system has been torn down.
--- a/src/applications/bmqbrkr/bmqbrkr.m.cpp
+++ b/src/applications/bmqbrkr/bmqbrkr.m.cpp
@@ -119,61 +119,6 @@ struct TaskEnvironment {
 /// Raw pointer to the TaskEnvironment, used so that the signal handler, and
 /// the assert handler can access the needed data.
 static TaskEnvironment* s_taskEnv_p = 0;
-
-// forward declaration
-static void onProcessedAdminCommand(const bsl::string&       prefix,
-                                    const bsl::string&       command,
-                                    const bsls::Types::Int64 start,
-                                    int                      rc,
-                                    const bsl::string&       results);
-
-/// Handler functor for incoming commands
-struct MTrapHandler {
-    TaskEnvironment* d_taskEnv_p;
-
-    explicit MTrapHandler(TaskEnvironment* taskEnv)
-    : d_taskEnv_p(taskEnv)
-    {
-        // PRECONDITIONS
-        BSLS_ASSERT(d_taskEnv_p);
-    }
-
-    void operator()(const bsl::string& prefix, bsl::istream& stream) const
-    {
-        BSLS_ASSERT_SAFE(d_taskEnv_p);
-
-        bsl::string cmd;
-
-        bsl::getline(stream, cmd);
-        cmd.erase(0, 1);  // cmd starts by a space, remove it
-
-        if (bdlb::String::areEqualCaseless(prefix, "EXIT")) {
-            d_taskEnv_p->d_shutdownSemaphore.post();
-            // Return now because as a consequence of the Task::shutdown()
-            // call, the BALL log system has been torn down.
-            return;  // RETURN
-        }
-        else if (bdlb::String::areEqualCaseless(prefix, "CMD")) {
-            const bsls::Types::Int64 start =
-                bmqsys::Time::highResolutionTimer();
-            d_taskEnv_p->d_app.object().enqueueCommand(
-                "MTRAP",
-                cmd,
-                bdlf::BindUtil::bind(
-                    onProcessedAdminCommand,
-                    prefix,
-                    cmd,
-                    start,
-                    bdlf::PlaceHolders::_1,    // rc
-                    bdlf::PlaceHolders::_2));  // commandExecResults
-            return;                            // RETURN
-        }
-
-        BALL_LOG_ERROR << "Unknown command '" << prefix << "'\n"
-                       << "Send 'HELP' to see a list of commands supported by "
-                       << "bmqbrkr.";
-    }
-};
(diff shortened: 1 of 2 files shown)


src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -471,15 +471,26 @@ initializeApplication(BSLA_MAYBE_UNUSED bsl::ostream& errorDescription,
...
                           task.allocatorStore().get("Application"));
 
+    // Local functor replacing Bind
+    struct MTrapHandler {
+        TaskEnvironment* d_taskEnv_p;
+
+        explicit MTrapHandler(TaskEnvironment* taskEnv)
+        : d_taskEnv_p(taskEnv)
+        {
+        }
```
678098:
```suggestion
        {
            // PRECONDITIONS
            BSLS_ASSERT(d_taskEnv_p);
        }
```

It is good to check that the provided ptr is not NULL, to ensure its usage is valid after

src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -471,15 +471,26 @@ initializeApplication(BSLA_MAYBE_UNUSED bsl::ostream& errorDescription,
...
+
+        void operator()(const bsl::string& prefix, bsl::istream& stream) const
+        {
+            onMTrap(d_taskEnv_p, prefix, stream);
```
678098: Possible to move `onMTrap` code to this operator and remove `onMTrap` fully

src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -471,15 +471,26 @@ initializeApplication(BSLA_MAYBE_UNUSED bsl::ostream& errorDescription,
...
+        }
+    };
```
678098: There is a C++03 build issue, could you try moving this functor out of this function scope to global scope in `bmqbrkr.m.cpp`?

src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -471,15 +471,26 @@ initializeApplication(BSLA_MAYBE_UNUSED bsl::ostream& errorDescription,
                           task.allocatorStatContext(),
                           task.allocatorStore().get("Application"));
 
+    // Local functor replacing Bind
```
678098:
```suggestion
    /// Handler functor for incoming commands
```

678098 (commented):
Looks good, just need to ensure C++03 build

28arnab:
@678098 I removed 'static' from onProcessedAdminCommand to avoid internal linkage issues with Bind and also removed the forward declaration since it was no longer needed.

However, I’m still unable to build the project locally. The errors appear to be unrelated to my changes, as I see linker errors involving ball::Log_Stream (string_view) and bsls_libraryfeatures_CPP17_ABI. I also tried checking out an earlier commit that previously built successfully, but the issue persists.

This makes me suspect a possible environment or build configuration issue on my side, but I’m not sure where the mismatch is coming from.

Here is a snippet of the error:
image


Could you help confirm if this is expected with newer toolchains or if I might be missing a clean rebuild step?

src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -120,6 +120,60 @@ struct TaskEnvironment {
 /// the assert handler can access the needed data.
 static TaskEnvironment* s_taskEnv_p = 0;
 
+// forward declaration
+static void onProcessedAdminCommand(const bsl::string&       prefix,
+                                    const bsl::string&       command,
+                                    const bsls::Types::Int64 start,
+                                    int                      rc,
+                                    const bsl::string&       results);
```
678098: @28arnab I think the issue is with forward-declaring `onProcessedAdminCommand` in unnamed namespace that is different from global namespace where `onProcessedAdminCommand` is defined.

Actually, it is not necessary to have a forward declaration. You can move `struct MTrapHandler` below, to the place where `static void onMTrap` was originally defined, it will be defined after `onProcessedAdminCommand`, so no forward declaration is needed. Could you try this?

678098:
Hi @28arnab, are you building the project in Linux or on Mac?
How exactly do you build it? Do you use `bin/build-[platform].sh` or do you use cmake presets build with vcpkg?

28arnab:
> Hi @28arnab, are you building the project in Linux or on Mac? How exactly do you build it? Do you use `bin/build-[platform].sh` or do you use cmake presets build with vcpkg?

OS: Ubuntu 24.04.4 LTS x86_64 
Kernel: 6.8.0-107-generic
Shell: bash 5.2.21 

I am trying to build the project using the bin/build-ubuntu.sh for the past 2 days removed and recloned the repo even the main branch isnt compiling. same error of linkers only time it compiled was the 1st commit which passed for all the checks except c++03. also there was no linker problem at that time.

src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -336,45 +335,70 @@ static void onProcessedAdminCommand(const bsl::string&       prefix,
...
+    : d_taskEnv_p(taskEnv)
+    {
+        // PRECONDITIONS
+        BSLS_ASSERT(d_taskEnv_p);
+    }
+
+    void operator()(const bsl::string& prefix, bsl::istream& stream) const
+    {
+        BSLS_ASSERT_SAFE(d_taskEnv_p);
+
```
678098, line 352:
```suggestion
```

This check is not necessary now, we already check that the provided `d_taskEnv_p` is not NULL in constructor. We don't plan to modify the pointer during the lifetime of the functor and can rely on its value being not NULL.

This principle saves us some CPU time: we check the pointer just once, on functor construction, and rely on this being non-NULL on each execution of the functor.

src/python/blazingmq/dev/it/cluster_util.py
```diff
@@ -184,6 +185,38 @@ def clean_storage_output(output_str: str) -> str:
     return json.dumps(data, indent=2)
 
 
+def _wait_for_passive_advisories(leader: Broker, cluster: Cluster, timeout: int = 10):
```
678098, line 188: The changes in this file doesn't belong to the PR and needs to be excluded

src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -336,45 +335,70 @@ static void onProcessedAdminCommand(const bsl::string&       prefix,
...
+            // unblock and call Application::stop(), which stops thread pools
+            // before pending jobs are processed, causing commands like
+            // domainReconfigure to never execute. The following sleep ensures
+            // the thread pool has time to pick up queued jobs before we
+            // initiate shutdown. See GitHub issue #678098.
+            static const int k_SHUTDOWN_DELAY_MS = 100;  // milliseconds
+            bsl::cout << "Waiting " << k_SHUTDOWN_DELAY_MS
+                      << "ms for pending commands to be dequeued...\n"
+                      << bsl::flush;
+            usleep(k_SHUTDOWN_DELAY_MS * 1000);  // Convert ms to microseconds
```
678098, line 374: This change needs to be reverted for three reasons.

1. This is a behaviour change. Pure refactoring should not change the behaviour: this makes changes safer and easier to accept.

2. The functor approach doesn't create any new race condition or any new problem. The existing behaviour was expected.

3. Timeout as a synchronization is fragile and non-working solution in general. It might "fix" something sometimes, but is not a good practice. "wait" in the code typically means that we don't know **when** something should happen or don't have control over it, and it is not a good sign to not have control over the program - so, it should be avoided.

678098 (commented):
Hi @28arnab, have some new comments

28arnab:
Hi @678098, 1 test is failing due to no response within the time limit of the test. How to solve this like if there is response at all or response is late.

678098 (approved):
Looks good, IT failure is unrelated

678098:
Thank you for contribution @28arnab !

28arnab:
> Thank you for contribution @28arnab !

Thank you for your guidance @678098 !

PR #1292, opened 2026-04-10
Fix[mqbnet_elector.cpp]: reduce logs on quorum mismatch

678098:
Have tons of logs like this:
```
10APR2026_19:32:59.992 (140567480620800) WARN mqbnet_elector.cpp:646 #ELECTOR_LEADER_HEARTBEAT LEADER received LEADER_HEARTBEAT with stale term [1] from node [4]. Current term [1]. Ignoring this heartbeat, and sending leader heartbeat to the sender node.
```

### DIFF
--- a/src/groups/mqb/mqbnet/mqbnet_elector.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_elector.cpp
@@ -642,7 +642,7 @@ void ElectorStateMachine::applyLeaderHeartbeatEventToLeader(
     bsls::Types::Uint64        term,
     int                        sourceNodeId)
 {
-    if (d_term >= term) {
+    if (d_term > term) {
         BALL_LOG_WARN
             << "#ELECTOR_LEADER_HEARTBEAT "
             << "LEADER received LEADER_HEARTBEAT with stale term [" << term
--- a/src/groups/mqb/mqbnet/mqbnet_elector.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_elector.cpp
@@ -30,6 +30,7 @@
 #include <mqbcfg_clusterquorummanager.h>
 
 // BDE
+#include <ball_logthrottle.h>
 #include <bdlb_print.h>
 #include <bdlb_string.h>
 #include <bdlb_stringrefutil.h>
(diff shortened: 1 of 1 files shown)


hallfox (approved):

src/groups/mqb/mqbnet/mqbnet_elector.cpp
```diff
@@ -642,7 +642,7 @@ void ElectorStateMachine::applyLeaderHeartbeatEventToLeader(
     bsls::Types::Uint64        term,
     int                        sourceNodeId)
 {
-    if (d_term >= term) {
+    if (d_term > term) {
```
kaikulimu, line 645: I did some reading on [`elector algorithm`](https://bloomberg.github.io/blazingmq/docs/architecture/election_rsm/#updated-algorithm-using-term). The original implementation seems correct. If I am a leader with `d_term = N`, and I receive a leader heartbeat with term `N`, it does look stale because there cannot be another leader with the same term as me.

When you say you see tons of logs of form:
```
10APR2026_19:32:59.992 (140567480620800) WARN mqbnet_elector.cpp:646 #ELECTOR_LEADER_HEARTBEAT LEADER received LEADER_HEARTBEAT with stale term [1] from node [4]. Current term [1]. Ignoring this heartbeat, and sending leader heartbeat to the sender node.
```
Where do you see these logs? And what's the reason they occur? Either there is a bug, and we changed the heartbeat logic in an undocumented way.

kaikulimu (changes requested):
Need more clarity

src/groups/mqb/mqbnet/mqbnet_elector.cpp
678098: I observe this in a scenario where I set quorum to 1 for one node to make it a leader. The rest of the cluster have quorum values preventing them from becoming leaders.

678098: I will investigate the logs then, to find the exact scenario.
This is FSM mode by the way

kaikulimu: Intuitively, it seems likely to happen in quorum-1 case. Say east1 becomes leader with term 1, quorum 1, and no other nodes know about it. When other nodes come up, they will propose as candidate also with term 1. Then east1 will fire this warning.

Harmless in the quorum-1 case, but if a normal quorum cluster experiences this, then it's a real issue.

678098: What about keeping the original condition, but making the logs throttled?

kaikulimu: That is a good idea! Let's do this.

kaikulimu (approved):

PR #1298, opened 2026-04-13
Feat: Introduce `BMQ_ASSERT_LEVEL` cache variable

pniedzielski:
Rather than invariably setting the assert level we compile at based on the build mode (release or non-release), this patch introduces a new cache variable `BMQ_ASSERT_LEVEL`, which can override the default assert level computed based on the build mode.  For explicitness’s sake, we set this to SAFE in our CMakePresets, but users can now override this.

Supercedes: #1243

### DIFF
--- a/etc/presets/default.json
+++ b/etc/presets/default.json
@@ -10,7 +10,7 @@
                 "CMAKE_EXPORT_COMPILE_COMMANDS": "1",
                 "CMAKE_BUILD_TYPE": "Debug",
                 "BDE_BUILD_TARGET_64": "1",
-                "BDE_BUILD_TARGET_SAFE": true,
+                "BMQ_ASSERT_LEVEL": "SAFE",
                 "CMAKE_CXX_STANDARD": "23",
                 "BDE_BUILD_TARGET_CPP23": "ON",
                 "CMAKE_INSTALL_LIBDIR": "lib64"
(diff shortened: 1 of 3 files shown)


etc/presets/default.json
```diff
@@ -9,10 +9,9 @@
             "cacheVariables": {
                 "CMAKE_EXPORT_COMPILE_COMMANDS": "1",
                 "CMAKE_BUILD_TYPE": "Debug",
-                "BDE_BUILD_TARGET_64": "1",
-                "BDE_BUILD_TARGET_SAFE": true,
-                "CMAKE_CXX_STANDARD": "23",
-                "BDE_BUILD_TARGET_CPP23": "ON",
+                "BMQ_ASSERT_LEVEL": "SAFE",
+                "CMAKE_CXX_STANDARD": "17",
+                "BDE_BUILD_TARGET_CPP17": "17",
```
hallfox: Why did the standard change?

pniedzielski: I made a mistake when resolving merge conflicts with the latest `main`, sorry.  Lemme fix.

pniedzielski: Fixed with latest push; only touched `BDE_BUILD_TARGET_SAFE`.

hallfox (approved):

PR #1302, opened 2026-04-14
Fix[mqb]: write failure transitions REOPENING->PENDING

dorjesinpo:
⏺ When the connection flaps during queue reopen, a channel write error causes the queue to transition to CLOSED and NACK all pending messages. The fix
  splits e_REQ_NOT_SENT into e_REQ_WRITE_ERROR (transport) and e_REQ_BAD (logical). On write error during reopen, the queue goes to PENDING instead of
  CLOSED, preserving pending messages for the next reconnect cycle.


pniedzielski:
`bmqimp_brokersession.t` is only failing in CI without the fix in #1301.  Tested locally to verify.

pniedzielski (approved):

PR #1303, opened 2026-04-14
Refactor[mqb] make authenticate() plugin API allocator aware

chrisbeard:
**Describe your changes**
`authenticate()` is expected to populate an output object. In BDE-style, output allocations are controlled by the caller, so the caller needs to explicitly pass an allocator. This change makes the memory ownership clear to plugin authors.

### DIFF
--- a/src/groups/mqb/mqbauthn/mqbauthn_testauthenticator.cpp
+++ b/src/groups/mqb/mqbauthn/mqbauthn_testauthenticator.cpp
@@ -81,10 +81,8 @@ TestAuthenticationResult::lifetimeMs() const
 // -----------------------
 
 TestAuthenticator::TestAuthenticator(
-    const mqbcfg::AuthenticatorPluginConfig* config,
-    bslma::Allocator*                        allocator)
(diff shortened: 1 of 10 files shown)


src/groups/mqb/mqbauthn/mqbauthn_testauthenticator.cpp
```diff
@@ -82,9 +82,8 @@ TestAuthenticationResult::lifetimeMs() const
 
 TestAuthenticator::TestAuthenticator(
     const mqbcfg::AuthenticatorPluginConfig* config,
-    bslma::Allocator*                        allocator)
-: d_allocator_p(allocator)
-, d_isStarted(false)
+    BSLA_MAYBE_UNUSED bslma::Allocator* allocator)
```
hallfox: If this class doesn't allocate we should probably not even take an allocator here.

hallfox (commented):
lgtm and fixes some existing issues with the authn implementation we had

hallfox (approved):

PR #1305, opened 2026-04-14
Remove: Unused executor adaptors for BDE types

hallfox:
We don't use these components and don't plan to


pniedzielski (dismissed):
🥳

pniedzielski (dismissed):

pniedzielski (approved):

PR #1310, opened 2026-04-16
Fix[mqb]: Make sure drain is complete before TCP stop

dorjesinpo:
`mqbnet::Channel::closeChannel()` is non-blocking.  When `Application` continues to stop, it calls `CPSessionFactory::stop()` which calls `StatChannelFactory::stop()` which results in TCP disconnects.  This creates a race with draining `mqbnet::Channel`.

Solution:
Add `mqbnet::Channel::stop()` which blocks (`join`) until the thread exits.  `mqbnet::Channel::d_isStopping` is a must to stop the channel.
Call `mqbnet::Channel::stop()` after calling non-blocking `mqbnet::Channel::requestToStop()` - to drain in parallel.

### DIFF
--- a/src/groups/mqb/mqbnet/mqbnet_channel.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_channel.cpp
@@ -80,12 +80,12 @@ Channel::Channel(bdlbb::BlobBufferFactory* blobBufferFactory,
              bsls::BlockGrowth::BSLS_CONSTANT,
              d_allocators.get("ItemPool"))
 , d_buffer(1024, allocator)
-, d_doStop(false)
+, d_isStopped(false)
 , d_state(e_RESET)
 , d_description(name + " - ", d_allocator_p)
 , d_name(name, d_allocator_p)
 , d_stats()
-, d_isClosing(false)
+, d_isStopping(false)
 {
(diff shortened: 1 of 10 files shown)


dorjesinpo:
`test_poison_fanout_crash_one_app` is unrelated issue.  Will be another PR

src/groups/mqb/mqbnet/mqbnet_channel.cpp
```diff
@@ -103,9 +103,28 @@ Channel::Channel(bdlbb::BlobBufferFactory* blobBufferFactory,
...
+    }
+    BSLS_ASSERT_SAFE(d_isStopped.load());
+}
+
+void Channel::stop()
+{
+    BSLS_ASSERT_SAFE(d_isStopping.load());
+
+    {
+        bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);  // LOCK
```
hallfox, line 119: What is this mutex protecting?

dorjesinpo: We want to signal only when waiting.
The signal is lost when happens between lines 840-849

hallfox (approved):

PR #1312, opened 2026-04-16
Fix[bmqp]: Remove dependency on bmqp from bmqa header

pniedzielski:
Only bmqa, bmqt, and bmqpi are part of the public interface of libbmq, so the introduction of a dependency on a bmqp header file causes clients to fail to compile.  Luckily, the issue is only an imported typedef, of which we have a local copy anyway.  This patch removes the dependency on bmqp by switching to the local typedef.


dorjesinpo (approved):
thank you

PR #1313, opened 2026-04-16
UT[mqbc::StorageMgr.t]: Verify ReplicaDataRequestPushs are sent 

kaikulimu:
Fixes failing UT on main.


src/groups/mqb/mqbc/mqbc_storagemanager.t.cpp
```diff
@@ -2943,6 +2951,14 @@ static void test18_primaryHealingWatchdogRetry()
                      4U);
     BMQTST_ASSERT_EQ(storageManager.partitionHealthState(k_PARTITION_ID),
                      mqbc::PartitionFSM::State::e_PRIMARY_HEALING_STG2);
+
+    NodeIdToSeqNumMap destinationReplicas;
+    destinationReplicas.insert(bsl::make_pair(reqIdToNodeIdMap.at(1), seqNum));
```
pniedzielski, line 2956: It would be great if we had an interface like `vector`'s `operator[]` that `ASSERT`s immediately, rather than throws an exception.  But `unordered_map`'s `operator[]` doesn't do that, so I guess this is the best we can do.

pniedzielski (approved):

PR #1315, opened 2026-04-16
Fix[mqb]: proxies must copy rda

dorjesinpo:
`storePushIfProxy` doesn't copy the RDA counter from the PUSH message into the storage's `AppMessage`. As the result, it has `isUnlimited = true`. Then, the message stays in the proxy's redelivery list and it is a race between redelivering the poisonous message on reconnect and primary purging it.


dorjesinpo:
IT failure is unrelated and the fix is https://github.com/bloomberg/blazingmq/pull/1316

pniedzielski (approved):

PR #1316, opened 2026-04-17
Fix[mqb]: Re-read counter after de-configure response

dorjesinpo:
Concurrent cookie rollback (`onOpenQueueConfirmationCookieReleased`) may decrement the same count as `dropHandleDispatched` resulting in double decrement.
Solution: re-read current counts from handle->subStreamInfos() upon de-configure response.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
@@ -2579,7 +2579,7 @@ void ClusterQueueHelper::onGetDomain(
     mqbi::Domain*                       domain,
     const bmqp_ctrlmsg::ControlMessage& request,
     mqbc::ClusterNodeSession*           requester,
-    const int                           peerInstanceId)
+    const int                           requesterId)
 {
     // executed by *ANY* thread
 
@@ -2591,7 +2591,7 @@ void ClusterQueueHelper::onGetDomain(
                               domain,
                               request,
                               requester,
-                              peerInstanceId),
+                              requesterId),
         d_cluster_p);
 }
 
@@ -2600,7 +2600,7 @@ void ClusterQueueHelper::onGetDomainDispatched(
     mqbi::Domain*                       domain,
     const bmqp_ctrlmsg::ControlMessage& request,
     mqbc::ClusterNodeSession*           requester,
-    const int                           peerInstanceId)
+    const int                           requesterId)
 {
     // executed by the cluster *DISPATCHER* thread
 
@@ -2680,7 +2680,7 @@ void ClusterQueueHelper::onGetDomainDispatched(
                               bdlf::PlaceHolders::_4,  // confCookie
                               request,
                               requester,
-                              peerInstanceId));
+                              requesterId));
 }
 
 void ClusterQueueHelper::onGetQueueHandle(
@@ -2740,8 +2740,8 @@ void ClusterQueueHelper::onGetQueueHandleDispatched(
             << ", because either the peer is down, or new instance "
             << "of the peer has come up. Peer node status: "
             << requester->nodeStatus()
-            << ", initial peerInstanceId: " << requesterId
-            << ", current peerInstanceId: "
+            << ", initial requesterId: " << requesterId
+            << ", current requesterId: "
             << requester->handleRequesterContext()->requesterId() << ".";
         return;  // RETURN
     }
--- a/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
@@ -2713,7 +2713,7 @@ void ClusterQueueHelper::onGetQueueHandleDispatched(
     const mqbi::OpenQueueConfirmationCookieSp& confirmationCookie,
     const bmqp_ctrlmsg::ControlMessage&        request,
     mqbc::ClusterNodeSession*                  requester,
-    const int                                  peerInstanceId)
+    const int                                  requesterId)
 {
     // executed by the cluster *DISPATCHER* thread
 
@@ -2722,7 +2722,7 @@ void ClusterQueueHelper::onGetQueueHandleDispatched(
     BSLS_ASSERT_SAFE(request.choice().isOpenQueueValue());
     BSLS_ASSERT_SAFE(d_cluster_p->isClusterMember());
 
-    if (peerInstanceId != requester->peerInstanceId() ||
+    if (requesterId != requester->handleRequesterContext()->requesterId() ||
         requester->nodeStatus() == bmqp_ctrlmsg::NodeStatus::E_UNAVAILABLE) {
         // Either a new instance of the peer is up, or old instance is no
         // longer available (ie, channel with old instance went down but has
@@ -2740,9 +2740,9 @@ void ClusterQueueHelper::onGetQueueHandleDispatched(
             << ", because either the peer is down, or new instance "
             << "of the peer has come up. Peer node status: "
             << requester->nodeStatus()
-            << ", initial peerInstanceId: " << peerInstanceId
-            << ", current peerInstanceId: " << requester->peerInstanceId()
-            << ".";
+            << ", initial peerInstanceId: " << requesterId
+            << ", current peerInstanceId: "
+            << requester->handleRequesterContext()->requesterId() << ".";
         return;  // RETURN
     }
 
@@ -2789,8 +2789,8 @@ void ClusterQueueHelper::onGetQueueHandleDispatched(
             << d_cluster_p->description()
             << ": Reused handle when processing peer "
             << "openQueueRequest [requester: " << requester->description()
-            << ", request: " << request
-            << ", peerInstanceId: " << peerInstanceId << "].";
+            << ", request: " << request << ", requesterId: " << requesterId
+            << "].";
 
         BSLS_ASSERT_SAFE(queueHandle == iter->second.d_handle_p);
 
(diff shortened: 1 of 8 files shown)


dorjesinpo:
`test_redelivery_on_primary_node_crash` failure is unrelated
`test_sync_if_leader_missed_rollover` failure is unrelated and the fix is https://github.com/bloomberg/blazingmq/pull/1310

src/groups/mqb/mqbblp/mqbblp_queue.cpp
```diff
@@ -77,21 +77,45 @@ struct Counter {
...
     const bmqp_ctrlmsg::Status&,
-    const bmqp_ctrlmsg::StreamParameters&,
-    mqbi::Queue*                               queue,
-    mqbi::QueueHandle*                         handle,
-    const bmqp_ctrlmsg::QueueHandleParameters& handleParameters,
-    const bsl::shared_ptr<Counter>&            counter)
+    const bmqp_ctrlmsg::StreamParameters& streamParameters,
+    mqbi::Queue*                          queue,
+    mqbi::QueueHandle*                    handle,
+
```
pniedzielski: Extra newline here

src/groups/mqb/mqbblp/mqbblp_queue.cpp
```diff
@@ -77,21 +77,45 @@ struct Counter {
...
+    BSLS_ASSERT_SAFE(appId != bmqp::ProtocolUtil::k_NULL_APP_ID);
 
-    queue->releaseHandle(handle,
-                         handleParameters,
-                         isFinal,
-                         mqbi::QueueHandle::HandleReleasedCallback());
+    if (info.d_counts.d_readCount == 0) {
+        // Already fully released by cookie rollback — nothing to do.
+    }
```
pniedzielski, line 98: I think this is missing a `return;`.  Else we still release the handle.

src/groups/mqb/mqbblp/mqbblp_queue.cpp
```diff
@@ -77,21 +77,45 @@ struct Counter {
...
+    mqbi::QueueHandle*                    handle,
+
+    const bsl::shared_ptr<Counter>& counter)
 {
     BSLS_ASSERT_SAFE(queue);
     BSLS_ASSERT_SAFE(handle);
 
-    const bool isFinal = (counter->decrement() == 0);
+    // Re-read current counts from the handle because cookie rollback may have
+    // already decremented some counts between de-confgigure request and now.
```
pniedzielski: `de-configure`

pniedzielski (changes requested):

dorjesinpo:
This PR is incomplete.  We need to handle the other racer - `onOpenQueueConfirmationCookieReleased`

pniedzielski:
Taking a look...

src/groups/mqb/mqbc/mqbc_clusternodesession.cpp
```diff
@@ -171,6 +162,36 @@ bool ClusterNodeSession::removePartitionSafe(int partitionId)
...
+    const bsl::shared_ptr<bmqst::StatContext>& statContext)
+{
+    // executed by the *DISPATCHER* thread
+
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(inDispatcherThread());
+    d_queueHandleRequesterContext_sp.createInplace(d_allocator_p,
+                                                   d_allocator_p);
+
+    d_queueHandleRequesterContext_sp->setClient(this)
```
pniedzielski, line 191: Comment, no action needed: This fluent builder pattern here didn't fully pass the sniff test to me, because the default object constructed is not valid (hence, fluent pattern is an anti-pattern).  This made me look into `QueueHandleRequesterContext` a little more, and I found this lovely contradiction:

```
/// Value-semantic type representing the context of a client, requester of a
/// queue handle.
class QueueHandleRequesterContext {
```
and
```
/// Copy constructor and assignment operator are not implemented.
QueueHandleRequesterContext&
operator=(const QueueHandleRequesterContext&) BSLS_CPP11_DELETED;
```

😮‍💨

From inspection of where it's used, I think we can very easily remove the invalid state from `QueueHandleRequesterContext` and make this type safe and impossible to misuse.

Let's get this change in as it is for the functionality change, and I'll prepare a separate PR that simplifies this.

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.h
```diff
@@ -720,7 +720,7 @@ class ClusterQueueHelper BSLS_KEYWORD_FINAL
     /// (in the specified `request`).  If the specified `status` is SUCCESS,
     /// the request was success and the specified `queueHandle` contains the
     /// handle representing the queue that was allocated for this specified
-    /// `requester` session having the specified `peerInstanceId`, and the
+    /// `requester` session having the specified `requesterId`, and the
```
pniedzielski, line 723: This rename wasn't done completely:

```
src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
2582:    const int                           peerInstanceId)
2594:                              peerInstanceId),
2603:    const int                           peerInstanceId)
2683:                              peerInstanceId));
2743:            << ", initial peerInstanceId: " << requesterId
2744:            << ", current peerInstanceId: "
src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.h
697:                     int                                 peerInstanceId);
701:    /// specified `requester` with the specified `peerInstanceId`.  If the
710:                               const int peerInstanceId);
```

I think the log message particularly is confusing.

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -3727,16 +3727,18 @@ void ClusterQueueHelper::restoreStateCluster(int partitionId)
...
+    if (d_clusterData_p->membership().selfNodeStatus() !=
+        bmqp_ctrlmsg::NodeStatus::E_AVAILABLE) {
         BALL_LOG_INFO << d_cluster_p->description()
                       << ": Not going ahead with restoring partition state "
                       << "because self is not AVAILABLE.  Self status: "
                       << d_clusterData_p->membership().selfNodeStatus();
         return;  // RETURN
     }
 
+    // TODO: revisit
```
pniedzielski, line 3741: Can we capture this in an issue with some context?  We have almost 200 TODOs in the codebase; this will get lost.

src/groups/mqb/mqbc/mqbc_clusternodesession.cpp
```diff
@@ -171,6 +162,36 @@ bool ClusterNodeSession::removePartitionSafe(int partitionId)
...
+        .setIdentity(identity)
+        .setDescription(bsl::string(description(), d_allocator_p))
+        .setIsClusterMember(true)
+        .setRequesterId(
+            mqbi::QueueHandleRequesterContext ::generateUniqueRequesterId())
```
pniedzielski: `mqbi::QueueHandleRequesterContext::generateUniqueRequesterId()`

src/groups/mqb/mqbc/mqbc_clusternodesession.cpp
```diff
@@ -171,6 +162,36 @@ bool ClusterNodeSession::removePartitionSafe(int partitionId)
...
+
+void ClusterNodeSession::createQueueHandleRequesterContextImpl(
+    const bmqp_ctrlmsg::ClientIdentity&        identity,
+    const bsl::shared_ptr<bmqst::StatContext>& statContext)
+{
+    // executed by the *DISPATCHER* thread
+
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(inDispatcherThread());
+    d_queueHandleRequesterContext_sp.createInplace(d_allocator_p,
```
pniedzielski, line 188: This is a little subtle.  `createQueueHandleRequesterContext` pulls out the `ClientIdentity` and `StatContext` out of the old queue handle requester context, passes them into `createQueueHandleRequesterContext` by `const&`, and then calls `createInplace`.  Are the objects these references point to still valid?  Depends on whether there's any other `shared_ptr` to the old handle requester context, and if not, whether `createInplace` calls the destructor of the existing object before it calls the constructor of the new object.  This isn't guaranteed, but either way it's very subtle.

I think it's better to avoid that subtle reasoning, since it's fragile next time we change this.  Let's add some copies before we `createInplace`:

```
// Copy before `createInplace` destroys the old object.
const bmqp_ctrlmsg::ClientIdentity identity = d_queueHandleRequesterContext_sp->identity();
const bsl::shared_ptr<bmqst::StatContext> statContext = d_queueHandleRequesterContext_sp->statContext();
```

src/groups/mqb/mqbc/mqbc_clusternodesession.h
```diff
@@ -276,6 +279,8 @@ class ClusterNodeSession : public mqbi::DispatcherClient,
     bmqu::GateKeeper& gatePut();
     bmqu::GateKeeper& gateConfirm();
 
+    void createQueueHandleRequesterContext();
```
pniedzielski, line 284: Doxygen docs ^

pniedzielski (changes requested):

src/groups/mqb/mqbc/mqbc_clusternodesession.cpp
pniedzielski: On further thought, we can put this copy in `createQueueHandleRequesterContext`, since the `...Impl` function is called in the constructor too.

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
dorjesinpo: Created a card "Revisit logic around NodeStatus, ElectorInfoLeaderStatus, PrimaryStatus"

pniedzielski (approved):

PR #1323, opened 2026-04-21
Misc: Add convenience script for cpp03 files

hallfox:
This script is meant to be run for developers working in a checkout and are getting annoyed by the timestamp being generated by the `sim_cpp11_features.pl` script that gets run during the CMake configure step.


bin/skip-cpp03-in-workspace.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
678098, line 1: I don't see how this script is mentioned during build. Is it found and run automatically somehow?

678098 (approved):

PR #1326, opened 2026-04-21
CI: Run lint workflow when PR is edited

pniedzielski:
This patch causes the lint workflow to be rerun whenever the PR is edited in the GitHub interface, such as editing its description or title.  In particular, this is meant to fix an issue we see where editing the title of a PR does not fix failing a PR title check. After applying this patch, modifying the title of the PR will kick off another run of the linting workflow, which will succeed if the new PR title matches the correct format.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -8,6 +8,7 @@ on:
       - "synchronize"
       - "labeled"
       - "unlabeled"
+      - "edited"
 
 jobs:
   commits_check_job:


pniedzielski:
https://github.com/bloomberg/blazingmq/pull/1315, at least at the moment, is an example of this failure.  Originally it used the commit message as the title, which did not pass the check, and manually editing the title did not rerun the check.

We could alternatively split out this one check into its own workflow that runs on `"edited"` to avoid running all the other linter checks whenever a PR title is changed.  But, the lint checks all run very, very quickly, and doing it the way this patch does avoids cluttering up the interface in https://github.com/bloomberg/blazingmq/actions with additional workflow runs.

hallfox (approved):

PR #1329, opened 2026-04-23
Feat[mqb]: store `AppConfig` value, not a pointer to config

678098:
`mqbcfg::BrokerConfig` only stores a pointer to an external configuration object. Due to this, the configuration object must be kept alive in runtime, or the pointer becomes invalid.
It is possible to make a copy of a provided configuration and remove the restriction of keeping configuration object alive.

### DIFF
--- a/src/groups/mqb/mqbcfg/mqbcfg_brokerconfig.cpp
+++ b/src/groups/mqb/mqbcfg/mqbcfg_brokerconfig.cpp
@@ -16,8 +16,16 @@
 // mqbcfg_brokerconfig.cpp                                            -*-C++-*-
 #include <mqbcfg_brokerconfig.h>
 
+// MQB
+#include <mqbcfg_messages.h>
+
+// BMQ
+#include <bmqu_singletonallocator.h>
+
 // BDE
+#include <bslmt_once.h>
 #include <bsls_assert.h>
+#include <bsls_objectbuffer.h>
 
 namespace BloombergLP {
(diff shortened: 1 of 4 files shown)


src/groups/mqb/mqbcfg/mqbcfg_brokerconfig.cpp
```diff
@@ -26,6 +34,17 @@ namespace {
...
 
+mqbcfg::AppConfig* getImpl()
+{
+    static bsls::ObjectBuffer<mqbcfg::AppConfig> appConfig;
+    BSLMT_ONCE_DO
+    {
+        new (static_cast<void*>(appConfig.buffer()))
+            mqbcfg::AppConfig(bmqu::SingletonAllocator::allocator());
+    }
+    return &appConfig.object();
```
hallfox: Do we want to have the actual objects managed by the singleton allocator? e.g.

```cpp
static mqbcfg::AppConfig* appConfig_p;
BSLMT_ONCE_DO {
  bslma::Allocator* alloc = bmqu::SingletonAllocator::allocator();
  appConfig_p = bslma::AllocatorUtil::newObject<mqbcfg::AppConfig>(alloc);
}
return *appConfig_p;
```

678098: Done

hallfox (approved):

PR #1330, opened 2026-04-23
Fix[mqba_clientsession]: fix deadlock on shutdown

678098:
`ClientSession::tearDownImpl` calls `d_self.invalidate();`
If `tearDownImpl` is executed before `ClientSession::initiateShutdownDispatched`, it blocks on `invalidate` because we keep a shared pointer `d_self` for `initiateShutdownDispatched`: deadlock.

### DIFF
--- a/src/groups/mqb/mqba/mqba_clientsession.cpp
+++ b/src/groups/mqb/mqba/mqba_clientsession.cpp
@@ -922,13 +922,12 @@ void ClientSession::onHandleConfiguredDispatched(
     sendPacketDispatched(d_state.d_schemaEventBuilder.blob(), true);
 }
 
-void ClientSession::initiateShutdownDispatched(const ShutdownCb& callback)
+void ClientSession::initiateShutdownDispatched()
 {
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqba/mqba_clientsession.cpp
```diff
@@ -2719,43 +2719,48 @@ void ClientSession::initiateShutdown(const ShutdownCb& callback)
...
     else {
+        // Use a weak ref in the dispatched event to avoid deadlock with
+        // tearDownImpl's d_self.invalidate() when both events are queued
+        // on the same client dispatcher thread.
         dispatcher()->execute(
-            bdlf::BindUtil::bindS(
-                d_state.d_allocator_p,
-                bdlf::MemFnUtil::memFn(
-                    &ClientSession::initiateShutdownDispatched,
-                    d_self.acquire()),
```
678098, line 2750: This `shared_ptr` copy might block us from calling `d_self.invalidate`

hallfox (approved):

PR #1335, opened 2026-04-27
Perf[bmqma]: faster allocation limits checks

678098:
# Description

- Remove recursive `onAllocationChange` checks. These checks cannot be inlined and called per every node in `CountingAllocator` tree path. In reality, we only configure one limit for the root node, and we won't configure node-specific limits.
- Add a component `AllocationLimitChecker`, one instance of which is shared between all `CountingAllocator` nodes in a tree. This component tracks all counted allocations and might trigger an alarm.
- On each allocation, only update the shared `AllocationLimitChecker`. It is one atomic update per allocation, not multiple like before.
- Replace the atomic `d_allocationLimit` with a non-atomic variable. This should make the checks run faster. Use the second-layer bool atomic to prevent retriggering the alarm.

# Benchmarks (amd64, Linux)

Note that `bmqma::CountingAllocator::onAllocationChange` is directly replaced by `bmqma::CountingAllocator::AllocationLimitChecker::update`:

**Before**

Screenshot 2026-04-27 at 19 01 25

**After**

Screenshot 2026-04-27 at 18 49 26

Allocation limit checks took 4.5% of all stack frames, now they take 0.77% of all stack frames.

### DIFF
--- a/src/groups/bmq/bmqma/bmqma_countingallocator.cpp
+++ b/src/groups/bmq/bmqma/bmqma_countingallocator.cpp
@@ -182,69 +182,60 @@ void CountingAllocator::configureStatContextTableInfoProvider(
         .zeroString("");
 }
 
-void CountingAllocator::onAllocationChange(bsls::Types::Int64 deltaValue)
-{
-    const bsls::Types::Uint64 totalAllocated = d_allocated.addRelaxed(
-        deltaValue);
-
-    if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(totalAllocated >
-                                              d_allocationLimit)) {
-        BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
-        const bsls::Types::Uint64 uint64Max =
-            bsl::numeric_limits<bsls::Types::Uint64>::max();
-        if (d_allocationLimit.swap(uint64Max) != uint64Max) {
-            // Use an atomic swap to only invoke the callback the first time
-            // limit is crossed.  Swap 'allocationLimit' with 'Uint64::max',
-            // which will disable all further maximum allocation checks.
-
-            BSLS_ASSERT_SAFE(d_allocationLimitCb);
-            // If d_allocationLimit was set, 'd_allocationLimitCb' must be
-            // a valid callback
-            d_allocationLimitCb();
-        }
-    }
-
-    // Propagate allocation change to parent, if any
-    if (d_parentCounting_p) {
-        d_parentCounting_p->onAllocationChange(deltaValue);
-    }
-}
-
 CountingAllocator::CountingAllocator(const bslstl::StringRef& name,
                                      bslma::Allocator*        allocator)
-: d_statContext_mp()
-, d_allocator_p(bslma::Default::allocator(allocator))
-, d_parentCounting_p(0)
-, d_allocated(0)
-, d_allocationLimit(bsl::numeric_limits<bsls::Types::Uint64>::max())
-// Disable allocation limit by default
+: d_allocator_p(bslma::Default::allocator(allocator))
+, d_statContext_mp()
+, d_limitChecker_sp()
 {
-    CountingAllocator* ca = dynamic_cast<CountingAllocator*>(d_allocator_p);
-    if (ca) {
-        d_allocator_p = ca->d_allocator_p;
-        if (ca->d_statContext_mp) {
-            d_statContext_mp = ca->d_statContext_mp->addSubcontext(
+    CountingAllocator* parent = dynamic_cast<CountingAllocator*>(
+        d_allocator_p);
+
+    if (parent) {
+        // `d_allocator_p` should never be a CountingAllocator.
+        // We support this invariant up the allocators tree and refer to the
+        // parent's `d_allocator_p`.
+        d_allocator_p     = parent->d_allocator_p;
+        d_limitChecker_sp = parent->d_limitChecker_sp;
+
+        if (parent->d_statContext_mp) {
+            d_statContext_mp = parent->d_statContext_mp->addSubcontext(
                 bmqst::StatContextConfiguration(name, allocator));
-            d_parentCounting_p = ca;
         }
     }
+    else {
+        // This is a root CountingAllocator: construct the original limit
+        // checker. All nested allocators will keep a reference to it.
+        d_limitChecker_sp = bsl::allocate_shared<AllocationLimitChecker>(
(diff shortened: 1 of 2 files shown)


src/groups/bmq/bmqma/bmqma_countingallocator.cpp
```diff
@@ -360,7 +350,39 @@ void CountingAllocator::deallocate(void* address)
...
+}
+
+void CountingAllocator::AllocationLimitChecker::update(
+    bsls::Types::Int64 deltaValue)
+{
+    const bsls::Types::Uint64 totalAllocated = d_allocated.addRelaxed(
+        deltaValue);
+
+    if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(
+            d_allocationLimit < totalAllocated && !d_alarmTriggered)) {
```
678098, line 363: In most cases this should short-circuit on the first check where we do not check atomic. This should help with performance.

hallfox (approved):

src/groups/bmq/bmqma/bmqma_countingallocator.cpp
678098: By the way, I run a benchmark comparing 2 approaches.

```

/// Mimics AllocationLimitChecker with non-atomic d_allocationLimit
struct NonAtomicLimitChecker {
    bsls::AtomicUint64  d_allocated;
    bsls::Types::Uint64 d_allocationLimit;
    bsls::AtomicBool    d_alarmTriggered;

    NonAtomicLimitChecker()
    : d_allocated(0)
    , d_allocationLimit(bsl::numeric_limits<bsls::Types::Uint64>::max())
    , d_alarmTriggered(false)
    {
    }

    void update(bsls::Types::Int64 deltaValue)
    {
        const bsls::Types::Uint64 totalAllocated =
            d_allocated.addRelaxed(deltaValue);

        if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(
                d_allocationLimit < totalAllocated && !d_alarmTriggered)) {
            d_alarmTriggered.swap(true);
        }
    }
};

/// Mimics the old design with atomic d_allocationLimit and a single
/// condition
struct AtomicLimitChecker {
    bsls::AtomicUint64 d_allocated;
    bsls::AtomicUint64 d_allocationLimit;

    AtomicLimitChecker()
    : d_allocated(0)
    , d_allocationLimit(bsl::numeric_limits<bsls::Types::Uint64>::max())
    {
    }

    void update(bsls::Types::Int64 deltaValue)
    {
        const bsls::Types::Uint64 totalAllocated =
            d_allocated.addRelaxed(deltaValue);

        if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(
                totalAllocated > d_allocationLimit)) {
            const bsls::Types::Uint64 uint64Max =
                bsl::numeric_limits<bsls::Types::Uint64>::max();
            d_allocationLimit.swap(uint64Max);
        }
    }
};

BSLA_MAYBE_UNUSED
static void testN2_performance_limitCheck()
// ------------------------------------------------------------------------
// PERFORMANCE - limit check (atomic vs non-atomic)
//
// Concerns:
//   Measure the performance difference between checking an allocation
//   limit stored as a non-atomic value (current design) versus an
//   atomic value (previous design) in the fast path of
//   AllocationLimitChecker::update.
//
// Plan:
//   1. Create two checker-like objects: one using a non-atomic limit
//      with two conditions (current), one using an atomic limit with a
//      single condition (previous).
//   2. Run a tight loop of update() calls that stay under the limit
//      (the common fast path).
//   3. Compare timings.
//
// Testing:
//   Performance of non-atomic vs atomic d_allocationLimit reads.
// ------------------------------------------------------------------------
{
    bmqtst::TestHelper::printTestName("PERFORMANCE - limit check"
                                      " (atomic vs non-atomic)");

    const bsls::Types::Int64 k_NUM_ITERATIONS = 10000000;  // 10M

    bsls::Types::Int64 timeNonAtomic = 0;
    {
        PV("--------------------------------");
        PV("Non-atomic limit (current)");
        PV("--------------------------------");

        NonAtomicLimitChecker checker;

        bsls::Types::Int64 start = bsls::TimeUtil::getTimer();
        for (bsls::Types::Int64 i = 0; i < k_NUM_ITERATIONS; ++i) {
            checker.update(1);
        }
        bsls::Types::Int64 end = bsls::TimeUtil::getTimer();

        timeNonAtomic = end - start;
        bsl::cout << "    Non-atomic limit: " << k_NUM_ITERATIONS
                  << " updates in "
                  << bmqu::PrintUtil::prettyTimeInterval(timeNonAtomic)
                  << " (avg " << timeNonAtomic / k_NUM_ITERATIONS
                  << " ns/call)\n";
    }

    bsls::Types::Int64 timeAtomic = 0;
    {
        PV("--------------------------------");
        PV("Atomic limit (previous)");
        PV("--------------------------------");

        AtomicLimitChecker checker;

        bsls::Types::Int64 start = bsls::TimeUtil::getTimer();
        for (bsls::Types::Int64 i = 0; i < k_NUM_ITERATIONS; ++i) {
            checker.update(1);
        }
        bsls::Types::Int64 end = bsls::TimeUtil::getTimer();

        timeAtomic = end - start;
        bsl::cout << "    Atomic limit:     " << k_NUM_ITERATIONS
                  << " updates in "
                  << bmqu::PrintUtil::prettyTimeInterval(timeAtomic)
                  << " (avg " << timeAtomic / k_NUM_ITERATIONS
                  << " ns/call)\n";
    }

    PV("");
    bsl::cout << "    Ratio (atomic / non-atomic): "
              << (static_cast<double>(timeAtomic) /
                  static_cast<double>(timeNonAtomic))
              << " (>1 means non-atomic is faster)\n";
}
```

678098: # Results

## Darwin M2

```
    Non-atomic limit: 10000000 updates in 49.25 ms (avg 4 ns/call)
    Atomic limit:     10000000 updates in 73.90 ms (avg 7 ns/call)
    Ratio (atomic / non-atomic): 1.50054 (>1 means non-atomic is faster)
```

## Linux amd64

```
    Non-atomic limit: 10000000 updates in 55.60 ms (avg 5 ns/call)
    Atomic limit:     10000000 updates in 75.89 ms (avg 7 ns/call)
    Ratio (atomic / non-atomic): 1.36508 (>1 means non-atomic is faster)
```

PR #1339, opened 2026-04-28
Cleanup[mqbc::PartitionStateTable]: Declare REAPPLY_SELF_PRIMARY/REPLICA events 

kaikulimu:
## Summary   
                                                                                                                                             
  - Introduce `REAPPLY_SELF_PRIMARY` / `REAPPLY_SELF_REPLICA` partition FSM events, distinct from `DETECT_SELF_PRIMARY` / `DETECT_SELF_REPLICA`, to cleanly separate CSL-originated role detection from reapply retries (watchdog timeout, error recovery).
  - Add a gatekeeper check in `enqueuePartitionFSMEventDispatched()` to drop redundant `DETECT_SELF_*` events when primary + leaseId is identical to current state. `REAPPLY_SELF_*` events bypass this check since they intentionally carry the same identity to retry a failed healing attempt.
  - Prohibit `mqbc::StorageManager::setPrimaryForPartition` — in FSM mode, `d_partitionInfoVec` is managed exclusively by the Partition FSM's `do_setPrimary` action. Remove `setPrimaryForPartitionDispatched`.                                                                        
  - Remove the legacy `setPartitionPrimary` test helper which called `setPrimaryForPartition`.

PR #1341, opened 2026-04-29
Cleanup[mqbc::ClusterStateTable]: Declare REAPPLY_LDR/FOL events

kaikulimu:
## Summary           
  - Similar changes to #1339.                                                                                                                                                                                                                            
  - Introduce `REAPPLY_LDR` / `REAPPLY_FOL` cluster FSM events, distinct from `SLCT_LDR` / `SLCT_FOL`, to cleanly separate  Elector-originated role selection from reapply retries (watchdog timeout).                                                                                     
  - Wire `do_reapplySelectLeader` and `do_reapplySelectFollower` to emit the new `REAPPLY_*` events instead of reusing `SLCT_*` events.                                                                                                         
  - Add a gatekeeper in `applyFSMEvent()` to drop redundant `SLCT_LDR` when the node is already leader.


src/groups/mqb/mqbc/mqbc_clusterstatetable.h
```diff
@@ -133,82 +133,88 @@ struct ClusterStateTableEvent {
...
     /// Enumeration used to distinguish among different type of event.
     enum Enum {
-        /// Self is transitioned as leader, or the transistion is being
-        /// reapplied due to a failure condition such as watchdog firing.
+        /// Self is transitioned as leader by the Elector.
         e_SLCT_LDR = 0,
 
-        /// Self is transitioned as follower, or the transistion is being
-        /// reapplied due to a failure condition such as watchdog firing.
+        /// Self is transitioned as follower by the Elector.
```
678098, line 139:
```suggestion
```
Possible to group 2 enums with 1 shared comment, the same thing we did in the previous PR

src/groups/mqb/mqbc/mqbc_clusterstatetable.h
```diff
@@ -133,82 +133,88 @@ struct ClusterStateTableEvent {
...
         e_SLCT_FOL = 1,
 
+        /// Generated by the reapply mechanism (watchdog timeout) to retry
+        /// leader healing.
+        e_REAPPLY_LDR = 2,
+
+        /// Generated by the reapply mechanism (watchdog timeout) to retry
+        /// follower healing.
```
678098, line 147:
```suggestion
```
Same, possible to group enums together with 1 comment

src/groups/mqb/mqbc/mqbc_clusterstatetable.h
```diff
@@ -435,10 +441,19 @@ class ClusterStateTable
             SLCT_LDR,
             startWatchDog_storeSelfLSN_sendFollowerLSNRequests_checkLSNQuorum,
             LDR_HEALING_STG1);
+        CST_CFG(
+            UNKNOWN,
+            REAPPLY_LDR,
+            startWatchDog_storeSelfLSN_sendFollowerLSNRequests_checkLSNQuorum,
```
678098, line 447: I am thinking if we can refactor such code with something like

```suggestion
            {startWatchDog, storeSelfLSN, sendFollowerLSNRequests, checkLSNQuorum},
```

So basically we can have small building block functions and can chain them in a sequence. Can use initializer lists if they are available, or think of something else. This allows us to get rid of long function names.

If you think it is good, we can do it in the future

678098 (approved):
Looks good, have some small comments

src/groups/mqb/mqbc/mqbc_clusterstatetable.h
kaikulimu: Yeah, in the future

PR #1346, opened 2026-04-30
Cleanup[bmqt::AuthnCallback]: immutable interface

678098:
- Remove setters from `bmqt::AuthnCallback`
- Require `bsl::string_view` in `bmqt::AuthnCallback` constructor to avoid constructing an intermediate `bsl::string` when `const char *` is provided, in cases like `bmqt::AuthnCallback("ANONYMOUS", data);`

### DIFF
--- a/src/groups/bmq/bmqt/bmqt_authncredential.cpp
+++ b/src/groups/bmq/bmqt/bmqt_authncredential.cpp
@@ -95,20 +95,6 @@ AuthnCredential::~AuthnCredential()
     // NOTHING
 }
 
-// MANIPULATORS
-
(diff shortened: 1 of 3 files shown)


678098:
I don't have strong opinion about changing the `AuthnCredentialCb` from

```cpp
    typedef bsl::function<bsl::optional<AuthnCredential>(bsl::ostream& error)>
        AuthnCredentialCb;
```

to

```cpp
    typedef bsl::function<bsl::optional<const AuthnCredential>(bsl::ostream& error)>
        AuthnCredentialCb;
```

`AuthnCredential` is immutable by design after this PR and writing `const` everywhere might be too much

pniedzielski (approved):
👍🏻 this is a good change

PR #1347, opened 2026-04-30
Fix[mqb]: removing erroneous assertion

dorjesinpo:
`ClusterQueueHelper::subtractCounters` should not assert `qinfo->d_queue_sp`
It can be `0` if queue creation failed and `subtractCounters` is part of rollback


678098 (approved):

dorjesinpo:
Will not merge yet.  Looking at IT failures

PR #1350, opened 2026-04-30
Remove: Emacs mode variable settings

pniedzielski:
Each file in the repository has a needless `-*-C++-*-`, which used to be used to ensure that Emacs opened all C++ files (including header files) in c++-mode rather than c-mode.  But, Emacs only looks in the first two lines of the file for these mode variable settings, so now that they are after the copyright notice, they serve no purpose.  This patch removes the comment with the filename and the mode variable setting in each file that is not generated from an XSD schema.


kaikulimu (approved):
999 lines deleted. Grate job :)

PR #1351, opened 2026-05-01
Fix[mqb]: race between recoveredQueuesCb and setActivePrimary

dorjesinpo:
`FileStore::setActivePrimary` and `StorageUtil::recoveredQueuesCb` can happen in any order
So, issuing implicit Receipts from both (used to be done in `FileStore::setActivePrimary` only)

And reverting https://github.com/bloomberg/blazingmq/pull/1265 as it is a partial fix.

2.  Unrelated.  `QueueHandle::configureDispatched` can be called _after_ `QueueHandle::clearClientDispatched` and needs to survive that.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
@@ -1121,11 +1121,12 @@ void QueueHandle::configureDispatched(
         d_queue_sp->configureHandle(this, streamParameters, configuredCb);
     }
     else {
-        BALL_LOG_ERROR
-            << "#CLIENT_IMPROPER_BEHAVIOR "
-            << "Attempting to configure handle without a client. Queue '"
-            << d_queue_sp->uri() << "', stream params: " << streamParameters
-            << ", handlePtr '" << this << "'.";
+        BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
+
+        BALL_LOG_ERROR << "#CLIENT_IMPROPER_BEHAVIOR "
+                       << "Attempting to configure handle [" << this
+                       << "] without a client, queue '" << d_queue_sp->uri()
+                       << "', stream params: [" << streamParameters << "].";
 
         bmqp_ctrlmsg::Status status(d_allocator_p);
         status.category() = bmqp_ctrlmsg::StatusCategory::E_UNKNOWN;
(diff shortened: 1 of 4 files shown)


src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
```diff
@@ -1111,15 +1111,30 @@ void QueueHandle::configureDispatched(
...
                   << ", id: " << id()
                   << ", appId: " << streamParameters.appId()
                   << "] with streamParameters: " << streamParameters;
 
-    d_queue_sp->configureHandle(this, streamParameters, configuredCb);
+    if (BSLS_PERFORMANCEHINT_PREDICT_LIKELY(d_haveClient)) {
+        d_queue_sp->configureHandle(this, streamParameters, configuredCb);
+    }
+    else {
+        BALL_LOG_ERROR
```
678098:
```suggestion
        BSLS_PERFORMANCEHINT_HINT_UNLIKELY;

        BALL_LOG_ERROR
```

src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
```diff
@@ -1111,15 +1111,30 @@ void QueueHandle::configureDispatched(
...
+            << "#CLIENT_IMPROPER_BEHAVIOR "
+            << "Attempting to configure handle without a client. Queue '"
+            << d_queue_sp->uri() << "', stream params: " << streamParameters
+            << ", handlePtr '" << this << "'.";
```
678098: `streamParameters` might be very long, possibly containing hundreds of subscriptions.
Printing `this` after `streamParameters` makes debug more difficult.
Since this log belongs to `QueueHandle`, why not printing it first?

src/groups/mqb/mqbs/mqbs_filestore.h
```diff
@@ -638,6 +638,7 @@ class FileStore BSLS_KEYWORD_FINAL : public DataStore {
     /// Send previously generated Replication Receipt to the specified `node`
     /// using the specified `nodeContext`.
     void sendReceipt(mqbnet::ClusterNode* node, NodeContext* nodeContext);
+    void sendImplicitReceipt();
```
678098, line 661: Missing docs for the newly added function.

678098 (approved):
Looks good

PR #1354, opened 2026-05-02
Fix[mqbblp::Domain]: fix thread-unsafe config write from admin thread

678098:
`Domain::config` is read from multiple threads.
Admin thread might reconfigure a domain, updating `d_config` while it is being read.
This might result in the following crash:

```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f21b9e58ee5 in __GI_abort () at abort.c:79
#2  0x0000000000abdfbf in BloombergLP::bsls::AssertImpUtil::failByAbort ()
    at /groups/bsl/bsls/bsls_assertimputil.cpp:91
#3  0x0000000000b218e3 in bmqAssertHandler (
    comment=0x1d125ec "!d_config.isNull()", 
    file=0x1d12600 
"/blazingmq/src/groups/mqb/mqbblp/mqbblp_domain.h
", line=362)
    at 
/blazingmq/src/applications/bmqbrkr/bmqbrkr.m.cpp:185
#4  0x0000000001614737 in BloombergLP::bsls::Assert::invokeHandler (
    violation=...)
    at /groups/bsl/bsls/bsls_assert.cpp:252
#5  0x0000000000d67cd4 in BloombergLP::mqbblp::Domain::config (
```

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_domain.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_domain.cpp
@@ -101,10 +101,10 @@ bool validateSubscriptionExpression(bsl::ostream& errorDescription,
 
 /// Validates a domain configuration. If `previousDefn` is provided, also
 /// checks that the implied reconfiguration is also valid.
-int validateConfig(bsl::ostream& errorDescription,
-                   const bdlb::NullableValue<mqbconfm::Domain>& previousDefn,
-                   const mqbconfm::Domain&                      newConfig,
-                   bslma::Allocator*                            allocator)
+int validateConfig(bsl::ostream&           errorDescription,
+                   const mqbconfm::Domain* previousDefn,
+                   const mqbconfm::Domain& newConfig,
+                   bslma::Allocator*       allocator)
 {
     enum RcEnum {
         // Value for the various RC error categories
@@ -164,13 +164,13 @@ int validateConfig(bsl::ostream& errorDescription,
 
     // 2. Check compatibility between old/new configurations,
     // if old configuration exists
-    if (previousDefn.isNull()) {
+    if (0 == previousDefn) {
         // First time configure, nothing more to validate
         return rc_SUCCESS;  // RETURN
     }
 
     // Validate properties of new configurations relative to old ones.
-    const mqbconfm::Domain& previousCfg = previousDefn.value();
+    const mqbconfm::Domain& previousCfg = *previousDefn;
 
     // Reconfiguring the routing mode is not allowed.
     if (previousCfg.mode().selectionId() != newConfig.mode().selectionId()) {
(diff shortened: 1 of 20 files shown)


src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -320,7 +325,7 @@ int Domain::configure(bsl::ostream&           errorDescription,
     }
 
     // Return early if there are no changes.
-    if (oldConfig && (finalConfig == oldConfig.value())) {
+    if (oldConfig && (finalConfig == *oldConfig)) {
```
hallfox, line 328: There is no lock on access to the old pointer, is it safe to do a comparison now? The old method made a copy (although without getting a lock.)

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -674,7 +685,7 @@ int Domain::processCommand(mqbcmd::DomainResult*        result,
             options.setSpacesPerLevel(2);
 
             BSLA_MAYBE_UNUSED const int rc = encoder.encode(out,
-                                                            d_config.value(),
+                                                            *cfgSnapshot,
```
hallfox, line 688: Same here, are we sure this is safe to read?

src/groups/mqb/mqbblp/mqbblp_domain.cpp
678098: We never modify the contents of the configuration under the shared pointer -- it is `const mqbconfm::Domain`. It is also safe to make copies of the existing shared pointers or release them from multiple threads.

So it is safe to access the old pointer and its contents. It just contains the old snapshot of configuration.

src/groups/mqb/mqbblp/mqbblp_domain.cpp
678098: Yes, this is the point.
We can only construct immutable configuration objects and can only access the pointer by locking a mutex.
It is fine that the current configuration snapshot stored in `mqbblp::Domain` changes, because:
- We still hold a copy of a `shared_ptr`.
- Even if our config snapshot is outdated, we enqueue cluster/queue dispatcher events to propagate the changes in configuration after `mqbblp::Domain::configure` call

src/groups/mqb/mqbblp/mqbblp_domain.cpp
678098: Also, on `mqbblp::Domain::configure` we create a replacement `shared_ptr` pointing to another config object.

dorjesinpo (approved):
Thank you

PR #1355, opened 2026-05-04
Feat[bmq]: get by without blob sp pool

dorjesinpo:
Why insist on `BlobSpPool*` for `AckEventBuilder` and `PushEventBuilder`?

`MockSessionUtil` does not and should have it.

### DIFF
--- a/src/groups/bmq/bmqa/bmqa_mocksession.cpp
+++ b/src/groups/bmq/bmqa/bmqa_mocksession.cpp
@@ -138,12 +138,12 @@ struct BlobSpCreatorF {
     : d_bufferFactory_p(bufferFactory)
     , d_allocator_p(allocator)
     {
+        BSLS_ASSERT_SAFE(d_bufferFactory_p);
     }
     bsl::shared_ptr<bdlbb::Blob> operator()()
     {
-        return bsl::shared_ptr<bdlbb::Blob>(
-            new (*d_allocator_p) bdlbb::Blob(d_bufferFactory_p, d_allocator_p),
-            d_allocator_p);
+        return bsl::allocate_shared<bdlbb::Blob>(d_allocator_p,
+                                                 d_bufferFactory_p);
     }
 };
 
--- a/src/groups/bmq/bmqa/bmqa_mocksession.cpp
+++ b/src/groups/bmq/bmqa/bmqa_mocksession.cpp
@@ -128,6 +128,133 @@ UriCorrIdToQueueMap& uriCorrIdToQueues(B& buffer)
 {
     return reinterpret_cast<UriCorrIdToQueueMap&>(*(buffer.buffer()));
 }
+
+struct BlobSpCreatorF {
+    bdlbb::BlobBufferFactory* d_bufferFactory_p;
+    bslma::Allocator*         d_allocator_p;
+
+    explicit BlobSpCreatorF(bdlbb::BlobBufferFactory* bufferFactory,
+                            bslma::Allocator*         allocator)
+    : d_bufferFactory_p(bufferFactory)
+    , d_allocator_p(allocator)
+    {
+    }
+    bsl::shared_ptr<bdlbb::Blob> operator()()
+    {
+        return bsl::shared_ptr<bdlbb::Blob>(
+            new (*d_allocator_p) bdlbb::Blob(d_bufferFactory_p, d_allocator_p),
+            d_allocator_p);
+    }
+};
+
+Event createAckEventImpl(bmqp::AckEventBuilder& ackBuilder,
+                         const bsl::vector<MockSessionUtil::AckParams>& acks,
+                         bdlbb::BlobBufferFactory* bufferFactory,
+                         bslma::Allocator*         allocator)
+{
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(!acks.empty());
+    BSLS_ASSERT_SAFE(bufferFactory);
+
+    /// Event impl shared pointer to access
+    /// the pimpl of `bmqa::Event`.
+    typedef bsl::shared_ptr<bmqimp::Event> EventImplSp;
+
+    /// Queue impl shared pointer to access
+    /// the pimpl of `bmqa::QueueId`.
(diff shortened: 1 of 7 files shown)


dorjesinpo:
wrt IT failure - https://github.com/bloomberg/blazingmq/pull/1356

src/groups/bmq/bmqa/bmqa_mocksession.h
```diff
@@ -694,15 +694,23 @@ struct MockSessionUtil {
     /// Create and return an `Event` configured as a message event of type
     /// `e_ACK` with the specified `acks` params using the specified
     /// `bufferFactory` and the specified `allocator` to supply memory.
+    static Event createAckEvent(const bsl::vector<AckParams>& acks,
+                                bdlbb::BlobBufferFactory*     bufferFactory,
+                                bslma::Allocator*             allocator);
     static Event createAckEvent(const bsl::vector<AckParams>& acks,
```
pniedzielski, line 703: Doxygen doesn't copy docs for overloads, so this overload will end up being undocumented and not clearly linked to the other overload.  Can we add Doxygen comments for this and `createPushEvent` below?

pniedzielski (dismissed):
This way of doing it seems okay to me.

src/groups/bmq/bmqa/bmqa_mocksession.cpp
```diff
@@ -129,6 +129,133 @@ UriCorrIdToQueueMap& uriCorrIdToQueues(B& buffer)
...
+                            bslma::Allocator*         allocator)
+    : d_bufferFactory_p(bufferFactory)
+    , d_allocator_p(allocator)
+    {
+    }
+    bsl::shared_ptr<bdlbb::Blob> operator()()
+    {
+        return bsl::shared_ptr<bdlbb::Blob>(
+            new (*d_allocator_p) bdlbb::Blob(d_bufferFactory_p, d_allocator_p),
+            d_allocator_p);
```
678098: `bsl::allocate_shared` is:
- more clear
- doesn't have explicit `new`
- impossible to misuse by providing different allocators to `new` and to `shared_ptr`

src/groups/bmq/bmqa/bmqa_mocksession.cpp
```diff
@@ -129,6 +129,133 @@ UriCorrIdToQueueMap& uriCorrIdToQueues(B& buffer)
...
+
+struct BlobSpCreatorF {
+    bdlbb::BlobBufferFactory* d_bufferFactory_p;
+    bslma::Allocator*         d_allocator_p;
+
+    explicit BlobSpCreatorF(bdlbb::BlobBufferFactory* bufferFactory,
+                            bslma::Allocator*         allocator)
+    : d_bufferFactory_p(bufferFactory)
+    , d_allocator_p(allocator)
+    {
```
678098, line 140: Missing precondition check for `d_bufferFactory_p` being non-null. We can do it here early

src/groups/bmq/bmqa/bmqa_mocksession.cpp
678098: Also, when `new` is used with `d_allocator_p`, it might crash if `d_allocator_p == 0`
`bsl::allocate_shared` handles null allocator

src/groups/bmq/bmqp/bmqp_pusheventbuilder.cpp
```diff
@@ -42,6 +42,19 @@ bool isDefaultSubQueueInfo(const Protocol::SubQueueInfosArray& subQueueInfos)
            subQueueInfos[0].id() == Protocol::k_DEFAULT_SUBSCRIPTION_ID;
 }
 
+struct BlobSpCreatorF {
+    bmqp::BlobPoolUtil::BlobSpPool* d_blobSpPool_p;
+
+    explicit BlobSpCreatorF(bmqp::BlobPoolUtil::BlobSpPool* blobSpPool)
+    : d_blobSpPool_p(blobSpPool)
+    {
+    }
```
678098, line 51: Missing check that `d_blobSpPool_p` is not null.
We have this check in event builder, but this functor is not complete on its own without this check.

src/groups/bmq/bmqp/bmqp_pusheventbuilder.h
```diff
@@ -164,6 +164,8 @@ class PushEventBuilder {
     /// set BlobBufferFactory since we might want to expand the built Blob
     /// dynamically.
     PushEventBuilder(BlobSpPool* blobSpPool_p, bslma::Allocator* allocator);
```
678098, line 165: We can remove the old constructor (expecting `BlobSpPool* blobSpPool_p`) and keep the one that expects `BlobSpCreator` only.

For this, we need to:
- Remove dependency on `bmqp::BlobPoolUtil` from event builders.
- Move `struct BlobSpCreatorF` to `bmqp_blobpoolutil` and use it from the code that constructs event builders.

678098: Might name it `BlobSpPooledCreatorF`

dorjesinpo: This is public, may break somewhere

src/groups/bmq/bmqp/bmqp_ackeventbuilder.cpp
```diff
@@ -30,21 +30,60 @@
...
 namespace bmqp {
 
+namespace {
+
+struct BlobSpCreatorF {
+    bmqp::BlobPoolUtil::BlobSpPool* d_blobSpPool_p;
+
+    explicit BlobSpCreatorF(bmqp::BlobPoolUtil::BlobSpPool* blobSpPool)
+    : d_blobSpPool_p(blobSpPool)
+    {
```
678098, line 40:
```suggestion
    {
        BSLS_ASSERT_SAFE(d_blobSpPool_p);
```
The functor itself is incomplete because it allows constructing itself with null ptr, that results in null ptr dereference

678098 (changes requested):

678098 (approved):

678098 (approved):

PR #1359, opened 2026-05-05
Remove: `flushAtShutdown` broker configuration

pniedzielski:
This pull request removes the `flushAtShutdown` broker configuration flag.  It is safe to do this because we never BER encode our broker configuration, just JSON-encode it, and we ignore configuration flags we don't know about.

### DIFF
--- a/docker/cluster/config/clusters.json
+++ b/docker/cluster/config/clusters.json
@@ -50,7 +50,6 @@
             ],
             "partitionConfig": {
                 "name": "local",
-                "flushAtShutdown": true,
                 "location": "/var/local/bmq/storage",
                 "maxArchivedFileSets": 0,
                 "maxDataFileSize": 268435456,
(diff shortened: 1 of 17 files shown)


pniedzielski:
`DataStore` takes a `flush` parameter to its `close()` function.  This PR makes it clear that in two places, we do not flush on close:

1. `mqbc::StorageManager` (line 3276): When storage is out of sync with the primary, we close and move archive the storage.  We don't flush with whatever the old storage was.

2. `mqbblp::StorageManager` (line 265): In local, non-clustered storage, when a buffered storage event fails to apply during recovery, we close the storage.  We don't flush with the partially applied event here.

In each of these cases, we could flush without sacrificing correctness.

kaikulimu:
@pniedzielski CI looks quite broken. Can you take a look?

kaikulimu (dismissed):
LGTM

PR #1296, opened 2026-04-13
Feat[mqb]: removing registerQueueReplica

dorjesinpo:
Replica always runs in legacy mode

PR #1361, opened 2026-05-06
Fix[mqb]: count without the flag is not valid

dorjesinpo:
If `bmqp_ctrlmsg::QueueHandleParameters` contain a count (for example `d_adminCount`) without corresponding flag, this gets the handle stuck after `QueueHandleCatalog::releaseHandleHelper` .
As the result, queue is not deleted and `DomainMgr::stop` gets stuck

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_queueutil.cpp
+++ b/src/groups/bmq/bmqp/bmqp_queueutil.cpp
@@ -46,17 +46,26 @@ bool QueueUtil::isValid(const bmqp_ctrlmsg::QueueHandleParameters& parameters)
             return false;  // RETURN
         }
     }
+    else if (parameters.readCount() > 0) {
+        return false;  // RETURN
+    }
 
     if (bmqt::QueueFlagsUtil::isWriter(parameters.flags())) {
         if (parameters.writeCount() <= 0) {
             return false;  // RETURN
         }
     }
+    else if (parameters.writeCount() > 0) {
+        return false;  // RETURN
+    }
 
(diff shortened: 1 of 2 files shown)


dorjesinpo:
IT failures
`test_add_remove_add_app[multi_node_fsm-strong_consistency]` - https://github.com/bloomberg/blazingmq/pull/1364

`test_load_domain_authorize_open_post[multi_node_fsm-strong_consistency]` - supposedly https://github.com/bloomberg/blazingmq/pull/1296

`test_puts_retransmission.py::TestPutsRetransmission::test_shutdown_replica[multi_node_fsm-strong_consistency]` - https://github.com/bloomberg/blazingmq/pull/1316

src/groups/bmq/bmqp/bmqp_queueutil.cpp
```diff
@@ -46,18 +46,27 @@ bool QueueUtil::isValid(const bmqp_ctrlmsg::QueueHandleParameters& parameters)
             return false;  // RETURN
         }
     }
+    else if (parameters.readCount() > 0) {
+        return false;  // RETURN
+    }
```
678098, line 51: What do you think about this approach?
```suggestion
    const bool isReader = bmqt::QueueFlagsUtil::isReader(parameters.flags());
    const bool hasReadCount = parameters.readCount() > 0;
    
    if (!(isReader == hasReadCount)) {
        return false;  // RETURN
    }
```

This offers less branching

678098 (approved):
Possible to simplify, otherwise looks good

PR #1362, opened 2026-05-06
Fix[mqb]: disable reconnecting on shutdown

dorjesinpo:
`TCPSessionFactory::stopListening()` prevents new incoming connections.
Outgoing connections should be disabled as well.

### DIFF
--- a/src/groups/bmq/bmqio/bmqio_reconnectingchannelfactory.cpp
+++ b/src/groups/bmq/bmqio/bmqio_reconnectingchannelfactory.cpp
@@ -184,6 +184,10 @@ ReconnectingChannelFactory_ConnectHandle::properties() const
 void ReconnectingChannelFactory::scheduleConnect(
     const ConnectHandlePtr& handle)
 {
+    if (!d_isReconnectEnabled) {
+        return;  // RETURN
+    }
+
     if (!handle->d_endpoints.empty()) {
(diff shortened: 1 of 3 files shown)


src/groups/bmq/bmqio/bmqio_reconnectingchannelfactory.cpp
```diff
@@ -440,6 +440,10 @@ void ReconnectingChannelFactory::onChannelDown(const ConnectHandlePtr& handle)
     // We only monitor autoReconnect channel for the down event
     BSLS_ASSERT_SAFE(handle->d_options.autoReconnect());
 
+    if (!d_isReconnectEnabled) {
+        return;  // RETURN
+    }
```
678098: There is a second call to `scheduleConnect` in void `ReconnectingChannelFactory::connectResultCb` that happens if channel never goes up in the first place.
Don't we need to handle it too?

dorjesinpo: makes sense.  Amended

678098 (approved):
Looks good

chrisbeard:
This seems like it fixed a regression that was introduced somewhere between 0.95.6 and 0.95.8, do we have an explanation for that?

PR #1365, opened 2026-05-06
Fix[mqbc::ClusterStateMgr]: Only cancel requests initiated by Cluster FSM 

kaikulimu:
## Summary
  - `ClusterStateManager::do_cancelRequests` currently calls `cancelAllRequests()` with no filter on the shared `ClusterData::requestManager`, which cancels  **all** outstanding requests — including those from other components sharing the same `RequestManager` instance. This is too broad and can interfere with unrelated in-flight operations.
  - This PR adds a `componentId` dimension to `bmqp::RequestManager` so that each component can tag its requests and cancel only its own. The Cluster FSM now tags its requests with `e_CLUSTER_FSM` and `do_cancelRequests` only cancels those.
  - The existing `groupId`-based cancellation (used for per-node cancellation when a node goes down) is unchanged and orthogonal to the new `componentId` axis.

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_requestmanager.h
+++ b/src/groups/bmq/bmqp/bmqp_requestmanager.h
@@ -861,15 +861,15 @@ class RequestManager {
     /// behavior is undefined if `groupId` is `NO_GROUP_ID`.  The
     /// corresponding response callbacks will be invoked in the order in
     /// which requests were sent.
-    void cancelAllRequests(const RESPONSE& reason, int groupId);
+    void cancelGroupRequests(const RESPONSE& reason, int groupId);
 
     /// Cancel all outstanding requests tagged with the specified
     /// `componentId`, with the specified `reason` response description.
     /// The behavior is undefined if `componentId` is `e_NO_COMPONENT_ID`.
     /// The corresponding response callbacks will be invoked in the order
     /// in which requests were sent.
-    void cancelAllRequests(const RESPONSE&                 reason,
-                           RequestManagerComponentId::Enum componentId);
+    void cancelComponentRequests(const RESPONSE&                 reason,
+                                 RequestManagerComponentId::Enum componentId);
 };
 
 // ============================================================================
@@ -1525,7 +1525,7 @@ void RequestManager<REQUEST, RESPONSE>::cancelAllRequests(
 }
 
 template <class REQUEST, class RESPONSE>
-void RequestManager<REQUEST, RESPONSE>::cancelAllRequests(
+void RequestManager<REQUEST, RESPONSE>::cancelGroupRequests(
     const RESPONSE& reason,
     int             groupId)
 {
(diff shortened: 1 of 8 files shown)


src/groups/bmq/bmqp/bmqp_requestmanager.h
```diff
@@ -1496,15 +1590,29 @@ void RequestManager<REQUEST, RESPONSE>::cancelAllRequestsImpl(
...
+        BALL_LOG_INFO << "Canceling requests belonging to component '"
+                      << componentId << "' (" << requestsCopy.size()
+                      << " items) with " << reason << ".";
+    }
+    else {
+        BALL_LOG_INFO << "Canceling requests belonging to group '" << groupId
+                      << "' and component '" << componentId << "' ("
+                      << requestsCopy.size() << " items) with " << reason
+                      << ".";
+    }
```
hallfox, line 1615: I think instead of checking all combinations of component and group values, we should just have a log for each case, e.g.

```c++
if (hasGroup) {
  BALL_LOG_INFO << "Cancelling requests belonging to group ...";
}
if (hasComponent) {
  BALL_LOG_INFO << "Cancelling requests belonging to component ...";
}
if (!hasGroup || !hasComponent) {
  BALL_LOG_INFO << "Canceling all requests ...";
}
```

src/groups/bmq/bmqp/bmqp_requestmanager.h
```diff
@@ -817,6 +862,14 @@ class RequestManager {
...
     /// which requests were sent.
     void cancelAllRequests(const RESPONSE& reason, int groupId);
+
+    /// Cancel all outstanding requests tagged with the specified
+    /// `componentId`, with the specified `reason` response description.
+    /// The behavior is undefined if `componentId` is `e_NO_COMPONENT_ID`.
+    /// The corresponding response callbacks will be invoked in the order
+    /// in which requests were sent.
+    void cancelAllRequests(const RESPONSE&                 reason,
+                           RequestManagerComponentId::Enum componentId);
```
hallfox: The difference in types between the groupId overload and this overload feel so easy to get wrong silently from implicit type conversion. Can we rename these to not be overloads,  e.g. `cancelComponentRequests` and `cancelGroupRequests`?

src/groups/bmq/bmqp/bmqp_requestmanager.h
kaikulimu: If both `hasGroup` and `hasComponent` are true, we cancel requests simultaneously belonging to that group and that component. To reduce confusion, I updated the last case to say:
```
        BALL_LOG_INFO
            << "Canceling requests simultaneously belonging to group '"
            << groupId << "' and component '" << componentId << "' ("
            << requestsCopy.size() << " items) with " << reason << ".";
```

hallfox (approved):

PR #1367, opened 2026-05-06
Style: Remove anonymous namespace from header

pniedzielski:
This patch removes the anonymous namespace containing `k_INVALID_PARTITION_ID` from mqbc_storageutil’s header, moving it instead to the utility struct contained within that file.

This patch also documents why there are two `k_INVALID_PARTITION_ID` constants in different components that seemingly have different values.

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_storageutil.h
+++ b/src/groups/mqb/mqbc/mqbc_storageutil.h
@@ -92,18 +92,19 @@ class ClusterNode;
 
 namespace mqbc {
 
-namespace {
-/// Constant used to represent an invalid partition id.
-const unsigned int k_INVALID_PARTITION_ID =
-    bsl::numeric_limits<unsigned int>::max();
-}  // close unnamed namespace
-
 // ==================
 // struct StorageUtil
 // ==================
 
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqbc/mqbc_storageutil.h
```diff
@@ -92,18 +92,19 @@ class ClusterNode;
...
 // ==================
 
 /// Generic utilities for storage related operations.
 struct StorageUtil {
+  public:
+    // PUBLIC CONSTANTS
+
+    /// Constant used to represent an invalid partition id.
+    static const unsigned int k_INVALID_PARTITION_ID =
+        bsl::numeric_limits<unsigned int>::max();
```
hallfox: a function call cannot appear in a constant-expression

kaikulimu: Not sure why it only complained after Patrick moved it around though

pniedzielski: should have built in C++03 before submitting :|

kaikulimu (approved):

PR #1368, opened 2026-05-06
Refactor[MQB]: better usage of domain config snapshots

678098:
Follow up for the PR https://github.com/bloomberg/blazingmq/pull/1354

- Removed unused `#include <bdlb_nullablevalue.h>`
- `mqbblp::Domain::config`: check pointer after unlocking the mutex
- `mqbc::ClusterUtil::assignQueue`: make exactly one snapshot of domain configuration for state consistency (was: 4)
- `mqbs::FileBackedStorage`: pass `const mqbconfm::Domain& config` to constructor to avoid making 4 independent snapshots of domain configuration and keep its state consistent


src/groups/mqb/mqbblp/mqbblp_domain.h
```diff
@@ -365,12 +364,16 @@ inline const bsl::string& Domain::name() const
...
+    bsl::shared_ptr<const mqbconfm::Domain> copy;
 
-    // PRECONDITIONS
-    BSLS_ASSERT_SAFE(d_config_sp);
+    {
+        bslmt::ReadLockGuard<bslmt::ReaderWriterMutex> guard(&d_configLock);
+        copy = d_config_sp;
+    }
 
-    return d_config_sp;
```
678098, line 373: The existing implementation is fine, but we can have a finer scope for lock and can show that we intend to make a copy

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -2497,13 +2497,13 @@ void StorageUtil::registerQueueAsPrimary(const mqbi::Cluster*    cluster,
...
-                fs,
-                appIdKeyPairs,
-                domain->config()->mode().isFanoutValue()));
+            (*event_sp).setCallback(
+                bdlf::BindUtil::bind(updateQueuePrimaryDispatched,
+                                     storageSp.get(),
+                                     storagesLock,
+                                     fs,
+                                     appIdKeyPairs,
+                                     true /* isFanout */));
```
hallfox, line 2506: Why do we know this is a fanout?

678098: Line 2489: `if (queueMode.isFanoutValue()) {`

hallfox (approved):

PR #1294, opened 2026-04-11
UT: Reduce stack frame noise in crash dumps by scoping heavy objects

SrinathhSatuluri:
#1286

When the broker crashes during `run()`, the main thread's stack trace is
flooded with large local variables that are no longer needed, making
crash triage harder. This PR limits the lifetime of three heavy objects:

1. Wrap `specTable` and `commandLine` in a scoped block in `main()` so
  they destruct before `run()`.
2. Reset `d_config` in `TaskEnvironment` to a default-constructed state
  before entering `run()`.
3. Remove the cached `const mqbcfg::TaskConfig& d_config` member from
  `Task` and pass it as a parameter to `initialize()` instead, which is
  the only place it was used.

**Testing performed**

1. Verified the broker builds and starts successfully with all three
  changes applied.
2. Confirmed CLI argument parsing still works correctly after scoping
  `specTable`/`commandLine`.
3. Confirmed `Task::initialize()` receives and uses the config parameter
  identically to the previous cached reference.

closes: #1286

PR #1369, opened 2026-05-07
UT[bmqbrkr]: Move d_configJson and d_config out of TaskEnvironment

### LINKED ISSUE #1319: Reduce stack footprint of TaskEnvironment on main thread
Follow-up to #1286 (merged in PR #1294 ). This PR scoped specTable/commandLine and reset d_config before run() to cut down main-thread stack noise, but looking at the crash dump, the taskEnv variable in main() is still pretty big. The maintainer mentioned we could address this in a separate PR.

**Reference comment**: https://github.com/bloomberg/blazingmq/pull/1294#issuecomment-4247786800

From what I can see in the dump, the main things taking up space are d_configJson (a bsl::string holding the raw config JSON, 3452 bytes, kept around for the whole lifetime of taskEnv) and d_config, which is an inline mqbcfg::Configuration with taskConfig, appConfig, network interfaces, dispatcher config, plugins, auth, etc., all embedded directly. The d_task and d_app buffers also seem to add a chunk.

**Some ideas that might work:**

- Putting d_configJson and/or d_config behind a ManagedPtr or unique_ptr so they live on the heap instead of inline in taskEnv
- Clearing d_configJson once parsing is done, if it's not needed after that
- Heap-allocating TaskEnvironment itself so only a pointer ends up on main's stack

would appreciate any guidance. Happy to work on this if it sounds good.

SrinathhSatuluri:
Closes #1319

This is a follow-up to #1294 with the lifetime change from #1329 in place. Both `d_configJson` and `d_config` are now locals in `main()` inside a scope that ends before `run()`, so they're destroyed before the broker starts its main work. That cuts down on what shows up under `taskEnv` in a crash dump.

The function signatures around config init were updated so they take only what they need. `getConfig` writes to caller pointers for the JSON string and the `Configuration`, and `initializeTask` takes a `TaskConfig` directly. `updateHistFile` reads from `BrokerConfig::get()` since the local config is out of scope by the time `run()` calls it. The config dump log line moved out of `initializeApplication` into `main()`.

This was unblocked by #1329, which made `BrokerConfig` store a value copy of `AppConfig` instead of a pointer to caller storage. With that in place, the local `Configuration` in `main()` can safely be destroyed after `BrokerConfig::set()` without dangling the global.

### DIFF
--- a/src/applications/bmqbrkr/bmqbrkr.m.cpp
+++ b/src/applications/bmqbrkr/bmqbrkr.m.cpp
@@ -90,13 +90,6 @@ struct TaskEnvironment {
     bsl::string d_bmqPrefix;
     // BMQ_PREFIX directory path
 
-    bsl::string d_configJson;
-    // JSON content ouput of the generated
-    // config
-
-    mqbcfg::Configuration d_config;
-    // De-serialized config
-
     bsls::ObjectBuffer<m_bmqbrkr::Task> d_task;
     // The Task
 
@@ -233,13 +226,14 @@ static void printStartStopTrace(const bslstl::StringRef& action,
 /// Retrieve the config from the `bmqbrkrcfg.json` file in the specified
 /// `configDir` file.
 ///
-/// Return 0 on success and store the JSON content, at well as its
-/// associated de-serialized object part in the corresponding field of the
-/// specified `taskEnv`, or return a non-zero error code and populate the
-/// specified `errorDescription` with a description of the error.
-static int getConfig(bsl::ostream&      errorDescription,
-                     TaskEnvironment*   taskEnv,
-                     const bsl::string& configDir)
+/// Return 0 on success and store the JSON content in the specified
+/// `configJson` and the de-serialized object in the specified `config`,
+/// or return a non-zero error code and populate the specified
+/// `errorDescription` with a description of the error.
+static int getConfig(bsl::ostream&          errorDescription,
+                     bsl::string*           configJson,
+                     mqbcfg::Configuration* config,
+                     const bsl::string&     configDir)
 {
     enum RcEnum {
         // Enum for the various RC error categories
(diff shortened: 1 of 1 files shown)


SrinathhSatuluri:
Hello @678098. Just finished scoping the locals in `main()`. One thing I wanted to highlight is that I noticed `initializeApplication` has `BSLA_MAYBE_UNUSED bsl::ostream& errorDescription` and only ever returns 0, which makes the `if (rc != 0)` block at its call site in `main()` unreachable. The parameter and return code look vestigial. Is it worth opening a separate cleanup issue to drop them, or is the current shape kept that way on purpose?

src/applications/bmqbrkr/bmqbrkr.m.cpp
```diff
@@ -280,26 +274,25 @@ static int getConfig(bsl::ostream&      errorDescription,
...
         return rc_DECODE_FAILED;  // RETURN
     }
 
-    taskEnv->d_config.appConfig().etcDir() = configDir;
-    taskEnv->d_config.appConfig().brokerVersion() =
-        bmqbrkrscm::Version::versionAsInt();
-    mqbcfg::BrokerConfig::set(taskEnv->d_config.appConfig());
+    config->appConfig().etcDir()        = configDir;
+    config->appConfig().brokerVersion() = bmqbrkrscm::Version::versionAsInt();
+    mqbcfg::BrokerConfig::set(config->appConfig());
```
678098: @SrinathhSatuluri could you rebase your branch to the latest main?
I have fixed a small issue in this PR: https://github.com/bloomberg/blazingmq/pull/1371

Basically you need to move `mqbcfg::BrokerConfig::set` just before the `initializeTask(errorDescription, &taskEnv, config.taskConfig());` call

SrinathhSatuluri: Done, thanks for the heads up on #1371. Rebased and moved `BrokerConfig::set` to right before the `initializeTask` call.

SrinathhSatuluri:
Hi @678098, looks like this branch is a few commits behind main. Should I rebase again, or is it fine as is until merge?

678098 (approved):
Looks good

678098:

PR #1370, opened 2026-05-07
Fix[bmqbrkr, bmq]: init allocation limits early

678098:
The existing broker code broke the contract of bmqma::CountingAllocator by configuring allocation limits after some allocations were done. Also, allocation limits are set after the log controller is started, resulting in a race condition.

Address this by setting the limits early.

### DIFF
--- a/src/applications/bmqbrkr/m_bmqbrkr_task.cpp
+++ b/src/applications/bmqbrkr/m_bmqbrkr_task.cpp
@@ -94,8 +94,11 @@ void onAllocationLimit(bsls::Types::Uint64 limit)
 // class Task_AllocatorManager
 // ---------------------------
 
-Task_AllocatorManager::Task_AllocatorManager(mqbcfg::AllocatorType::Value type)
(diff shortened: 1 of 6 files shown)


src/applications/bmqbrkr/m_bmqbrkr_task.cpp
```diff
@@ -229,8 +234,9 @@ void Task::onLogCommand(const bsl::string& prefix, bsl::istream& input)
 }
 
 Task::Task(const bsl::string&           bmqPrefix,
-           mqbcfg::AllocatorType::Value allocatorType)
-: d_allocatorManager(allocatorType)
+           mqbcfg::AllocatorType::Value allocatorType,
+           bsls::Types::Uint64          allocationLimit)
+: d_allocatorManager(allocatorType, allocationLimit)
```
678098, line 239: So now we will always set allocation limits before any allocations

hallfox (approved):

PR #1373, opened 2026-05-07
Fix[mqb]: thread-safe access to PartitionInfo

678098:
Make `mqbi::StorageManager_PartitionInfo::d_primaryStatus` atomic.

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_storagemanager.cpp
+++ b/src/groups/mqb/mqbc/mqbc_storagemanager.cpp
@@ -69,7 +69,7 @@ const bsls::Types::Int64 k_NS_PER_MESSAGE =
 
 const int k_GC_MESSAGES_INTERVAL_SECONDS = 5;
 
-bool isPrimaryActive(const mqbi::StorageManager_PartitionInfo pinfo)
+bool isPrimaryActive(const mqbi::StorageManager_PartitionInfo& pinfo)
 {
     return pinfo.primaryStatus() == bmqp_ctrlmsg::PrimaryStatus::E_ACTIVE;
 }
--- a/src/groups/mqb/mqbi/mqbi_storagemanager.h
+++ b/src/groups/mqb/mqbi/mqbi_storagemanager.h
@@ -51,6 +51,7 @@
 #include <bsl_utility.h>
 #include <bsl_vector.h>
 #include <bslma_managedptr.h>
+#include <bsls_atomic.h>
 #include <bsls_cpp11.h>
 
 namespace BloombergLP {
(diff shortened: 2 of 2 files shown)


src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -68,11 +68,6 @@ const bsls::Types::Int64 k_NS_PER_MESSAGE =
 
 const int k_GC_MESSAGES_INTERVAL_SECONDS = 5;
 
-bool isPrimaryActive(const mqbi::StorageManager_PartitionInfo pinfo)
```
678098, line 72: Value copy here by the way

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -3660,9 +3655,17 @@ bool StorageManager::allPartitionsAvailable() const
...
+        static bool
+        isPrimaryActive(const mqbi::StorageManager_PartitionInfo& pinfo)
+        {
+            return pinfo.isActive();
+        }
+    };
+
     return bsl::all_of(d_partitionInfoVec.cbegin(),
                        d_partitionInfoVec.cend(),
-                       &isPrimaryActive);
```
678098, line 3693: This code was executed from any partition thread. Other partition thread might modify its partition status at the same time, leading to TSAN observed error

src/groups/mqb/mqbi/mqbi_storagemanager.h
```diff
@@ -102,13 +102,21 @@ class StorageManager_PartitionInfo {
 
     bmqp_ctrlmsg::PrimaryStatus::Value d_primaryStatus;
 
+    bsls::AtomicBool d_isActive;
```
kaikulimu: Should we make `d_primaryStatus` itself atomic instead? Having two variable with very similar functionality, one being non-atomic and one being atomic, can be misused easily.

src/groups/mqb/mqbi/mqbi_storagemanager.h
```diff
@@ -122,6 +130,8 @@ class StorageManager_PartitionInfo {
     unsigned int primaryLeaseId() const;
 
     bmqp_ctrlmsg::PrimaryStatus::Value primaryStatus() const;
+
+    bool isActive() const;
```
kaikulimu: If we decide to keep both `primaryStatus()` and `isActive()`, we need to comment that only `isActive()` is thread-safe. Still, people might not read the comment and call `primaryStatus()` accidentally.

kaikulimu (changes requested):

kaikulimu (approved):

PR #1376, opened 2026-05-08
Fix: Pass `allocator` to `AllocationLimitChecker`

pniedzielski:
The `d_allocationLimitCb` `bsl::function` allocates using the global allocator on Solaris, causing bmqma_countingallocatorutil.t.cpp CASE 3 to fail with `'_defAlloc.numBlocksTotal()' (2) == '0' (0)`.  This patch passes an allocator down into that function to fix this.

### DIFF
--- a/src/groups/bmq/bmqma/bmqma_countingallocator.h
+++ b/src/groups/bmq/bmqma/bmqma_countingallocator.h
@@ -145,15 +145,20 @@ class CountingAllocator BSLS_KEYWORD_FINAL : public bslma::Allocator {
         AllocationLimitCallback d_allocationLimitCb;
 
       public:
+        // TRAITS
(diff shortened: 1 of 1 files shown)


src/groups/bmq/bmqma/bmqma_countingallocator.cpp
```diff
@@ -206,6 +206,7 @@ CountingAllocator::CountingAllocator(const bslstl::StringRef& name,
         // This is a root CountingAllocator: construct the original limit
         // checker. All nested allocators will keep a reference to it.
         d_limitChecker_sp = bsl::allocate_shared<AllocationLimitChecker>(
+            d_allocator_p,
             d_allocator_p);
```
678098, line 209: Let's add a nested trait declaration (uses bslma allocator) to avoid providing the same allocator to `bsl::allocate_shared` twice?

678098 (approved):

PR #1377, opened 2026-05-08
Fix[mqbi::StorageMessageAttributes]: incorrect order of args

678098:
`refCount` and `appDataLen` order is incorrect in `mqbi::StorageMessageAttributes` constructor declaration. Fix it and add a UT that checks `mqbi::StorageMessageAttributes` class API.

### DIFF
--- a/src/groups/mqb/mqbi/mqbi_storage.h
+++ b/src/groups/mqb/mqbi/mqbi_storage.h
@@ -215,8 +215,8 @@ class StorageMessageAttributes {
 
     StorageMessageAttributes(
         bsls::Types::Uint64                  arrivalTimestamp,
-        unsigned int                         appDataLen,
         unsigned int                         refCount,
+        unsigned int                         appDataLen,
         const bmqp::MessagePropertiesInfo&   messagePropertiesInfo,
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqbi/mqbi_storage.h
```diff
@@ -215,8 +215,8 @@ class StorageMessageAttributes {
 
     StorageMessageAttributes(
         bsls::Types::Uint64                  arrivalTimestamp,
-        unsigned int                         appDataLen,
         unsigned int                         refCount,
+        unsigned int                         appDataLen,
```
678098, line 219: Compare with definition:

https://github.com/bloomberg/blazingmq/blob/277e3c1e9faae1753b4dc645837a1ba85ee6022e/src/groups/mqb/mqbi/mqbi_storage.h#L810-L823

678098: All codebase usages are correct, for example:

https://github.com/bloomberg/blazingmq/blob/277e3c1e9faae1753b4dc645837a1ba85ee6022e/src/groups/mqb/mqbblp/mqbblp_remotequeue.cpp#L446-L451

hallfox (approved):

PR #1381, opened 2026-05-11
Fix[mqbc::StorageManager]: Cancel stale PFSM requests

kaikulimu:
## Summary                     
  - The Partition FSM had no request cancellation — when a healing session was abandoned (role change, reset, stop), in-flight requests remained outstanding and stale responses could be processed by a new session.                                                                                 
  - Adds `do_cancelRequests` to the Partition FSM, mirroring the pattern established for the Cluster FSM in PR #1365.                                                                                             
  - Tags all 5 StorageManager request send sites with a partition-specific `componentId` (`partitionFSM(partitionId)`) so cancellation is scoped to the affected partition only, without disrupting other partitions' healing.                                                               
  - Depends on the `componentId` enum-to-int refactor from PR #1378.

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_partitionstatetable.h
+++ b/src/groups/mqb/mqbc/mqbc_partitionstatetable.h
@@ -814,6 +814,37 @@ class PartitionStateTable
 // class PartitionStateTableActions
 // --------------------------------
 
+// MACROS
+
+/// Generate a composite action implementation that calls exactly the 4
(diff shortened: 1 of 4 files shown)


hallfox (commented):
Can we add some tests for the partition state table actions and storage manager action?

kaikulimu:
Instead of adding a test that do_A_B_C_D calls do_A to do_D, I defined a macro to guarantee correctness. Let me know what you think

hallfox (approved):

PR #1384, opened 2026-05-11
CI: Upload clang-format patch on failure

pniedzielski:
This patch updates our CI to upload a clang-format patch as a build artifact when the clang-format lint check fails.  This prevents contributors from manually copying and pasting the patch, which could introduce errors, or from needing exactly the same version of clang-format installed locally that we use in CI.


pniedzielski:
@kaikulimu as you requested :)

kaikulimu (approved):
LGTM

PR #1386, opened 2026-05-11
Fix: Remove attempts to test out of range enums

hallfox:
There are more test cases where we attempt to cast to an enum from an out of range value and test that messages are still process correctly. This is undefined behavior, and likely need to approach handling parsing differently if we want to test our system handles garbage values correctly. Most of our APIs do not have a reasonable expectation to work if the underlying data is invalid.

### DIFF
--- a/src/groups/bmq/bmqimp/bmqimp_messagedumper.t.cpp
+++ b/src/groups/bmq/bmqimp/bmqimp_messagedumper.t.cpp
@@ -190,11 +190,6 @@ struct Tester BSLS_CPP11_FINAL {
     // is undefined unless 'command' is a valid dump command (as indicated
     // by the return code of 'MessageDumper::parseCommand').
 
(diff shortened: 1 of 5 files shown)


src/groups/bmq/bmqimp/bmqimp_messagedumper.t.cpp
```diff
@@ -1309,9 +1285,10 @@ static void test5_reset()
         Tester      tester(bmqtst::TestHelperUtil::allocator());
         const Test& test = k_DATA[idx];
 
-        // 1. Process various DumpMessages commands and verify that dumping of
```
pniedzielski, line 1312: This and those below look like a stray changes that got swept up in the PR, or am I missing something?

pniedzielski (commented):
I am so on board with this.

OSX CI check fails because a function becomes unused, so we can delete some more code here too.

pniedzielski (approved):
🥳

PR #1392, opened 2026-05-12
Fix: Ignore `-Wunneeded-member-function`

pniedzielski:
This warning fires on `BSLMF_NESTED_TRAIT_DECLARATION` in anonymous namespaces when building with clang with modern C++ versions, but it is necessary specifically for C++03 versions.  In this case, the warning is too aggressive, so this patch turns off the warning, rather than changing the source code to do one thing in C++03 and another in later C++ versions.


pniedzielski:
This was introduced in https://github.com/bloomberg/blazingmq/pull/1389 and https://github.com/bloomberg/blazingmq/pull/1391, and we merged those with this failure intentionally to avoid blocking a release.  But now CI is 💚

hallfox (approved):

PR #1382, opened 2026-05-11
Cleanup[mqbc, mqbblp]: Normalize LSN terminology in logs and comments

kaikulimu:
## Summary                                                                                                                                 
  - Normalize all log messages and comments referring to Leader Sequence Number to consistently use the abbreviation **"LSN"** instead of the
   6+ variants previously scattered across the codebase (`seqNum`, `sequenceNumber`, `leaderMessageSeq`, `leader message sequence`,          
  `leader-msg-sequence number`, `leader sequence number`).  
  - Inconsistent terminology makes grepping production logs during incidents unreliable — a single canonical term eliminates that problem.   
  - This is the first of two PRs; a follow-up will normalize Partition Sequence Number (PSN) terminology.

PR #1393, opened 2026-05-12
Cleanup[mqbc, mqbblp]: Normalize PSN terminology in logs and comments

kaikulimu:
## Summary                                                                                                                                    
  - Normalize all "partition sequence number" / "sequence number" / "seqNum" terminology to the abbreviation **PSN** across broker logs and comments, following the same pattern established for LSN in PRs #1382 and #1387.                                                              
  - Add a `PrintPSN` streamable struct and `printPSN()` factory in `mqbs_filestoreprintutil.h` so all PSN values are printed in a consistent `[ primaryLeaseId = X sequenceNumber = Y ]` format.                                                                                              
  - Rename `ReceiveDataContext` members from `d_beginSeqNum`/`d_endSeqNum`/`d_currSeqNum` to `d_beginPSN`/`d_endPSN`/`d_currPSN`, and rename `recoverSeqNum()` to `recoverPSN()` to match the normalized terminology.

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_storageutil.cpp
+++ b/src/groups/mqb/mqbc/mqbc_storageutil.cpp
@@ -23,6 +23,7 @@
 #include <mqbi_queueengine.h>
 #include <mqbnet_cluster.h>
 #include <mqbs_datastore.h>
+#include <mqbs_filestoreprintutil.h>
 #include <mqbs_filestoreprotocol.h>
 #include <mqbs_filestoreutil.h>
 #include <mqbs_filesystemutil.h>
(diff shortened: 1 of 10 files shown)


src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -3853,7 +3853,7 @@ void StorageUtil::forceIssueAdvisoryAndSyncPt(mqbc::ClusterData*   clusterData,
         BALL_LOG_ERROR << clusterData->identity().description()
                        << "Partition [" << fs->config().partitionId()
                        << "]: failed to force-issue SyncPt, rc: " << rc
-                       << ", current partition sequence number: " << psn;
+                       << ", current PSN: " << psn;
```
pniedzielski: Should this log statement and the one above use `printPSN` now rather than constructing a fake news `bmqp_ctrlmsg`?

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -1691,15 +1684,15 @@ int FileStore::recoverMessages(QueueKeyInfoMap*     queueKeyInfoMap,
                     << partitionDesc()
                     << "Encountered a sync point during backward journal "
                        "iteration"
-                    << " with zero sequenceNum, current sequenceNum: "
```
pniedzielski, line 1694: Does it make sense to print the full PSN here?  (and maybe also in the `if` statement above, where we check the primary lease ID part?)

pniedzielski (changes requested):
`printPSN` is so nice

src/groups/mqb/mqbc/mqbc_storageutil.cpp
kaikulimu: yes

src/groups/mqb/mqbs/mqbs_filestore.cpp
kaikulimu: yes

pniedzielski (approved):

PR #1396, opened 2026-05-13
Fix[mqbs::FileStore]: Reject rollover if self not primary

kaikulimu:
Fix for Internal Ticket 184187997

  ## Summary
  - **Fix crash**: Reject the ROLLOVER admin command on replicas. Previously, running `STORAGE PARTITION <id> ROLLOVER` on a non-primary node crashed the broker with `BSLS_ASSERT_SAFE(d_isPrimary)` in `FileStore::issueSyncPoint()`, then corrupted storage causing cascading failures on restart.
  - **Refactor**: Change `StorageManager::processCommand` return type from `int` to `void` — callers already use `result->isError()` and never checked the return code.
  - **Integration tests**: Add `test_rollover_journal.py` covering primary rollover success, replica rollover rejection with cluster health verification, and journal file cleanup. Refactor shared PROLOGUE/EPILOGUE logic into `cluster_util.rollover_queues_and_apps_test()`.

### DIFF
--- a/src/groups/mqb/mqbmock/mqbmock_storagemanager.cpp
+++ b/src/groups/mqb/mqbmock/mqbmock_storagemanager.cpp
@@ -245,11 +245,11 @@ void StorageManager::applyForEachQueue(
     // NOTHING
 }
 
-int StorageManager::processCommand(
+void StorageManager::processCommand(
     BSLA_MAYBE_UNUSED mqbcmd::StorageResult* result,
     BSLA_MAYBE_UNUSED const mqbcmd::StorageCommand& command)
 {
-    return 0;
+    return;
 }
 
 void StorageManager::gcUnrecognizedDomainQueues()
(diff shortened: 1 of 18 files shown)


src/groups/mqb/mqbmock/mqbmock_storagemanager.cpp
```diff
@@ -245,11 +245,11 @@ void StorageManager::applyForEachQueue(
     // NOTHING
 }
 
-int StorageManager::processCommand(
+void StorageManager::processCommand(
     BSLA_MAYBE_UNUSED mqbcmd::StorageResult* result,
     BSLA_MAYBE_UNUSED const mqbcmd::StorageCommand& command)
 {
-    return 0;
+    return;
```
678098, line 252:
```suggestion
    // NOTHING
```

678098 (approved):

PR #1400, opened 2026-05-13
fix consumer invalidation while iterating

dorjesinpo:
when invalidating while iterating, need to advance _before_ invalidation

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
@@ -979,8 +979,7 @@ void RelayQueueEngine::applyConfiguration(App_State&        app,
     Routers::Consumers& consumers = app.routing()->d_consumers;
 
     for (Routers::Consumers::const_iterator itConsumer = consumers.begin();
-         itConsumer != consumers.end();
-         ++itConsumer) {
+         itConsumer != consumers.end();) {
         Routers::Consumer& consumer = consumers.value(itConsumer);
         mqbi::QueueHandle* handle   = itConsumer->first;
 
@@ -988,8 +987,8 @@ void RelayQueueEngine::applyConfiguration(App_State&        app,
             // This can happen with out-of-order Configure responses as in the
             // case when network disconnects with two concurrent requests and
             // the second gets (error) response before the first one gets
-            // cancelled.
-            itConsumer->second.lock()->invalidate();
+            // cancelled.  Advance before invalidate() erases the element.
+            (itConsumer++)->second.lock()->invalidate();
 
             app.routing()->finalize();
 
@@ -997,6 +996,7 @@ void RelayQueueEngine::applyConfiguration(App_State&        app,
         }
 
         consumer.registerSubscriptions(handle);
+        ++itConsumer;
     }
 
     BMQ_LOGTHROTTLE_INFO << "For queue '" << d_queueState_p->uri() << "', "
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -979,24 +979,24 @@ void RelayQueueEngine::applyConfiguration(App_State&        app,
...
         Routers::Consumer& consumer = consumers.value(itConsumer);
         mqbi::QueueHandle* handle   = itConsumer->first;
 
         if (!d_queueState_p->handleCatalog().hasHandle(handle)) {
             // This can happen with out-of-order Configure responses as in the
             // case when network disconnects with two concurrent requests and
             // the second gets (error) response before the first one gets
-            // cancelled.
-            itConsumer->second.lock()->invalidate();
+            // cancelled.  Advance before invalidate() erases the element.
```
678098: Does this mean that `consumers` collection is possibly modified during iteration? If so, I would rather refactor the code to avoid this entirely.

1. A vector of updates (in most cases, it will be empty).
```
bsl::vector<bsl::weak_ptr<mqbblp::Routers::Consumers::Item> > invalid(d_allocator_p);
```

2. Put `weak_ptr`-s in the original loop to this vector without calling `invalidate/finalize`

3. After the loop, iterate through `invalid`, lock weak pointers, call `invalidate`

4. In the end, call `app.routing()->finalize();` just once (note that it is called multiple times possibly!)

In general, I think that having a loop with advance in loop syntax itself is also more safe from coding standards: `for (int i = 0; i < smth; ++i)` is more safe than variations like `for (int i = 0; i < smth;)` where `i` is incremented **somewhere** or in multiple places

678098 (changes requested):

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -988,16 +990,22 @@ void RelayQueueEngine::applyConfiguration(App_State&        app,
...
         consumer.registerSubscriptions(handle);
     }
+    if (!invalid.empty()) {
+        for (InvalidItems::const_iterator cit = invalid.cbegin();
+             cit != invalid.cend();
+             ++cit) {
+            cit->lock()->invalidate();
+        }
+        app.routing()->finalize();
+    }
```
678098, line 1008: Looking at this again, can we just call `invalidate` in place (without even populating `invalid` and call `app.routing()->finalize();` once here after the loop?

We can have a bool flag instead:

```
if (hasInvalidItems) {
    app.routing()->finalize();
}
```

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -988,16 +990,22 @@ void RelayQueueEngine::applyConfiguration(App_State&        app,
             // This can happen with out-of-order Configure responses as in the
             // case when network disconnects with two concurrent requests and
             // the second gets (error) response before the first one gets
-            // cancelled.
-            itConsumer->second.lock()->invalidate();
-
-            app.routing()->finalize();
+            // cancelled.  Advance before invalidate() erases the element.
```
678098: Maybe we can remove the comment about advancing

dorjesinpo: amended

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
dorjesinpo: We can, but separate collection is not a bad idea, let's keep it

678098 (approved):
Looks good

PR #1401, opened 2026-05-13
Fix[test]: need to have post timeout longer than 15sec

dorjesinpo:
`test_client_timeout_reopen[multi_node-eventual_consistency]` can fail if all actions take longer than 15 secs while the next reconnect timeout is 16secs

### DIFF
--- a/src/integration-tests/test_client_timeout.py
+++ b/src/integration-tests/test_client_timeout.py
@@ -123,13 +123,15 @@ def test_client_timeout_reopen(multi_node: Cluster, domain_urls: tc.DomainUrls):
             broker.resume()
     producer_broker.wait_healthy()
 
-    # Post after short timeout should result in successful Ack
+    # Post after short timeout should result in successful Ack.
+    # Use a longer timeout to allow the producer's reconnection backoff
+    # (which grows exponentially during the suspended period) to expire.
     assert (
-        producer.post(uri_priority, ["1"], wait_ack=True, succeed=True)
+        producer.post(uri_priority, ["1"], wait_ack=True, succeed=True, timeout=30)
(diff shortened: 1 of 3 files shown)


dorjesinpo:
`test_open_queue_while_cluster_blips_quorum_leader_change[multi_node_fsm-strong_consistency]` - https://github.com/bloomberg/blazingmq/pull/1404

src/integration-tests/test_client_timeout.py
```diff
@@ -123,13 +123,15 @@ def test_client_timeout_reopen(multi_node: Cluster, domain_urls: tc.DomainUrls):
...
+    # Use a longer timeout to allow the producer's reconnection backoff
+    # (which grows exponentially during the suspended period) to expire.
     assert (
-        producer.post(uri_priority, ["1"], wait_ack=True, succeed=True)
+        producer.post(uri_priority, ["1"], wait_ack=True, succeed=True, timeout=30)
         == Client.e_SUCCESS
     )
     assert (
-        producer.post(uri_priority2, ["2"], wait_ack=True, succeed=True)
+        producer.post(uri_priority2, ["2"], wait_ack=True, succeed=True, timeout=30)
```
678098, line 134: Just to clear, is it possible to reduce reconnection timeout somewhere as an alternative to increasing the timeout here?

678098 (approved):
Have 1 comment, can merge

src/integration-tests/test_client_timeout.py
dorjesinpo: Not sure.  That would be `ReconnectingChannelFactoryConfig::d_lastRecoveredStrongConsistency` which seems like never changed

PR #1404, opened 2026-05-15
Fix[mqb]: Skip creating storage when there is no journal record

dorjesinpo:
In the case when new queue makes to CSL but not to the journal, do NOT create storage.  Create storage when queue gets open for real, to make sure QueueCreationRecord gets replicated.


dorjesinpo:
`test_strong_consistency.py::TestStrongConsistency::test_kill_post_start[multi_node-strong_consistency]` - https://github.com/bloomberg/blazingmq/pull/1351

kaikulimu (approved):

PR #1405, opened 2026-05-15
Perf[bmqu::Blob]: do not create tmp alias when not needed

678098:
We waste CPU cycles on incrementing/decrementing atomics in temporary shared pointers in `BlobUtil::appendToBlob`. This is avoidable.

This makes `bmqu::BlobUtil::appendToBlob` ~2 times faster.

# Primary

**Before**

Screenshot 2026-05-15 at 16 04 53

**After**

Screenshot 2026-05-15 at 16 43 47

# Proxy

**Before**

Screenshot 2026-05-15 at 16 02 43

**After**

Screenshot 2026-05-15 at 16 43 05

### DIFF
--- a/src/groups/bmq/bmqu/bmqu_blob.cpp
+++ b/src/groups/bmq/bmqu/bmqu_blob.cpp
@@ -328,14 +328,22 @@ int BlobUtil::appendToBlob(bdlbb::Blob*       dest,
 
     // Now we just have to append new buffers to 'dest'
     while (localSection.start().buffer() < localSection.end().buffer()) {
-        const BlobPosition&         pos    = localSection.start();
-        const bdlbb::BlobBuffer&    srcBuf = src.buffer(pos.buffer());
-        const bsl::shared_ptr<char> buf(srcBuf.buffer(),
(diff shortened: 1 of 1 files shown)


src/groups/bmq/bmqu/bmqu_blob.cpp
```diff
@@ -328,14 +328,22 @@ int BlobUtil::appendToBlob(bdlbb::Blob*       dest,
...
-                                        srcBuf.data() + pos.byte());
-        const bdlbb::BlobBuffer     newBuf(buf,
-                                       bufferSize(src, pos.buffer()) -
-                                           pos.byte());
-        dest->appendDataBuffer(newBuf);
+        const BlobPosition&      pos    = localSection.start();
+        const bdlbb::BlobBuffer& srcBuf = src.buffer(pos.buffer());
+        const int size = bufferSize(src, pos.buffer()) - pos.byte();
+
+        if (pos.byte() == 0) {
```
678098, line 335: This branch is similar to the one we already have for the last data buffer (line 360)

dorjesinpo (approved):
thank you

PR #1410, opened 2026-05-16
Refactor[mqbi]: Return bsl::string_view from DispatcherClient::description

### LINKED ISSUE #1063: Refactor: return `bsl::string_view` from `mqbi::DispatcherClient::description`

SrinathhSatuluri:
Closes #1063

Changes `mqbi::DispatcherClient::description()` and
`mqbnet::Session::description()` to return `bsl::string_view`. Both
interfaces required the change because `ClientSession` and `AdminSession`
inherit from both with a single override.

At call sites that pass the result to `const bsl::string&` parameters
(function arguments, bind captures, assignments to `bsl::string&`
accessors), the return value is explicitly converted via
`bsl::string(description())`. Streaming sites (`<<`) require no change
since `bsl::string_view` has its own `operator<<`.

The lifetime contract is documented on both interfaces.


src/groups/mqb/mqba/mqba_clientsession.cpp
```diff
@@ -1284,7 +1284,9 @@ void ClientSession::closeQueueCb(
     //       and the process of this event by the dispatcher should not add it
     //       to the flush list.
     handle->queue()->dispatcher()->execute(
-        bdlf::BindUtil::bind(&finalizeClosedHandle, description(), handle),
+        bdlf::BindUtil::bind(&finalizeClosedHandle,
+                             bsl::string(description()),
```
678098:
```suggestion
                             bsl::string(description(), d_allocator_p),
```
Should probably use the allocator

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -176,8 +176,9 @@ void Cluster::startDispatched(bsl::ostream* errorDescription, int* rc)
 
     // Start a StatMonitorSnapshotRecorder to track system stats during
     // recovery
-    mqbstat::StatMonitorSnapshotRecorder statRecorder(description() + ": ",
-                                                      d_allocator_p);
+    mqbstat::StatMonitorSnapshotRecorder statRecorder(
+        bsl::string(description()) + ": ",
```
678098:
```suggestion
        bsl::string(description(), d_allocator_p) + ": ",
```

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -3120,7 +3121,7 @@ void Cluster::loadClusterStatus(mqbcmd::ClusterResult* result)
     mqbcmd::ClusterStatus& clusterStatus = result->makeClusterStatus();
 
     clusterStatus.name()        = d_clusterData.identity().name();
-    clusterStatus.description() = description();
+    clusterStatus.description() = bsl::string(description());
```
678098:
```suggestion
    clusterStatus.description() = bsl::string(description(), d_allocator_p);
```

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
```diff
@@ -1135,7 +1135,7 @@ void ClusterProxy::loadClusterStatus(mqbcmd::ClusterResult* out)
     mqbcmd::ClusterProxyStatus& clusterProxyStatus =
         out->makeClusterProxyStatus();
 
-    clusterProxyStatus.description() = description();
+    clusterProxyStatus.description() = bsl::string(description());
```
678098:
```suggestion
    clusterProxyStatus.description() = bsl::string(description(), d_allocator_p);
```

src/groups/mqb/mqbblp/mqbblp_queuesessionmanager.cpp
```diff
@@ -519,7 +519,7 @@ QueueSessionManager::QueueSessionManager(
 
     d_requesterContext_sp->setClient(dispatcherClient)
         .setIdentity(clientIdentity)
-        .setDescription(dispatcherClient->description())
+        .setDescription(bsl::string(dispatcherClient->description()))
```
678098:
```suggestion
        .setDescription(bsl::string(dispatcherClient->description(), allocator))
```

src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
678098, line 399: Looking at the call chain in the `mqbc::StorageUtil::updateQueueStorageDispatched` call, a string object is never binded to anything, it is only used for logging.
We can safely change the argument type to `bsl::string_view` and avoid constructing a string here entirely.

src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
```diff
@@ -1128,7 +1128,7 @@ void StorageManager::resetQueue(const bmqt::Uri& uri,
         bdlf::BindUtil::bind(&mqbc::StorageUtil::resetQueueDispatched,
                              &d_storages[partitionId],
                              d_storageLockVec[partitionId].get(),
-                             fs->description(),
+                             bsl::string(fs->description()),
```
678098:
```suggestion
                             bsl::string(fs->description(), d_allocator_p),
```

src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
```diff
@@ -1306,7 +1306,7 @@ int StorageManager::start(bsl::ostream& errorDescription)
     BALL_LOG_INFO_BLOCK
     {
         printRecoveryBanner(BALL_LOG_OUTPUT_STREAM,
```
678098, line 1308: Looking at the name of the function, it is for logging only, so there are no binds. We can safely change `printRecoveryBanner` to accept `bsl::string_view` arg and avoid constructing an intermediate string here

src/groups/mqb/mqbc/mqbc_clusternodesession.cpp
```diff
@@ -59,7 +59,7 @@ ClusterNodeSession::ClusterNodeSession(
 
     d_queueHandleRequesterContext_sp->setClient(this)
         .setIdentity(identity)
-        .setDescription(description())
+        .setDescription(bsl::string(description()))
```
678098:
```suggestion
        .setDescription(bsl::string(description(), allocator))
```

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -160,7 +160,7 @@ void StorageManager::queueCreationCb(int                     partitionId,
             &d_storages[partitionId],
             d_storageLockVec[partitionId].get(),
             d_domainFactory_p,
-            fs->description(),
+            bsl::string(fs->description()),
```
678098: `updateQueueStorageDispatched` doesn't seem to be binded anywhere, or having any bind in its implementation, so the arg type can possibly be changed to `bsl::string_view` to avoid constructing an intermediate string

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -4171,7 +4171,7 @@ void StorageManager::resetQueue(const bmqt::Uri& uri,
         bdlf::BindUtil::bind(&StorageUtil::resetQueueDispatched,
                              &d_storages[partitionId],
                              d_storageLockVec[partitionId].get(),
-                             fs->description(),
+                             bsl::string(fs->description()),
```
678098:
```suggestion
                             bsl::string(fs->description(), d_allocator_p),
```

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -347,7 +347,7 @@ int StorageUtil::updateQueuePrimaryRaw(mqbs::ReplicatedStorage* storage,
 
         rc = addVirtualStoragesInternal(storage,
                                         addedIdKeyPairs,
-                                        fs->description(),
+                                        bsl::string(fs->description()),
```
678098: `addVirtualStoragesInternal` uses description only for logging, and it is not binded - safe to change the arg type to `bsl::string_view`

src/groups/mqb/mqbnet/mqbnet_multirequestmanager.h
```diff
@@ -542,12 +542,12 @@ MultiRequestManager<REQUEST, RESPONSE, TARGET>::targetDescription(
 }
 
 template <class REQUEST, class RESPONSE, class TARGET>
-inline const bsl::string&
+inline bsl::string
```
678098:
```suggestion
inline bsl::string_view
```
Typically, it is not good to have getters that construct an intermediate value (if this is a memory-heavy object, like string), because it harms performance.

For this function, I think it is safe to change the return type to `bsl::string_view` and avoid constructing and returning an intermediate string.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -677,11 +677,11 @@ void TCPSessionFactory::initialConnectionComplete(
         // schedule channelStateCb/poolSessionStateCb/onClose/tearDown
         channel->close();
 
-        logOpenSessionTime(session->description(), channel);
+        logOpenSessionTime(bsl::string(session->description()), channel);
```
678098: The arg type for `logOpenSessionTime` might be changed to `bsl::string_view` to avoid constructing an intermediate string

src/groups/mqb/mqbnet/mqbnet_transportmanager.cpp
```diff
@@ -171,7 +171,7 @@ bool TransportManager::processSession(
         // Handle incoming proxy connection
         cluster->onProxyConnectionUp(session->channel(),
                                      peerIdentity,
-                                     session->description());
+                                     bsl::string(session->description()));
```
678098: I think `onProxyConnectionUp` interface might be changed to accept `bsl::string_view`.
After this, in `mqbblp::Cluster::onProxyConnectionUp` there is a `bind`, and a string might be constructed there, using the `mqbblp::Cluster` allocator

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -2939,7 +2939,7 @@ StorageUtil::createQueueStorageImpl(mqbs::FileStore*        fs,
 
     if (0 != addVirtualStoragesInternal(rs_sp.get(),
                                         appIdKeyPairs,
-                                        fs->description(),
+                                        bsl::string(fs->description()),
```
678098: The arg type might be changed to `bsl::string_view`

src/groups/mqb/mqbc/mqbc_storageutil.cpp
```diff
@@ -2846,7 +2846,7 @@ void StorageUtil::createQueueStorageAsReplica(
                                  bdlf::PlaceHolders::_2,  // domain*
                                  &domain,
                                  &latch,
-                                 fs->description(),
+                                 bsl::string(fs->description()),
```
678098: `onDomain` uses the provided description only for logging an error, so it is a rare code path. It's a pity we have to construct a string that we might not even use. This is not the PR's problem, this pre-existed in main (bind of `const bsl::string&` still copies a string).

What about changing the signature of `mqbc::StorageManager` to accept `const FileStore*` instead?

```
    static void onDomain(const bmqp_ctrlmsg::Status& status,
                         mqbi::Domain*               domain,
                         mqbi::Domain**              out,
                         bslmt::Latch*               latch,
                         const bsl::string&          description,
                         const bsl::string&          domainName);
```

turns into

```
    static void onDomain(const bmqp_ctrlmsg::Status& status,
                         mqbi::Domain*               domain,
                         mqbi::Domain**              out,
                         bslmt::Latch*               latch,
                         const FileStore*            location_p,
                         const bsl::string&          domainName);
```

And we can call `location_p->description()` for logging, if there is an error.

678098 (commented):
Have some comments:
1. Possible to provide allocators to construct strings
2. Possible to change arg/return types somewhere to `bsl::string_view` to avoid constructing a string

src/groups/mqb/mqbc/mqbc_storageutil.cpp
SrinathhSatuluri: I like the idea of deferring the description() call to the error path, but I want to verify that the FileStore outlives the bound callback before passing a raw pointer into the binder. I was thinking of adding the allocator here for now. Would you prefer the redesign in this PR as a separate issue?

678098: Looking at the call chain, it seems that everything happens in the same thread. We don't sent the binded callback to another thread that might outlive this `FileStore`'s own thread and `FileStore` object.
This can be done in the same PR

SrinathhSatuluri: Thank you for the comments! I see now that string_view is borrowed and bsl::string is owned, so for that reason the conversion should occur only where ownership is truly required and not reflexively at the point of access.  

Also while making fixes I noticed RequestManager::sendRequest takes const bsl::string& but copies it into a member. Its parameter could be bsl::string_view, pushing the ownership boundary to where the copy actually occurs. Is this worth looking at with a separate issue, as it is not in this PR's scope?

678098: If this is not a regression from the existing state in `main`, it can be done separately, any variant is good here

SrinathhSatuluri: Pushed the updates: allocator threading at the flagged call sites, string_view parameter changes, onDomain redesign, and MultiRequestManager return type. Ready for another look when you have time

678098 (approved):
Looks good

678098:

PR #1411, opened 2026-05-18
Fix[tweakgen]: Make the script work

pniedzielski:
This pull request contains a number of changes that enable this script to even be run so we can regenerate `blazingmq/dev/it/tweaks/generated.py`.

### DIFF
--- a/src/python/bin/schemagen
+++ b/src/python/bin/schemagen
@@ -23,7 +23,7 @@ ERROR
 fi
 
 copyright=$(cat <<COPYRIGHT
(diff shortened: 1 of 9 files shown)


pniedzielski:
This is particularly rough.  The new xsdata version uses Python 3.10+ features, and we don't have a requirements.txt to pin a particular version (apparently intentionally??).  We need Python 3.8 support for the moment, and we run with Python 3.8 in CI.

I'll put this work on hold while we figure that out.

PR #1412, opened 2026-05-18
Refactor[mbqdomaincfg]: Rename mqbconfm to mqbdomaincfg

pniedzielski:
This pull request performs a mechanical rename of all instances of mqbconf/mqbconfm in the repo to mqbdomaincfg.  This name is much clearer and less likely to be confused with mqbcfg, the broker configuration.

This patch also regenerates the code Python schemas files, the Python tweak files, and the C++ messages files using the new name.

The only non-mechanical change in this patch is updating the new mqbdomaincfg package documentation, which required slightly more intensive changes to be useful.  Still, this patch only performs those textual changes; it does not change the format to Doxygen like we ultimately want to.


pniedzielski:
For the moment, this PR also includes the changes in #1411.  They should be merged first.

PR #1413, opened 2026-05-18
Fix[mqbc::StorageMgr]: Replica check irreconcilable primary data before send

kaikulimu:
Follow-up to #1163 

 ## Summary                                                                                                                                 
  - In `do_sendDataToPrimary`, the replica now checks the primary's PSN against its own `highestSeqNums` before sending data chunks.  If the primary's PSN is irreconcilable (extra records under a lease ID that exceeds what the replica has ever seen), the replica replies with a failure `ReplicaDataResponsePull` carrying a new `e_IRRECONCILABLE_DATA` error code instead of blindly sending data.                       
  - On the primary side, upon receiving this irreconcilable failure response, the primary drops its own storage, resets its PSN to (0,0), and re-sends `ReplicaDataRequestPull` to pull the full partition data from the replica.                                                       
  - This complements the existing primary-side irreconcilable detection in `determineDataDestinations` (Push/Drop path) by adding the symmetric replica-side detection in the Pull path.


src/integration-tests/test_fsm_partition_sync.py
```diff
@@ -628,3 +633,150 @@ def test_sync_after_replicas_missed_or_extra_records(
...
+    storage and re-pulls from (0,0).
+    - Start cluster
+    - Put 2 messages
+    - Stop Replica 1, 2, and 3
+    - Put 2 more messages
+    - Stop the original primary
+    - Start all nodes except the original primary.  Wait for the second primary
+      to heal the cluster
+    - Put 2 more messages
+    - Simulate: original primary joins the cluster and syncs CSL, but crashes
```
dorjesinpo, line 654: why do we need the original primary crash?  Does it not already has irreconcilable messages (3 and 4)?

dorjesinpo (commented):
It is a bit strange that Replica decides that Primary needs to remove its storage.  Having Primary making that kind decisions seems more natural.  
Question, `highestSeqNums` on Replica and Primary, do they always agree?  And if not, should it be the Primary version to win?

kaikulimu:
> It is a bit strange that Replica decides that Primary needs to remove its storage. Having Primary making that kind decisions seems more natural. Question, `highestSeqNums` on Replica and Primary, do they always agree? And if not, should it be the Primary version to win?

Let me explain the logic:
One important fact is that primary is not necessarily the node with the most up-to-date sequence number. Primary gathers a quorum of sequence numbers and decides that a replica has the highest, it requests replica to send data chunks. Say primary has data (1,1) to (1,15), while replica has (1,1) to (1,10) and (2,1) to (2,10). Primary's (1,11) to (1,15) are considered irreconcilable data and should be dropped. Previously, replica had no way to tell and would blindly send (2,1) to (2,10). The `highestSeqNums` map on primary would be `{1: 15}`; on replica would be `{1: 10, 2: 10}` -- so they do not always agree. And replica version will win. With the highestSeqNums check, replica would be able to tell primary to about the irreconcilable data, while postponing data chunks.

You can think of it this way: Primary decides to trust highest seqNum replica --> Replica detects irreconcilable data and informs primary --> Primary decides to drop data and request from replica again.

src/integration-tests/test_fsm_partition_sync.py
kaikulimu: If we don't do this step, old primary will have electorTerm 1, while other nodes will have electorTerm 2. Then old primary can never win election to become new primary. We want to let old primary join in with electorTerm 2's leader, **but** we want to keep Messages 3 and 4 irreconcilable. Effectively, we want old primary to sync CSL to electorTerm 2, but do not sync partition files.

dorjesinpo (approved):

PR #1417, opened 2026-05-19
Misc[mqbs::FileStore]: Check rc when closing

kaikulimu:


src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -3460,13 +3457,11 @@ int FileStore::close(FileSet& fileSetRef, bool flush)
...
                 << BMQTSK_ALARMLOG_END;
-            rcFinal = (rcFinal == rc_FLUSH_FAILURE)
-                          ? rc_FLUSH_AND_CLOSE_FAILURES
-                          : rc_CLOSE_FAILURE;
+            return rc * 10 + rc_QLIST_CLOSE_FAILURE;  // RETURN
         }
     }
 
-    return rcFinal;
+    return rc_SUCCESS;
```
678098, line 3464: Should we also return non zero rc here if flush failed, but files were closed?

678098 (approved):
One comment, otherwise good

src/groups/mqb/mqbs/mqbs_filestore.cpp
kaikulimu: No, not harmful to simply alarm and still close with zero rc.

PR #1418, opened 2026-05-19
Fix[bmqp]: Specialize `getPropertyAsStringOr` for `string_view`

pniedzielski:
Commit 87ddacbe40 introduced a performance optimization for message properties where string properties may be deserialized as `bsl::string_view` rather than `bsl::string` when the property data is contiguous in memory.  However, this commit only modified `getPropertyAsString()` to use the new helper to convert `string_view` back to `string` when checking the property’s data type.  It did not modify the sister function `getPropertyAsStringOr()`.

This patch performs the same change to `getPropertyAsStringOr()` as we made in `getPropertyAsString()`.  The function bodies now mirror each other.  In addition, we add test cases that can reproduce such a failure in both functions without the fix.’

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_messageproperties.h
+++ b/src/groups/bmq/bmqp/bmqp_messageproperties.h
@@ -1056,7 +1056,16 @@ inline const bsl::string&
 MessageProperties::getPropertyAsStringOr(bsl::string_view   name,
                                          const bsl::string& value) const
 {
-    return getPropertyOr(name, value);
+    PropertyMapIter it = findProperty(name);
+    if (it == d_properties.end()) {
+        return value;
+    }
+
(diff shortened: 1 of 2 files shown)


pniedzielski:
IT failures

`test_cluster_node_shutdown.py::TestClusterNodeShutdown::test_open_queue_while_cluster_blips_quorum_leader_change[multi_node_fsm-strong_consistency]`

`test_restart_between_modes.py::test_restart_between_Legacy_and_FSM_basic[single_node_fsm-eventual_consistency-to_fsm_quorum_1_then_start_all/from_legacy]`

appear to be unrelated to this change---nothing to do with string message properties.

dorjesinpo (approved):
thank you

PR #1426, opened 2026-05-20
Fix[bmqc::MQTP]: wait each thread to set its `threadId`

678098:
- Fix race condition in `bmqc::MultiQueueThreadPool` where `queueThreadId()` could return 0 if the worker thread hadn't started yet.
- Refactor `QueueInfo` to use `shared_ptr` ownership, removing copy semantics and `reset()`
- Add `bslmt::Latch` to `ProcessQueueJob` so `start()` blocks until all threads have set their `d_threadId`

Fixes the following crash in IT:
```
  #6  bmqAssertHandler (comment="!inDispatcherThread()",
       file="mqbblp_cluster.cpp", line=2207)                                           
  #7  bsls::Assert::invokeHandler(bsls::AssertViolation const&)                        
  #8  mqbblp::Cluster::initiateShutdown (this=0x555e8e51ef90, callback=...)            
       at mqbblp_cluster.cpp:2207                                                      
  #9  mqba::Application::stop (this=0x7ffe3ed7fa00)                                    
       at mqba_application.cpp:500 
```

The crash happened because the app continues constructing objects such as clusters without waiting for MQTP thread to assign `threadId` first. In this case, cluster dispatcher client doesn't have the correct `threadId`: it has `ANY` thread id, making all positive `inDispatcherThread()` pass, and making one negative `!inDispatcherThread()` not pass.

The test:
`test_appids.py::test_old_data_new_app[single_node-strong_consistency]`

Failed with:
`blazingmq.dev.it.process.proc.ProcessExitError: single: process exited with return   
  code -6 (Aborted)`

### DIFF
--- a/src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
+++ b/src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
@@ -166,6 +166,7 @@
 #include <bslma_default.h>
 #include <bslma_usesbslmaallocator.h>
 #include <bslmf_nestedtraitdeclaration.h>
+#include <bslmt_latch.h>
 #include <bslmt_threadutil.h>
 #include <bslmt_timedsemaphore.h>
 #include <bsls_assert.h>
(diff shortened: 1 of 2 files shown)


dorjesinpo:
The IT failure is interesting case - https://github.com/bloomberg/blazingmq/pull/1464

dorjesinpo (approved):
thank you

PR #1180, opened 2026-03-06
Feat[MQB]: Json printer can print clients and clusters stat context data 

bbpetukhov:
**Describe your changes**
- Stat printer can output stat contexts snapshot as ascii table (as before) and also in JSON format to stat log file. TABLE/JSON/TABLE_AND_JSON options can be selected in broker config file.
- JSON stat printer outputs stats also for clients, channels and allocators stat contexts. JSON structure changed to be list of flat objects.
- Response of JSON admin command STAT SHOW changed accordingly. 

**Testing performed**
- Updated admin command (STAT SHOW) integration tests

PR #1430, opened 2026-05-20
Feat[mqbcfg]: support stats encoding setting

678098:
- `mqbcfg`: add (but do not use yet) settings for extended stats printing.
- default to table print **and** json print.
- regenerate schema sources (c++, python).

https://github.com/bloomberg/blazingmq/pull/1180

### DIFF
--- a/src/python/blazingmq/schemas/mqbcfg.py
+++ b/src/python/blazingmq/schemas/mqbcfg.py
@@ -894,52 +894,17 @@ class ResolvedDomain:
     )
 
 
-@dataclass
-class StatsPrinterConfig:
-    print_interval: int = field(
-        default=60,
-        metadata={
-            "name": "printInterval",
-            "type": "Element",
-            "namespace": "http://bloomberg.com/schemas/mqbcfg",
-            "required": True,
-        },
-    )
-    file: Optional[str] = field(
-        default=None,
-        metadata={
-            "type": "Element",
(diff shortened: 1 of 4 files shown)


src/python/blazingmq/schemas/mqbcfg.py
```diff
@@ -2025,6 +2096,14 @@ class AuthenticatorConfig:
...
         },
     )
+    credential_provider: Optional[CredentialProviderConfig] = field(
+        default=None,
+        metadata={
+            "name": "credentialProvider",
+            "type": "Element",
+            "namespace": "http://bloomberg.com/schemas/mqbcfg",
+        },
+    )
```
678098, line 2106: This was also autogenerated

src/python/blazingmq/schemas/mqbcfg.py
```diff
@@ -1628,6 +1593,62 @@ class StatPluginConfigPrometheus:
     )
 
 
+@dataclass
+class StatsPrinterConfig:
```
678098, line 1597: For some reason, xsd generator shuffled some sources which resulted in a huge diff

src/python/blazingmq/schemas/mqbcfg.py
chrisbeard: That's fine, this was just merged earlier today: https://github.com/bloomberg/blazingmq/pull/1372

chrisbeard (approved):

PR #1434, opened 2026-05-20
Feat[mqbauthn]: integrate CredentialProvider with AuthenticationController

chrisbeard:
This PR contains the first subset of changes in https://github.com/bloomberg/blazingmq/pull/1428 for easier reviewing.

- Move credential provider management into the `AuthenticationController`, which already handles authenticator plugin factory logic.
- Add `BasicCredentialProvider`, corresponding to `BasicAuthenticator`, and register as a built-in credential provider.

### DIFF
--- a/src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
+++ b/src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
@@ -26,10 +26,13 @@
 #include <mqbcfg_brokerconfig.h>
 #include <mqbcfg_messages.h>
 #include <mqbplug_authenticator.h>
+#include <mqbplug_credentialprovider.h>
 #include <mqbplug_pluginfactory.h>
+#include <mqbplug_plugintype.h>
 
 // BDE
 #include <ball_log.h>
+#include <bdlb_nullablevalue.h>
 #include <bdlb_string.h>
 #include <bsl_ostream.h>
 #include <bsl_string.h>
@@ -37,6 +40,7 @@
 #include <bsl_unordered_map.h>
 #include <bsl_unordered_set.h>
 #include <bsl_vector.h>
+#include <bslmf_movableref.h>
 
 namespace BloombergLP {
 namespace mqbauthn {
@@ -123,7 +127,9 @@ int AuthenticationController::initializeAuthenticators(
 
     // Step 1: Collect all available plugin factories (built-in + external)
     PluginFactories pluginFactories(d_allocator_p);
-    if (int rc = collectAvailablePluginFactories(&pluginFactories)) {
+    if (int rc = collectAvailablePluginFactories(
+            mqbplug::PluginType::e_AUTHENTICATOR,
+            &pluginFactories)) {
         return rc * 10 + rc_COLLECT_FACTORIES_FAILED;  // RETURN
     }
 
@@ -164,17 +170,17 @@ int AuthenticationController::initializeAuthenticators(
 }
 
 int AuthenticationController::collectAvailablePluginFactories(
-    PluginFactories* pluginFactories)
+    mqbplug::PluginType::Enum type,
+    PluginFactories*          pluginFactories)
 {
     // Collect external plugin factories loaded by PluginManager
-    d_pluginManager_p->get(mqbplug::PluginType::e_AUTHENTICATOR,
-                           pluginFactories);
+    d_pluginManager_p->get(type, pluginFactories);
 
     // Add built-in plugin factories
     bsl::vector<mqbplug::PluginInfo>::const_iterator it =
         d_builtInPluginLibrary->plugins().cbegin();
     for (; it != d_builtInPluginLibrary->plugins().cend(); ++it) {
-        if (it->type() == mqbplug::PluginType::e_AUTHENTICATOR) {
+        if (it->type() == type) {
             pluginFactories->insert(it->factory().get());
         }
     }
@@ -381,6 +387,70 @@ int AuthenticationController::ensureDefaultAuthenticator(
     return 0;
 }
 
+int AuthenticationController::initializeCredentialProvider(
+    bsl::ostream& errorDescription)
+{
+    enum RcEnum {
+        rc_SUCCESS          = 0,
+        rc_COLLECT_FAILED   = -1,
+        rc_PLUGIN_NOT_FOUND = -2,
+        rc_START_FAILED     = -3
+    };
+
+    const bdlb::NullableValue<mqbcfg::CredentialProviderConfig>& credProvCfg =
+        mqbcfg::BrokerConfig::get().authentication().credentialProvider();
+
+    if (credProvCfg.isNull()) {
+        return rc_SUCCESS;  // RETURN
+    }
+
+    const bsl::string& providerName = credProvCfg.value().name();
+
+    BALL_LOG_INFO << "Configuring credential provider '" << providerName
+                  << "'";
+
+    PluginFactories factories(d_allocator_p);
+    if (int rc = collectAvailablePluginFactories(
+            mqbplug::PluginType::e_CREDENTIAL_PROVIDER,
+            &factories)) {
+        return rc * 10 + rc_COLLECT_FAILED;  // RETURN
+    }
+
+    for (PluginFactories::const_iterator it = factories.cbegin();
+         it != factories.cend();
+         ++it) {
+        mqbplug::CredentialProviderPluginFactory* candidate =
+            dynamic_cast<mqbplug::CredentialProviderPluginFactory*>(*it);
+        if (!candidate) {
+            continue;
+        }
+        bslma::ManagedPtr<mqbplug::CredentialProvider> provider =
+            candidate->create(d_allocator_p);
+        if (provider) {
+            d_credentialProvider_mp = bslmf::MovableRefUtil::move(provider);
+            break;
+        }
+    }
+
+    if (!d_credentialProvider_mp) {
+        errorDescription << "CredentialProvider plugin '" << providerName
+                         << "' not found";
+        return rc_PLUGIN_NOT_FOUND;  // RETURN
+    }
+
+    int rc = d_credentialProvider_mp->start(errorDescription);
+    if (rc != 0) {
+        return rc * 10 + rc_START_FAILED;  // RETURN
+    }
+
+    d_credentialFunc = d_credentialProvider_mp->load();
+
+    BALL_LOG_INFO << "Credential provider '" << providerName << "' started";
+
+    return rc_SUCCESS;
+}
+
+// CREATORS
 AuthenticationController::AuthenticationController(
     mqbplug::PluginManager* pluginManager,
     bslma::Allocator*       allocator)
@@ -388,6 +458,8 @@ AuthenticationController::AuthenticationController(
       bslma::ManagedPtrUtil::allocateManaged<mqbauthn::PluginLibrary>(
           allocator))
 , d_pluginManager_p(pluginManager)
+, d_credentialProvider_mp()
+, d_credentialFunc()
 , d_isStarted(false)
 , d_allocator_p(allocator)
 {
@@ -397,10 +469,11 @@ int AuthenticationController::start(bsl::ostream& errorDescription)
 {
     enum RcEnum {
         // Enum for the various RC error categories
-        rc_SUCCESS             = 0,
-        rc_ALREADY_STARTED     = -1,
-        rc_INVALID_CONFIG      = -2,
-        rc_INIT_AUTHENTICATORS = -3
+        rc_SUCCESS                  = 0,
+        rc_ALREADY_STARTED          = -1,
+        rc_INVALID_CONFIG           = -2,
+        rc_INIT_AUTHENTICATORS      = -3,
+        rc_INIT_CREDENTIAL_PROVIDER = -4
     };
 
     if (d_isStarted) {
@@ -420,7 +493,10 @@ int AuthenticationController::start(bsl::ostream& errorDescription)
         return rc * 10 + rc_INIT_AUTHENTICATORS;
     }
 
-    // Initialize Authenticators from plugins
+    rc = initializeCredentialProvider(errorDescription);
+    if (rc != 0) {
+        return rc * 10 + rc_INIT_CREDENTIAL_PROVIDER;
+    }
 
     d_isStarted = true;
     return rc_SUCCESS;
@@ -435,6 +511,13 @@ void AuthenticationController::stop()
     BALL_LOG_INFO << "Stopping AuthenticationController";
 
     d_isStarted = false;
+
+    // Stop credential provider
+    if (d_credentialProvider_mp) {
+        d_credentialProvider_mp->stop();
+        d_credentialProvider_mp.reset();
+        d_credentialFunc = mqbplug::CredentialProvider::CredentialFunc();
+    }
 
     // Stop all authenticators
     for (AuthenticatorMap::iterator it = d_authenticators.begin();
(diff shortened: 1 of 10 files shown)


src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -164,17 +170,17 @@ int AuthenticationController::initializeAuthenticators(
 }
 
 int AuthenticationController::collectAvailablePluginFactories(
-    PluginFactories* pluginFactories)
+    mqbplug::PluginType::Enum type,
+    PluginFactories*          pluginFactories)
```
678098, line 174:
```suggestion
    PluginFactories*          pluginFactories_p,
    mqbplug::PluginType::Enum type)
```
`pluginFactories` is an output arg so it should be earlier

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -164,17 +170,17 @@ int AuthenticationController::initializeAuthenticators(
...
 {
     // Collect external plugin factories loaded by PluginManager
```
678098, line 179:
```suggestion
    // PRECONDITIONS
    BSLS_ASSERT_SAFE(pluginFactories_p);

    // Collect external plugin factories loaded by PluginManager
```

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
678098: On the other hand, `d_pluginManager_p->get(type, pluginFactories);` already implements this non-BDE order. Possible to change both places

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -381,13 +387,79 @@ int AuthenticationController::ensureDefaultAuthenticator(
...
+        return rc * 10 + rc_COLLECT_FAILED;  // RETURN
+    }
+
+    for (PluginFactories::const_iterator it = factories.cbegin();
+         it != factories.cend();
+         ++it) {
+        mqbplug::CredentialProviderPluginFactory* candidate =
+            dynamic_cast<mqbplug::CredentialProviderPluginFactory*>(*it);
+        if (!candidate) {
+            continue;
```
678098: We should probably log an error, it is not okay to have a factory that is retrieved from `mqbplug::PluginType::e_CREDENTIAL_PROVIDER` but is not castable to `mqbplug::CredentialProviderPluginFactory`.

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.h
```diff
@@ -84,6 +85,14 @@ class AuthenticationController {
     /// Held, not owned.
     mqbplug::PluginManager* d_pluginManager_p;
 
+    /// Credential provider for outbound authentication.  Null if not
+    /// configured.
+    bslma::ManagedPtr<mqbplug::CredentialProvider> d_credentialProvider_mp;
+
+    /// Credential function loaded from the provider.  Empty if no provider
+    /// is configured.
+    mqbplug::CredentialProvider::CredentialFunc d_credentialFunc;
```
678098: `mqbplug::CredentialProvider::CredentialFunc` implements a concrete implementation of a possible credential callback function interface, that is `bsl::function<bsl::optional<mqbplug::AuthnCredential>(bsl::ostream& error)>`. We don't have to expose `mqbplug::CredentialProvider::CredentialFunc` in the header as we might switch to lambda or something else in the future, and we might just assign `mqbplug::CredentialProvider::CredentialFunc` to a `bsl::function` in cpp file.

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.h
```diff
@@ -84,6 +85,14 @@ class AuthenticationController {
     /// Held, not owned.
     mqbplug::PluginManager* d_pluginManager_p;
 
+    /// Credential provider for outbound authentication.  Null if not
+    /// configured.
+    bslma::ManagedPtr<mqbplug::CredentialProvider> d_credentialProvider_mp;
```
678098, line 90: This is a pointer type, together with `bsl::function` change below it is possible to have a forward declaration to `mqbplug::CredentialProvider` and remove `#include <mqbplug_credentialprovider.h>` from the header.

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.h
678098: Also, we might rename the field `d_credentialFunc` -> `d_credentialCb`, to make it closer to `authenticationCb`, and to hide the concrete implementation detail that we use functor for this.

src/groups/mqb/mqbauthn/mqbauthn_basiccredentialprovider.h
```diff
@@ -0,0 +1,164 @@
...
+        bsl::string d_password;
+
+      public:
+        // CREATORS
+
+        /// Create a `CredentialFunctor` with the specified `mechanism`,
+        /// `username`, and `password`.
+        CredentialFunctor(const bsl::string& mechanism,
+                          const bsl::string& username,
+                          const bsl::string& password);
```
678098: Possible to init from string views to avoid constructing intermediate strings. `CredentialFunctor(k_MECHANISM, d_username, d_password);` creates a string for `k_MECHANISM`.
```suggestion
        explicit CredentialFunctor(bsl::string_view  mechanism,
                                   bsl::string_view  username,
                                   bsl::string_view  password
                                   bslma::Allocator* allocator = 0);
```

```
    BSLMF_NESTED_TRAIT_DECLARATION(CredentialFunctor,
                                   bslma::UsesBslmaAllocator)
```

678098: This also allows to use allocator in `BasicCredentialProvider::CredentialFunctor::operator()`

src/groups/mqb/mqbauthn/mqbauthn_basiccredentialprovider.cpp
```diff
@@ -0,0 +1,185 @@
...
+}
+
+BasicCredentialProvider::~BasicCredentialProvider()
+{
+    BSLS_ASSERT_OPT(!d_isStarted &&
+                    "stop() must be called before destroying this object");
+}
+
+// MANIPULATORS
+mqbplug::CredentialProvider::CredentialFunc BasicCredentialProvider::load()
```
678098: Can also return `bsl::function` here

src/groups/mqb/mqbauthn/mqbauthn_basiccredentialprovider.cpp
```diff
@@ -0,0 +1,185 @@
...
+    // NOTHING
+}
+
+bslma::ManagedPtr<mqbplug::CredentialProvider>
+BasicCredentialProviderPluginFactory::create(bslma::Allocator* allocator)
+{
+    const bdlb::NullableValue<mqbcfg::CredentialProviderConfig>& providerCfg =
+        mqbcfg::BrokerConfig::get().authentication().credentialProvider();
+
+    if (providerCfg.isNull()) {
```
678098, line 170: This check is similar to the one happening in `mqbauthn_authenticationcontroller.cpp`:

```
    const bdlb::NullableValue<mqbcfg::CredentialProviderConfig>& credProvCfg =
        mqbcfg::BrokerConfig::get().authentication().credentialProvider();

    if (credProvCfg.isNull()) {
        return rc_SUCCESS;  // RETURN
    }
```

The question is, should `AuthenticationController` be responsible for checking it?

src/groups/mqb/mqbauthn/mqbauthn_basiccredentialprovider.cpp
```diff
@@ -0,0 +1,185 @@
...
+        return bslma::ManagedPtr<mqbplug::CredentialProvider>();  // RETURN
+    }
+
```
678098, line 173: Possible to make a const ref to `providerCfg.value()` here to simplify next checks

src/groups/mqb/mqbauthn/mqbauthn_basiccredentialprovider.cpp
```diff
@@ -0,0 +1,185 @@
...
+        }
+        else if (it->key() == "password") {
+            password = it->value().stringVal();
+        }
+    }
+
+    return bslma::ManagedPtr<mqbplug::CredentialProvider>(
+        new (*allocator)
+            BasicCredentialProvider(username, password, allocator),
+        allocator);
```
678098:
```suggestion
    return bslma::ManagedPtrUtil::allocateManaged<mqbauthn::BasicCredentialProvider>(allocator, username, password);
```

This allows to remove `allocator = bslma::Default::allocator(allocator);` line before, since every `allocator` use here is checked against 0 in class implementations

src/groups/mqb/mqbauthn/mqbauthn_basiccredentialprovider.h
```diff
@@ -0,0 +1,164 @@
...
+                            bslma::Allocator* allocator = 0);
+
+    /// Destructor.
+    ~BasicCredentialProvider() BSLS_KEYWORD_OVERRIDE;
+
+    // MANIPULATORS
+
+    /// Return a credential-providing function.  The behavior is undefined
+    /// unless `start()` has been called successfully.
+    CredentialFunc load() BSLS_KEYWORD_OVERRIDE;
```
678098: Changing this to `bsl::function` allows (possibly) to declare/define functor in cpp file, removing it from header. Not sure if this will compile on C++03/Solaris though

src/groups/mqb/mqbauthn/mqbauthn_basiccredentialprovider.cpp
```diff
@@ -0,0 +1,185 @@
...
+namespace mqbauthn {
+
+const char* BasicCredentialProvider::k_NAME      = "BasicCredentialProvider";
+const char* BasicCredentialProvider::k_MECHANISM = "BASIC";
+
+// ------------------------
+// class CredentialFunctor
+// ------------------------
+
+BasicCredentialProvider::CredentialFunctor::CredentialFunctor(
```
678098: If `CredentialFunctor` is not a part of public API (removed from the header, or moved to `private` and returned only as `bsl::function` to external users, we can remove `mechanism` passing fully and use `k_MECHANISM` that never changes.

src/groups/mqb/mqbauthn/mqbauthn_basiccredentialprovider.cpp
```diff
@@ -0,0 +1,185 @@
...
+#include <bslma_allocator.h>
+#include <bslma_default.h>
+#include <bslma_managedptr.h>
+#include <bsls_assert.h>
+
+namespace BloombergLP {
+namespace mqbauthn {
+
+const char* BasicCredentialProvider::k_NAME      = "BasicCredentialProvider";
+const char* BasicCredentialProvider::k_MECHANISM = "BASIC";
```
678098:
```suggestion
const bsl::string_view BasicCredentialProvider::k_NAME      = "BasicCredentialProvider";
const bsl::string_view BasicCredentialProvider::k_MECHANISM = "BASIC";
```
Using `bsl::string_view` gives the bonus of having its length instantly instead of having to call `strlen` in string operations

src/groups/mqb/mqbauthn/mqbauthn_basiccredentialprovider.h
```diff
@@ -0,0 +1,164 @@
...
+// =============================
+// class BasicCredentialProvider
+// =============================
+
+class BasicCredentialProvider : public mqbplug::CredentialProvider {
+  public:
+    // PUBLIC CLASS DATA
+    static const char* k_NAME;
+    static const char* k_MECHANISM;
+
```
678098:
```suggestion
```
Do we need to have these statics in the header? Possible to keep it in cpp file in unnamed namespace.

If we need to know plugin names and mechanisms, we should implement virtual getters in the base class and override them in all implementations. Having these statics declared in each class separately doesn't give a generic way of accessing them for any random plugin.

src/groups/mqb/mqbauthn/mqbauthn_basiccredentialprovider.h
```diff
@@ -0,0 +1,164 @@
...
+    // TRAITS
+    BSLMF_NESTED_TRAIT_DECLARATION(BasicCredentialProvider,
+                                   bslma::UsesBslmaAllocator)
+
+    // CREATORS
+
+    /// Create a `BasicCredentialProvider` with the specified `username`
+    /// and `password`, using the optionally specified `allocator` to supply
+    /// memory.
+    BasicCredentialProvider(bsl::string_view  username,
```
678098:
```suggestion
    explicit BasicCredentialProvider(bsl::string_view  username,
```

src/groups/mqb/mqbauthn/mqbauthn_basiccredentialprovider.h
```diff
@@ -0,0 +1,164 @@
...
+    /// description of any failure encountered.
+    int start(bsl::ostream& errorDescription) BSLS_KEYWORD_OVERRIDE;
+
+    /// Stop the provider.
+    void stop() BSLS_KEYWORD_OVERRIDE;
+};
+
+// ========================================
+// class BasicCredentialProviderPluginFactory
+// ========================================
```
678098, line 117:
```suggestion
// ==========================================
// class BasicCredentialProviderPluginFactory
// ==========================================
```

678098 (commented):
Have some comments

src/groups/mqb/mqbauthn/mqbauthn_basiccredentialprovider.h
chrisbeard: Only `k_NAME` is needed, it's used to register the plugin.

src/groups/mqb/mqbauthn/mqbauthn_basiccredentialprovider.cpp
```diff
@@ -0,0 +1,214 @@
...
+    bslma::Allocator* d_allocator_p;
+
+  public:
+    // TRAITS
+    BSLMF_NESTED_TRAIT_DECLARATION(BasicCredentialFunctor,
+                                   bslma::UsesBslmaAllocator)
+
+    // CREATORS
+    BasicCredentialFunctor(const bsl::string& username,
+                           const bsl::string& password,
```
678098:
```suggestion
    BasicCredentialFunctor(bsl::string_view username,
                           bsl::string_view password,
```

This is needed so the following call doesn't create intermediate strings:
```
mqbplug::CredentialProvider::CredentialCb BasicCredentialProvider::load()
{
    return BasicCredentialFunctor(d_username, d_password, d_allocator_p);
}
```

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
```diff
@@ -499,5 +583,18 @@ AuthenticationController::anonymousCredential()
...
+// ACCESSORS
+bool AuthenticationController::hasCredentialProvider() const
+{
+    return d_credentialProvider_mp.get() != 0;
+}
+
+const mqbplug::CredentialProvider::CredentialCb&
+AuthenticationController::credentialCb() const
+{
+    BSLS_ASSERT_SAFE(hasCredentialProvider());
```
678098, line 600:
```suggestion
    BSLS_ASSERT_SAFE(hasCredentialProvider());
    BSLS_ASSERT_SAFE(d_credentialCb);
```

We typically either have both or don't have both at the same time, but what if we break it one day and have non-zero ManagedPtr and null `d_credentialCb`. Worth checking `d_credentialCb` itself because it is going to be called

678098 (dismissed):

src/groups/mqb/mqbauthn/mqbauthn_authenticationcontroller.cpp
chrisbeard: Yes, good point

678098 (approved):
Looks good

PR #1435, opened 2026-05-20
Misc[mqbc]: Assert primary non-null before dereferencing

kaikulimu:
Assertion is better than segmentation fault. Nuff said.


678098 (approved):

678098:
Agree, segfaults are not nice

PR #1436, opened 2026-05-20
Fix: Do not append empty message property

pniedzielski:
This appears to be a pre-existing bug that was copied from binary message properties to string message properties.  It is within contract to construct a message with both empty binary message property values and empty string message value (in fact, our Python SDK tests these cases).  Internally when streaming out a message property, we call `BlobUtil::append`.  In the case of `bsl::string`s, the empty string is really a buffer containing the NUL byte, so `str.data()` is a pointer to the NUL byte, and has well-defined behavior, even though the `append` never appends that NUL byte.  For an empty `bsl::vector<char>`, though, `vec.data()` is immediate undefined behavior. When compiled with safe asserts on, as we have recently started doing, this results in a crash.

To make matters worse, the recent change to allow `bsl::string_view` message properties now means we can get this same issue with strings, because a `bsl::string_view` need not point to a NUL-terminated string; it is always illegal to call `data` on an empty `string_view` for the same reason as above.

The same issue in principle applies when streaming an empty message property in, both for binary message properties and now string message properties.  If we stream in an empty binary message property, we construct a vector of length `0` and then dereference it to pass it to `readNBytes`, resulting in undefined behavior (and with safe asserts, a broker crash).

And now, if we stream in an empty string message property, a similar issue can surface now where we attempt to read out past the end of the message property blob buffer.

This patch fixes all four cases: the wire protocol allows length zero message property values (which were already in contract and tested my the Python SDK), but on each end, we only construct in memory representations without invoking undefined behavior.  Additionally, this patch adds a minimal test that only passes when all four of the above changes have been made, showing that each really is necessary.

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_messageproperties.cpp
+++ b/src/groups/bmq/bmqp/bmqp_messageproperties.cpp
@@ -104,16 +104,20 @@ class PropertyValueStreamOutVisitor {
 
     void operator()(const bsl::string_view& value)
     {
-        bdlbb::BlobUtil::append(d_blob_p,
-                                value.data(),
-                                static_cast<int>(value.length()));
+        if (!value.empty()) {
+            bdlbb::BlobUtil::append(d_blob_p,
+                                    value.data(),
+                                    static_cast<int>(value.length()));
+        }
     }
 
     void operator()(const bsl::vector<char>& value)
     {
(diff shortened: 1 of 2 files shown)


pniedzielski:
This change shows we need ITs for message properties, we should have caught this much earlier.

dorjesinpo (approved):
thank you

PR #1440, opened 2026-05-21
IT[test_wipe_partitions.py]: Add more scenarios

kaikulimu:
## Summary
  - Add comprehensive integration tests for node recovery after wiping partition storage files, covering single-partition wipe, CSL-only wipe, all-partition wipe, majority-node wipe, qlist-only wipe, stale replica + wipe scenarios, and single-node wipe.
  - Existing tests (`test_wipe_single_partition_files`, `test_wipe_all_partitions_and_csl`) are enhanced to verify message delivery through the healed replica, not just journal file comparison.
  - FSM-only tests (`test_wipe_storage_all_but_one_node`, `test_wipe_partitions_stale_replica`) use staggered startup to ensure the surviving node wins the leader election before wiped nodes join.

### DIFF
--- a/src/integration-tests/test_wipe_partitions.py
+++ b/src/integration-tests/test_wipe_partitions.py
@@ -396,3 +396,91 @@ def test_wipe_qlist_all_nodes(
 
     consumer.confirm(uri_priority, "*", succeed=True)
     consumer.close(uri_priority, succeed=True)
+
+
+@tweak.cluster.queue_operations.shutdown_timeout_ms(100)
+def test_wipe_partitions_stale_replica(
+    fsm_multi_node: Cluster,
+    sc_domain_urls: tc.DomainUrls,
+) -> None:
+    """
+    Test recovery when a stale replica (which missed messages) coexists
+    with wiped replicas.
+    - start cluster
+    - post initial messages
+    - stop one replica (it becomes stale)
+    - post more messages
+    - stop the entire cluster
+    - wipe partition files from two other nodes, including the original primary
+    - restart all nodes, but make sure the surviving replica is part of the
+      initial leader quorum
+    - verify all nodes recover and a consumer can receive all messages
+
+    NOTE: This test is run in FSM mode only; in legacy mode, the cluster cannot
+    heal when the majority of nodes have incorrect storage.
+    """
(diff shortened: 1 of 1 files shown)


src/integration-tests/test_wipe_partitions.py
```diff
@@ -330,3 +330,69 @@ def test_wipe_storage_all_but_one_node(
...
         surviving_node.name, wiped_node.name, cluster
     )
+
+
+@tweak.cluster.queue_operations.shutdown_timeout_ms(100)
+def test_wipe_qlist_all_nodes(
+    multi_node: Cluster,
+    domain_urls: tc.DomainUrls,
+) -> None:
+    """
```
pniedzielski, line 340: Will this behavior stay true even after we switch fully to CSL?

src/integration-tests/test_wipe_partitions.py
```diff
@@ -484,3 +484,72 @@ def test_wipe_partitions_stale_replica(
...
     )
+
+
+@tweak.cluster.queue_operations.shutdown_timeout_ms(100)
+def test_single_node_wipe_partition_keep_csl(
+    single_node: Cluster,
+    domain_urls: tc.DomainUrls,
+) -> None:
+    """
+    Test that a single-node cluster can recover after partition files are
```
pniedzielski, line 495: Does this test really verify this?  In the final step, as soon as any client connects, the queue will become assigned, and we should be able to open it (i.e., queue is functional).  Is that correct?

src/integration-tests/test_wipe_partitions.py
kaikulimu: Yes, FSM mode uses CSL fully. As per discussion with Vitaly, we agreed that only primary will interact with the CSL, and replicas trust the partition files blindly. The primary will notice that CSL has extra queues compared to the partition files, but will not create them due to #1404

src/integration-tests/test_wipe_partitions.py
kaikulimu: The test verifies that the old messages will be lost but the new instance of the queue will be functional. Please let me know if you have any suggestion to improve the test.

pniedzielski (approved):

PR #1441, opened 2026-05-21
Refactor[mqbstat]: extract file logging from `mqbstat::TablePrinter`

678098:
`TablePrinter` was doing three jobs: formatting stats into tables, deciding when to print, and managing file I/O (rotation, cleanup). This change splits those responsibilities:

- `SnapshotTracker` (new private class in `StatController`) - owns the countdown logic that determines when periodic stats output should fire. Exposes `willPrintOnNextSnapshot()` so callers can prepare (e.g., snapshot allocator contexts)  before the print triggers.
- `StatsFileLogger` (new component) — owns the `ball::FileObserver2`, file rotation config, and log cleanup. Accepts a `PrinterCb` callback so it doesn't know about `TablePrinter` directly.
- `TablePrinter` (simplified) — pure formatter. Takes a stat contexts map and history size, prints to any `bsl::ostream`. No config, no file I/O, no start/stop lifecycle. 
- `StatController` — orchestrates: initializes the tracker, conditionally creates the file logger, binds `TablePrinter::printStats` as the callback passed to `logStats`. 

This is a preparation PR to extend stats logging with Json. We will be able to reuse `StatsFileLogger` and `SnapshotTracker` and avoid duplicating this logics in `JsonPrinter`.

https://github.com/bloomberg/blazingmq/pull/1180

### DIFF
--- a/src/groups/mqb/mqbstat/mqbstat_statsfilelogger.cpp
+++ b/src/groups/mqb/mqbstat/mqbstat_statsfilelogger.cpp
@@ -1,0 +1,134 @@
+// Copyright 2026 Bloomberg Finance L.P.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <mqbstat_statsfilelogger.h>
+
+#include <mqbscm_version.h>
+// MQB
+#include <mqbcfg_brokerconfig.h>
+#include <mqbcfg_messages.h>
+
+// BDE
+#include <ball_context.h>
+#include <ball_log.h>
+#include <ball_logfilecleanerutil.h>
+#include <ball_recordattributes.h>
+#include <ball_recordstringformatter.h>
+#include <ball_transmission.h>
+#include <bdls_processutil.h>
+#include <bdlt_datetime.h>
+#include <bdlt_epochutil.h>
+#include <bsl_ctime.h>
+#include <bsl_ostream.h>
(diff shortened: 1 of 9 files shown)


src/groups/mqb/mqbstat/mqbstat_statsfilelogger.cpp
```diff
@@ -0,0 +1,134 @@
...
+    attributes.setCategory(k_LOG_CATEGORY);
+    attributes.setSeverity(ball::Severity::e_INFO);
+
+    attributes.clearMessage();
+    bsl::ostream os(&attributes.messageStreamBuf());
+    os << "===== ===== ===== ===== ===== ===== ===== ===== ===== =====\n"
+       << "Stats id: " << statId << " @ " << now << "\n"
+       << "===== ===== ===== ===== ===== ===== ===== ===== ===== =====\n";
+
+    printerCb(os);
```
678098, line 126: `TablePrinter::print` is called here (passed in `printerCb`), `JsonPrinter::print` will also be called here in the following PRs

src/groups/mqb/mqbstat/mqbstat_statsfilelogger.cpp
```diff
@@ -0,0 +1,134 @@
...
+    attributes.setFileName(__FILE__);
+    attributes.setLineNumber(__LINE__);
+    attributes.setCategory(k_LOG_CATEGORY);
+    attributes.setSeverity(ball::Severity::e_INFO);
+
+    attributes.clearMessage();
+    bsl::ostream os(&attributes.messageStreamBuf());
+    os << "===== ===== ===== ===== ===== ===== ===== ===== ===== =====\n"
+       << "Stats id: " << statId << " @ " << now << "\n"
+       << "===== ===== ===== ===== ===== ===== ===== ===== ===== =====\n";
```
678098, line 124: We should probably not print this for `Json`s, will fix it in `Json`s support PR

src/groups/mqb/mqbstat/mqbstat_statsfilelogger.cpp
chrisbeard: Json may get interesting here. Right now, this prints as one giant log line, if I understand this BALL record logic above? In JSON, we were talking about splitting that up so that logs are individual, per-stat.

chrisbeard (approved):
Looks good, this cleans up the TablePrinter significantly which I think is a good sign for the next steps.

PR #1442, opened 2026-05-21
Fix[mqbc]: Stop d_miscWorkThreadPool in StorageManager::stop()

### LINKED ISSUE #1437: StorageManager::stop() does not stop d_miscWorkThreadPool
`StorageManager::d_miscWorkThreadPool` is constructed and started via `StorageUtil::assignPartitionDispatcherThreads`, but `StorageManager::stop()` never calls `d_miscWorkThreadPool.stop()`.

During shutdown, `stop()` tears down partitions, dispatchers, and recovery managers while the pool may still have in-flight tasks referencing those members, creating a potential use-after-destroy. `mqbblp::Cluster` correctly pairs `start()` and `stop()` for its equivalent pool.

The code:

https://github.com/bloomberg/blazingmq/blob/main/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L3990-L4035

SrinathhSatuluri:
Closes: #1437

`StorageManager::d_miscWorkThreadPool` is started via `StorageUtil::assignPartitionDispatcherThreads` but never stopped in `StorageManager::stop()`. During shutdown, `stop()` tears down partition file stores, unregisters FSM observers, and cancels scheduler events while the pool may still have in-flight tasks referencing those resources, creating a potential use-after-destroy.

The fix adds `d_miscWorkThreadPool.stop()` after the `d_isStarted = false` guard and before the partition teardown loops, ensuring all in-flight tasks complete while the resources they depend on are still valid. 

A unit test is included that verifies the thread pool is no longer started after `StorageManager::stop()` returns.

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_storagemanager.cpp
+++ b/src/groups/mqb/mqbc/mqbc_storagemanager.cpp
@@ -4011,6 +4011,8 @@ void StorageManager::stop()
 
     d_isStarted = false;
 
+    d_miscWorkThreadPool.stop();
+
     for (size_t pid = 0; pid < d_fileStores.size(); ++pid) {
         mqbs::FileStore* fs = d_fileStores[pid].get();
         BSLS_ASSERT_SAFE(fs);
--- a/src/groups/mqb/mqbc/mqbc_storagemanager.cpp
+++ b/src/groups/mqb/mqbc/mqbc_storagemanager.cpp
@@ -4011,8 +4011,6 @@ void StorageManager::stop()
 
     d_isStarted = false;
 
-    d_miscWorkThreadPool.stop();
-
     for (size_t pid = 0; pid < d_fileStores.size(); ++pid) {
         mqbs::FileStore* fs = d_fileStores[pid].get();
         BSLS_ASSERT_SAFE(fs);
@@ -4045,6 +4043,8 @@ void StorageManager::stop()
                              this,
                              bdlf::PlaceHolders::_1,    // partitionId
                              bdlf::PlaceHolders::_2));  // latch
+
+    d_miscWorkThreadPool.stop();
 }
 
 void StorageManager::initializeQueueKeyInfoMap(
--- a/src/groups/mqb/mqbc/mqbc_storagemanager.h
+++ b/src/groups/mqb/mqbc/mqbc_storagemanager.h
@@ -1077,6 +1077,11 @@ class StorageManager BSLS_KEYWORD_FINAL
     /// THREAD: Must be called from the cluster dispatcher thread or the
     ///         partition dispatcher thread for the specified `partitionId`.
     bool isWatchdogActive(int partitionId) const;
+
+    /// Return whether the miscellaneous work thread pool is currently
+    /// started.  This accessor is meant to be used for unit testing
+    /// purposes.
+    bool isMiscWorkThreadPoolStarted() const;
 };
 
 // ============================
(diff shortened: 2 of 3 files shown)


src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -4011,6 +4011,8 @@ void StorageManager::stop()
 
     d_isStarted = false;
 
+    d_miscWorkThreadPool.stop();
```
678098: I think it is safer to move it after `StorageUtil::stop`

678098: The idea is that it is totally fine for `FileStore`-s to use misc thread pool until the moment we stop them. After this, they should not be using misc thread pool, and we can safely stop it.

src/groups/mqb/mqbc/mqbc_storagemanager.h
```diff
@@ -1077,6 +1077,11 @@ class StorageManager BSLS_KEYWORD_FINAL
     /// THREAD: Must be called from the cluster dispatcher thread or the
     ///         partition dispatcher thread for the specified `partitionId`.
     bool isWatchdogActive(int partitionId) const;
+
+    /// Return whether the miscellaneous work thread pool is currently
+    /// started.  This accessor is meant to be used for unit testing
+    /// purposes.
+    bool isMiscWorkThreadPoolStarted() const;
```
678098: Looking back at this, I am not sure if we really need to make a unit test for this. There are several problems with it:
1. It extends `mqbc::StorageManager` class API with `isMiscWorkThreadPoolStarted` accessor that is useful only in this one test and is not used anywhere else.
2. It is 50+ lines to check just 1 line, that is also obviously fine.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
SrinathhSatuluri: I see now that the pool should outlive its producers, so the FileStores can enqueue work freely until they're drained. One thing I wanted to clarify: does bdlmt::FixedThreadPool::stop() drain already-enqueued jobs before returning, or could in-flight work be dropped?

src/groups/mqb/mqbc/mqbc_storagemanager.h
SrinathhSatuluri: Will remove the isMiscWorkThreadPoolStarted() accessor and test. As you said for this one line, I suppose the unit test is not necessary, but for lifecycle fixes like this where unit testing requires exposing internals, would the integration test suite be the better layer to catch regressions?

678098: In such a case, adding a precondition to a destructor is enough.
In other cases, sanitizers can help to find a use-after-free situation.
A good thing about sanitizers is that they can catch a problem during execution of another test.

678098 (approved):
Looks good

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
678098: From bdlmt docs:
```
    /// Disable enqueuing jobs on this thread pool, wait until all active
    /// and pending jobs complete, and join all processing threads.  If the
    /// thread pool was not already started (`isStarted()` is `false`), this
    /// method has no effect.  At the completion of this method,
    /// `false == isStarted()`.
    void stop();
```

https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/bdlmt__fixedthreadpool_8h_source.html

So all pending jobs will be drained.

678098:
@SrinathhSatuluri thank you for contribution!

PR #1443, opened 2026-05-21
Fix[mqbc::StorageMgr]: --d_numPartitionsRecoveredFully when out of healed

kaikulimu:
## Summary
  - Fix `d_numPartitionsRecoveredFully` double-counting bug: if a partition transitions healed → unhealed → healed, the counter was incremented twice for the same partition, causing the "all partitions healed" check to either never trigger or produce incorrect results in `allPartitionsAvailable()`.
  - Add `onTransitionOutOfHealed` observer callback to `PartitionFSMObserver`, which fires when a partition leaves `PRIMARY_HEALED` or `REPLICA_HEALED`. `StorageManager` uses this to decrement the counter.
  - Guard first-time healing logic (`printStorageRecoveryCompletion`, GC event scheduling, `d_recoveryStatusCb`) with `BSLMT_ONCE_DO` to prevent re-execution if the counter re-reaches the target after a heal/unheal/re-heal cycle.

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_storagemanager.cpp
+++ b/src/groups/mqb/mqbc/mqbc_storagemanager.cpp
@@ -329,42 +329,35 @@ void StorageManager::onPartitionRecovery(int partitionId)
         d_recoveryStartTimes[partitionId]);
     BALL_LOG_INFO << out.str();
 
-    if (++d_numPartitionsRecoveredFully ==
+    if (++d_numPartitionsRecoveredFully !=
         static_cast<int>(d_fileStores.size())) {
-        BSLMT_ONCE_DO
-        {
-            // First time healing logic
+        BSLS_ASSERT_SAFE(d_numPartitionsRecoveredFully <
+                         static_cast<int>(d_fileStores.size()));
+        return;  // RETURN
+    }
 
-            out.reset();
-            const bool success =
-                mqbs::StoragePrintUtil::printStorageRecoveryCompletion(
-                    out,
-                    d_fileStores,
-                    d_clusterData_p->identity().description());
-            BALL_LOG_INFO << out.str();
+    BSLMT_ONCE_DO
+    {
+        // First time healing logic
 
-            if (success) {
-                // All partitions have opened successfully.  Schedule a
-                // recurring event in StorageMgr, which will in turn enqueue an
-                // event in each partition's dispatcher thread for GC'ing
-                // expired messages as well as cleaning history.
+        out.reset();
+        const bool success =
+            mqbs::StoragePrintUtil::printStorageRecoveryCompletion(
+                out,
+                d_fileStores,
+                d_clusterData_p->identity().description());
+        BALL_LOG_INFO << out.str();
 
-                d_clusterData_p->scheduler().scheduleRecurringEvent(
-                    &d_gcMessagesEventHandle,
-                    bsls::TimeInterval(k_GC_MESSAGES_INTERVAL_SECONDS),
-                    bdlf::BindUtil::bind(&StorageManager::forceFlushFileStores,
-                                         this));
-
-                // Even though Cluster FSM and all Partition FSMs are now
-                // healed, we must check that all partitions have an active
-                // primary before transitioning ourself to E_AVAILABLE.
-                if (allPartitionsAvailable()) {
-                    d_recoveryStatusCb(0);
-                }
+        if (success) {
+            // Even though Cluster FSM and all Partition FSMs are now
+            // healed, we must check that all partitions have an active
+            // primary before transitioning ourself to E_AVAILABLE.
+            if (allPartitionsAvailable()) {
+                d_recoveryStatusCb(0);
             }
-            else {
-                d_recoveryStatusCb(-1);
-            }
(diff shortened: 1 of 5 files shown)


src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -3831,6 +3835,26 @@ void StorageManager::onTransitionToReplicaHealed(
...
+void StorageManager::onTransitionOutOfHealed(
+    int               partitionId,
+    BSLA_MAYBE_UNUSED PartitionStateTableState::Enum oldState)
+{
+    // executed by *QUEUE_DISPATCHER* thread associated with 'partitionId'
+
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(d_fileStores[partitionId]->inDispatcherThread());
+    BSLS_ASSERT_SAFE(0 <= partitionId &&
+                     partitionId < static_cast<int>(d_fileStores.size()));
```
678098:
```suggestion
    BSLS_ASSERT_SAFE(0 <= partitionId &&
                     partitionId < static_cast<int>(d_fileStores.size()));
    BSLS_ASSERT_SAFE(d_fileStores[partitionId]->inDispatcherThread());
```
There will be a segfault on the first assert if partitionId is out of range, need to check it first

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -330,37 +331,40 @@ void StorageManager::onPartitionRecovery(int partitionId)
 
     if (++d_numPartitionsRecoveredFully ==
         static_cast<int>(d_fileStores.size())) {
```
678098:
```suggestion
if (++d_numPartitionsRecoveredFully !=
static_cast<int>(d_fileStores.size())) {
    return;  // RETURN
}
```

I do believe that early return here is super useful to reduce indentation levels of the entire function below

678098: Alternatively, you can extract the code from BSLMT_ONCE_DO to a separate function `onFirstTimeHealing`

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -330,37 +331,40 @@ void StorageManager::onPartitionRecovery(int partitionId)
...
+
+                // Even though Cluster FSM and all Partition FSMs are now
+                // healed, we must check that all partitions have an active
+                // primary before transitioning ourself to E_AVAILABLE.
+                if (allPartitionsAvailable()) {
+                    d_recoveryStatusCb(0);
+                }
+            }
+            else {
+                d_recoveryStatusCb(-1);
```
678098: What happens with the broker if we enter this branch and call `d_recoveryStatusCb(-1);`? Note that we are in BSLMT_ONCE_DO block and we wont' be able to schedule recurring event - ever. The broker will never be able to recover on itself

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -330,37 +331,40 @@ void StorageManager::onPartitionRecovery(int partitionId)
...
+                // All partitions have opened successfully.  Schedule a
+                // recurring event in StorageMgr, which will in turn enqueue an
+                // event in each partition's dispatcher thread for GC'ing
+                // expired messages as well as cleaning history.
+
+                d_clusterData_p->scheduler().scheduleRecurringEvent(
+                    &d_gcMessagesEventHandle,
+                    bsls::TimeInterval(k_GC_MESSAGES_INTERVAL_SECONDS),
+                    bdlf::BindUtil::bind(&StorageManager::forceFlushFileStores,
+                                         this));
```
678098: Why don't we just schedule it on `StorageManager::start` and stop it on `StorageManager::stop`? We can (and probably should) make `StorageManager::forceFlushFileStores` robust to to unheathy cluster.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: If we call `d_recoveryStatusCb(-1);`, [the broker is terminated](https://github.com/bloomberg/blazingmq/blob/7db465bb9ddf8d821edd987b63893e77f8eaba60/src/groups/mqb/mqbblp/mqbblp_cluster.cpp#L1634-L1643). Therefore, no need to worry about anything that happens afterwards.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: Good idea. The [`if (!fs->isOpen()) {` check](https://github.com/bloomberg/blazingmq/blob/7db465bb9ddf8d821edd987b63893e77f8eaba60/src/groups/mqb/mqbc/mqbc_storageutil.cpp#L3426-L3428) in `StorageManager::forceFlushFileStores` already makes it robust to to unhealthy cluster.

678098 (approved):
Looks good

PR #1445, opened 2026-05-22
Fix[bmqio]: cancel inflight reconnecting on shutdown

dorjesinpo:
[https://github.com/bloomberg/blazingmq/pull/1362](https://github.com/bloomberg/blazingmq/pull/1362) was not enough because it allows already scheduled connects.
We need to cancel them at the time of `ReconnectingChannelFactory::disableReconnect()`


678098 (approved):
Looks good

PR #1446, opened 2026-05-22
Fix[bmqeval]: guard against UB in expression evaluation

chrisbeard:
Integer division or modulus by zero in subscription evaluation causes UB, which can result in a SIGFPE on x86. Add check in NumBinaryOperation::evaluate to return a new error code, e_ARITHMETIC, in these cases.

Existing unit tests mixed evaluation failures and errors, indicating the latter by convention. This has been explicitly split into separate test cases that check errors for the correct error code.

### DIFF
--- a/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.h
+++ b/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.h
@@ -103,8 +103,9 @@ struct ErrorType {
         e_NAME             = -1,
         e_TYPE             = -2,
         e_BINARY           = -3,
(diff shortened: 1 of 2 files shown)


678098 (approved):

SrinathhSatuluri:
Hello @chrisbeard, I was looking into this and from my understanding, the SIGFPE here comes from the IDIV instruction itself faulting and not from anything language-level. IDIV also faults when the quotient overflows the register, which I think only happens when the dividend is INT64_MIN and the divisor is -1. Would that be a concern here too since it's the same hardware path?

PR #1447, opened 2026-05-26
Feat[mqba,mqbnet]: add AuthenticationClient for outbound broker-to-broker authn

chrisbeard:
This PR contains the second subset of changes in https://github.com/bloomberg/blazingmq/pull/1428 for easier reviewing, follows from https://github.com/bloomberg/blazingmq/pull/1434.

Add `mqbnet::AuthenticationClient` protocol and `mqba::AuthenticationClient`
implementation. The client obtains credentials from a `CredentialProvider`,
sends an AuthenticationRequest, processes the AuthenticationResponse, and
schedules reauthentication at 80% of the credential lifetime.

### DIFF
--- a/src/groups/mqb/mqba/mqba_authenticationclient.cpp
+++ b/src/groups/mqb/mqba/mqba_authenticationclient.cpp
@@ -74,7 +74,6 @@ AuthenticationClient::AuthenticationClient(
     BSLS_ASSERT_SAFE(channel);
     BSLS_ASSERT_SAFE(d_blobSpPool_p);
     BSLS_ASSERT_SAFE(d_scheduler_p);
-    BSLS_ASSERT_SAFE(d_allocator_p);
 }
 
 AuthenticationClient::~AuthenticationClient()
@@ -196,11 +195,11 @@ int AuthenticationClient::handleResponse(
     }
 
     BALL_LOG_INFO << "Authentication successful [peer: " << channel.get()
-                  << "], [lifetimeMs: "
+                  << "], lifetimeMs: "
                   << (authnResponse.lifetimeMs().isNull()
                           ? "N/A"
-                          : bsl::to_string(authnResponse.lifetimeMs().value()))
-                  << "]";
+                          : bsl::to_string(
+                                authnResponse.lifetimeMs().value()));
 
     // If a lifetime is provided, schedule a reauthentication request to be
     // sent at 80% of the lifetime before the connection expires.
--- a/src/groups/mqb/mqba/mqba_authenticationclient.cpp
+++ b/src/groups/mqb/mqba/mqba_authenticationclient.cpp
@@ -47,7 +47,7 @@ void AuthenticationClient::onReauthenticate()
 {
     // executed by the *SCHEDULER* thread
 
-    bmqu::MemOutStream errStream;
+    bmqu::MemOutStream errStream(d_allocator_p);
     const int          rc = authenticate(errStream);
     if (rc != 0) {
         BALL_LOG_ERROR << "Reauthentication failed: " << errStream.str();
@@ -66,10 +66,7 @@ void AuthenticationClient::scheduleReauthentication(
     {
         bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
 
-        if (d_reauthHandle) {
-            d_scheduler_p->cancelEvent(&d_reauthHandle);
-        }
-
+        d_scheduler_p->cancelEvent(&d_reauthHandle);
         d_scheduler_p->scheduleEvent(
             &d_reauthHandle,
             bsls::TimeInterval(bmqu::Time::nowMonotonicClock())
@@ -78,6 +75,7 @@ void AuthenticationClient::scheduleReauthentication(
                                  this));
     }
 
+    // Guaranteed to exist, the caller has already checked
     bsl::shared_ptr<bmqio::Channel> channel = d_channel_wp.lock();
     BALL_LOG_INFO << "Scheduled reauthentication"
                   << " [peer: " << channel.get() << "]" << " in " << reauthMs
@@ -132,7 +130,7 @@ int AuthenticationClient::authenticate(bsl::ostream& errorDescription)
     }
 
     // Obtain credentials
-    bmqu::MemOutStream                      credErrStream;
+    bmqu::MemOutStream                      credErrStream(d_allocator_p);
     bsl::optional<mqbplug::AuthnCredential> credential = d_credentialCb(
         credErrStream);
 
@@ -143,7 +141,7 @@ int AuthenticationClient::authenticate(bsl::ostream& errorDescription)
     }
 
     BALL_LOG_INFO << "Sending AuthenticationRequest"
-                  << " [peer: " << channel.get() << "]" << " with mechanism '"
+                  << " [peer: " << channel.get() << "] with mechanism '"
                   << credential->mechanism() << "'";
 
     // Build AuthenticationRequest
@@ -210,9 +208,15 @@ int AuthenticationClient::handleResponse(
     }
 
     bsl::shared_ptr<bmqio::Channel> channel = d_channel_wp.lock();
+    if (!channel) {
+        // This should never happen. The owner of this client should keep
+        // the channel alive while the response is being processed.
+        BALL_LOG_WARN << "Channel is no longer available";
+        return rc_SUCCESS;  // RETURN
+    }
 
-    BALL_LOG_INFO << "Authentication successful" << " [peer: " << channel.get()
-                  << "]" << " [lifetimeMs: "
+    BALL_LOG_INFO << "Authentication successful [peer: " << channel.get()
+                  << "], [lifetimeMs: "
                   << (authnResponse.lifetimeMs().isNull()
                           ? "N/A"
                           : bsl::to_string(authnResponse.lifetimeMs().value()))
@@ -231,10 +235,7 @@ void AuthenticationClient::stop()
 
     bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
 
-    if (d_reauthHandle) {
-        d_scheduler_p->cancelEvent(&d_reauthHandle);
-    }
-
+    d_scheduler_p->cancelEvent(&d_reauthHandle);
     d_channel_wp.reset();
 }
 
--- a/src/groups/mqb/mqba/mqba_authenticationclient.cpp
+++ b/src/groups/mqb/mqba/mqba_authenticationclient.cpp
@@ -74,11 +74,17 @@ AuthenticationClient::AuthenticationClient(
     BSLS_ASSERT_SAFE(channel);
     BSLS_ASSERT_SAFE(d_blobSpPool_p);
     BSLS_ASSERT_SAFE(d_scheduler_p);
+
+    d_gateKeeper.open();
 }
 
 AuthenticationClient::~AuthenticationClient()
 {
-    stop();
+    d_gateKeeper.close();
+    d_scheduler_p->cancelEventAndWait(&d_reauthHandle);
+
+    bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
+    d_channel_wp.reset();
 }
 
 // MANIPULATORS
@@ -165,6 +171,13 @@ int AuthenticationClient::handleResponse(
         rc_AUTHENTICATION_FAILURE = -2
     };
 
+    // Prevent scheduling reauthentication if destructor has been called.
+    // The destructor will wait for this function to return.
+    bmqu::GateKeeper::Status gateStatus(d_gateKeeper);
+    if (!gateStatus.isOpen()) {
+        return rc_SUCCESS;  // RETURN
+    }
+
     if (!response.isAuthenticationResponseValue()) {
         errorDescription << "Expected AuthenticationResponse but received: "
                          << response;
@@ -228,17 +241,6 @@ int AuthenticationClient::handleResponse(
     }
 
     return rc_SUCCESS;
-}
-
-void AuthenticationClient::stop()
-{
-    // executed by *ANY* thread (during channel teardown)
-
-    // Cancel future events, synchronize with any currently executing callback
-    d_scheduler_p->cancelEventAndWait(&d_reauthHandle);
-
-    bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
-    d_channel_wp.reset();
 }
 
 }  // close package namespace
--- a/src/groups/mqb/mqba/mqba_authenticationclient.cpp
+++ b/src/groups/mqb/mqba/mqba_authenticationclient.cpp
@@ -54,34 +54,6 @@ void AuthenticationClient::onReauthenticate()
     }
 }
 
-void AuthenticationClient::scheduleReauthentication(
-    bsls::Types::Int64 lifetimeMs)
-{
-    // executed by the *IO* thread
-
-    // Reauthenticate at 80% of the lifetime to give margin
-    const bsls::Types::Int64 reauthMs = static_cast<bsls::Types::Int64>(
-        static_cast<double>(lifetimeMs) * 0.8);
-
-    {
-        bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
-
-        d_scheduler_p->cancelEvent(&d_reauthHandle);
-        d_scheduler_p->scheduleEvent(
-            &d_reauthHandle,
-            bsls::TimeInterval(bmqu::Time::nowMonotonicClock())
-                .addMilliseconds(reauthMs),
-            bdlf::BindUtil::bind(&AuthenticationClient::onReauthenticate,
-                                 this));
-    }
-
-    // Guaranteed to exist, the caller has already checked
-    bsl::shared_ptr<bmqio::Channel> channel = d_channel_wp.lock();
-    BALL_LOG_INFO << "Scheduled reauthentication"
-                  << " [peer: " << channel.get() << "]" << " in " << reauthMs
-                  << " ms (lifetimeMs: " << lifetimeMs << ")";
-}
-
 // CREATORS
 AuthenticationClient::AuthenticationClient(
     const mqbplug::CredentialProvider::CredentialCb& credentialCb,
@@ -123,7 +95,11 @@ int AuthenticationClient::authenticate(bsl::ostream& errorDescription)
         rc_WRITE_FAILURE      = -4
     };
 
-    bsl::shared_ptr<bmqio::Channel> channel = d_channel_wp.lock();
+    bsl::shared_ptr<bmqio::Channel> channel;
+    {
+        bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
+        channel = d_channel_wp.lock();
+    }
     if (!channel) {
         errorDescription << "Channel is no longer available";
         return rc_CHANNEL_GONE;  // RETURN
@@ -207,7 +183,11 @@ int AuthenticationClient::handleResponse(
         return rc_AUTHENTICATION_FAILURE;  // RETURN
     }
 
-    bsl::shared_ptr<bmqio::Channel> channel = d_channel_wp.lock();
+    bsl::shared_ptr<bmqio::Channel> channel;
+    {
+        bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
+        channel = d_channel_wp.lock();
+    }
     if (!channel) {
         // This should never happen. The owner of this client should keep
         // the channel alive while the response is being processed.
@@ -222,8 +202,30 @@ int AuthenticationClient::handleResponse(
                           : bsl::to_string(authnResponse.lifetimeMs().value()))
                   << "]";
 
+    // If a lifetime is provided, schedule a reauthentication request to be
+    // sent at 80% of the lifetime before the connection expires.
     if (authnResponse.lifetimeMs().has_value()) {
-        scheduleReauthentication(authnResponse.lifetimeMs().value());
+        const bsls::Types::Int64 lifetimeMs =
+            authnResponse.lifetimeMs().value();
+
+        const bsls::Types::Int64 reauthMs = static_cast<bsls::Types::Int64>(
+            static_cast<double>(lifetimeMs) * 0.8);
+
+        {
+            bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
+
+            d_scheduler_p->cancelEvent(&d_reauthHandle);
+            d_scheduler_p->scheduleEvent(
+                &d_reauthHandle,
+                bsls::TimeInterval(bmqu::Time::nowMonotonicClock())
+                    .addMilliseconds(reauthMs),
+                bdlf::BindUtil::bind(&AuthenticationClient::onReauthenticate,
+                                     this));
+        }
+
+        BALL_LOG_INFO << "Scheduled reauthentication"
+                      << " [peer: " << channel.get() << "] in " << reauthMs
+                      << " ms (lifetimeMs: " << lifetimeMs << ")";
     }
 
     return rc_SUCCESS;
@@ -233,9 +235,10 @@ void AuthenticationClient::stop()
 {
     // executed by *ANY* thread (during channel teardown)
 
-    bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
+    // Cancel future events, synchronize with any currently executing callback
+    d_scheduler_p->cancelEventAndWait(&d_reauthHandle);
 
-    d_scheduler_p->cancelEvent(&d_reauthHandle);
+    bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
     d_channel_wp.reset();
 }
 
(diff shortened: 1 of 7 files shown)


src/groups/mqb/mqbnet/mqbnet_authenticationclient.h
```diff
@@ -0,0 +1,76 @@
...
+/// @brief Provide a protocol for a client-side authenticator.
+///
+/// @bbref{mqbnet::AuthenticationClient} is a protocol for the client side of
+/// broker-to-broker authentication for a single connection.  An
+/// implementation sends an @bbref{bmqp_ctrlmsg::AuthenticationRequest}
+/// on an outbound channel and processes the resulting
+/// @bbref{bmqp_ctrlmsg::AuthenticationResponse}.
+
+// BMQ
+#include <bmqp_ctrlmsg_messages.h>
```
678098, line 30: This header is heavy, not necessary to include it in all sources that include this interface. Can replace its usage with a forward declaration to `bmqp_ctrlmsg::AuthenticationMessage` that is only used as a reference.

src/groups/mqb/mqbnet/mqbnet_authenticationclient.h
```diff
@@ -0,0 +1,76 @@
...
+
+// BDE
+#include <bsl_memory.h>
```
678098:
```suggestion
```
Not used here

src/groups/mqb/mqba/mqba_authenticationclient.h
```diff
@@ -0,0 +1,157 @@
...
+/// outbound connection to a broker.  It obtains credentials from a
+/// @bbref{mqbplug::CredentialProvider::CredentialFunc} and sends an
+/// @bbref{bmqp_ctrlmsg::AuthenticationRequest} to the remote broker.
+
+// MQB
+#include <mqbnet_authenticationclient.h>
+#include <mqbplug_credentialprovider.h>
+
+// BMQ
+#include <bmqio_channel.h>
```
678098: Can be replaced with a forward declaration.
Only referenced here as pointer

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,242 @@
...
+// --------------------------
+// class AuthenticationClient
+// --------------------------
+
+// PRIVATE MANIPULATORS
+void AuthenticationClient::onReauthenticate()
+{
+    // executed by the *SCHEDULER* thread
+
+    bmqu::MemOutStream errStream;
```
678098:
```suggestion
    bmqu::MemOutStream errStream(d_allocator_p);
```

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,242 @@
...
+    };
+
+    bsl::shared_ptr<bmqio::Channel> channel = d_channel_wp.lock();
+    if (!channel) {
+        errorDescription << "Channel is no longer available";
+        return rc_CHANNEL_GONE;  // RETURN
+    }
+
+    // Obtain credentials
+    bmqu::MemOutStream                      credErrStream;
```
678098:
```suggestion
    bmqu::MemOutStream                      credErrStream(d_allocator_p);
```

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,242 @@
...
+    const bsls::Types::Int64 reauthMs = static_cast<bsls::Types::Int64>(
+        static_cast<double>(lifetimeMs) * 0.8);
+
+    {
+        bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
+
+        if (d_reauthHandle) {
+            d_scheduler_p->cancelEvent(&d_reauthHandle);
+        }
+
```
678098:
```suggestion
        d_scheduler_p->cancelEvent(&d_reauthHandle);
```
`cancelEvent` is a no-op when event handle is null(undefined).

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,242 @@
...
+            bsls::TimeInterval(bmqu::Time::nowMonotonicClock())
+                .addMilliseconds(reauthMs),
+            bdlf::BindUtil::bind(&AuthenticationClient::onReauthenticate,
+                                 this));
+    }
+
+    bsl::shared_ptr<bmqio::Channel> channel = d_channel_wp.lock();
+    BALL_LOG_INFO << "Scheduled reauthentication"
+                  << " [peer: " << channel.get() << "]" << " in " << reauthMs
+                  << " ms (lifetimeMs: " << lifetimeMs << ")";
```
678098: `channel` sp might be a null pointer here, if `lock()` failed.
If it is null, it is better to log something else.
Also, if it is null at the start of this function, we don't need to schedule anything.

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,242 @@
...
+        credErrStream);
+
+    if (!credential.has_value()) {
+        errorDescription << "Failed to obtain credentials: "
+                         << credErrStream.str();
+        return rc_CREDENTIAL_FAILURE;  // RETURN
+    }
+
+    BALL_LOG_INFO << "Sending AuthenticationRequest"
+                  << " [peer: " << channel.get() << "]" << " with mechanism '"
```
678098:
```suggestion
                  << " [peer: " << channel.get() << "] with mechanism '"
```

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+    if (authnResponse.status().category() !=
+        bmqp_ctrlmsg::StatusCategory::E_SUCCESS) {
+        errorDescription << "Authentication failed: "
+                         << authnResponse.status().message()
+                         << " [code: " << authnResponse.status().code() << "]";
+        return rc_AUTHENTICATION_FAILURE;  // RETURN
+    }
+
+    bsl::shared_ptr<bmqio::Channel> channel = d_channel_wp.lock();
```
678098: If `channel` is null here, might need to do/log something else

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+    BALL_LOG_INFO << "Authentication successful" << " [peer: " << channel.get()
```
678098:
```suggestion
    BALL_LOG_INFO << "Authentication successful [peer: " << channel.get()
```

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,242 @@
...
+                  << "]" << " [lifetimeMs: "
```
678098:
```suggestion
                  << ", lifetimeMs: "
```

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+void AuthenticationClient::stop()
+{
+    // executed by *ANY* thread (during channel teardown)
+
+    bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
+
+    if (d_reauthHandle) {
+        d_scheduler_p->cancelEvent(&d_reauthHandle);
+    }
```
678098:
```suggestion
    d_scheduler_p->cancelEvent(&d_reauthHandle);
```
No op even if null, documented behaviour of `cancelEvent`

src/groups/mqb/mqbnet/mqbnet_authenticationclient.h
chrisbeard: It is, but `mqbnet_authenticator.h` already includes the header so it doesn't make much of a difference in practice.

678098: Looking at `mqbnet_authentcator.h`, it can also forward declare `bmqp_ctrlmsg_messages.h`

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,243 @@
...
+}
+
+void AuthenticationClient::stop()
+{
+    // executed by *ANY* thread (during channel teardown)
+
+    bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
+
+    d_scheduler_p->cancelEvent(&d_reauthHandle);
+    d_channel_wp.reset();
```
678098: Note that `d_channel_wp.reset()` is a write operation on the same weak_ptr object that is read in multiple places without locking a mutex, it is not thread safe.
There are a couple of alternatives what to do, but I think a good one is to replace mutex usage with `bmqu_atomicgate`.

src/groups/mqb/mqbnet/mqbnet_authenticationclient.h
678098: However, we have a problem with `bsl::optional<mqbcfg::Credential>` in `mqbnet_authentcator.h`: optional requires type knowledge, so it is not possible to forward declare the type used in it. Maybe it is a design overlook

src/groups/mqb/mqba/mqba_authenticationclient.h
```diff
@@ -0,0 +1,161 @@
...
+/// outbound connection to a broker.  It obtains credentials from a
+/// @bbref{mqbplug::CredentialProvider::CredentialFunc} and sends an
+/// @bbref{bmqp_ctrlmsg::AuthenticationRequest} to the remote broker.
+
+// MQB
+#include <mqbnet_authenticationclient.h>
+#include <mqbplug_credentialprovider.h>
+
+// BMQ
+
```
678098, line 35:
```suggestion
```

src/groups/mqb/mqba/mqba_authenticationclient.t.cpp
```diff
@@ -0,0 +1,741 @@
...
+        default: {
+            cerr << "WARNING: CASE '" << _testCase << "' NOT FOUND." << endl;
+            bmqtst::TestHelperUtil::testStatus() = -1;
+        } break;
+        }
+
+        bmqu::Time::shutdown();
+    }
+
+    TEST_EPILOG(bmqtst::TestHelper::e_DEFAULT);
```
678098: With this setting, allocators checks are ignored

src/groups/mqb/mqba/mqba_authenticationclient.t.cpp
```diff
@@ -0,0 +1,741 @@
...
+    resp.status().code()     = code;
+    resp.status().message()  = message;
+    return msg;
+}
+
+int decodeAuthenticationEvent(bmqp_ctrlmsg::AuthenticationMessage* output,
+                              const bdlbb::Blob&                   blob,
+                              bslma::Allocator*                    allocator)
+{
+    bmqp::Event event(&blob, allocator);
```
678098, line 135:
```suggestion
    BSLS_ASSERT(output);

    bmqp::Event event(&blob, allocator);
```

src/groups/mqb/mqba/mqba_authenticationclient.t.cpp
```diff
@@ -0,0 +1,741 @@
...
+}
+
+bsl::optional<mqbplug::AuthnCredential>
+successCredentialCb(bsl::ostream&      error,
+                    const bsl::string& mechanism,
+                    const bsl::string& payload)
+{
+    (void)error;
+    bsl::vector<char> data(payload.begin(), payload.end());
+    return mqbplug::AuthnCredential(mechanism, data);
```
678098:
```suggestion
    bsl::vector<char> data(payload.begin(), payload.end(), bmqtst::TestHelperUtil::allocator());
    return mqbplug::AuthnCredential(mechanism, data, bmqtst::TestHelperUtil::allocator());
```

src/groups/mqb/mqba/mqba_authenticationclient.t.cpp
```diff
@@ -0,0 +1,741 @@
...
+    // CREATORS
+    explicit TestBench(bslma::Allocator* allocator)
+    : d_bufferFactory(256, allocator)
+    , d_blobSpPool(bdlf::BindUtil::bind(&createBlob,
+                                        &d_bufferFactory,
+                                        bdlf::PlaceHolders::_1,
+                                        bdlf::PlaceHolders::_2),
+                   1024,
+                   allocator)
+    , d_channel(new bmqio::TestChannel(allocator))
```
678098:
```suggestion
    , d_channel_sp(bsl::allocate_shared<bmqio::TestChannel>(allocator))
```

src/groups/mqb/mqba/mqba_authenticationclient.t.cpp
```diff
@@ -0,0 +1,741 @@
...
+    {
+        d_scheduler.cancelAllEventsAndWait();
+        d_scheduler.stop();
+    }
+
+    // MANIPULATORS
+    mqba::AuthenticationClient*
+    createClient(const mqbplug::CredentialProvider::CredentialCb& credCb)
+    {
+        return new (*d_allocator_p) mqba::AuthenticationClient(credCb,
```
678098: `new (*d_allocator_p)` causes a segfault if `d_allocator_p` is null. The second problem is that `mqba::AuthenticationClient* createClient(...)` API is only safe when its user manages the raw pointer correctly: e.g. assigns it to a smart pointer type right away.

Both issues can be addressed by constructing and returning a smart pointer right away with an appropriate helper, both `ManagedPtr` or `shared_ptr` are fine:

```
bsl::shared_ptr<mqba::AuthenticationClient>
    createClient(const mqbplug::CredentialProvider::CredentialCb& credCb)
    {
        return bsl::allocate_shared<mqba::AuthenticationClient>(d_allocator_p, ... 
```

src/groups/mqb/mqba/mqba_authenticationclient.t.cpp
```diff
@@ -0,0 +1,741 @@
...
+    BMQTST_ASSERT_EQ(rc, 0);
+
+    // 2)
+    tb.d_testClock.d_timeSource.advanceTime(
+        bsls::TimeInterval().addMilliseconds(7999));
+    BMQTST_ASSERT_EQ(tb.d_channel->numWriteCalls(), writeCountAfterAuth);
+
+    // 3)
+    tb.d_testClock.d_timeSource.advanceTime(
+        bsls::TimeInterval().addMilliseconds(1));
```
678098, line 445: From bdlmt
```
// Also note that `bdlt::EventSchedulerTestTimeSource::advanceTime`
// synchronizes with the dispatching thread.  Specifically, `advanceTime` does
// not return until the dispatcher processes all events triggered by the change
// in current time.
```
Would be nice to have a comment in-place describing this behaviour

678098 (commented):
More comments

src/groups/mqb/mqba/mqba_authenticationclient.cpp
chrisbeard: Valid, and second issue, this should really use `cancelEventAndWait`. Will fix both.

src/groups/mqb/mqba/mqba_authenticationclient.t.cpp
chrisbeard: Added global allocator check

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,246 @@
...
+, d_mutex()
+, d_reauthHandle()
+, d_allocator_p(allocator)
+{
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(d_credentialCb);
+    BSLS_ASSERT_SAFE(channel);
+    BSLS_ASSERT_SAFE(d_blobSpPool_p);
+    BSLS_ASSERT_SAFE(d_scheduler_p);
+    BSLS_ASSERT_SAFE(d_allocator_p);
```
678098:
```suggestion
```
Possible to allow `NULL` `d_allocator_p` because it is only passed to constructors where it defaults to the `bslma::Default` allocator: no direct `new` calls that might segfault with null allocator

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,246 @@
...
+    }
+    if (!channel) {
+        // This should never happen. The owner of this client should keep
+        // the channel alive while the response is being processed.
+        BALL_LOG_WARN << "Channel is no longer available";
+        return rc_SUCCESS;  // RETURN
+    }
+
+    BALL_LOG_INFO << "Authentication successful [peer: " << channel.get()
+                  << "], [lifetimeMs: "
```
678098:
```suggestion
                  << ", lifetimeMs: "
```

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+    return rc_SUCCESS;
+}
+
+void AuthenticationClient::stop()
+{
+    // executed by *ANY* thread (during channel teardown)
+
+    // Cancel future events, synchronize with any currently executing callback
+    d_scheduler_p->cancelEventAndWait(&d_reauthHandle);
```
678098: There is a thread race between `AuthenticationClient::stop()` and `AuthenticationClient::handleResponse()`.

`d_reauthHandle` can be possibly modified from 2 separate threads at the same time:
- the mutex in `handleResponse()` doesn't cover `d_reauthHandle` cancellation in `stop()`
- the mutex in `cancelEventAndWait()` (in `stop`) is not locked in `cancelEvent` (in `handleResponse()`)

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,246 @@
...
+        errorDescription << "Authentication failed: "
+                         << authnResponse.status().message()
+                         << " [code: " << authnResponse.status().code() << "]";
+        return rc_AUTHENTICATION_FAILURE;  // RETURN
+    }
+
+    bsl::shared_ptr<bmqio::Channel> channel;
+    {
+        bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
+        channel = d_channel_wp.lock();
```
678098, line 201: The following scenario:

Thread1: AuthenticationClient::handleResponse -> `channel` shared_ptr is taken
Thread2: AuthenticationClient::stop -> `d_reauthHandle` is cancelled, `d_channel_wp` is reset
Thread1: AuthenticationClient::handleResponse (continues) -> `d_reauthHandle` is scheduled again

This results in `d_reauthHandle` being non-null after `stop`

src/groups/mqb/mqba/mqba_authenticationclient.t.cpp
```diff
@@ -0,0 +1,731 @@
...
+    bsls::TimeInterval monotonicClock() { return d_timeSource.now(); }
+    bsls::Types::Int64 highResTimer()
+    {
+        return d_timeSource.now().totalNanoseconds();
+    }
+};
+
+struct SuccessCredentialCb {
+    const char* d_mechanism;
+    const char* d_payload;
```
678098:
```suggestion
    const bsl::string_view d_mechanism;
    const bsl::string_view d_payload;
```

src/groups/mqb/mqba/mqba_authenticationclient.t.cpp
```diff
@@ -0,0 +1,731 @@
...
+
+    SuccessCredentialCb(const char* mechanism, const char* payload)
```
678098:
```suggestion
    SuccessCredentialCb(bsl::string_view mechanism, bsl::string_view payload)
```

src/groups/mqb/mqba/mqba_authenticationclient.t.cpp
```diff
@@ -0,0 +1,731 @@
...
+    : d_mechanism(mechanism)
+    , d_payload(payload)
+    {
+    }
+
+    bsl::optional<mqbplug::AuthnCredential> operator()(bsl::ostream&) const
+    {
+        bslma::Allocator* alloc = bmqtst::TestHelperUtil::allocator();
+        bsl::vector<char> data(d_payload,
+                               d_payload + bsl::strlen(d_payload),
```
678098:
```suggestion
        bsl::vector<char> data(d_payload.begin(),
                               d_payload.end(),
```
Safer with `bsl::string_view` due to no explicit pointer arithmetics

src/groups/mqb/mqba/mqba_authenticationclient.t.cpp
```diff
@@ -0,0 +1,731 @@
...
+        msg.makeAuthenticationResponse();
+    resp.status().category() = bmqp_ctrlmsg::StatusCategory::E_SUCCESS;
+    resp.status().code()     = 0;
+    resp.status().message()  = "OK";
+    resp.lifetimeMs()        = lifetimeMs;
+    return msg;
+}
+
+bmqp_ctrlmsg::AuthenticationMessage
+makeFailureResponse(const bsl::string& message, int code)
```
678098:
```suggestion
makeFailureResponse(bsl::string_view message, int code)
```

src/groups/mqb/mqba/mqba_authenticationclient.t.cpp
```diff
@@ -0,0 +1,731 @@
...
+// ------------------------------------------------------------------------
+{
+    bmqtst::TestHelper::printTestName("AUTHENTICATE SUCCESS");
+
+    bslma::Allocator* alloc = bmqtst::TestHelperUtil::allocator();
+    TestBench         tb(alloc);
+
+    // 1)
+    bslma::ManagedPtr<mqba::AuthenticationClient> client = tb.createClient(
+        SuccessCredentialCb("BASIC", "user:pass"));
```
678098: `BASIC`, `user:pass` is used 11 times, might be worth having a `bsl::string_view` constants

678098: Alternatively these might be provided as default args to `SuccessCredentialCb` constructor

src/groups/mqb/mqba/mqba_authenticationclient.t.cpp
```diff
@@ -0,0 +1,731 @@
...
+    // 2)
+    tb.d_channel.reset();
+
+    // 3)
+    bmqu::MemOutStream errStream(alloc);
+    int                rc = client->authenticate(errStream);
+    BMQTST_ASSERT_EQ(rc, -1);
+    BMQTST_ASSERT(bsl::string(errStream.str(), alloc)
+                      .find("Channel is no longer available") !=
+                  bsl::string::npos);
```
678098: What about this here and in other places:
```suggestion
    BMQTST_ASSERT(errStream.str().starts_with("Channel is no longer available"));
```

- no npos mentioned
- no intermediate string construction
- less code

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,247 @@
...
+    const bmqp_ctrlmsg::AuthenticationMessage& response)
+{
+    // executed by the *IO* thread
+
+    enum RcEnum {
+        rc_SUCCESS                = 0,
+        rc_INVALID_AUTHN_RESPONSE = -1,
+        rc_AUTHENTICATION_FAILURE = -2
+    };
+
```
678098, line 173: What do you think about having a larger scope for `bmqu::GateKeeper`?
Note that `GateKeeper` is relatively cheap: a few atomic operations only.
There is no harm in acquiring it every time.
This makes semantics simpler: `if destructor was called, we must finish the entire function scope and don't start it again`
Placing the gate early prevents from accessing any fields that might be in the middle of destruction.
```suggestion
    // Drop the response if destructor has been called.
    // The destructor will wait for the rest of this scope to finish.
    bmqu::GateKeeper::Status gateStatus(d_gateKeeper);
    if (!gateStatus.isOpen()) {
        errorDescription << "AuthenticationClient " << this << " is stopping, ignoring response";
        return rc_CLOSING;  // RETURN
    }


```

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -0,0 +1,247 @@
...
+    d_gateKeeper.open();
+}
+
+AuthenticationClient::~AuthenticationClient()
+{
+    d_gateKeeper.close();
+    d_scheduler_p->cancelEventAndWait(&d_reauthHandle);
+
+    bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);
+    d_channel_wp.reset();
```
678098, line 87:
```suggestion
```
Placing `AtomicGate` in the beginning of `AuthenticationClient::authenticate` ensures that after closing no other thread is not accessing `d_channel_wp`: 1. no need to explicitly clear it 2. it is cleared anyway as the destructor finishes its work

678098 (commented):
have one last proposal

678098 (approved):
Looks good

PR #1452, opened 2026-05-27
Perf[mqbs::FileStore::aliasMessage]: move shared_ptrs on last use

678098:
Avoid unnecessary atomic reference-count increments/decrements in `aliasMessage` by moving `shared_ptr` and `BlobBuffer` objects that are at their last use point. This eliminates up to 5 redundant atomic operations per call on a hot path.

Saves some time on subscriptions evaluation path

Screenshot 2026-05-27 at 13 53 45


dorjesinpo (approved):
thank you

PR #1456, opened 2026-05-28
Perf[mqbs]: remove aliased buffer deleters pool

678098:
Replace `AliasedBufferDeleterSpPool` with a single `shared_ptr` per `mqbs::FileSet`.
As a result, we don't modify atomics for multiple `shared_ptr`s and do not perform object pool operations. Instead, we mostly modify atomics for 1 `shared_ptr` (associated with the current `FileStore`)

# Changes

- The pool of `AliasedBufferDeleter` objects and its associated struct are removed.
- Instead, each FileSet holds a `d_aliasedChunk_sp` token — a `shared_ptr<FileSet>` with a custom deleter that calls `gc()`.
- All aliased `shared_ptr<char>` instances are created via the aliasing constructor from this token.
- The token's `use_count()` replaces `d_aliasedBlobBufferCount` for tracking outstanding references, and the custom deleter handles deferred `gc` when the last alias dies.
- `d_inlineGc` flag prevents the deleter from firing after inline `gc` has already handled cleanup.

# Benchmarks

**Before**

Screenshot 2026-05-28 at 14 04 05

**After**

Screenshot 2026-05-28 at 14 57 23

This PR saves at least ~1.5% of all CPU frames spent on `SharedObjectPool::getObject` for `AliasedBufferDeleters`, also saves some time on returning elements to this pool.

It removes ~1% of all CPU frames spent on other calls related to `AliasedBufferDeleter` (saving ~2.5% in total).
All this is gone (1.5% on the top is `SharedObjectPool::getObject`):

Screenshot 2026-05-28 at 15 25 11

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_fileset.h
+++ b/src/groups/mqb/mqbs/mqbs_fileset.h
@@ -32,6 +32,7 @@
 #include <mqbu_storagekey.h>
 
 // BDE
+#include <bsl_memory.h>
 #include <bsl_string.h>
 #include <bslma_allocator.h>
 #include <bslma_usesbslmaallocator.h>
@@ -87,12 +88,18 @@ struct FileSet BSLS_CPP11_FINAL {
 
     bool d_fileSetRolloverPolicyAlarm;
 
-    bsls::AtomicInt64 d_aliasedBlobBufferCount;
-    // Number of payload blob buffers
-    // referencing to the mapped data file
-    // of this file set *PLUS* one.  See
-    // notes in ctor of this component for
-    // more details.
+    /// `true` if this FileSet is garbage-collected on rollover, `false` if
+    /// GC is expected on the last alias destruction.
+    bsls::AtomicBool d_inlineGc;
+
+    /// The shared alias used to keep track of the current number of records
+    /// that still reference this FileSet.  FileSet cannot be safely GCed
+    /// unless all references are gone.
+    bsl::shared_ptr<FileSet> d_aliasedChunk_sp;
+
+    /// A weak_ptr pointing at the aliased chunk.
+    /// Guaranteed to be never reset: safe to get `use_count`.
+    bsl::weak_ptr<FileSet> d_aliasedChunk_wp;
 
     bslma::Allocator* d_allocator_p;
 
@@ -107,6 +114,11 @@ struct FileSet BSLS_CPP11_FINAL {
 
     // CREATORS
     FileSet(FileStore* store, bslma::Allocator* allocator);
+
+    // ACCESSORS
+    /// Return the snapshot of the remaining number of references to this
+    /// FileSet.
+    long numReferences() const { return d_aliasedChunk_wp.use_count(); }
 };
 
 // ============================================================================
@@ -135,20 +147,12 @@ inline FileSet::FileSet(FileStore* store, bslma::Allocator* allocator)
 , d_outstandingBytesQlist(0)
 , d_journalFileAvailable(true)
 , d_fileSetRolloverPolicyAlarm(false)
-, d_aliasedBlobBufferCount(1)  // See note below explaining value of '1'
+, d_inlineGc(false)
+, d_aliasedChunk_sp()
+, d_aliasedChunk_wp()
 , d_allocator_p(allocator)
 {
     BSLS_ASSERT(allocator);
-
-    // The reason 'd_aliasedBlobBufferCount' is initialized with 1 instead of 0
-    // is that we don't want the gc logic to kick in for the active file set
-    // everytime its count goes to zero.  While there is a check in gc logic
-    // which does not gc the file set if its active one, its still expensive to
-    // invoke it (in the dispatcher thread) everytime count goes to 0.  In the
-    // worst case, the count can go to 0 after everytime a message is deleted
-    // from the store.  Here's how: assume that message flow is such that there
-    // is only one outstanding message at any time, which is deleted after
-    // consumer confirms it.
 }
 
 }  // close package namespace
--- a/src/groups/mqb/mqbs/mqbs_filestore.cpp
+++ b/src/groups/mqb/mqbs/mqbs_filestore.cpp
@@ -297,39 +297,6 @@ void printNextEventRecord(bsl::ostream&                       stream,
 
 }  // close unnamed namespace
 
-// -------------------------------------
-// struct FileStore_AliasedBufferDeleter
-// -------------------------------------
-
-// CREATORS
-FileStore_AliasedBufferDeleter::FileStore_AliasedBufferDeleter()
-: d_fileSet_p(0)
-{
-    // NOTHING
-}
-
-// MANIPULATORS
-void FileStore_AliasedBufferDeleter::setFileSet(FileSet* fileSet)
-{
-    d_fileSet_p = fileSet;
-    ++d_fileSet_p->d_aliasedBlobBufferCount;
-}
-
-void FileStore_AliasedBufferDeleter::reset()
-{
-    // executed by *ANY* thread
-
-    // PRECONDITIONS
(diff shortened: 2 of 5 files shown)


src/groups/mqb/mqbs/mqbs_fileset.h
```diff
@@ -135,20 +133,11 @@ inline FileSet::FileSet(FileStore* store, bslma::Allocator* allocator)
 , d_outstandingBytesQlist(0)
 , d_journalFileAvailable(true)
 , d_fileSetRolloverPolicyAlarm(false)
-, d_aliasedBlobBufferCount(1)  // See note below explaining value of '1'
```
678098, line 138: `bool` atomic `d_inlineGc` handles special cases when decrement might reach 0. I would prefer to refactor the code to get rid of the `bool` atomic too, but I want to keep the changes minimal and as safe as possible

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -2694,14 +2666,22 @@ int FileStore::create(FileSetSp* fileSetSp)
...
+                                   d_config,
+                                   partitionDesc(),
+                                   d_qListAware,
+                                   d_allocator_p);
+    if (0 == rc) {
+        FileSet* fs = fileSetSp->get();
+        fs->d_aliasedChunk_sp.reset(
+            fs,
+            bdlf::BindUtil::bind(&FileStore::gc, this, fs));
+    }
```
678098, line 2700: We set up `d_aliasedChunk_sp` in both places where we construct a new `FileSet`. There is no time in between when `FileSet` can be used with unset `d_aliasedChunk_sp`

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -2952,7 +2926,9 @@ int FileStore::rolloverImpl(bsls::Types::Uint64 timestamp)
                            "ROLLOVER - STEP 2 (TRUNCATE)");
     }
 
-    if (0 == --activeFileSet->d_aliasedBlobBufferCount) {
+    if (activeFileSet->d_aliasedChunk_sp.unique()) {
+        activeFileSet->d_inlineGc = true;
```
678098, line 2949: In this branch, we perform inline gc, and we don't even care anymore if `d_aliasedChunk_sp` custom destructor is called or not: `d_inlineGc` disables it. We don't even have to call reset for this `shared_ptr`.

src/groups/mqb/mqbs/mqbs_fileset.h
```diff
@@ -87,12 +88,9 @@ struct FileSet BSLS_CPP11_FINAL {
 
     bool d_fileSetRolloverPolicyAlarm;
 
-    bsls::AtomicInt64 d_aliasedBlobBufferCount;
-    // Number of payload blob buffers
-    // referencing to the mapped data file
-    // of this file set *PLUS* one.  See
-    // notes in ctor of this component for
-    // more details.
+    bsls::AtomicBool d_inlineGc;
```
kaikulimu, line 93: Docs

src/groups/mqb/mqbs/mqbs_fileset.h
```diff
@@ -87,12 +88,9 @@ struct FileSet BSLS_CPP11_FINAL {
...
+
+    bsl::shared_ptr<FileSet> d_aliasedChunk_sp;
```
kaikulimu, line 98: Docs

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -5417,7 +5383,9 @@ int FileStore::close(bool flush, bool archive)
     // Failure to truncate is non-fatal.  No need to check rc.
     truncate(activeFileSet);
 
-    if (0 == --activeFileSet->d_aliasedBlobBufferCount) {
+    if (activeFileSet->d_aliasedChunk_sp.unique()) {
```
kaikulimu: https://github.com/bloomberg/bde/blob/69b40730210662c30d5cbd0d8997a920016573f0/groups/bsl/bslstl/bslstl_sharedptr.h#L2909-L2916
BDE warns against calling `unique()`. Is there a safer alternative?

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -2952,7 +2925,9 @@ int FileStore::rolloverImpl(bsls::Types::Uint64 timestamp)
                            "ROLLOVER - STEP 2 (TRUNCATE)");
     }
 
-    if (0 == --activeFileSet->d_aliasedBlobBufferCount) {
+    if (activeFileSet->d_aliasedChunk_sp.unique()) {
```
kaikulimu: BDE [warns against](https://github.com/bloomberg/bde/blob/69b40730210662c30d5cbd0d8997a920016573f0/groups/bsl/bslstl/bslstl_sharedptr.h#L2909-L2916) calling `unique()`. Is there a safer alternative?

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -2694,14 +2665,22 @@ int FileStore::create(FileSetSp* fileSetSp)
...
+                                   d_config.partitionId(),
+                                   d_config,
+                                   partitionDesc(),
+                                   d_qListAware,
+                                   d_allocator_p);
+    if (0 == rc) {
+        FileSet* fs = fileSetSp->get();
+        fs->d_aliasedChunk_sp.reset(
+            fs,
+            bdlf::BindUtil::bind(&FileStore::gc, this, fs));
```
kaikulimu, line 2698: Should we move this block of code inside `FileStoreUtil::create`?

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -3000,8 +2976,8 @@ int FileStore::rolloverImpl(bsls::Types::Uint64 timestamp)
         BALL_LOG_OUTPUT_STREAM << partitionDesc()
                                << "All data files and alias counts: \n";
         for (unsigned int index = 0; index < d_fileSets.size(); ++index) {
-            const FileSet*     fs    = d_fileSets[index].get();
-            bsls::Types::Int64 count = fs->d_aliasedBlobBufferCount;
+            const FileSet* fs    = d_fileSets[index].get();
+            long           count = fs->d_aliasedChunk_sp.use_count();
```
kaikulimu: Should this be `fs->d_aliasedChunk_sp.use_count()` or `fs->d_aliasedChunk_sp.use_count() - 1`?

kaikulimu (changes requested):
Reviewed

src/groups/mqb/mqbs/mqbs_filestore.cpp
678098: `FileStore::gc` is private. We can move if we provide a `gc` callback to it

src/groups/mqb/mqbs/mqbs_filestore.cpp
678098: `unique` is good enough, but it can be replaced with `use_count`. Still, due to the nature of this object, it is just a snapshot that might change, but it will not cause any problems. The same thing with any atomic counter.

src/groups/mqb/mqbs/mqbs_filestore.cpp
678098: Also note that in recovery path we do not call `FileStoreUtil::create`, we build `FileStore` in place, field after field.

src/groups/mqb/mqbs/mqbs_filestore.cpp
678098: The log below handles -1 case

kaikulimu (approved):
LGTM

PR #1457, opened 2026-05-28
Fix[mqbblp]: replace double-throttled LOGTHROTTLE macros with BALL_LOG

### LINKED ISSUE #1451: RequestPermission() wrapping LOGTHROTTLE macros suppresses log output
`BMQ_LOGTHROTTLE_*` macros expand to `BALL_LOGTHROTTLE_*`, which internally constructs a static `bdlmt::Throttle` per call site. When a `requestPermission()` check on a per-instance throttle guards one of these macros, the log message must pass two independent throttle gates before being emitted. The outer throttle controls whether the call site is reached, and the inner throttle independently decides whether to emit. Since neither gate is aware of the other, messages are suppressed more aggressively than either throttle alone intends.

SrinathhSatuluri:
Closes #1451

`BMQ_LOGTHROTTLE_INFO` expands to `BALL_LOGTHROTTLE_INFO` which carries its own internal static `bdlmt::Throttle`. When these macros are called inside a `requestPermission()` guard, the message passes through two independent throttle gates in series. The outer gate may grant permission, but the inner one can still suppress the log entry, effectively silencing output that was already meant to be rate-limited by `requestPermission()` alone.

This replaces `BMQ_LOGTHROTTLE_INFO` with `BALL_LOG_INFO` at four sites across `mqbblp_queuehandle.cpp` and `mqbblp_relayqueueengine.cpp` where `requestPermission()` already controls the throttle. The same files use `requestPermission()` with `BALL_LOG_WARN` correctly at other call sites, confirming this was unintentional.

The `mqbstat_statcontroller.cpp` site was intentionally left unchanged. That call uses `!requestPermission()`, which inverts the semantics. It logs when the throttle denies permission, then rate-limits the warning separately. The two throttles serve distinct purposes there and are not redundant.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_queuehandle.cpp
@@ -996,9 +996,8 @@ void QueueHandle::deliverMessage(
             // needed we way have to change this method to return a tri-value
             // (delivered, non-delivered, skipped).
             if (d_throttledDroppedPutMessages.requestPermission()) {
-                BMQ_LOGTHROTTLE_INFO << "Queue '" << d_queue_sp->description()
-                                     << "' skipping duplicate PUSH "
-                                     << iter.guid();
+                BALL_LOG_INFO << "Queue '" << d_queue_sp->description()
+                              << "' skipping duplicate PUSH " << iter.guid();
             }
             continue;  // CONTINUE
         }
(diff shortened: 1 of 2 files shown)


678098 (approved):
Looks good

678098:
@SrinathhSatuluri thank you for contribution!

PR #1462, opened 2026-05-30
Fix[mqb]: expirePendingMessagesDispatched after d_self.invalidate()

dorjesinpo:
`convertToLocal` destructs `RemoteQueue` which may have `expirePendingMessagesDispatched` scheduled resulting in use after free.
The fix is to make `expirePendingMessagesDispatched` `weakMemFn`


dorjesinpo:
IT failure - https://github.com/bloomberg/blazingmq/pull/1465

678098 (approved):

PR #1463, opened 2026-05-30
Fix[mqb]: handle concurrent Cluster::initiateShutdown

dorjesinpo:
`Cluster::initiateShutdown` can be called by `Cluster::terminate` _and_ by `Application::stop`
Making sure the logic executes only once.
(A symptom was `mqbnet_channel.cpp:129 Assertion failed: rc == 0`)

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_cluster.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_cluster.cpp
@@ -90,9 +90,18 @@ namespace BloombergLP {
 namespace mqbblp {
 
 namespace {
+
 const double k_LOG_SUMMARY_INTERVAL = 60.0 * 5;  // 5 minutes
 
 const double k_QUEUE_GC_INTERVAL = 60.0;  // 1 minute
+
+struct ChainNoOp {
+    template <class T>
+    void operator()(const T& completionCb) const
+    {
+        completionCb();
+    }
+};
 
 }  // close unnamed namespace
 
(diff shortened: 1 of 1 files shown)


dorjesinpo:
IT failure - https://github.com/bloomberg/blazingmq/pull/1466

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -562,6 +571,17 @@ void Cluster::initiateShutdownDispatched(const VoidFunctor& callback)
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(inDispatcherThread());
 
+    if (d_isStopping) {
+        BALL_LOG_INFO << "Cluster: [name: '" << name()
+                      << "'] shutdown already in progress, skipping";
```
678098:
```suggestion
        BALL_LOG_INFO << "Cluster [" << name()
                      << "] shutdown already in progress, skipping";
```
This is how cluster info is logged in other places

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
```diff
@@ -562,6 +571,17 @@ void Cluster::initiateShutdownDispatched(const VoidFunctor& callback)
...
+
+        // Append a no-op link so 'callback' fires when the in-progress
+        // shutdown chain completes.  If the chain has already finished,
+        // the appended link executes immediately.
+        d_shutdownChain.appendInplace(ChainNoOp(), callback);
```
678098, line 581: Is an intermediate functor needed?
Why not using the 1-arg overload:
```
    template <class OP_CALLBACK>
    void appendInplace(BSLS_COMPILERFEATURES_FORWARD_REF(OP_CALLBACK)
                           opCallback);
```

```
d_shutdownChain.appendInplace(callback);
```
?

678098 (changes requested):

src/groups/mqb/mqbblp/mqbblp_cluster.cpp
dorjesinpo: `appendInplace` expects  `opCallback` to take `coCallback` as an argument.
Even if `coCallback` is `noOp`

678098 (approved):

PR #1464, opened 2026-05-30
Fix[mqb]: Concurrent setChannel and resetChannel

dorjesinpo:
An existing connection can get disconnected and then connected in a different IO thread.
In this case, new `setChannel` can be called before old `resetChannel` resulting in the old `resetChannel` resetting  the new connection.
Need to detect stale `resetChannel` (by the channel, for example)

### DIFF
--- a/src/groups/mqb/mqbnet/mqbnet_cluster.h
+++ b/src/groups/mqb/mqbnet/mqbnet_cluster.h
@@ -135,8 +135,12 @@ class ClusterNode {
     /// or if it is already reading.
     virtual bool enableRead() = 0;
 
-    /// Reset the channel associated to this node.
-    virtual ClusterNode* resetChannel() = 0;
+    /// Reset the channel associated to this node.  The specified
+    /// `closedChannel` identifies the channel being closed; if the node
+    /// already has a different (newer) channel, this reset is stale and
+    /// will be ignored.
+    virtual ClusterNode*
+    resetChannel(const bsl::shared_ptr<bmqio::Channel>& closedChannel) = 0;
 
     /// Enqueue the specified message `blob` of the specified `type`to be
     /// written to the channel associated to this node.  Return 0 on
--- a/src/groups/mqb/mqbnet/mqbnet_clusterimp.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_clusterimp.cpp
@@ -123,10 +123,7 @@ bool ClusterNodeImp::enableRead()
 ClusterNode* ClusterNodeImp::resetChannel(
     const bsl::shared_ptr<bmqio::Channel>& closedChannel)
 {
-    d_channel.resetChannel(closedChannel);
-
-    bsl::shared_ptr<bmqio::Channel> currentSp = d_channel.channel();
-    if (currentSp) {
+    if (!d_channel.resetChannel(closedChannel)) {
         // The channel was not reset (stale close), skip cleanup.
         return this;  // RETURN
     }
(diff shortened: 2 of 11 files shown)


dorjesinpo:
@678098 Could you please take a look at `test_journal_full_recovery.py::test_journal_full_purge_recovery[multi_node-eventual_consistency]` failure - `blazingmq.dev.it.process.proc.ProcessExitError: east2: process exited with return code 10`?

The problem seems to be sequence gap at the time of purge due to previous sync point failure because of `PARTITION_READONLY`

dorjesinpo:
`test_restart_between_legacy_and_fsm_purge_queue_app` - https://github.com/bloomberg/blazingmq/pull/1467

src/groups/mqb/mqbnet/mqbnet_cluster.h
```diff
@@ -135,8 +135,12 @@ class ClusterNode {
     /// or if it is already reading.
     virtual bool enableRead() = 0;
 
-    /// Reset the channel associated to this node.
-    virtual ClusterNode* resetChannel() = 0;
+    /// Reset the channel associated to this node.  The specified
+    /// `closedChannel` identifies the channel being closed; if the node
+    /// already has a different (newer) channel, this reset is stale and
+    /// will be ignored.
```
678098, line 141:
```suggestion
    /// `closedChannel` identifies the channel being closed.
```
This is implementation detail. I am sure it should not be mentioned in the interface class. Less sure whether it should also be mentioned in implementation override.

src/groups/mqb/mqbnet/mqbnet_clusterimp.cpp
```diff
@@ -120,9 +120,17 @@ bool ClusterNodeImp::enableRead()
...
 
-ClusterNode* ClusterNodeImp::resetChannel()
+ClusterNode* ClusterNodeImp::resetChannel(
+    const bsl::shared_ptr<bmqio::Channel>& closedChannel)
 {
-    d_channel.resetChannel();
+    d_channel.resetChannel(closedChannel);
+
+    bsl::shared_ptr<bmqio::Channel> currentSp = d_channel.channel();
+    if (currentSp) {
```
678098:
```suggestion
    if (!d_channel.resetChannel(closedChannel)) {
```
What about returning `bool` on reset failure?
This avoids making intermediate useless copies of the stored shared_ptr and avoids locking the mutex again on `d_channel.channel()` call

src/groups/mqb/mqbnet/mqbnet_mockcluster.cpp
```diff
@@ -103,9 +103,15 @@ bool MockClusterNode::enableRead()
...
 
-ClusterNode* MockClusterNode::resetChannel()
+ClusterNode* MockClusterNode::resetChannel(
+    const bsl::shared_ptr<bmqio::Channel>& closedChannel)
 {
-    d_channel.resetChannel();
+    d_channel.resetChannel(closedChannel);
+
+    bsl::shared_ptr<bmqio::Channel> currentSp = d_channel.channel();
+    if (currentSp) {
```
678098:
```suggestion
    if (!d_channel.resetChannel(closedChannel)) {
```
`bool` return also works here fine

678098 (commented):
Might simplify calls by removing shared_ptr copies and mutex lock

src/groups/mqb/mqbnet/mqbnet_cluster.h
dorjesinpo: Well, the `bool` change does need some explanation.
Identifying and ignoring stale reset deserves mentioning in the contract.

678098 (approved):
Looks good

PR #1467, opened 2026-05-31
Fix[test]: post must match queue in the ACK

dorjesinpo:
client `post` must match the queue when waiting for ACK.
otherwise, it can capture previous ACK (from a batch) and that breaks synchronization in tests


dorjesinpo:
IT - https://github.com/bloomberg/blazingmq/pull/1471

jll63 (approved):

PR #1468, opened 2026-06-01
Fix[mqb]: release event handlers before dtors

dorjesinpo:
Need to cancel `QueueConsumptionMonitor::d_alarmEventHandle` before destructing.
Otherwise, if `maxIdleTimeSec` is long enough, `BSLS_ASSERT_SAFE(!d_alarmEventHandle)` fires in dtor.

Similar with `RootQueueEngine::d_flowControlEventHandle` (not used now).

Also, handle the extra hop not covered by `cancelEventAndWait` in a generic way.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_domain.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_domain.cpp
@@ -265,7 +265,7 @@ Domain::Domain(const bsl::string&                     name,
 , d_teardownCb()
 , d_teardownRemoveCb()
 , d_mutex()
-, d_numQueues(0)
+, d_numQueues(1)
 {
     if (d_cluster_sp->isRemote()) {
         // In a remote domain, we don't care about monitoring, so disable it
@@ -409,8 +409,6 @@ void Domain::teardown(const mqbi::Domain::TeardownCb& teardownCb)
         BALL_LOG_INFO << "Stopping domain '" << d_name << "' having "
                       << d_queues.size() << " registered queues.";
 
-        d_numQueues.add(1);
-
         d_teardownCb = teardownCb;
         d_state      = e_STOPPING;
 
@@ -434,8 +432,6 @@ void Domain::teardownRemove(const TeardownCb& teardownCb)
 
         BALL_LOG_INFO << "Removing domain '" << d_name << "' having "
                       << d_queues.size() << " registered queues.";
-
-        d_numQueues.add(1);
 
         d_teardownRemoveCb = teardownCb;
 
--- a/src/groups/mqb/mqbblp/mqbblp_domain.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_domain.cpp
@@ -55,21 +55,9 @@ namespace BloombergLP {
 namespace mqbblp {
 
 namespace {
-const char k_LOG_CATEGORY[] = "MQBBLP.DOMAIN";
 
 const char k_NODE_IS_STOPPING[]              = "Node is stopping";
 const char k_DOMAIN_IS_REMOVING_OR_REMOVED[] = "Domain is removing or removed";
-
-/// Validates an application subscription.
-/// This method does nothing.. it's just used so that we can control the
-/// destruction of the specified `queue` to happen once we guarantee the
-/// associated Dispatcher's queue has been drained and flushed.
-void queueHolderDummy(const bsl::shared_ptr<mqbi::Queue>& queue)
-{
-    BALL_LOG_SET_CATEGORY(k_LOG_CATEGORY);
-
-    BALL_LOG_INFO << "Deleted queue '" << queue->uri().canonical() << "'";
-}
 
 bool validateSubscriptionExpression(bsl::ostream& errorDescription,
                                     const mqbconfm::Expression& expression,
@@ -586,7 +574,10 @@ void Domain::unregisterQueue(mqbi::Queue* queue)
     bsl::shared_ptr<mqbi::Queue> queueSp(it->second);
     d_queues.erase(it);
 
-    queueSp->close(bdlf::BindUtil::bind(&queueHolderDummy, queueSp));
+    bslmt::Latch latch(1);
+    queueSp->close(bdlf::BindUtil::bind(&bslmt::Latch::arrive, &latch));
+
+    latch.wait();
 
     BALL_LOG_INFO << "Unregistered queue from domain '" << d_name
                   << "' [canonicalURI: " << queueSp->uri().canonical() << "]. "
--- a/src/groups/mqb/mqbblp/mqbblp_domain.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_domain.cpp
@@ -59,6 +59,7 @@ namespace {
 const char k_NODE_IS_STOPPING[]              = "Node is stopping";
 const char k_DOMAIN_IS_REMOVING_OR_REMOVED[] = "Domain is removing or removed";
 
+/// Validates an application subscription.
 bool validateSubscriptionExpression(bsl::ostream& errorDescription,
                                     const mqbconfm::Expression& expression,
                                     bslma::Allocator*           allocator)
@@ -264,6 +265,7 @@ Domain::Domain(const bsl::string&                     name,
 , d_teardownCb()
 , d_teardownRemoveCb()
 , d_mutex()
+, d_numQueues(0)
 {
     if (d_cluster_sp->isRemote()) {
         // In a remote domain, we don't care about monitoring, so disable it
@@ -400,22 +402,24 @@ void Domain::teardown(const mqbi::Domain::TeardownCb& teardownCb)
     // is atomic.  Otherwise, there is a chance due to race that 'teardownCb'
     // can be executed more than once.
 
-    bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);  // d_mutex LOCKED
+    QueueMap queues(d_allocator_p);
+    {
+        bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);  // d_mutex LOCKED
 
-    BALL_LOG_INFO << "Stopping domain '" << d_name << "' having "
-                  << d_queues.size() << " registered queues.";
+        BALL_LOG_INFO << "Stopping domain '" << d_name << "' having "
+                      << d_queues.size() << " registered queues.";
 
-    d_teardownCb = teardownCb;
-    d_state      = e_STOPPING;
+        d_numQueues.add(1);
 
-    if (d_queues.empty()) {
-        d_teardownCb(d_name);
-        d_teardownCb = bsl::nullptr_t();
-        d_state      = e_STOPPED;
-        return;  // RETURN
+        d_teardownCb = teardownCb;
+        d_state      = e_STOPPING;
+
+        queues.swap(d_queues);
     }
 
-    closeAllQueues();
+    closeAllQueues(queues);
+
+    subtractQueueCount();
 }
 
 void Domain::teardownRemove(const TeardownCb& teardownCb)
@@ -424,21 +428,23 @@ void Domain::teardownRemove(const TeardownCb& teardownCb)
     BSLS_ASSERT_SAFE(!d_teardownRemoveCb);
     BSLS_ASSERT_SAFE(teardownCb);
 
-    bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);  // d_mutex LOCKED
+    QueueMap queues(d_allocator_p);
+    {
+        bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);  // d_mutex LOCKED
 
-    BALL_LOG_INFO << "Removing domain '" << d_name << "' having "
-                  << d_queues.size() << " registered queues.";
+        BALL_LOG_INFO << "Removing domain '" << d_name << "' having "
+                      << d_queues.size() << " registered queues.";
 
-    d_teardownRemoveCb = teardownCb;
+        d_numQueues.add(1);
 
-    if (d_queues.empty()) {
-        d_teardownRemoveCb(d_name);
-        d_teardownRemoveCb = bsl::nullptr_t();
-        d_state            = e_STOPPED;
-        return;  // RETURN
+        d_teardownRemoveCb = teardownCb;
+
+        queues.swap(d_queues);
     }
 
-    closeAllQueues();
+    closeAllQueues(queues);
+
+    subtractQueueCount();
 }
 
 void Domain::openQueue(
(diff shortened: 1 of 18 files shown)


dorjesinpo:
IT - https://github.com/bloomberg/blazingmq/pull/1466

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -1276,7 +1276,7 @@ bmqt::GenericResult::Enum ClusterQueueHelper::sendReopenQueueRequest(
             << "ReopenQueue request: " << request->request() << ", rc: " << rc
             << ".";
 
-        setAsClosed(subQueueContext);
+        setStreamState(subQueueContext, SubQueueContext::k_CLOSED);
```
678098: These changes doesn't belong to this PR, they belong to
https://github.com/bloomberg/blazingmq/pull/1469

src/groups/mqb/mqbblp/mqbblp_rootqueueengine.cpp
```diff
@@ -556,24 +556,29 @@ int RootQueueEngine::initializeAppId(const bsl::string& appId,
     return 0;
 }
 
+void RootQueueEngine::close()
+{
+    if (d_flowControlEventHandle) {
+        d_scheduler_p->cancelEventAndWait(&d_flowControlEventHandle);
+    }
```
```suggestion
    d_scheduler_p->cancelEventAndWait(&d_flowControlEventHandle);
```
Not necessary, null event is handled there

src/groups/mqb/mqbmock/mqbmock_queue.h
```diff
@@ -230,10 +230,12 @@ class Queue : public mqbi::Queue {
     void dropHandle(mqbi::QueueHandle* handle,
                     bool doDeconfigure = true) BSLS_KEYWORD_OVERRIDE;
 
-    /// Close this queue.
+    /// Close this queue.  Call the specified `callback` when the queue does
+    /// not have any outstanding events.
```
678098:
```suggestion
    /// Close this queue.  Call the specified `callback` on close complete.
```

I don't see outstanding events in implementation, not sure what this comment means then. Anyways, better to avoid exposing concrete implementation details. Implementation might easily change one day, but the comment will stay incorrect.

src/groups/mqb/mqbblp/mqbblp_queue.cpp
```diff
@@ -354,6 +356,18 @@ void Queue::closeDispatched()
     }
 
     updateStats();
+
+    // Enqueue the holder AFTER 'engine->close()'.  Any two-hop callbacks
```
678098, line 358: Don't fully like the idea of complex relations between components and making 2 jumps just to close something. I think the code is fragile due to this, too easy to have it wrong. Don't have other solution right now though

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -592,21 +575,10 @@ void Domain::unregisterQueue(mqbi::Queue* queue)
...
-    //
-    // This code relies on queue being in an 'empty' state meaning it will NOT
-    // schedule any new dispatcher work.
-    d_dispatcher_p->execute(bdlf::BindUtil::bind(&queueHolderDummy, queueSp),
-                            queueSp.get(),
-                            mqbi::DispatcherEventType::e_DISPATCHER);
+    bslmt::Latch latch(1);
+    queueSp->close(bdlf::BindUtil::bind(&bslmt::Latch::arrive, &latch));
+
+    latch.wait();
```
678098: This `latch.wait()` means that the cluster dispatcher is fully blocked and non-operable until queue dispatcher thread finishes processing its queue until it gets to newly enqueued `closeDispatched`. And even this is not the end: cluster thread then waits for queue dispatcher thread to process all new events until it gets to the dispatcher Dispatcher event that contains the callback that calls `latch.arrive()`.

Unregistering queues will now block cluster thread for a long time. Data is also going through cluster thread, this means that data delivery will also be blocked now.

Another problem: if queue dispatcher thread decides to synchronize with cluser thread, they might block each other.

678098 (changes requested):

src/groups/mqb/mqbmock/mqbmock_queue.h
dorjesinpo: An example of outstanding event is `alarmEventDispatched` or `onFlowControlTimer` scheduled _after_ `cancelEventAndWait`.
This is important part of `close` contract - there must be no events.  I.e. the queue can be destructed.

src/groups/mqb/mqbblp/mqbblp_queue.cpp
dorjesinpo: Another way to work with `cancelEventAndWait` and dispatcher is `weakMemFn`

src/groups/mqb/mqbblp/mqbblp_domain.cpp
dorjesinpo: `synchronize` is not better, indeed.  We could make `unregisterQueue` async.  Check out the change
(`closeAllQueues` is not a concern, we do want to block there)

678098 (dismissed):

src/groups/mqb/mqbblp/mqbblp_domain.cpp
678098: `closeAllQueues` is totally fine to block, it is only executed on shutdown.
The current async `unregisterQueue` also looks good, it doesn't block data processing.
Approved

dorjesinpo:
`test_restart_between_Legacy_and_FSM_basic[single_node-strong_consistency-to_fsm/from_legacy]` - tentatively https://github.com/bloomberg/blazingmq/pull/1468

src/groups/mqb/mqbblp/mqbblp_domain.cpp
dorjesinpo: @678098 Not a good idea to async `unregisterQueue`. 
`Queue::close` is already async and we may end up stopping dispatcher before `Queue::closeDispatched`
Blocking `unregisterQueue` seems as a safer option.  It cannot take long time - if queue is unregistering, it does not have clients and therefore cannot have traffic.

678098: If dispatcher stop is the thing that requires extra attention, we can keep the number of references to the dispatched `unregisterQueue` async calls so we will not try to stop the dispatcher if there are any in-flight queue unregisters.

Blocking `unregisterQueue` definitely can take a long time, because it synchronizes not with one queue that doesn't have clients, it synchronizes with the entire queue dispatcher queue, meaning, the entire partition and all queues in that partition.

dorjesinpo: I don't share the concern about blocking unregisterQueue but let's try something else.  Check out new commit...

678098: [Tests / IT [single/legacy_mode/strong_consistency] (pull_request)](https://github.com/bloomberg/blazingmq/actions/runs/26907830641/job/79382340020?pr=1468)
Tests / IT [single/legacy_mode/strong_consistency] (pull_request)Failing after 4m

@dorjesinpo Test fails due to dispatcher event callback execution after MQTP has stopped

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -573,60 +581,63 @@ void Domain::unregisterQueue(mqbi::Queue* queue)
...
+            if (d_teardownCb) {
+                teardownCb   = d_teardownCb;
+                d_teardownCb = bsl::nullptr_t();
+                d_state      = e_STOPPED;
+            }
+            if (d_teardownRemoveCb) {
+                teardownRemoveCb   = d_teardownRemoveCb;
+                d_teardownRemoveCb = bsl::nullptr_t();
+                d_state            = e_STOPPED;
+            }
```
678098:
```suggestion
            bsl::swap(teardownCb, d_teardownCb);
            bsl::swap(teardownRemoveCb, d_teardownRemoveCb);
            d_state = e_STOPPED;
```

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -573,60 +581,63 @@ void Domain::unregisterQueue(mqbi::Queue* queue)
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(d_cluster_sp->inDispatcherThread());
 
-    bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);  // LOCK
+    bsl::shared_ptr<mqbi::Queue> queueSp;
+    mqbi::Domain::TeardownCb     teardownCb;        // no-op
+    mqbi::Domain::TeardownCb     teardownRemoveCb;  // no-op
+    size_t                       count = 0;
```
678098, line 563:
```suggestion
```

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -573,60 +581,63 @@ void Domain::unregisterQueue(mqbi::Queue* queue)
...
+        d_queues.erase(it);
+
+        // This code relies on one-way 'd_teardownCb' and 'd_teardownRemoveCb'.
+        // Call them when all _existing_ queues are gone, new queues are not a
+        // concern.
+        count = d_queues.size();
+
+        // Refer to note in 'teardown' routine to see why 'd_state' is updated
+        // while 'd_mutex' is acquired.
+        if (count == 0) {
```
678098:
```suggestion
        if (d_queues.empty()) {
```

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -573,60 +581,63 @@ void Domain::unregisterQueue(mqbi::Queue* queue)
...
 
-    bsl::shared_ptr<mqbi::Queue> queueSp(it->second);
-    d_queues.erase(it);
+        queueSp = bslmf::MovableRefUtil::move(it->second);
+        d_queues.erase(it);
+
+        // This code relies on one-way 'd_teardownCb' and 'd_teardownRemoveCb'.
+        // Call them when all _existing_ queues are gone, new queues are not a
+        // concern.
+        count = d_queues.size();
```
678098, line 581:
```suggestion
```

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -573,60 +581,63 @@ void Domain::unregisterQueue(mqbi::Queue* queue)
...
-            d_teardownCb = bsl::nullptr_t();
-            d_state      = e_STOPPED;
-        }
-        if (d_teardownRemoveCb) {
-            d_teardownRemoveCb(d_name);
-            d_teardownRemoveCb = bsl::nullptr_t();
-            d_state            = e_STOPPED;
-        }
-    }
+                  << count;
```
678098, line 593:
```suggestion
                  << d_queues.size();
```

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -412,25 +402,24 @@ void Domain::teardown(const mqbi::Domain::TeardownCb& teardownCb)
...
+        bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);  // d_mutex LOCKED
 
-    BALL_LOG_INFO << "Stopping domain '" << d_name << "' having "
-                  << d_queues.size() << " registered queues.";
+        BALL_LOG_INFO << "Stopping domain '" << d_name << "' having "
+                      << d_queues.size() << " registered queues.";
 
-    d_teardownCb = teardownCb;
-    d_state      = e_STOPPING;
+        d_numQueues.add(1);
```
678098: As far as I understand, you do `add`/`subtract` here to ensure that this thread will call the shutdown callbacks.

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -806,6 +774,54 @@ int Domain::processCommand(mqbcmd::DomainResult*        result,
...
+    subtractQueueCount();
+}
+
+void Domain::subtractQueueCount()
+{
+    bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);  // d_mutex LOCKED
+
+    --d_numQueues;
+
+    if (d_numQueues == 0) {
```
678098, line 799: The situation:
Queue A is registered, and unregistered in its domain D.
Domain D either has 0 or 1 queues. This domain will be continuously constructed/destructed, right?

To avoid this, `d_numQueues` needs to be initialized with 1, so domain keeps self-reference. This self-reference should be released when domain is stopped.

src/groups/mqb/mqbblp/mqbblp_domain.cpp
dorjesinpo: this code is outdated.

src/groups/mqb/mqbblp/mqbblp_domain.cpp
dorjesinpo: this code is outdated.

src/groups/mqb/mqbblp/mqbblp_domain.cpp
dorjesinpo: cannot access `d_queues` outside the lock

src/groups/mqb/mqbblp/mqbblp_domain.cpp
dorjesinpo: this is to skip checking if `d_numQueues == 0` outside `subtractQueueCount`

src/groups/mqb/mqbblp/mqbblp_domain.cpp
dorjesinpo: `d_numQueues == 0` without `tearDown` is no-op.

And `tearDown` is one-way.  Once called, there is no way back.

Initializing with `1` does not hurt though

src/groups/mqb/mqbblp/mqbblp_domain.cpp
```diff
@@ -412,25 +402,22 @@ void Domain::teardown(const mqbi::Domain::TeardownCb& teardownCb)
...
+        queues.swap(d_queues);
     }
 
-    for (QueueMap::iterator it = d_queues.begin(); it != d_queues.end();
-         ++it) {
-        it->second->close();
-    }
+    closeAllQueues(queues);
+
+    subtractQueueCount();
```
678098, line 420:
```suggestion
    // Release reference kept on initialization of `d_numQueues(1)`
    subtractQueueCount();
```

678098 (approved):

PR #1470, opened 2026-06-01
Fix[mqbc::StorageMgr]: Validate ReplicaDataRequestDrop

kaikulimu:
Validate partition Id and node primaryship instead of asserting

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_storagemanager.cpp
+++ b/src/groups/mqb/mqbc/mqbc_storagemanager.cpp
@@ -1202,11 +1202,51 @@ void StorageManager::processReplicaDataRequestDrop(
                      bmqp_ctrlmsg::ReplicaDataType::E_DROP);
 
     const int partitionId = replicaDataRequest.partitionId();
-    BSLS_ASSERT_SAFE(0 <= partitionId &&
-                     partitionId < static_cast<int>(d_fileStores.size()));
-    BSLS_ASSERT_SAFE(
-        source->nodeId() ==
-        d_clusterState_p->partitionsInfo().at(partitionId).primaryNodeId());
+    if (partitionId < 0 ||
+        partitionId >= static_cast<int>(d_fileStores.size())) {
+        BALL_LOG_ERROR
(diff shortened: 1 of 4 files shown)


src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -1202,11 +1202,51 @@ void StorageManager::processReplicaDataRequestDrop(
...
+
+        bmqp_ctrlmsg::ControlMessage controlMsg;
+        controlMsg.rId() = message.rId();
+
+        bmqp_ctrlmsg::Status& status = controlMsg.choice().makeStatus();
+        status.category()            = bmqp_ctrlmsg::StatusCategory::E_REFUSED;
+        status.code()                = mqbi::ClusterErrorCode::e_NO_PARTITION;
+        status.message()             = "Invalid partitionId";
+
+        d_clusterData_p->messageTransmitter().sendMessageSafe(controlMsg,
```
pniedzielski, line 1221: What happens on the primary in this case?  It looks like it enqueues `PartitionFSM::Event::e_FAIL_REPLICA_DATA_RSPN_DROP` but nothing happens.  Is that correct?  Do we have a test for that?

pniedzielski: Specifically I'm thinking, the misbehaving/false primary will continue as if everything is A-okay.  Especially for the invalid partition id, I want to make sure that's okay.

The replica should never make us change our view of the world, but will continuing on in this state break things?

kaikulimu: Will write the following UT: 
1. Replica starts in FSM mode
2. It goes to e_REPLICA_HEALING state
3. Receives ReplicaDataRequestDrop with invalid pid. Verify it sends failure response, and stays in e_REPLICA_HEALING
4. A put message comes in. Verify that we do not process the put

pniedzielski (approved):

PR #1476, opened 2026-06-03
Fix[mqb]: notify queue after pending close

dorjesinpo:
`ClusterQueueHelper` calls `RemoteQueue::onOpenUpstream` when it reopens queue so that `RemoteQueue` can retransmit pending PUTs.  But `ClusterQueueHelper` may receive Close queue while reopening in which case it buffers it in `d_pendingCloseRequests`.  When `ClusterQueueHelper` receives reopen response, it calls `onOpenUpstream` and _then_ sends pending close.  Resulting in RemoteQueue retransmitting PUT after close which may fail an assertion, if the close drops writeCount to 0.

Fix - first process pending close requests and if anything left, call `RemoteQueue::onOpenUpstream` (and send Configure request)

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
@@ -857,22 +857,26 @@ void ClusterQueueHelper::onQueueContextAssigned(
     }
 }
 
-void ClusterQueueHelper::finishReopening(QueueContext*        queueContext,
-                                         StreamsMap::iterator sqit)
+void ClusterQueueHelper::finishReopening(
+    QueueContext*                                queueContext,
+    StreamsMap::iterator                         sqit,
+    mqbnet::ClusterNode*                         activeNode,
+    const bsl::shared_ptr<PartitionReopenCycle>& cycle)
 {
     // executed by the cluster *DISPATCHER* thread
 
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(d_cluster_p->inDispatcherThread());
 
+    SubQueueContext&      subQueueContext    = sqit->value();
+    SubQueueContext::Enum state              = subQueueContext.d_state;
+    bool                  isValid            = true;
+    const unsigned int    upstreamSubQueueId = sqit->subId();
+
     bsl::vector<SubQueueContext::PendingClose> pendingClose(d_allocator_p);
-    pendingClose.swap(sqit->value().d_pendingCloseRequests);
-
-    const bool isOpen  = (sqit->value().d_state == SubQueueContext::k_OPEN);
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -910,9 +914,98 @@ void ClusterQueueHelper::finishReopening(QueueContext*        queueContext,
             // 'false' means 'sqit' is deleted (all counters are zeroes)
         }
     }
+    const bsls::Types::Uint64 generationCount = cycle->generationCount();
+
+    if (isValid && state == SubQueueContext::k_OPEN) {
+        // Remain valid after all pending close
+        // Did not erase subQueueContext.
```
678098:
```suggestion
        // Remains valid if subQueueContext was not erased
        // after processing all pending close requests.
```
Find the original comment confusing

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -1677,88 +1758,21 @@ void ClusterQueueHelper::onReopenQueueResponse(
...
+    BSLS_ASSERT_SAFE(appId == sqit->appId());
+
+    BSLS_ASSERT_SAFE(subQueueContext.d_state == SubQueueContext::k_REOPENING);
+
     subQueueContext.d_state = SubQueueContext::k_OPEN;
 
     BMQ_LOGTHROTTLE_INFO
         << d_cluster_p->description() << ": queue successfully reopened ["
         << requestContext->request()
         << "]. Attempting to send a configure-queue request now.";
```
678098:
```suggestion
        << "].";
```
Should not log `attempting here` if we are not attempting it here

678098 (changes requested):
Merge conflicts

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -1691,88 +1772,21 @@ void ClusterQueueHelper::onReopenQueueResponse(
...
+    setStreamState(&subQueueContext, SubQueueContext::k_OPEN);
 
-        // Abort restore of the state: the channel is no longer valid, we'll
-        // wait for a new one to be active and will restart restoring the state
-        // from the beginning.
-    }
+    BMQ_LOGTHROTTLE_INFO << d_cluster_p->description()
+                         << ": queue successfully reopened ["
+                         << requestContext->request()
+                         << "].";
```
678098, line 1789:
```suggestion
                         << requestContext->request() << "].";
```

678098 (approved):
Linter issue, otherwise is good

PR #1478, opened 2026-06-03
Refactor[mqbs::FileStore]: decouple `issueSyncPointInternal`

678098:
- Extract `issueSyncPointIfNeeded` from `issueSyncPointInternal` to use in periodic checks.
- Add `inDispatcherThread()` checks
- Remove `immediateFlush` arg for sync points because it is always true

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_filestore.cpp
+++ b/src/groups/mqb/mqbs/mqbs_filestore.cpp
@@ -1042,9 +1042,7 @@ int FileStore::openInRecoveryMode(bsl::ostream&    errorDescription,
                 ? fileSetSp->d_qlistFilePosition / bmqp::Protocol::k_WORD_SIZE
                 : 0;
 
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -4069,63 +4067,64 @@ void FileStore::issueSyncPointDispatched(BSLA_MAYBE_UNUSED int partitionId)
     // attempts to issue a SyncPt if applicable (see 'issueSyncPointCb'), which
     // means that there must be space for at least 2 journal records.
 
-    issueSyncPointInternal(SyncPointType::e_REGULAR,
-                           true);  // ImmediateFlush flag
+    issueSyncPointIfNeeded();
```
678098, line 4070: Return code is not used, so can return `void`

kaikulimu (approved):
LGTM

PR #1480, opened 2026-06-04
Fix[bmqp_protocolutil.h]: limit sequence size for `BerDecoder`

678098:
The default sequence size is ~1kkk that is not necessary.
Use blob length as upper bound for max sequence length that `BerDecoder` can read.
In the worst case the entire blob is a `vector<char>`, and it has a bit less elements than the number of bytes in blob.


hallfox (dismissed):

678098:
IT failure is unrelated

PR #1481, opened 2026-06-04
Test: Clean up fuzz test runner

pniedzielski:
This pull request cleans up the fuzz test runner.  As it stands now, the fuzz test `README` is only correct for some ways of building BlazingMQ, and not the default one on MacOS.  Furthermore, the fuzz tests have to be run in their own way, separate from the standard ctest UT and IT integration that we use.  By converting the fuzz tests to pytest, we can use the same integration between ctest and pytest that we use with integration tests, enabling fuzz tests to be run directly from a CMake-aware editor, like VSCode.


src/fuzz-tests/conftest.py
```diff
@@ -0,0 +1,91 @@
+# Copyright 2024 Bloomberg Finance L.P.
```
hallfox:
```suggestion
# Copyright 2026 Bloomberg Finance L.P.
```

src/fuzz-tests/test_fuzz.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2024 Bloomberg Finance L.P.
```
hallfox:
```suggestion
# Copyright 2026 Bloomberg Finance L.P.
```

hallfox (dismissed):

src/fuzz-tests/conftest.py
pniedzielski: Good catch.  I'll apply these two changes in a rebase and request a review again.

hallfox (approved):

PR #1482, opened 2026-06-04
Fix[mqb]: getting rid of resetQueue

dorjesinpo:
upon shutdown, domain may 'teardown', close queue and then shutdown stops dispatcher.  
meanwhile `ClusterQueueHelper::onQueueHandleDestroyedDispatched` attempts to 'deleteQueue' and calls 'resetQueue' which tries to use dispatcher which is already stopped.

Fix, get rid of 'resetQueue'  and make `closeDispatched' call `setQueue(0)`


src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -255,7 +255,7 @@ bool isConfigureErrorPermanent(
         return false;
     }
     default: {
-        BSLS_ASSERT_OPT_UNREACHABLE("Unknown error kind");
+        //        BSLS_ASSERT_OPT_UNREACHABLE("Unknown error kind");
```
```suggestion
        BSLS_ASSERT_OPT_UNREACHABLE("Unknown error kind");
```

678098 (approved):
Looks good

PR #1483, opened 2026-06-04
Fix[mqb]: reject `OpenQueue` with `adminCount != 0`

678098:


src/groups/bmq/bmqp/bmqp_queueutil.cpp
```diff
@@ -55,16 +55,12 @@ bool QueueUtil::isValid(const bmqp_ctrlmsg::QueueHandleParameters& parameters)
             return false;  // RETURN
         }
     }
-    else if (parameters.writeCount() > 0) {
```
678098, line 58: I made conditions more strict. Negative read/write counts are invalid anyway

chrisbeard (approved):

PR #1484, opened 2026-06-04
Fix[mqbstat]: Pass `shared_ptr` to `FileObserver2::publish`

pniedzielski:
`StatsFileLogger::logStats` passes a stack-allocated `ball::Record` by reference to `ball::FileObserver2::publish`.  `FileObserver2` only overrides the `shared_ptr<const Record>` overload of `publish`; it does not override the deprecated `const Record&` overload.  The base class `ball::Observer::publish(const Record&, const Context&)` default implementation contains `BSLS_ASSERT_OPT(false)`, so the broker crashes on the first stats print interval after startup.

The crash is delayed because `snapshotAndNotify` fires every `snapshotInterval` seconds, but `logStats` is only called when the snapshot counter reaches the `printInterval` (i.e., after `printInterval / snapshotInterval` snapshots).  With typical configuration values, this means the broker runs for about a minute before hitting the assertion.

This pull request wraps the record in a `bsl::shared_ptr` before calling `publish`, so that the call resolves to the `shared_ptr` overload that `FileObserver2` implements.

### DIFF
--- a/src/groups/mqb/mqbstat/mqbstat_statsfilelogger.cpp
+++ b/src/groups/mqb/mqbstat/mqbstat_statsfilelogger.cpp
@@ -31,6 +31,7 @@
 #include <bdlt_datetime.h>
 #include <bdlt_epochutil.h>
 #include <bsl_ctime.h>
+#include <bsl_memory.h>
 #include <bsl_ostream.h>
 #include <bslma_allocator.h>
 #include <bslmt_threadutil.h>
@@ -56,6 +57,7 @@ StatsFileLogger::StatsFileLogger(bsl::string_view       filePattern,
 : d_filePattern(filePattern, allocator)
 , d_statsLogFile(allocator)
 , d_statLogCleaner(eventScheduler, allocator)
+, d_allocator_p(allocator)
 {
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(eventScheduler);
@@ -107,8 +109,9 @@ void StatsFileLogger::logStats(const PrinterCb& printerCb)
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(printerCb);
 
-    ball::Record            record;
-    ball::RecordAttributes& attributes = record.fixedFields();
+    bsl::shared_ptr<ball::Record> record_sp =
+        bsl::allocate_shared<ball::Record>(d_allocator_p);
+    ball::RecordAttributes& attributes = record_sp->fixedFields();
     bdlt::Datetime          now;
     bdlt::EpochUtil::convertFromTimeT(&now, time(0));
     attributes.setTimestamp(now);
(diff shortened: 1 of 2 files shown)


pniedzielski:
Found this while running fuzz tests locally.

src/groups/mqb/mqbstat/mqbstat_statsfilelogger.cpp
```diff
@@ -124,8 +126,10 @@ void StatsFileLogger::logStats(const PrinterCb& printerCb)
 
     printerCb(os);
 
+    bsl::shared_ptr<const ball::Record> recordPtr =
+        bsl::allocate_shared<ball::Record>(d_allocator_p, record);
```
678098:
```suggestion
bsl::shared_ptr<ball::Record> record_sp =
    bsl::allocate_shared<ball::Record>(d_allocator_p);
ball::RecordAttributes& attributes = record_sp->fixedFields();
bdlt::Datetime          now;
bdlt::EpochUtil::convertFromTimeT(&now, time(0));
attributes.setTimestamp(now);
attributes.setProcessID(bdls::ProcessUtil::getProcessId());
attributes.setThreadID(bslmt::ThreadUtil::selfIdAsUint64());
attributes.setFileName(__FILE__);
attributes.setLineNumber(__LINE__);
attributes.setCategory(k_LOG_CATEGORY);
attributes.setSeverity(ball::Severity::e_INFO);

attributes.clearMessage();
bsl::ostream os(&attributes.messageStreamBuf());

printerCb(os);
```
What about constructing a record shared ptr right away?

src/groups/mqb/mqbstat/mqbstat_statsfilelogger.cpp
```diff
@@ -124,8 +126,10 @@ void StatsFileLogger::logStats(const PrinterCb& printerCb)
...
     d_statsLogFile.publish(
-        record,
+        recordPtr,
```
678098: Pity that it doesn't support move

src/groups/mqb/mqbstat/mqbstat_statsfilelogger.cpp
pniedzielski: Good idea, that's nicer.

src/groups/mqb/mqbstat/mqbstat_statsfilelogger.cpp
```diff
@@ -124,8 +126,26 @@ void StatsFileLogger::logStats(const PrinterCb& printerCb)
 
     printerCb(os);
 
+    bsl::shared_ptr<const ball::Record> recordPtr =
```
678098: It seems that the proposed patch did not work when added to a non-diff

pniedzielski: Manually fixed the patch application and letting build run.

678098 (approved):
Looks good

PR #1488, opened 2026-06-05
Build: fix the build with C++03

678098:
- Remove redundant `;`
- Pass pointers to `allocate_shared`, not references.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_routers.h
+++ b/src/groups/mqb/mqbblp/mqbblp_routers.h
@@ -710,7 +710,7 @@ class Routers {
         Priorities d_priorities;
         Consumers  d_consumers;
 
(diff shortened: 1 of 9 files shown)


src/groups/mqb/mqbblp/mqbblp_routers.h
```diff
@@ -848,19 +848,19 @@ inline Routers::Consumer::~Consumer()
 // struct Routers::AppContext
 // -----------------------------
 
-inline Routers::AppContext::AppContext(QueueRoutingContext& queue,
```
678098, line 851: non-const references don't work with `allocate_shared` when passed as argument. Also, BDE style requires providing non-const args as pointers.

kaikulimu (approved):

PR #1491, opened 2026-06-05
Fix[mqbblp]: correct operator precedence in Cluster confirm/reject handlers

### LINKED ISSUE #1477: Operator precedence bug in Cluster confirm/reject handlers silences corrupt event detection
The confirm and reject event handlers in `mqbblp_cluster.cpp` have a misplaced parenthesis in their iterator loops that appears to silently disable the corrupt event error path. The PUT and ACK handlers in the same file parenthesize correctly, assigning the return value of `next()` to `rc` before comparing it to 1. The CONFIRM and REJECT handlers have the closing paren in the wrong place, so `next() == 1` evaluates as a bool first, then assigns 0 or 1 to `rc`. If `next()` returns a negative error code for a corrupt event, the comparison yields `false`, `rc` becomes 0, and the `if (rc < 0)` block that logs the corruption and fires `BSLS_ASSERT_SAFE(false)` should never execute. It looks like corrupt confirm and reject events on the primary node would be silently swallowed.

SrinathhSatuluri:
Closes #1477

The confirm and reject iterator loops in `mqbblp_cluster.cpp` had a misplaced parenthesis causing `next() == 1` to evaluate as a bool before assignment to `rc`, making the `if (rc < 0)` corrupt event error path unreachable. Fixed to match the PUT and ACK handlers. Also corrected the confirm handler's throttle object from `d_throttledFailedRejectMessages` to `d_throttledFailedConfirmMessages`.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_cluster.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_cluster.cpp
@@ -1069,7 +1069,7 @@ void Cluster::onConfirmEvent(const mqbevt::ConfirmEvent& event)
     int msgNum = 0;
     int rc     = 0;
 
-    while ((rc = confIt.next() == 1)) {
+    while ((rc = confIt.next()) == 1) {
         const int          id    = confIt.message().queueId();
         const unsigned int subId = static_cast<unsigned int>(
             confIt.message().subQueueId());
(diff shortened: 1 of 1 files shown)


678098 (approved):

678098:

PR #1494, opened 2026-06-05
Refactor: move `bmqp::ProtocolUtil::QueueInfo` to a separate source

678098:
`bmqp::ProtocolUtil::QueueInfo` is only used in 3 headers in `mqb`.
It is easily extractable from `bmqp_protocolutil`, and this should be done to remove includes to heavy headers from transitive include trees.

This is a preparation PR to reduce include burden in the codebase, since `bmqp_protocolutil.h` is transitively included in 240+ source files.
In the future we might consider moving `bmqp::QueueInfo` to `mqb` package.


dorjesinpo (approved):
thank you

PR #1495, opened 2026-06-08
Fix: Add libssl dependency to build-ubuntu.sh

jakeski2015:


bin/.build-ubuntu.sh.swp
pniedzielski, line 1: I don't think we want this file, right?

pniedzielski (changes requested):

pniedzielski (approved):

kaikulimu:
Pog 🔥

PR #1505, opened 2026-06-09
Fix[mqbc::ClusterStateMger]: Apply CSL snapshot for out-of-date follower

kaikulimu:
## Summary
- Introduce `e_LDR_HEALING_STG3` to the Cluster FSM, splitting the old `LDR_HEALING_STG2` into two distinct stages: **STG2** (requesting CSL snapshot from follower) and **STG3** (waiting for the applied LeaderAdvisory to commit).
- In STG3, if a late-joining follower sends a `RegistrationRequest` or `FollowerLSNResponse`, the leader applies a new CSL LeaderAdvisory (reusing existing partition assignments). This prevents a race where a late-joining follower could miss the advisory broadcast and fail to heal.
- Introduce `e_FOL_WAITING` state: follower waits for RegistrationResponse before entering `FOL_HEALING`. While waiting, it rejects `FollowerLSNRequest`s to prevent the leader from receiving duplicate LSN info (once via RegistrationRequest, once via FollowerLSNResponse).
- Add `IsHealedCb` to the `ClusterStateLedger` interface: the `IncoreClusterStateLedger` now rejects `e_UPDATE` records (e.g., QueueAssignmentAdvisory) while the follower is still healing, preventing accidental premature transition to `FOL_HEALED`.
- Add `uncommittedAdvisories()` to the `ClusterStateLedger` interface, used by `do_applyCSLSelf` to detect whether partition assignments can be reused on re-entry.

## On `uncommittedAdvisories()`

`uncommittedAdvisories()` is a new **accessor** we exposed on the `ClusterStateLedger` interface:

```cpp
/// Load into the specified `out` the list of uncommitted advisories as
/// const references.
virtual void uncommittedAdvisories(ClusterMessageCRefList* out) const = 0;
```

Previously the set of applied-but-not-yet-committed advisories was purely private state inside `IncoreClusterStateLedger` (`d_uncommittedAdvisories`). This PR surfaces a read-only view of it so the `ClusterStateManager` can inspect what advisories are currently in flight without waiting for them to commit.

### Why `do_applyCSLSelf` needs it

With the new STG3 stage, `do_applyCSLSelf` can now be **re-entered** during the same leadership session: the leader first applies a LeaderAdvisory when it initially heals, and then applies *another* one if a late-joining follower shows up (via `RegistrationRequest`/`FollowerLSNResponse`) before the first advisory has committed.

On that second application, the leader must decide whether to **reuse the existing partition assignment** or compute a fresh one via `ClusterUtil::assignPartitions`. Re-assigning partitions would be incorrect here — the primaries have already been chosen for this term, and reassigning could shuffle them.

The leader already knows it can reuse the assignment if self is fully healed (a committed advisory exists). But in STG3 the first advisory has **not committed yet**, so `isSelfHealed()` is still false. The only evidence that partitions were already assigned in this term is the earlier advisory sitting in the uncommitted set — which is exactly what `uncommittedAdvisories()` exposes:

```cpp
ClusterStateLedger::ClusterMessageCRefList uncommitted(d_allocator_p);
d_clusterStateLedger_mp->uncommittedAdvisories(&uncommitted);

// We already have a valid partition assignment if either:
//   (a) self is healed (committed advisory exists), or
//   (b) we have an uncommitted LeaderAdvisory from our own elector term.
bool hasOwnUncommittedAdvisory = false;
for (size_t i = 0; i < uncommitted.size(); ++i) {
    const bmqp_ctrlmsg::ClusterMessageChoice& choice =
        uncommitted[i].get().choice();
    if (choice.isLeaderAdvisoryValue() &&
        choice.leaderAdvisory().sequenceNumber().electorTerm() ==
            d_clusterData_p->electorInfo().electorTerm()) {
        hasOwnUncommittedAdvisory = true;
        break;
    }
}

if (d_clusterFSM.isSelfLeader() &&
    (d_clusterFSM.isSelfHealed() || hasOwnUncommittedAdvisory)) {
    ClusterUtil::loadPartitionsInfo(...);   // reuse existing assignment
}
else {
    ClusterUtil::assignPartitions(...);     // first-time assignment
}
```

The elector-term check matters: an uncommitted advisory from a *previous* leader's term is stale (advisories are not cancelled on leader switch), so its partition assignment must not be reused. We iterate the list because such stale advisories from an old leader may precede ours.


src/groups/mqb/mqbc/mqbc_clusterstatemanager.cpp
```diff
@@ -295,6 +295,48 @@ void ClusterStateManager::do_applyCSLSelf(const EventWithMetadata& event)
...
+    ClusterStateLedger::ClusterMessageCRefList uncommitted(d_allocator_p);
+    d_clusterStateLedger_mp->uncommittedAdvisories(&uncommitted);
+
+    if (uncommitted.size() != 1) {
+        BMQTSK_ALARMLOG_ALARM("RECOVERY")
+            << d_clusterData_p->identity().description()
+            << ": Cannot send CSL patch to out-of-date follower because we "
+            << "expect to have exactly 1 uncommitted advisory, but saw "
+            << uncommitted.size() << BMQTSK_ALARMLOG_END;
+        return;  // RETURN
```
dorjesinpo: does this happen before or after RegistrationRequest?  Can the follower learn about errors?  What can it do?

src/groups/mqb/mqbc/mqbc_clusterstatetable.h
```diff
@@ -614,7 +623,7 @@ class ClusterStateTable
         CST_CFG(LDR_HEALING_STG2,
                 FOL_CSL_RSPN,
                 applyCSLSelf,
-                LDR_HEALING_STG2);
+                LDR_HEALING_STG3);
```
dorjesinpo, line 681: Is this because _any_ FollowerClusterStateResponse fills the gap? And there is exactly one?

src/groups/mqb/mqbc/mqbc_clusterstatetable.h
```diff
@@ -672,6 +681,58 @@ class ClusterStateTable
...
+                SLCT_FOL,
+                stopWatchdog_cleanupLSNs_cancelRequests_reapplyEvent,
+                UNKNOWN);
+        CST_CFG(LDR_HEALING_STG3,
+                FOL_LSN_RQST,
+                sendFailureFollowerLSNResponse,
+                LDR_HEALING_STG3);
+        CST_CFG(LDR_HEALING_STG3,
+                FOL_LSN_RSPN,
+                sendCSLPatch,
```
dorjesinpo: why not `applyCSLSelf`?

dorjesinpo (commented):
There is a possibility to update late replica in exactly the same way as in the `HEALED` state (snapshot) - that looks more consistent, although patching works too. 
Note that making it consistent the other way - by patching in the `HEALED` can work too but it may require a bigger change.

src/groups/mqb/mqbc/mqbc_clusterstatetable.h
kaikulimu: Changed from `sendCSLPatch ` to `applyCSLSelf `

src/groups/mqb/mqbc/mqbc_clusterstatemanager.cpp
kaikulimu: Not relevant anymore because `sendCSLPatch` has been removed.

src/groups/mqb/mqbc/mqbc_clusterstatetable.h
kaikulimu: Yes to both questions

src/groups/mqb/mqbc/mqbc_incoreclusterstateledger.cpp
```diff
@@ -1042,6 +1054,16 @@ int IncoreClusterStateLedger::applyImpl(const bdlbb::Blob&   event,
...
+        if (recordHeader->recordType() == ClusterStateRecordType::e_UPDATE &&
+            !d_isHealedCb()) {
+            BALL_LOG_WARN << description()
+                          << ": Ignoring e_UPDATE record with LSN = "
+                          << printLSN(lsn) << " from '"
+                          << source->nodeDescription()
+                          << "' because self follower is not healed.";
+            return rc_UPDATE_WHILE_NOT_HEALED;  // RETURN
+        }
+
```
dorjesinpo, line 1080: Here is tool generated report
But the matching e_COMMIT is not gated — only e_UPDATE is. So consider a node rejoining a healthy cluster (leader LDR_HEALED, healed majority):

  1. Node N is in FOL_WAITING/FOL_HEALING (isSelfHealed() == false).
  2. Normal traffic → leader applies a QueueAssignmentAdvisory (e_UPDATE, LSN 100) and writeAll()s it to everyone, including N.
  3. N rejects it (gate above) → nothing stored in d_uncommittedAdvisories, no ack sent. (intended)
  4. Leader reaches quorum from the healed majority → broadcasts the e_COMMIT (LSN 101, committing 100) via writeAll() — again to everyone, including N.
  5. N receives the commit. It is not gated, so it enters the commit branch (mqbc_incoreclusterstateledger.cpp:616):

  Consequences

  - leaderMessageSequence gets advanced (to 101) for a commit the node never applied. The node's persisted CSL LSN now runs ahead of the actual content of
  its ledger — it has neither record 100 nor 101. Before d5a3c9f, rc_ADVISORY_NOT_FOUND was effectively unreachable for a follower (it always stored the
  advisory before the commit arrived); the new gate makes it reachable, and the LSN side-effect at line 626 fires before the not-found bailout.
  - A guaranteed BALL_LOG_ERROR per committed update during every healing window — misleading noise for exactly the routine late-join case this feature
  targets.

  The healing snapshot usually papers over the LSN inconsistency (the leader mints a higher LSN via nextLeaderMessageSequence in do_applyCSLSelf, so the
  snapshot isn't seen as stale) — but relying on that is fragile, and the advanced-LSN-without-content state is a real latent hazard if the node is later
  consulted for its LSN or wins leadership.

dorjesinpo (commented):
would it be possible to move the `uncommittedAdvisories` to a new PR and add more description? A scenario can help.

Also, see one comment inline

kaikulimu:
A detailed explanation on the necessity of `uncommittedAdvisories()` has been added to the PR description.

src/groups/mqb/mqbc/mqbc_incoreclusterstateledger.cpp
kaikulimu: The commit titled `mqbc::IncoreCSL: Ignore commit of gated e_UPDATE` addresses this issue.

kaikulimu:
@dorjesinpo Evgenii will take over the review of this PR. Thanks so much to Evgenii!

kaikulimu:
Will be split into multiple smaller PRs

PR #1506, opened 2026-06-10
Fix[mqb]: race between stopListening and setChannel

dorjesinpo:
If a new connection happens before `TCPSessionFactory::stopListening()` and  the IO / authenticator thread is slow and calls  `mqbnet::Channel::setChannel` after `Cluster::initiateShutdown` has called `mqbnet::Channel::stop` which joins the channel thread.
Then, `mqbnet::Channel::setChannel` gets stuck holding IO thread.
The fix - do not call `mqbnet::Channel::setChannel` after `TCPSessionFactory::stopListening()`

### DIFF
--- a/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
@@ -818,7 +818,17 @@ void TCPSessionFactory::channelStateCallback(
         BSLS_ASSERT_SAFE(status);  // got a channel up, it must be success
         BSLS_ASSERT_SAFE(channel);
 
-        if (channel->peerUri().empty()) {
+        if (!d_isListening) {
+            BALL_LOG_INFO << "#SESSION_NEGOTIATION TCPSessionFactory '"
+                          << d_config.name()
+                          << "' has stopped listening and rejecting '"
+                          << channel.get() << "'";
+
+            bmqio::Status closeStatus(bmqio::StatusCategory::e_GENERIC_ERROR,
+                                      d_allocator_p);
+            channel->close(closeStatus);
+        }
+        else if (channel->peerUri().empty()) {
             BALL_LOG_ERROR << "#SESSION_NEGOTIATION "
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -719,31 +719,35 @@ void TCPSessionFactory::initialConnectionComplete(
...
-        bdlf::BindUtil::bind(&TCPSessionFactory::readCallback,
-                             this,
-                             bdlf::PlaceHolders::_1,  // status
-                             bdlf::PlaceHolders::_2,  // numNeeded
-                             bdlf::PlaceHolders::_3,  // blob
-                             info.get()));
+    bool result = false;
 
-    if (!result || !d_isListening) {
+    if (d_isListening) {
```
678098, line 724: Idea: what about replacing this atomic with atomic gate?
Making it impossible to even enter some parts of the code after we called `invalidate()`?
It looks like a more robust solution than trying to find correct timings to check atomics.

dorjesinpo: The problem is checking `d_isListening` after calling `d_resultCb`.  Atomic or gate, it does not matter

678098 (dismissed):

678098 (approved):

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -814,7 +818,17 @@ void TCPSessionFactory::channelStateCallback(
         BSLS_ASSERT_SAFE(status);  // got a channel up, it must be success
         BSLS_ASSERT_SAFE(channel);
 
-        if (channel->peerUri().empty()) {
+        if (!d_isListening) {
```
678098, line 821: @chrisbeard this is the entry point closest to ntc, when we just received a new channel. The idea is that we can just drop any new channels here early on if we are stopping.

PR #1511, opened 2026-06-10
Perf[mqbblp]: early return if no throttle to cancel

678098:
Avoid looking for handle via `findQueueHandleContext` if there is no throttle.
In other words, `tryCancelThrottle` should be no-op if there is no throttle scheduled.

Avoid spending time on this lookup if there is no throttle:
Screenshot 2026-06-10 at 19 17 20


src/groups/mqb/mqbblp/mqbblp_queueengineutil.cpp
```diff
@@ -1210,14 +1215,12 @@ void QueueEngineUtil_AppState::tryCancelThrottle(
     }
 
     queueHandleContext->d_timeLastMessageSent = 0;
```
678098, line 1217: The side-effect is that we don't set this `d_timeLastMessageSent` to 0 if there is no throttle, but it looks fine.

dorjesinpo (approved):
thank you

PR #1512, opened 2026-06-11
Fix[bmqeval]: guard against INT64_MIN / -1 quotient overflow

### LINKED ISSUE #1504: Fix[bmqeval]: guard against IDIV quotient overflow in integer division and modulus
#1446 guards against SIGFPE from division and modulus by zero in `NumBinaryOperation::evaluate`. The x86 IDIV instruction faults on a second condition: when the quotient overflows the destination register. `INT64_MIN / -1` and `INT64_MIN % -1` trigger this. The mathematical result is `INT64_MAX + 1`, which is undefined behavior in C++ and produces the same SIGFPE.

SrinathhSatuluri:
Closes #1504

This PR extends the existing zero-divisor guard in `NumBinaryOperation::evaluate` to also cover the case where the dividend is `INT64_MIN` and the divisor is `-1`. On x86, the IDIV instruction faults when the quotient overflows the destination register, resulting in SIGFPE.

Both division and modulus with these operands now return `e_ARITHMETIC`.

### DIFF
--- a/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.h
+++ b/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.h
@@ -58,6 +58,7 @@
 // BDE
 #include <bdld_datum.h>
 #include <bsl_functional.h>
+#include <bsl_limits.h>
 #include <bsl_memory.h>
 #include <bsl_string.h>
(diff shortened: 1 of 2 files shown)


678098 (approved):
Looks good

chrisbeard (approved):

PR #1514, opened 2026-06-11
CI: Add clang-tidy action

hallfox:
Attempting to add extra clang-tidy checks for CI. Currently it's only configured to run on diffs since I suspect we'd have multiple clang-tidy violations with the default checks.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -184,3 +184,23 @@ jobs:
         run: |
           export PYTHONPATH=${{ github.workspace }}/src/python:$PYTHONPATH
           git diff -U0 origin/main | lint-diffs
+
+  cpp-linter-check:
+    name: C++ Linter Check
+    runs-on: ubuntu-slim
+    steps:
+      - uses: actions/checkout@v5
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yaml
```diff
@@ -184,3 +184,23 @@ jobs:
         run: |
           export PYTHONPATH=${{ github.workspace }}/src/python:$PYTHONPATH
           git diff -U0 origin/main | lint-diffs
+
+  cpp-linter-check:
+    name: C++ Linter Check
+    runs-on: ubuntu-slim
+    steps:
+      - uses: actions/checkout@v5
+      - uses: cpp-linter/cpp-linter-action@v2
```
pniedzielski: For non-official actions, let's pin to a SHA, like https://github.com/bloomberg/blazingmq/blob/285e86c61f0abe9bdc91876de311186ac9781581/.github/workflows/publish.yaml#L24

pniedzielski (changes requested):
Looks good, just one nit

pniedzielski (approved):

PR #1516, opened 2026-06-11
Fix[bmqbrkr]: keep `bmqu::Time` initialized

678098:
There was a brief time window between `bmqu::Time::shutdown` and `scheduler->stop()` when `bmqu::Time` is not initialized. Any time call will cause a crash.

`bmqu::Time::initialize`/`bmqu::Time::shutdown` are refcounted so I extend the life scope of `bmqu::Time` to the `Task` lifetime

This PR fixes the following crash
```
    #6  0x00007f2d0eabb0da in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6                                         
    No symbol table info available.                                                                                   
    #7  0x00007f2d0eaa5a55 in std::terminate() () from /lib/x86_64-linux-gnu/libstdc++.so.6                           
    No symbol table info available.                                                                                   
    #8  0x00007f2d0eabb391 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6                                
    No symbol table info available.                                                                                   
    #9  0x000055a66f9ef222 in BloombergLP::bslstl::Function_Variadic<long long ()>::operator()() const [clone         
  .part.0] ()                                                                                                         
    No symbol table info available.                                                                                   
    #10 0x000055a66f9ef22b in BloombergLP::defaultDispatcherFunction(bsl::function<void ()> const&) [clone .cold] ()  
    No symbol table info available.                                                                                   
    #11 0x000055a670aac60d in BloombergLP::bdlmt::EventScheduler::dispatchEvents() ()                                 
    No symbol table info available. 
```

### DIFF
--- a/src/applications/bmqbrkr/m_bmqbrkr_task.cpp
+++ b/src/applications/bmqbrkr/m_bmqbrkr_task.cpp
@@ -25,6 +25,7 @@
 #include <bmqtsk_alarmlog.h>
 #include <bmqu_memoutstream.h>
 #include <bmqu_printutil.h>
+#include <bmqu_time.h>
 
 // BDE
 #include <bdlb_string.h>
@@ -277,6 +278,8 @@ int Task::initialize(bsl::ostream&             errorDescription,
     int                rc = rc_SUCCESS;
     bmqu::MemOutStream localError;
 
+    bmqu::Time::initialize();
+
     // ---------
     // Scheduler
     rc = d_scheduler.start();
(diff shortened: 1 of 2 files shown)


src/applications/bmqbrkr/m_bmqbrkr_task.cpp
```diff
@@ -379,6 +382,8 @@ void Task::shutdown()
     // Stop the scheduler
     d_scheduler.stop();
 
+    bmqu::Time::shutdown();
```
678098, line 385: The last `bmqu::Time::shutdown` happens after stopping the global scheduler

dorjesinpo (commented):
there are `Application::oneTimeInit()` / `Application::oneTimeShutdown()`.  How do they play along?

PR #1518, opened 2026-06-11
FIX: Broker now receives the PUT message sent in the fuzztests

jakeski2015:
**Describe your changes**
Turned off fuzz for event size and description. The broker wasn't recognizing PUT messages with malformed event descriptions and sizes. 

**Testing performed**
Manually viewing on wireshark. 

**Additional context**

Before Changes:
image

After:
image

### DIFF
--- a/src/python/blazingmq/dev/fuzztest/__init__.py
+++ b/src/python/blazingmq/dev/fuzztest/__init__.py
@@ -180,11 +180,13 @@ def wrap_event(
         endian=">",
         inclusive=True,
         length=NumBytes.WORD,
+        fuzzable=False,
     )
     event_desc = boofuzz.Bytes(
         name="event_desc",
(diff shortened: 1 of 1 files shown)


src/python/blazingmq/dev/fuzztest/__init__.py
```diff
@@ -180,11 +180,13 @@ def wrap_event(
         endian=">",
         inclusive=True,
         length=NumBytes.WORD,
+        fuzzable=False,
```
678098, line 183: I think it is the point of fuzzing to also fuzz event size.
If the broker handles it and rejects the event, it is good. If not, we need to find it - and this is why fuzzing is needed.
Note that we only fuzz 1 field at once. So if we fuzz other field of PUT event, we don't fuzz event size.

PR #1475, opened 2026-06-03
Fix[mqbc::StorageMgr]: Send primary lease id in PFSM state rqst/resp

kaikulimu:
> Fix for Internal Ticket 184450660

In `bmqp::ReplicaStateRequest` and similar PFSM state requests/responses, the `latestSequenceNumber()` refers to the latest PSN we retrieve from storage, but it does not necessarily equal our known primaryLeaseId. Then, we add a `primaryLeaseId()` field explicitly. We use this field to detect and drop stale events.

*Note*: This PR introduces backward incompatible changes to `bmqp_ctrlmsg.xsd`, but it is fine as long as we restart all nodes together for any FSM-enabled clusters.

PR #1521, opened 2026-06-11
Cleanup[bmqp_ctrlmsg.xsd]: Primary lease id in ReplicaDataRequest/Response

kaikulimu:
## Summary
  - Add a `primaryLeaseId` field to `ReplicaDataRequest` and `ReplicaDataResponse` control messages so that replicas can detect stale primaries during partition data synchronization.
  - Verify that the `primaryLeaseId` in responses matches the one sent in the corresponding request, catching mismatches early.
  - Pass `primaryLeaseId` through `PartitionFSMEventData` for all PULL, PUSH, DROP, and failure event paths so the partition FSM has the correct lease context for stale detection.

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_ctrlmsg.xsd
+++ b/src/groups/bmq/bmqp/bmqp_ctrlmsg.xsd
@@ -786,9 +786,10 @@
         This type represents a request sent to the replica by the primary to
         start the synchronization of data.
 
+        partitionId:         partition id for corresponding partition.
+        primaryLeaseId:      lease id of the current primary.
         replicaDataType:     type of request i.e. PULL, PUSH or DROP for
                              corresponding partition.
-        partitionId:         partition id for corresponding partition.
         beginSequenceNumber: Primary's begin sequence number for corresponding
                              partition for corresponding data chunks.
         endSequenceNumber:   Primary's end sequence number for corresponding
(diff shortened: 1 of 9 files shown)


src/groups/bmq/bmqp/bmqp_ctrlmsg.xsd
```diff
@@ -786,18 +786,20 @@
...
                              partition for corresponding data chunks.
         endSequenceNumber:   Primary's end sequence number for corresponding
                              partition for corresponding data chunks.
       </documentation>
     </annotation>
     <sequence>
-      <element name='replicaDataType'     type='tns:ReplicaDataType'/>
       <element name='partitionId'         type='int'/>
+      <element name='primaryLeaseId' type='unsignedInt'/>
+      <element name='replicaDataType'     type='tns:ReplicaDataType'/>
```
678098, line 802: So basically reordering fields is safe with json, but is not safe in BER, right?
@kaikulimu

kaikulimu: Yes, so it will break in FSM mode if we mix a broker having this PR merged with an older broker. For clusters running in FSM mode, we would have to restart the entire cluster when rolling out new broker there.

kaikulimu: We did the same restart after merging https://github.com/bloomberg/blazingmq/pull/1475, and things were fine.

678098 (approved):

PR #1522, opened 2026-06-12
Refactor[mqbnet]: less includes for `TCPSessionFactory`

678098:
- Replace heavy includes with forward declarations.
- Change some member fields to pointer types.
- Cache `TCPSessionFactory '<name>'` in `d_name` to use for shorter logging.
- Make logging consistent.

This PR makes `mqbnet_tcpsessionfactory.h` ~45 times cheaper to include.
Since it it included by 58 other source files, it removes ~500k lines during compilation.

### Before

```
Header                                               Lines      Cost  Direct Transitive
---------------------------------------------------------------------------------------
mqbnet_tcpsessionfactory.h                             751    145595       6      58

Codebase Stats
  Total source files:                             961
  Total own lines:                             651929
  Total transitive includes:                    23504
  Total include cost (lines):                34060783
```

### After

```
Header                                               Lines      Cost  Direct Transitive
---------------------------------------------------------------------------------------
mqbnet_tcpsessionfactory.h                             776      2655       6      58

Codebase Stats
  Total source files:                             961
  Total own lines:                             651947
  Total transitive includes:                    22200
  Total include cost (lines):                33497655
```

### DIFF
--- a/src/groups/bmq/bmqimp/bmqimp_application.cpp
+++ b/src/groups/bmq/bmqimp/bmqimp_application.cpp
@@ -32,6 +32,7 @@
 #include <bmqio_tcpendpoint.h>
 #include <bmqma_countingallocatorutil.h>
 #include <bmqp_crc32c.h>
+#include <bmqp_heartbeatmonitor.h>
 #include <bmqscm_version.h>
 #include <bmqst_statvalue.h>
 #include <bmqst_tableutil.h>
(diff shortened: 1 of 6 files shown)


hallfox (approved):

hallfox:
While I'm glad this helps, it also makes me wonder if the 58 other uses of this component would not be better benefited by depending on an interface for this, I suspect very few places actually need to depend on the actual implementation. This class is a well known mess for us to deal with.

678098:
> While I'm glad this helps, it also makes me wonder if the 58 other uses of this component would not be better benefited by depending on an interface for this, I suspect very few places actually need to depend on the actual implementation. This class is a well known mess for us to deal with.

Unfortunately, the existing design lacks isolation through interfaces. Even "interface" types such as `mqbi::Queue` reference heavy protocol headers such as `bmqp_ctrlmsg`. I am trying to change it gradually.

PR #1528, opened 2026-06-12
Fix: use blocking cancelEventAndWait

678098:


src/groups/mqb/mqbblp/mqbblp_storagemanager.cpp
```diff
@@ -1262,18 +1262,6 @@ int StorageManager::start(bsl::ostream& errorDescription)
...
-    // space, archive clean up, etc).
-    d_clusterData_p->scheduler().scheduleRecurringEvent(
-        &d_storageMonitorEventHandle,
-        bsls::TimeInterval(bdlt::TimeUnitRatio::k_SECONDS_PER_MINUTE),
-        bdlf::BindUtil::bind(&mqbc::StorageUtil::storageMonitorCb,
-                             &d_lowDiskspaceWarning,
-                             &d_isStarted,
-                             d_minimumRequiredDiskSpace,
-                             d_clusterData_p->identity().description(),
-                             d_clusterConfig.partitionConfig()));
```
678098, line 1275: Moved after the last possible `return`, otherwise event handle might leak

678098:
https://github.com/bloomberg/blazingmq/blob/6c4f6c65049b54d333ac3fdf436b9c8482e5577a/src/groups/mqb/mqba/mqba_authenticationclient.cpp#L229
Is the only place where I haven't changed it since it might be double-blocked on mutex.
And there is a waiting `cancelEventAndWait` in the destructor.

PR #1530, opened 2026-06-15
Build: disable C++03 autogen in base CMake preset

678098:
Set `SIM_CPP11=""` in the base CMake preset to prevent `sim_cpp11_features.pl` from running during local builds. The script embeds a timestamp in generated `*_cpp03.h` files, causing them to appear dirty after every build even when nothing changed.

CI presets (`gh-base`) are unaffected — they don't inherit from base, so C++03 codegen validation still runs in GitHub Actions.


678098:
Maybe this PR will not be needed soon

hallfox (approved):

PR #1536, opened 2026-06-15
Refactor: simplify stat includes

678098:
This PR removes 3.4kk (-10%) intermediate lines in preprocessed files during compilation.

# Before
```
Codebase Stats
  Total source files:                             956
  Total own lines:                             650373
  Total transitive includes:                    20797
  Total include cost (lines):                32926286
```

# After

```
Codebase Stats
  Total source files:                             956
  Total own lines:                             650380
  Total transitive includes:                    19255
  Total include cost (lines):                29446938
```

### DIFF
--- a/src/groups/mqb/mqbstat/mqbstat_statcontroller.h
+++ b/src/groups/mqb/mqbstat/mqbstat_statcontroller.h
@@ -27,24 +27,15 @@
 // config) dumping the stats to a dedicated log file.
 
 // MQB
-#include <mqbcmd_messages.h>
-#include <mqbstat_flatjsonprinter.h>
-#include <mqbstat_jsonprinter.h>
-#include <mqbstat_statmonitor.h>
-#include <mqbstat_statsfilelogger.h>
-#include <mqbstat_tableprinter.h>
-
 #include <bmqma_countingallocatorstore.h>
-#include <bmqst_statcontext.h>
 
 // BDE
 #include <ball_log.h>
 #include <bdlbb_blob.h>
 #include <bdlmt_throttle.h>
-#include <bdlmt_timereventscheduler.h>
 #include <bsl_functional.h>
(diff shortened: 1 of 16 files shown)


src/groups/mqb/mqbstat/mqbstat_statcontroller.h
```diff
@@ -356,9 +353,9 @@ class StatController {
     /// object.  Return zero on success or a nonzero value otherwise.
     /// The specified `encoding` parameter controls whether the result should
     /// be in a text format or in a JSON.
-    int processCommand(mqbcmd::StatResult*                  result,
-                       const mqbcmd::StatCommand&           command,
-                       const mqbcmd::EncodingFormat::Value& encoding);
+    int processCommand(mqbcmd::StatResult*        result,
+                       const mqbcmd::StatCommand& command,
+                       int                        encoding);
```
678098, line 350: Casting enum to int is the only way to have a forward declaration to `mqbcmd` here

hallfox: This kind of seems worse? Is weaking this type constraint worth removing the header include?

678098: To be fair, we do this almost all the time when we return exit codes from functions.
This is not a narrowing conversion, it is opposite.

678098: Replaced with 2 bool flags

hallfox: I think the untyped int is probably better than 2 bool flags if we're not going include this header at all.

678098: Returned int

hallfox (approved):

PR #1541, opened 2026-06-16
Test: Fuzz test for message properties

jakeski2015:
**Describe your changes**
Added a fuzz test that fuzzes over different fields in the first message property header and the message properties header of PUT messages. It opens a TCP connection to the broker, does the negotiation with the broker, and then fires PUT messages with many mutations of message property header fields. 

**Additional context**
The fuzz test currently does not fuzz test the actual properties, only some of the headers. Next steps are in the bottom of the put_message_properties.py file. This PR serves as a first PR towards building a more thorough and long running fuzz test.


src/python/blazingmq/schemas/broker.py
```diff
@@ -130,6 +130,33 @@ class PutHeaderFlags(IntFlag):
...
+            "subscriptions": [
+                {
+                    "sId": 0,
+                    "expression": {
+                        "version": "E_VERSION_1",
+                        "text": "x > 0",
+                    },
+                    "consumers": [
+                        {
+                            "maxUnconfirmedMessages": 0,
```
pniedzielski: Do I understand right that these values are defaults?  If so, let's make `maxUnconfirmedMessages` and `maxUnconfirmedBytes` something like 100 or such so that we're sent messages when we're not fuzzing these fields.

src/python/blazingmq/dev/fuzztest/put_message_properties.py
```diff
@@ -0,0 +1,419 @@
...
+    )
+    session.connect(put)
+
+    try:
+        session.fuzz(max_depth=2)
+    finally:
+        conn.shutdown()
+
+
+# todo: (specific to fuzz testing message properties)
```
pniedzielski: Probably best not to have `todo`s in our code when PRs get merged.  Definitely fine to keep them around locally.

src/python/blazingmq/dev/fuzztest/put_message_properties.py
```diff
@@ -0,0 +1,419 @@
...
+    )
+    event_contents = PaddingBlock(
+        "event_contents",
+        children=[event_desc, flags_field, message_size, message],
+    )
+
+    return [event_size, event_contents]
+
+
+class _PersistentConnection(boofuzz.TCPSocketConnection):
```
pniedzielski: This is a very useful class.  I think it's a good idea to pull it out into its own file as `class PersistentConnection`.

src/python/blazingmq/dev/fuzztest/put_message_properties.py
```diff
@@ -0,0 +1,419 @@
...
+    )
+
+    put = boofuzz.Request(
+        "Put",
+        children=(make_put_with_fuzzable_properties()),
+    )
+    session.connect(put)
+
+    try:
+        session.fuzz(max_depth=2)
```
pniedzielski: I wonder if there's a good way to pull this value out in some way, so we could run the same fuzz tests on each PR and nightly, but with different `max_depth`s?

pniedzielski (changes requested):

src/python/blazingmq/schemas/broker.py
```diff
@@ -130,6 +130,33 @@ class PutHeaderFlags(IntFlag):
...
+                    "sId": 0,
+                    "expression": {
+                        "version": "E_VERSION_1",
+                        "text": "x > 0",
+                    },
+                    "consumers": [
+                        {
+                            "maxUnconfirmedMessages": 100,
+                            "maxUnconfirmedBytes": 10000,
+                            "consumerPriority": -2147483648,
```
678098: The idea behind these schema types is that they represent valid messages by default, so, if messages are sent in the correct order, the state of the connection on the broker side changes all the time

pniedzielski: @jakeski2015 Does it make sense to have a separate consumer process rather than have the producer and consumer be the same process?  That would avoid needing to set these maxUnconfirmed* to something nonzero to actually test what you're trying to test.

jakeski2015: Yea I think it makes sense, and sounds reasonable. I still don't quite understand why we would need it though, unless the client won't try to consume it's own messages.

pniedzielski (approved):

PR #1542, opened 2026-06-17
Fix[bmqex::SystemExecutor]: remove static singleton

678098:
## Problem

The `bmqex::SystemExecutor_Context` singleton might be destroyed during static cleanup, causing an assertion failure if anything constructed a `bmqex::SystemExecutor` afterward (static destruction order fiasco). 

## Fix

Replace the singleton with reference-counted shared ownership via `bsl::shared_ptr`. 
 
## Changes

Lifetime management: 
- Context is now heap-allocated via `bsl::allocate_shared` and tracked by a `bsl::weak_ptr`
- Each `bmqex::SystemExecutor` holds a `bsl::shared_ptr<bmqex::SystemExecutor_Context>` — context lives as long as any executor exists 
- Bookkeeping (`weak_ptr`, `Mutex`) is heap-allocated inside `BSLMT_ONCE_DO` and intentionally never freed, making `acquireContext()` safe during static destruction 
 
Thread safety: 
- Each spawned thread stores a `bsl::shared_ptr` in its `ThreadData`, preventing context destruction while the thread runs 
- The `onThreadRun` function copies this `bsl::shared_ptr` to a local before destroying ThreadData, keeping the context alive
through `finalizeThread`
 
Simplifications:
- Removed `d_unjoinedThreads`, `d_threadExitCondition`, and the waiting loop in the destructor — unnecessary because the `bsl::shared_ptr` guarantees all threads have finalized before the destructor can run 
- Removed `bmqex::SystemExecutorContextGuard` class (redundant with `bsl::shared_ptr`) 
- Removed `singleton()`, `initSingleton()`, `shutdownSingleton()` — replaced by single `acquireContext()` factory 
- Destructor just joins the last completed thread

### DIFF
--- a/src/groups/bmq/bmqex/bmqex_executor.h
+++ b/src/groups/bmq/bmqex/bmqex_executor.h
@@ -161,12 +161,6 @@ class Executor_TargetBase {
   public:
     // ACCESSORS
 
-    /// Return `e1 == e2`, where `e1` and `e2` are contained executor
-    /// targets (of type `E`) of `*this` and `other` respectively. The
-    /// behavior is undefined unless `typeid(e1) == typeid(e2)` is `true`.
-    virtual bool
-    equal(const Executor_TargetBase& other) const BSLS_KEYWORD_NOEXCEPT = 0;
-
     /// Return a pointer to the contained executor target.
     virtual void*       target() BSLS_KEYWORD_NOEXCEPT       = 0;
     virtual const void* target() const BSLS_KEYWORD_NOEXCEPT = 0;
@@ -224,10 +218,6 @@ class Executor_Target : public Executor_TargetBase {
 
   public:
     // ACCESSORS
-    bool equal(const Executor_TargetBase& other) const BSLS_KEYWORD_NOEXCEPT
-        BSLS_KEYWORD_OVERRIDE;
-    // Implements 'Executor_TargetBase::equal()'.
-
     void* target() BSLS_KEYWORD_NOEXCEPT BSLS_KEYWORD_OVERRIDE;
 
     /// Implements `Executor_TargetBase::target()`.
@@ -293,11 +283,6 @@ class Executor_Box_SboImp {
     /// size of the on-stack buffer used to store the executor target.
     struct Dummy {
         void* d_padding[4];
-
-        bool operator==(const Dummy&) const BSLS_KEYWORD_NOEXCEPT
-        {
-            return false;
-        }
 
         void post(const bsl::function<void()>&) const
         {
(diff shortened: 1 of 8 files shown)


src/groups/bmq/bmqex/bmqex_systemexecutor.cpp
```diff
@@ -249,32 +114,64 @@ SystemExecutor_Context::allocator() const BSLS_KEYWORD_NOEXCEPT
...
+}
+
+// MANIPULATORS
+SystemExecutor& SystemExecutor::operator=(
+    bslmf::MovableRef<SystemExecutor> original) BSLS_KEYWORD_NOEXCEPT
+{
+    SystemExecutor& ref = bslmf::MovableRefUtil::access(original);
+    d_context_sp        = bslmf::MovableRefUtil::move(ref.d_context_sp);
+    d_threadAttributes  = bslmf::MovableRefUtil::move(ref.d_threadAttributes);
+    return *this;
```
hallfox: I don't think we need the allocator-extended move constructor here. Since both of these fields are `shared_ptr`, the default special constructors should suffice.

src/groups/bmq/bmqex/bmqex_systemexecutor.h
```diff
@@ -303,18 +255,31 @@ class SystemExecutor {
...
-    // submitted functors. Note that to meet the `noexcept` requirements
-    // for the executor copy and move constructors, this object is shared
-    // between instances.
+    // submitted functors.
     bsl::shared_ptr<bslmt::ThreadAttributes> d_threadAttributes;
 
+  private:
+    // NOT IMPLEMENTED
+    SystemExecutor(const SystemExecutor&) BSLS_KEYWORD_DELETED;
+    SystemExecutor& operator=(const SystemExecutor&) BSLS_KEYWORD_DELETED;
```
hallfox: Do we need to hide the copy constructors?

678098: We have 2 options:
1. Keep copy/assignment, then we cannot perform threads join check in `SystemExecutor` and have to do it in possibly shared `SystemExecutor_Context`.
2. Remove, then we have to update all code that copies `SystemExecutor` to pass a `shared_ptr` to it.

src/groups/bmq/bmqex/bmqex_systemexecutor.h
```diff
@@ -358,52 +324,20 @@ class SystemExecutor {
   public:
     // ACCESSORS
 
-    /// Return a pointer to the executor's associated tread attributes, or a
+    /// Return a shared pointer to this executor's execution context.
+    const bsl::shared_ptr<SystemExecutor_Context>&
+    context() const BSLS_KEYWORD_NOEXCEPT;
```
hallfox: Are we exposing this for tests?

678098: Yes. Thinking of this, some tests are useless and might be removed for simplification.

src/groups/bmq/bmqex/bmqex_executor.h
```diff
@@ -600,17 +571,6 @@ inline void Executor_Target<EXECUTOR>::move(void* dst) BSLS_KEYWORD_NOEXCEPT
 }
 
 // ACCESSORS
-template <class EXECUTOR>
-inline bool Executor_Target<EXECUTOR>::equal(
```
678098, line 604: It is not very useful to have `==` or `!=` for executors

hallfox (approved):

PR #1544, opened 2026-06-17
Fix: Remove `typename` outside of template context

pniedzielski:
In C++03, `typename` can only occur within a template context.  When used outside of template context, `typename` causes warnings or errors on C++03-only compilers, and are not necessary.  This patch removes them.


hallfox (approved):

pniedzielski:
Put retransmission test failure is not related to this change.

PR #1546, opened 2026-06-17
UT[bmqst::TableUtil]: modernize tests, remove unused

678098:


src/groups/bmq/bmqst/bmqst_tableutil.h
```diff
@@ -154,22 +45,6 @@ struct TableUtil {
...
-                              const TableInfoProvider&                info);
-
-    /// Print the specified `table` to the specified
-    /// `stream` as a csv with one line for the header, and one
-    /// line per row.
-    static void printCsv(bsl::ostream& stream, const BaseTable& table);
-
-    /// Print the specified `info` to the specified `stream` in CSV format
-    /// with one line for the header, and one line per row.
-    static void printCsv(bsl::ostream& stream, const TableInfoProvider& info);
```
678098, line 172: These 3 functions are not used, removed.
Also removed test for it

src/groups/bmq/bmqtst/bmqtst_testhelper.h
```diff
@@ -546,6 +546,9 @@
 //                                 TEST SHELL
 // ----------------------------------------------------------------------------
 #define TEST_PROLOG(F)                                                        \
+    /* NOLINTBEGIN(cppcoreguidelines-avoid-magic-numbers) */                  \
```
678098: `case 1: test1_` etc are magic numbers according to linter

src/groups/bmq/bmqtst/bmqtst_testhelper.h
```diff
@@ -546,6 +546,9 @@
...
+    /* NOLINTBEGIN(cppcoreguidelines-pro-bounds-pointer-arithmetic) */        \
```
678098: `const int _testCase                      = argc > 1 ? atoi(argv[1]) : 0;` is a pointer arithmetic "problem" according to linter

src/groups/bmq/bmqtst/bmqtst_testhelper.h
```diff
@@ -546,6 +546,9 @@
...
+    /* NOLINTBEGIN(bugprone-branch-clone) */                                  \
```
678098:
```
        if (bmqtst::TestHelperUtil::verbosityLevel() == 1) {                  \
            _logSeverity = ball::Severity::e_INFO;                            \
        }                                                                     \
        else if (bmqtst::TestHelperUtil::verbosityLevel() == 2) {             \
            _logSeverity = ball::Severity::e_DEBUG;                           \
        }                                                                     \
        else if (bmqtst::TestHelperUtil::verbosityLevel() >= 3) {             \
            _logSeverity = ball::Severity::e_TRACE;                           \
        } 
```
is a "bug" according to linter

src/groups/bmq/bmqtst/bmqtst_testhelper.h
hallfox: The alternative for pointer arithmetic here is supposed to be `bsl::span`, so to make this lint go away we'd need to convert `argv` into span like this:

```c++
bsl::span<const char*> argvSpan(argv, argc);
```

src/groups/bmq/bmqtst/bmqtst_testhelper.h
hallfox: That's really weird, the check is supposed to identify *identical* branches: https://clang.llvm.org/extra/clang-tidy/checks/bugprone/branch-clone.html

These are clearly not identical.

678098: I think it triggers to pattern
```
if (smth) {
   var = ...
} else {
   var = ...
}
```
Had something similar fixed in another PR

hallfox: It's supposed to only trigger if `var = ...` are the same in both branches. It's really bizarre that we're seeing it here. I'm not able to see the same issue when running clang-tidy on my machine. I think it's fine to leave the nolint in for the time being.

src/applications/bmqstoragetool/m_bmqstoragetool_commandprocessorfactory.t.cpp
```diff
@@ -97,7 +97,7 @@ static void test2_cslProcessorTest()
 //                                 MAIN PROGRAM
 // ----------------------------------------------------------------------------
 
-int main(int argc, char* argv[])
+int main(int argc, const char* argv[])
```
678098: `span` requires `const char**`, it is 202 files changed. @hallfox

678098: `benchmark` requires non-const `char**`

```
BENCHMARK_EXPORT void Initialize(int* argc, char** argv,
                                 void (*HelperPrintf)() = PrintDefaultHelp);
```

678098: I think this is not worth it

src/groups/bmq/bmqtst/bmqtst_testhelper.h
678098: This conflicts with libbenchmark initialize that expects `char**`

src/applications/bmqstoragetool/m_bmqstoragetool_commandprocessorfactory.t.cpp
hallfox: Can't we initialize a `span` from a `char**`?

hallfox (approved):

PR #1547, opened 2026-06-17
Feat: Authorizer plugin API

hallfox:
This breaks ground on the authorization plugin API. This is mainly up for review by the project maintainers to discuss API design.

The main focus is:
```C++
    virtual bool authorize(const Action&               action,
                           const AuthenticationResult& authnResult) = 0;
```

Where `Action` encapsulates a set of possible choices of resources and the actions that can be taken with those resources.

### DIFF
--- a/src/groups/mqb/CMakeLists.txt
+++ b/src/groups/mqb/CMakeLists.txt
@@ -20,8 +20,7 @@ set_property(SOURCE "mqba/mqba_application.cpp"
              APPEND
              PROPERTY COMPILE_DEFINITIONS "BMQ_BUILD_TYPE=${CMAKE_BUILD_TYPE}")
 
-set(MQB_PRIVATE_PACKAGES mqba mqbblp mqbc mqbcmd mqbconfm mqbevt mqbi mqbmock mqbnet mqbs mqbsi mqbsl mqbu)
-target_bmq_style_uor( mqb PRIVATE_PACKAGES ${MQB_PRIVATE_PACKAGES})
+target_bmq_style_uor( mqb )
 
 # Turn off warnings in generated code.
 if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "(Apple)?Clang")
--- a/src/groups/mqb/mqbauthz/mqbauthz_basicauthorizer.h
+++ b/src/groups/mqb/mqbauthz/mqbauthz_basicauthorizer.h
@@ -1,0 +1,104 @@
+// Copyright 2026 Bloomberg Finance L.P.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef INCLUDED_MQBAUTHN_BASICAUTHORIZER
+#define INCLUDED_MQBAUTHN_BASICAUTHORIZER
+
+/// @file mqbauthn_basicauthorizer.h
+///
+/// @brief Provide a basic authorizer plugin that uses username and
+/// password.
+///
+/// @bbref{mqbauthn::BasicAuthorizer} provides a built-in authorizer
+/// plugin that authorizes all actions. It's designed to represent the current
+/// authorization policy, which allows all clients access.
+///
+/// @bbref{mqbauthn::BasicAuthorizerPluginFactory} is the corresponding factory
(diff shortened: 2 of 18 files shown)


src/groups/mqb/mqbauthz/mqbauthz_basicauthorizer.h
```diff
@@ -0,0 +1,127 @@
...
+
+// FORWARD DECLARATION
+
+namespace mqbauthz {
+
+// ========================
+// class BasicAuthorizer
+// ========================
+
+class BasicAuthorizer : public mqbplug::Authorizer {
```
chrisbeard, line 55: Is this something like a default, built-in authorizer?

src/groups/mqb/mqbplug/mqbplug_authorizer.h
```diff
@@ -0,0 +1,565 @@
...
+                       const Action_ExecuteAdminCommand& rhs)
+{
+    return !(lhs == rhs);
+}
+
+// ===================
+// class Action
+// ===================
+
+class Action {
```
678098: I don't like that the sources autogenerated from xsd schemas are very large, however, this might be the place where it is justified.
What about moving `Action` and its types to a new `xsd`, then generating sources automatically?
We can also forward include this in this header to avoid including it everywhere.

src/groups/mqb/mqbplug/mqbplug_authorizer.g.cpp
```diff
@@ -0,0 +1,210 @@
...
+{
+    bsl::allocator<> alloc = bmqtst::TestHelperUtil::allocator();
+    mqbplug::Action_ExecuteAdminCommand obj1("test1");
+
+    EXPECT_TRUE(obj1 == obj1);
+    EXPECT_FALSE(obj1 != obj1);
+
+    mqbplug::Action_ExecuteAdminCommand obj2("test2");
+    EXPECT_TRUE(obj1 != obj2);
+    EXPECT_FALSE(obj1 == obj2);
```
678098: With xsd-autogenerated sources, we don't need to have this test file

chrisbeard (commented):
I like the API, it feels like it'll be practical from the plugin side.

Aside from the codegen/allocator bit, the only question I have is `BasicAuthorizer`; do we want to stick with that name for the time being? Maybe we revisit it when we get to the actual broker bits where the authorizer is plugged in by default.

hallfox:
@chrisbeard @678098 I added an alternate version of the actions API that use the XSD generator to create the action VSTs, along with an authorize API equivalent. Let me know what you think about going with this approach.

src/groups/mqb/mqbauthz/mqbauthz_basicauthorizer.h
```diff
@@ -0,0 +1,104 @@
...
+  public:
+    // CLASS DATA
+
+  private:
+    // CLASS-SCOPE CATEGORY
+    BALL_LOG_SET_CLASS_CATEGORY("MQBAUTHN.BASICAUTHORIZER");
+
```
678098, line 62:
```suggestion
```
There is nothing being logged here, can remove scope category

src/groups/mqb/mqbauthz/mqbauthz_basicauthorizer.h
```diff
@@ -0,0 +1,104 @@
...
+///
+/// @bbref{mqbauthn::BasicAuthorizerPluginFactory} is the corresponding factory
+/// classes for the Authorizer plugin.
+
+// MQB
+#include <mqbplug_authenticator.h>
+#include <mqbplug_authorizer.h>
+
+// BDE
+#include <ball_log.h>
```
678098, line 36:
```suggestion
```

src/groups/mqb/mqbplug/mqbplug_authorizer.h
```diff
@@ -0,0 +1,104 @@
...
+#include <bslstl_inplace.h>
+
+namespace BloombergLP {
+
+// FORWARD DECLARATION
+
+namespace mqbcfg {
+class AuthorizerPluginConfig;
+}
+
```
678098, line 44:
```suggestion
```
Not used in the header

src/groups/mqb/CMakeLists.txt
```diff
@@ -20,8 +20,7 @@ set_property(SOURCE "mqba/mqba_application.cpp"
              APPEND
              PROPERTY COMPILE_DEFINITIONS "BMQ_BUILD_TYPE=${CMAKE_BUILD_TYPE}")
 
-set(MQB_PRIVATE_PACKAGES mqba mqbblp mqbc mqbcmd mqbconfm mqbevt mqbi mqbmock mqbnet mqbs mqbsi mqbsl mqbu)
-target_bmq_style_uor( mqb PRIVATE_PACKAGES ${MQB_PRIVATE_PACKAGES})
```
chrisbeard, line 24: Only one question: what's the context for this changing?

chrisbeard: Ah, this is from old days when `mqb` was shipped as a library. Good by me.

chrisbeard (approved):

PR #1548, opened 2026-06-17
Fix[test]: flaky tests

dorjesinpo:
`set_app_ids` is NOT blocking and needs synchronization
`test_open_queue_while_cluster_blips_quorum_and_kill_all_non_leader` does not work with divergence

### DIFF
--- a/src/integration-tests/test_restart_between_modes.py
+++ b/src/integration-tests/test_restart_between_modes.py
@@ -840,7 +840,13 @@ def test_restart_between_legacy_and_fsm_add_remove_app(
     confirm_one_message(consumer, fanout_queue, "foo")
 
     current_app_ids = DEFAULT_APP_IDS + ["quux"]
-    cluster.set_app_ids(current_app_ids, du)
+    cluster.set_app_ids(current_app_ids, du, succeed=None)
+
+    # need to make sure the config has made it to the primary since this is a
+    # non-blocking op.
+    assert cluster.last_known_leader.capture(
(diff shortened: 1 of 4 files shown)


src/integration-tests/test_restart_between_modes.py
kaikulimu, line 1009: `set_app_ids` is also called here, and in other places of this Python file. Either we add the `cluster.last_known_leader.capture` to all the other places, or we modify `cluster::set_app_ids` to include the capture functionality. I believe modifying `cluster::set_app_ids` is the correct approach, with the benefit that all calls to `cluster::set_app_ids` across all Python tests will be synchronized.

kaikulimu: It seems you captured the `Adding VirtualStorage for ...` while ignored the `Removing VirtualStorage for ...`. Is there a particular reason?

kaikulimu (changes requested):

hiSandog:
For a PR named broadly around flaky tests, it would help reviewability to include a small table mapping each changed test to the suspected nondeterminism: timing, shared state, ordering, or resource cleanup. If any fix relies on longer waits, a repeated local run or stress target in CI would make it clearer that the flake is actually removed rather than made less likely. The best regression here is usually deterministic time/control hooks instead of more generous sleeps.

dorjesinpo:
closing in favor of blocking reconfigure admin command - implemented in raft refactoring as polling registered storages

PR #1549, opened 2026-06-18
CI: don't wait for full build to lint

678098:
CI: don't wait for full build to lint

- Moved clang-tidy job from build.yaml to `lint.yaml`, depending on configure-only step instead of full build             
- Split build step into separate `Configure` and `Build`, allowing configure to run standalone                              
- Fixed dependency caching: added `/opt/bb` to cache paths, included C++ standard in cache key                            
- GoogleTest: install to `/opt/bb`, dropped `sudo`

### DIFF
--- a/.github/workflows/build-ubuntu.yaml
+++ b/.github/workflows/build-ubuntu.yaml
@@ -1,7 +1,7 @@
 name: Build Ubuntu
 
 description: |
-  This is a reusable workflow to build BlazingMQ on Ubuntu. 
+  This is a reusable workflow to build BlazingMQ on Ubuntu.
   Optionally run unit tests if the target is "all.t".
   Given a ref and a target, this workflow will build the project and will save the build results both as artifact and in cache.
   The cache key is generated based on the ref and the target, and is returned via workflow output.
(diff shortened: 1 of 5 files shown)


hallfox (dismissed):

.github/workflows/dependencies.yaml
```diff
@@ -13,10 +13,6 @@ on:
         description: "GH Actions Cache key associated with the built dependencies"
         value: ${{ jobs.build_dependencies.outputs.cache_key }}
 
-concurrency:
-  group: global-${{ inputs.cxx_standard }}
-  cancel-in-progress: false
```
678098, line 18: https://github.com/orgs/community/discussions/5435
In short, this `concurrency` setting is what makes our jobs skipped sometimes without making them pending

.github/workflows/build-ubuntu.yaml
```diff
@@ -87,7 +87,9 @@ jobs:
       - uses: actions/cache/restore@v4
         if: steps.build-cache-restore-step.outputs.cache-hit != 'true' # Variable type is string, thus using quotes
         with:
-          path: deps
+          path: |
+            deps
+            /opt/bb
```
678098, line 92: We create this dir and install everything there, so by restoring it we can avoid calling `build_deps.sh` when we use this cache

.github/workflows/dependencies.yaml
```diff
@@ -29,13 +25,15 @@ jobs:
       - name: Generate dependencies cache key
         id: get-cache-key
         run: |
-          echo "cache_key=deps-`date +%y-%m-%d`-`cat docker/build_deps.sh | shasum`" >> $GITHUB_OUTPUT
+          echo "cache_key=deps-${{ inputs.cxx_standard }}-`date +%y-%m-%d`-`cat docker/build_deps.sh | shasum`" >> $GITHUB_OUTPUT
```
678098, line 28: There was a hidden bug here: collision between cpp23 and cpp03 deps

hallfox (approved):

PR #1550, opened 2026-06-18
CI: Add github action for running fuzz tests nightly

jakeski2015:


.github/workflows/nightly-fuzz-cron.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Download broker build
+        uses: actions/download-artifact@v5
+        with:
+          name: ${{ needs.build.outputs.artifact_key }}
+
+      - name: Prepare broker runtime files
+        run: |
+          BROKER_DIR=build/blazingmq/src/applications/bmqbrkr
+          find ./build -type f -name "*.tsk" -exec chmod +x {} +
+          cp src/applications/bmqbrkr/run "$BROKER_DIR/run"
```
pniedzielski, line 52: I think this is a good catch.  From the looks of it, `conftest.py` for the fuzz test does not double check that `./run` exists in the broker directory.  I think I used that as a crunch, because it also sets up the directory structure the broker needs.

For now, let's keep this in this GH Actions, but I think it's worth thinking about whether to make `conftest.py` more robust or not depend on `run`.

pniedzielski (approved):
Let's give this a shot.  We can try to run it tonight and see if it does what we expect.

PR #1551, opened 2026-06-18
Fix[bmqeval]: correct not_equal_to with single boolean literal in makeComparison

### LINKED ISSUE #1523: Boolean literal optimization in makeComparison does not account for not_equal_to
The one literal simplification in makeComparison applies equal_to semantics for both equal_to and not_equal_to. A subscription expression like expr != true will match when the property is true rather than when it is false.

The isEquality gate at bmqeval_simpleevaluator.h:1082 enters the optimization for both operators, and the both literals case at line 1093 correctly dispatches through Op<bool>(). But the one literal branches at lines 1100 and 1109 return the expression unchanged when the literal is true and negate it when the literal is false, which is correct for equal_to but inverted for not_equal_to.

The parser at bmqeval_simpleevaluatorparser.y:147 confirms NE reaches this path.

SrinathhSatuluri:
Closes #1523

makeComparison optimizes boolean literal comparisons at compile time. The two-literal case correctly delegates to Op<bool>(), but the one-literal case was shared across both operators under a single isEquality branch, hardcoding equal_to semantics regardless of the actual Op. This meant expr != true silently simplified to expr rather than !expr, with no compilation error to surface the problem.

The fix splits isEquality into two self-contained branches with the one-literal returns inverted in the not_equal_to block so that a true literal produces Not(expr) and a false literal returns expr directly.

### DIFF
--- a/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.h
+++ b/src/groups/bmq/bmqeval/bmqeval_simpleevaluator.h
@@ -1072,20 +1072,17 @@ CompilationContext::makeComparison(ExpressionPtr a, ExpressionPtr b)
     }
 
     // Handle comparisons with boolean literals at compile time.
-    //   - for 'expr = true' and 'true = expr' return 'expr'
-    //   - for 'expr = false' and 'false = expr' return '!expr'
+    //   - for 'expr == true'  and 'true == expr'  return 'expr'
+    //   - for 'expr == false' and 'false == expr' return '!expr'
+    //   - for 'expr != true'  and 'true != expr'  return '!expr'
+    //   - for 'expr != false' and 'false != expr' return 'expr'
     //   - if both sides are boolean literals, return
-    //   'BooleanLiteral(a && b)'
+    //   'BooleanLiteral(Op(a, b))'
 
     typedef SimpleEvaluator::BooleanLiteral BooleanLiteral;
     typedef SimpleEvaluator::Not            Not;
 
-    // Booleans can only be compared for (in)equality.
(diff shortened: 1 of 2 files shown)


src/groups/bmq/bmqeval/bmqeval_simpleevaluator.h
```diff
@@ -1117,6 +1114,38 @@ CompilationContext::makeComparison(ExpressionPtr a, ExpressionPtr b)
...
+
+        if (boolean_a) {
+            if (boolean_b) {
+                return ExpressionPtr(
+                    new (*d_allocator) BooleanLiteral(
+                        Op<bool>()(boolean_a->value(), boolean_b->value())),
+                    d_allocator);  // RETURN
+            }
+
+            if (boolean_a->value()) {
```
SrinathhSatuluri, line 1131: Returns are inverted here compared to the equal_to block above

src/groups/bmq/bmqeval/bmqeval_simpleevaluator.t.cpp
```diff
@@ -542,6 +542,10 @@ static void test3_evaluation()
         {"true && b_true", true},
         {"b_true == true", true},
         {"b_true < true", false},
+        {"b_true != true", false},
+        {"b_true != false", true},
+        {"b_false != true", true},
+        {"b_false != false", false},
```
678098, line 548: Run these test cases locally and they are all failing in the main branch indeed

678098 (approved):
Looks good

678098:

PR #1556, opened 2026-06-22
feat: Dev/storagewriter

dorjesinpo:
⏺ Summary of storagetool functionality changes:

  1. Journal JSON output always shows details — `SearchResultFactory` (not just when `--details` is specified).
  2. Exact-match searches always show details — `SearchExactMatchResult` (`--guid`, `--seqnum`, `--offset`) now always prints full record details in all modes.
  Previously without `--details` only printed GUIDs.
  3. Payload dump supported in JSON mode — removed the `"not supported for Json mode"` restriction. In JSON mode, payload is embedded as a hex "Payload" field in the record.
  In human mode, payload is still dumped separately.
  4. CSL JSON output always includes the advisory payload — no `--details` needed.
  5. CSL advisory payload is proper JSON — Record field is a nested JSON object via `baljsn::Encoder`.
  6. Keys are hex in CSL JSON — instead of base64.
  7. Null AppKey is "0000000000" — instead of `"** NULL **"`.
  
  
  ⏺ bmqstoragewriter — writes BlazingMQ storage files from storagetool JSON output.

  Usage: bmqstoragewriter [options]

  Options:
  - -i|--input — journal JSON input
  - -c|--csl-input — CSL JSON input
  - -o|--output-dir — output directory (default: .)
  - -p|--partition — partition ID (default: 0)
  - --prefix — filename prefix (default: bmq_)

  Behavior:
  1. Creates journal/data/qlist files if they don't exist (with proper headers).
  2. If `--csl-input` provided: writes CSL records to a .bmq_csl file and populates an in-memory queue cache (QueueKey → URI, AppIds).
  3. If `--input` provided: appends records to journal/data/qlist files. For `QUEUE_OP` `CREATION`, looks up queue URI and AppIds from cache; falls back to hex key as URI if not
  in cache.
  4. CSL is processed before journal so the cache is available.
  5. If no options are provided, prints help.
  6. If neither `--input` nor `--csl-input` is provided (regardless of other options), just creates empty journal/data/qlist files.
  7. Appending to existing CSL file warns if input LogId differs from file's LogId.

### DIFF
--- a/src/applications/bmqstoragetool/integration-tests/data/queueop_result.txt
+++ b/src/applications/bmqstoragetool/integration-tests/data/queueop_result.txt
@@ -8,7 +8,7 @@
     Timestamp       : 29OCT2024_14:01:06.000000
     Epoch           : 1730210466
     QueueKey        : 26DACDC974
-    AppKey          : ** NULL **
+    AppKey          : 0000000000
     QueueOpType     : CREATION
     QLIST OffsetWords: 9
 
--- a/src/applications/bmqstoragetool/integration-tests/test_journalfile.py
+++ b/src/applications/bmqstoragetool/integration-tests/test_journalfile.py
@@ -77,8 +77,9 @@ def test_short_json(storagetool, journal_file):
         assert res.returncode == EX_OK
         json_res = json.loads(res.stdout)
         assert json_res["MessageRecords"] == "2"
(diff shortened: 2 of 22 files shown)


kaikulimu (approved):
Since bmqstoragetool is not production critical, I only did a brief review.

PR #1557, opened 2026-06-23
Perf[bmqc]: stop MQTP queues in parallel

### LINKED ISSUE #1028: Improvement: stop MQTP queues in parallel
`bmqc::MultiQueueThreadPool::stop` iterates over all processing queues and enqueues a monitor event to request stop, then wait each queue to stop serially.
It might be more efficient to split this loop in two loops.
In the first loop, send a monitor event to the queue and disable push.
In the second loop, wait for all queues to finish using the shared `TimedSemaphore`.

Implementing this causes some tests to fail. We might have hidden assumptions how the stop is done, and it's worth investigating.

The code:

https://github.com/bloomberg/blazingmq/blob/47a89d55abf9dc0ad0707bc0afab74c757033f6c/src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h#L1161-L1182

SrinathhSatuluri:
Closes: #1028

Split the serial stop loop in `MultiQueueThreadPool::stop()` into two phases so all queues drain concurrently rather than sequentially. The first loop signals every queue (pushBack NULL, disablePushBack), the second waits on each queue's `d_finished` semaphore. This is safe because each queue's shutdown is self-contained: its own semaphore, ref count, and backing SingleConsumerQueue with no cross-queue state. The ref count nets to zero per queue (+1 enqueue monitor, -1 release initial reference), so finish order is irrelevant.

### DIFF
--- a/src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
+++ b/src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
@@ -865,6 +865,7 @@ inline void MultiQueueThreadPool<TYPE>::stop()
             &d_warningMonitorEventHandle);
     }
 
+    // Signal all queues to stop.
     for (size_t i = 0; i < d_queues.size(); ++i) {
         QueueInfo& info = *d_queues[i];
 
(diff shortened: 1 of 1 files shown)


678098 (approved):
Looks good

678098:
The shutdown issue in sanitizers build is fixed here: https://github.com/bloomberg/blazingmq/pull/1563

678098:

PR #1562, opened 2026-06-23
Refactor[bmq,mqb]: un-template request managers

678098:
- `bmqp::RequestManager` and `mqbnet::MultiRequestManager` always have the same template args `REQUEST == RESPONSE == bmqp_ctrlmsg::ControlMessage`. The only reason why it is a template is for not including `#include <bmqp_ctrlmsg_messages.h>`. However, we still include this header. Due to this, template args are not doing anything other than making the code more complex.
- It is possible to remove `#include <bmqp_ctrlmsg_messages.h>` in both class sources, but this should be done not with templates, but with forward declarations and code organization. This PR is a preparation for this.
- Simplify arg types: `RequestManagerType::RequestSp` -> `RequestSp`

### DIFF
--- a/src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
+++ b/src/groups/bmq/bmqimp/bmqimp_brokersession.cpp
@@ -1011,9 +1011,8 @@ void BrokerSession::QueueFsm::setQueueState(
     queue->setState(value);
 }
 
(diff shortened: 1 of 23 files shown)


pniedzielski (approved):

pniedzielski:
This is a Good Change.

PR #1571, opened 2026-06-24
Refactor: User authentication credentials callback

hallfox:
This change removes the ostream& parameter from the user provided authentication credentials callback. This parameter was originally intended to allow the user to provide a message about why credentials failed to be provided in the case of an error, but this was just immediately logged by us. Instead of doing that, we are preferring the user log any helpful error messages themselves.


hallfox:
The linter failure with cppcoreguidelines-init-variables on `nulloptCb` is junk, that check isn't even supposed to fire on non-primitive types.

chrisbeard (approved):

pniedzielski (approved):

PR #1572, opened 2026-06-24
Remove: bmqvt_rcdescriptionerror

hallfox:
This component is unused.


chrisbeard (approved):

678098 (approved):
We can probably simplify it more.
`bmqvt_valueorerror` is only used in `mqba_domainmanager` and might be replaced with a local struct
`bmqvt_propertybag` is used only in `bmqio` and in components that work with `bmqio`, we can just move `bmqvt_propertybag` -> `bmqio_propertybag`
This allows to remove `bmqvt` package fully

hallfox:
I have dreams of making `bmqvt_valueorerror` more specifically useful for BMQ code, but as of right now it's basically an allocator-aware `std::expected`-lite, and we definitely don't use it prolifically. `bmqvt_propertybag` scares me.

PR #1573, opened 2026-06-25
Fix[mqbnet]: data race on d_isReading and d_isReadEnabled in ClusterImp

### LINKED ISSUE #1568: Data race on d_isReading and d_isReadEnabled in mqbnet::ClusterImp
ClusterNodeImp::d_isReading is a plain bool written from the transport thread in setChannel, the dispatcher thread in enableRead, and the IO thread in resetChannel, with a check-then-act in enableRead that allows concurrent calls to both pass the guard and double-enable the channel. ClusterImp::d_isReadEnabled has a similar issue, written without the mutex in enableRead but read outside d_mutex in notifyObserversOfNodeStateChange.

SrinathhSatuluri:
Closes #1568

`ClusterNodeImp::d_isReading` is a plain `bool` written from the transport thread in `setChannel`, the dispatcher thread in `enableRead`, and the IO thread in `resetChannel`. The check-then-act in `enableRead` has a TOCTOU window that allows concurrent calls to both pass the guard and double-enable the channel. `ClusterImp::d_isReadEnabled` is written without `d_mutex` in `enableRead` but read outside `d_mutex` in `notifyObserversOfNodeStateChange`. Convert both to `bsls::AtomicBool` and close the TOCTOU with `testAndSwap`.

@678098

### DIFF
--- a/src/groups/mqb/mqbnet/mqbnet_clusterimp.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_clusterimp.cpp
@@ -94,7 +94,7 @@ bool ClusterNodeImp::enableRead()
 
     BSLS_ASSERT_SAFE(d_readCb);
 
-    if (d_isReading) {
+    if (d_isReading.testAndSwap(false, true) != false) {
         return true;  // RETURN
     }
 
@@ -109,10 +109,10 @@ bool ClusterNodeImp::enableRead()
 
         channelSp->close();
 
+        d_isReading = false;
+
         return false;  // RETURN
     }
-
-    d_isReading = true;
 
     BALL_LOG_INFO << nodeDescription() << ": reading from the channel '"
                   << channelSp->peerUri() << "'";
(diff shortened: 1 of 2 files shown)


src/groups/mqb/mqbnet/mqbnet_clusterimp.cpp
```diff
@@ -109,11 +109,11 @@ bool ClusterNodeImp::enableRead()
 
         channelSp->close();
 
+        d_isReading = false;
```
SrinathhSatuluri, line 112: `testAndSwap` sets `d_isReading` to true before `read()` executes, so we reset on failure. This has the side effect of blocking concurrent `enableRead` calls during the `read()` window.

src/groups/mqb/mqbnet/mqbnet_clusterimp.cpp
```diff
@@ -94,7 +94,7 @@ bool ClusterNodeImp::enableRead()
 
     BSLS_ASSERT_SAFE(d_readCb);
 
-    if (d_isReading) {
+    if (d_isReading.testAndSwap(false, true) != false) {
```
678098:
```suggestion
    if (d_isReading.swap(true)) {
```

This might be a shorter alternative. This is not a performance critical code so there is no need to optimize cache writes

678098 (dismissed):
Looks good

678098 (approved):

678098:

PR #1576, opened 2026-06-26
Fix[mqbblp]: race between replica recovery and open queue

678098:
Replica only checks that primary has recovered its partition before sending open queue upstream.
If replica's own partition is not recovered fully, it might receive open queue response and try to open a queue on a partition that is not operable yet, leading to recovery failures.

recovery_race_diagram

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
@@ -1145,10 +1145,23 @@ void ClusterQueueHelper::processOpenQueueRequest(
             }
         }
         else {
-            // We are a replica for the queue, make sure it has a unique
-            // upstream id associated, if we haven't already done so, assign
-            // one now.
+            // We are a replica for the queue.  If the local partition
+            // hasn't finished recovery, buffer the request until
+            // restoreState() drains it after recovery completes.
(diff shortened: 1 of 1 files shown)


src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
```diff
@@ -1145,10 +1145,23 @@ void ClusterQueueHelper::processOpenQueueRequest(
             }
         }
         else {
-            // We are a replica for the queue, make sure it has a unique
-            // upstream id associated, if we haven't already done so, assign
-            // one now.
+            // We are a replica for the queue.  If the local partition
```
kaikulimu: Would be more inline with the check above if we did `if (d_clusterData.membership().selfNodeStatus() != bmqp_ctrlmsg::NodeStatus::E_AVAILABLE)`.

We could combine this check with above to say `if (hasActiveAvailablePrimary(pid) && isSelfAvailable) {`

kaikulimu (changes requested):

src/groups/mqb/mqbblp/mqbblp_clusterqueuehelper.cpp
678098: Added a broader check for node-level availability as you proposed

kaikulimu (approved):

PR #1577, opened 2026-06-26
Test: Added fuzztest that fuzzes the PUT message property values.

jakeski2015:
The original fuzz test doesn't account for the actual message properties, it only fuzzes over the fields. This test should cover potentially more interesting cases where the message property is a string of length 0, a super large string, large integers, etc. Max depth 2 is on to explore additional combinations of the fields and property values.

### DIFF
--- a/src/python/blazingmq/dev/fuzztest/put_message_properties.py
+++ b/src/python/blazingmq/dev/fuzztest/put_message_properties.py
@@ -82,8 +82,11 @@ def _make_mph_fields(prefix, prop_type, name_len, middle_value, fuzzable=False):
 
 def make_message_properties_area(
     num_properties: int = 2,
-    fuzz_header: bool = False,
     fuzz_mph: bool = False,
+    fuzz_data: bool = False,
+    fuzz_byte0: bool = False,
+    fuzz_area_words: bool = False,
+    fuzz_num_properties: bool = False,
     new_style: bool = False,
 ) -> boofuzz.Request:
     """
@@ -105,18 +108,18 @@ def make_message_properties_area(
             name="properties_hdr_byte0",
             default_value=packed_byte0,
             full_range=True,
-            fuzzable=fuzz_header,
+            fuzzable=fuzz_byte0,
         ),
         boofuzz.Byte(
             name="area_words_upper",
             default_value=(area_words >> 16) & 0xFF,
             fuzzable=False,
         ),
-        boofuzz.Word(  # double check
+        boofuzz.Word(
             name="area_words_lower",
             default_value=area_words & 0xFFFF,
             endian=">",
-            fuzzable=fuzz_header,
+            fuzzable=fuzz_area_words,
         ),
         boofuzz.Byte(
             name="properties_hdr_reserved",
@@ -127,7 +130,7 @@ def make_message_properties_area(
             name="num_properties",
             default_value=num_properties,
             full_range=True,
-            fuzzable=fuzz_header,
+            fuzzable=fuzz_num_properties,
         ),
     ]
 
@@ -147,9 +150,7 @@ def make_message_properties_area(
                 prop_type,
                 len(name),
                 middle,
-                fuzzable=(
-                    fuzz_mph and i == 0
-                ),  # only fuzz the first message property header to save time, might want to fuzz more
+                fuzzable=fuzz_mph,
             )
         )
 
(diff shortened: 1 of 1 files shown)


src/python/blazingmq/dev/fuzztest/put_message_properties.py
```diff
@@ -343,13 +409,45 @@ def fuzz_properties(host: str, port: int, max_depth: int = 2) -> None:
...
+            fuzz_header=False,
+            fuzz_mph=False,
+            fuzz_data=True,
+            fuzz_byte0=True,
+            fuzz_area_words=True,
+        ),
+    )
+
+
+# todo: (specific to fuzz testing message properties)
```
pniedzielski: I think we should make sure not to have TODOs in the code.  We can keep track of these elsewhere.

src/python/blazingmq/dev/fuzztest/put_message_properties.py
```diff
@@ -84,6 +86,9 @@ def make_message_properties_area(
     num_properties: int = 2,
     fuzz_header: bool = False,
     fuzz_mph: bool = False,
+    fuzz_data: bool = False,
```
pniedzielski: Is there a reason to have this much flexibility?  Maybe just being able to turn on and off message property fuzzing is good enough.  I'd expect if we fuzz message properties we fuzz everything, generally, especially if this is only running nightly.

src/python/blazingmq/dev/fuzztest/put_message_properties.py
```diff
@@ -313,21 +327,73 @@ def make_put_with_fuzzable_properties(
     return [event_size, event_contents]
 
 
-def fuzz_properties(host: str, port: int, max_depth: int = 2) -> None:
-    """
-    Launch a long-running fuzzing session targeting message properties in PUT
-    messages at depth 2 (all pairs of fuzzable fields).
-    NOTE: This currently only fuzzes the first property's header fields to limit the combinatorial explosion, but can be extended to fuzz more.
+class _PersistentConnection(boofuzz.TCPSocketConnection):
```
pniedzielski: Why did we duplicate this here?  Isn't it in `blazingmq.dev.fuzztest.persistent_connection`?

src/python/blazingmq/dev/fuzztest/put_message_properties.py
```diff
@@ -343,13 +409,45 @@ def fuzz_properties(host: str, port: int, max_depth: int = 2) -> None:
...
-    put = boofuzz.Request(
-        "Put",
-        children=(make_put_with_fuzzable_properties()),
-    )
+    put = boofuzz.Request("Put", children=put_sequence)
     session.connect(put)
 
     try:
-        session.fuzz(max_depth=max_depth)
+        session.fuzz(max_depth=2)
```
pniedzielski: Could still be worth passing in `max_depth` like before.

src/python/blazingmq/schemas/broker.py
```diff
@@ -130,7 +130,7 @@ class PutHeaderFlags(IntFlag):
     },
 }
 
-CONFIGURE_STREAM_WITH_CONSUMER_SUBSCRIPTION_SCHEMA: SchemaDescription = {
+CONFIGURE_STREAM_PROPERTY_EXPRESSION_SCHEMA: SchemaDescription = {
```
pniedzielski: I think the name we had before is probably better.

src/python/blazingmq/schemas/broker.py
```diff
@@ -163,9 +163,9 @@ class PutHeaderFlags(IntFlag):
         "qId": 0,
         "streamParameters": {
             "subIdInfo": {"subId": 0, "appId": "__default"},
-            "maxUnconfirmedMessages": 0,
```
pniedzielski, line 166: How come we're changing this?  I think these should stay as the default, since they run on all fuzz tests that configure queue.

pniedzielski (changes requested):
Initial pass through

src/python/blazingmq/dev/fuzztest/put_message_properties.py
pniedzielski: I think this comment still holds.  Maybe we do want to have this much flexibility, I'm not sure.  But `fuzz_mph`, `fuzz_data`, `fuzz_byte0`, `fuzz_area_words`, and `fuzz_num_properties` form a lot of knobs we can turn.  Do we need that many?

pniedzielski (approved):

pniedzielski (dismissed):
Merge conflict here, double-check it, but this looks good to me.

pniedzielski (approved):
Will just let CI run, then I'll merge

PR #1578, opened 2026-06-29
IT: Added fuzztest for PUT message property fields to nightly github action workflow 

jakeski2015:
Changed the yaml file to include a target to the fuzztest, and added a matrix entry for the PUT message property fields fuzztest. Also added broker logs to artifact.


.github/workflows/nightly-fuzz-cron.yaml
```diff
@@ -26,17 +26,29 @@ jobs:
...
+          - request: confirm
+            pytest_target: "test_fuzz.py -k 'test_fuzz_request[confirm]'"
+          - request: close_queue
+            pytest_target: "test_fuzz.py -k 'test_fuzz_request[close_queue]'"
+          - request: disconnect
+            pytest_target: "test_fuzz.py -k 'test_fuzz_request[disconnect]'"
+          - request: negotiation_bypass_authentication
+            pytest_target: "test_fuzz.py -k 'test_fuzz_request[negotiation_bypass_authentication]'"
+          - request: put_message_properties
+            pytest_target: "test_fuzz_put_message_properties.py -k 'test_fuzz_put_message_properties'"
```
pniedzielski: Might be worth doing something like this to reduce duplication: `pytest test_file.py::test_func[arg1,arg2]`.
https://docs.pytest.org/en/stable/how-to/usage.html

Or maybe try to put test_fuzz_put_message_properties in test_fuzz.py?

Crucially the way this PR does it, the run UI will include both of these two variables in the output, which makes it very hard to see what the test case really is.  Compared with before, it will just have the request type.

pniedzielski (changes requested):

pniedzielski (approved):

pniedzielski:
I'll let CI run then merge this

PR #1580, opened 2026-06-30
Fix: Add missing definition for createSessionEvent

hallfox:
Missed this implementation in #1579. Added a section in the test driver to ensure both overloads are used.

### DIFF
--- a/src/groups/bmq/bmqa/bmqa_mocksession.t.cpp
+++ b/src/groups/bmq/bmqa/bmqa_mocksession.t.cpp
@@ -278,6 +278,24 @@ static void test1_staticMethods()
             bmqt::SessionEventType::e_CONNECTED,
             bmqt::CorrelationId(1),
             0,
+            errorDescription,
+            bmqtst::TestHelperUtil::allocator());
+        bmqa::SessionEvent sessionEvent = event.sessionEvent();
+
+        BMQTST_ASSERT_EQ(sessionEvent.type(),
+                         bmqt::SessionEventType::e_CONNECTED);
+        BMQTST_ASSERT_EQ(sessionEvent.statusCode(), 0);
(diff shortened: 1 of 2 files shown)


src/groups/bmq/bmqa/bmqa_mocksession.t.cpp
```diff
@@ -290,6 +308,7 @@ static void test1_staticMethods()
         BMQTST_ASSERT_EQ(sessionEvent.correlationId(), bmqt::CorrelationId(1));
     }
 
+
```
pniedzielski: `clang-format` warns about this

src/groups/bmq/bmqa/bmqa_mocksession.t.cpp
```diff
@@ -340,6 +359,29 @@ static void test1_staticMethods()
         bmqa::QueueId       queueId(1);
         bmqt::CorrelationId corrId(1);
 
+        bmqa::Event event = bmqa::MockSessionUtil::createQueueSessionEvent(
```
pniedzielski, line 361: Was this a test case we forgot from before?

pniedzielski (commented):
LGTM, I'll approve after the clang-format issue is fixed.

src/groups/bmq/bmqa/bmqa_mocksession.t.cpp
chrisbeard:
```suggestion
```

src/groups/bmq/bmqa/bmqa_mocksession.t.cpp
hallfox: No, we just changed it to use the extra parameter.

chrisbeard (approved):

PR #1582, opened 2026-07-01
Test: Added fuzztest for Event::Initialize method

jakeski2015:
This fuzz test found an input that fails during safe and debug builds, which could potentially crash the broker if it was reached. However, in practice this code path isn't reachable because bmqio::ChannelUtil::handleRead frames incoming packets before they reach bmqp::Event. The framing layer reads the first 4 bytes of the stream as a 4 byte length field (EventHeader.d_fragmentBitAndLength), then accumulates bytes until the blob is exactly that declared length before passing it to Event::initialize(). This guarantees the assertion d_header->length() == blob->length() at bmqp_event.h:316 always holds in production.

It may be worthwhile to edit this fuzz test to meet this precondition. Mainly putting this here because I didn't see anything about this precondition documented anywhere in the code.

### DIFF
--- a/src/standalones/s_bmqfuzz/s_bmqfuzz_bmqp_event_initialize.fuzz.cpp
+++ b/src/standalones/s_bmqfuzz/s_bmqfuzz_bmqp_event_initialize.fuzz.cpp
@@ -1,0 +1,34 @@
+// Copyright 2026 Bloomberg Finance L.P.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


src/standalones/s_bmqfuzz/s_bmqfuzz_bmqp_event_initialize.fuzz.cpp
```diff
@@ -0,0 +1,34 @@
+// Copyright 2025 Bloomberg Finance L.P.
```
pniedzielski: `2026`

pniedzielski:
Linter failure fixed by #1586.

pniedzielski (approved):

PR #1583, opened 2026-07-01
refactor: enhance m_bmqstoragetool_parameters.h

### LINKED ISSUE #1507: Refactor: enhance `m_bmqstoragetool_parameters.h`
https://github.com/bloomberg/blazingmq/blob/main/src/applications/bmqstoragetool/m_bmqstoragetool_parameters.h

This class has a few places that might be improved:

1. Minimize includes

These includes are not needed in header, since they are not used there:

```
// MQB
#include <mqbs_datafileiterator.h>
#include <mqbs_filestoreprotocol.h>
#include <mqbs_journalfileiterator.h>
#include <mqbs_mappedfiledescriptor.h>

// BMQ
#include <bmqu_stringutil.h>
```

Other includes (in BDE section) might also be unused in the header.
If they are not used, they should be removed.
It is possible that other headers rely on includes in this file, so they should be updated to include needed directly.

2. There is a block of constants declared in the header. These constants are only used internally in `cpp` file and should be removed from the class API:

```
    /// Record types constants
    static const char* k_ALL_TYPE;
    static const char* k_MESSAGE_TYPE;
    static const char* k_QUEUEOP_TYPE;
    static const char* k_JOURNALOP_TYPE;
    static const char* k_CSL_ALL_TYPE;
    static const char* k_CSL_SNAPSHOT_TYPE;
    static const char* k_CSL_UPDATE_TYPE;
    static const char* k_CSL_COMMIT_TYPE;
    static const char* k_CSL_ACK_TYPE;
    /// Print modes constants
    static const char* k_HUMAN_MODE;
    static const char* k_JSON_PRETTY_MODE;
    static const char* k_JSON_LINE_MODE;
```

The constants should go to unnamed namespace in its `cpp` file.
Also, the type should be changed to `const bsl::string_view`, for example (other existing constant in the codebase):
```
/// Control Sequence Introducer
const bsl::string_view k_CSI = "\x1b[";
```

3. Classes `Parameters`, `CommandLineArguments` declared in the source miss allocator traits. Should have traits similar to (other existing example in the codebase):

```
    // TRAITS
    BSLMF_NESTED_TRAIT_DECLARATION(CslSearchOffsetDecorator,
                                   bslma::UsesBslmaAllocator)
```

4. Class API accepts allocator on construction.
However, many methods also require allocator, for example:
```
    bool validate(bsl::string* error, bslma::Allocator* allocator = 0);
    void validateCslModeArgs(bsl::ostream&     stream,
                             bslma::Allocator* allocator = 0);
    void validateQueueNames(bslma::Allocator* allocator = 0) const;
```

A more clear way is to keep `d_allocator_p` field in each class that require it, init it in constructor and use it in the corresponding methods. Allocator arg can be removed from them after this.

Mauricio0704:
Closes #1507

- Remove unused includes from the header and add direct includes to printer.h and searchresult.h
- Move record-type and print-mode constants into an unnamed namespace in the .cpp as bsl::string_view
- Add allocator traits to CommandLineArguments and Parameters
- Store d_allocator_p in each class and remove allocator argument used in other functions

@678098 could you take a look when you have a moment? Thanks!

### DIFF
--- a/src/applications/bmqstoragetool/m_bmqstoragetool_printer.h
+++ b/src/applications/bmqstoragetool/m_bmqstoragetool_printer.h
@@ -27,6 +27,10 @@
 #include <m_bmqstoragetool_messagedetails.h>
 #include <m_bmqstoragetool_parameters.h>
 
+// MQB
+#include <mqbs_datafileiterator.h>
+#include <mqbs_journalfileiterator.h>
+
 // BMQ
 #include <bmqt_messageguid.h>
 
--- a/src/applications/bmqstoragetool/m_bmqstoragetool_printer.h
+++ b/src/applications/bmqstoragetool/m_bmqstoragetool_printer.h
@@ -27,10 +27,6 @@
 #include <m_bmqstoragetool_messagedetails.h>
 #include <m_bmqstoragetool_parameters.h>
 
-// MQB
-#include <mqbs_datafileiterator.h>
-#include <mqbs_journalfileiterator.h>
-
 // BMQ
 #include <bmqt_messageguid.h>
 
@@ -43,6 +39,12 @@
 #include <bslma_managedptr.h>
 
 namespace BloombergLP {
+
+// FORWARD DECLARATIONS
+namespace mqbs {
+class DataFileIterator;
+class JournalFileIterator;
+}
 
 namespace m_bmqstoragetool {
 
--- a/src/applications/bmqstoragetool/m_bmqstoragetool_searchresult.h
+++ b/src/applications/bmqstoragetool/m_bmqstoragetool_searchresult.h
@@ -61,7 +61,9 @@
 #include <m_bmqstoragetool_printer.h>
 
 // MQB
+#include <mqbs_datafileiterator.h>
 #include <mqbs_filestoreprotocolprinter.h>
+#include <mqbs_journalfileiterator.h>
 
 // BDE
 #include <bsl_iostream.h>
--- a/src/applications/bmqstoragetool/m_bmqstoragetool_searchresult.h
+++ b/src/applications/bmqstoragetool/m_bmqstoragetool_searchresult.h
@@ -60,10 +60,7 @@
 #include <m_bmqstoragetool_payloaddumper.h>
 #include <m_bmqstoragetool_printer.h>
 
-// MQB
-#include <mqbs_datafileiterator.h>
 #include <mqbs_filestoreprotocolprinter.h>
-#include <mqbs_journalfileiterator.h>
 
 // BDE
 #include <bsl_iostream.h>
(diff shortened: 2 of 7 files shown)


kaikulimu:
@dorjesinpo Will this conflict with your storage writer?

src/applications/bmqstoragetool/m_bmqstoragetool_printer.h
```diff
@@ -27,6 +27,10 @@
 #include <m_bmqstoragetool_messagedetails.h>
 #include <m_bmqstoragetool_parameters.h>
 
+// MQB
+#include <mqbs_datafileiterator.h>
+#include <mqbs_journalfileiterator.h>
+
```
678098:
```suggestion
```
These includes might be replaced with forward declarations because the types are only used as pointers in this header

src/applications/bmqstoragetool/m_bmqstoragetool_searchresult.h
```diff
@@ -61,7 +61,9 @@
 #include <m_bmqstoragetool_printer.h>
 
 // MQB
+#include <mqbs_datafileiterator.h>
 #include <mqbs_filestoreprotocolprinter.h>
+#include <mqbs_journalfileiterator.h>
```
678098:
```suggestion
#include <mqbs_filestoreprotocolprinter.h>
```
These 2 includes can also be replaced with forward declarations to reduce the burden of including too much

Mauricio0704:
Thanks for the feedback. I've pushed the changes. The MQB iterators are now forward-declared in the headers, and the includes moved to the .cpp files that dereference them.

678098:
Hi @Mauricio0704, the UI shows unresolved merge conflicts. Could you rebase your branch to the latest main and solve them?

Mauricio0704:
Hi @678098, I resolved the conflict by dropping the JSON-mode validation that #1556 removed. The PR shows mergeable now. Thanks!

src/applications/bmqstoragetool/m_bmqstoragetool_searchresult.h
```diff
@@ -77,7 +76,12 @@
 #include <bsls_keyword.h>
 
 namespace BloombergLP {
-namespace m_bmqstoragetool {
+
```
678098, line 79:
```suggestion
namespace m_bmqstoragetool {

```

Mauricio0704: Thanks. Restored the m_bmqstoragetool namespace after the mqbs forward declarations so those stay in BloombergLP::mqbs, matching printer.h

678098 (approved):

678098:
Thank you for contribution @Mauricio0704!

kaikulimu:
Thank you @Mauricio0704 !

PR #1584, opened 2026-07-01
Build: fetch_git bde last in `build_deps.sh`

678098:
`bde` repo clone might take more time than the other repos, make sure to clone it last.


pniedzielski (approved):
Good idea

PR #1587, opened 2026-07-06
Misc[mqbs::FileStore]: Add logging for record writes

kaikulimu:
## Summary
  - Add INFO-level logging to the record-write methods in `mqbs::FileStore` to improve observability of storage operations on both the primary and replica paths.
  - Only journal/queue-op-style records are logged. High-frequency **message-related** records (MessageRecord, ConfirmRecord, DeletionRecord) are intentionally **not** logged at all to avoid flooding logs — no throttling mechanism is introduced.
  - Each log line consistently carries PSN (primary lease id + sequence number), journal offset, and record-type-specific fields (queueKey, appKey, guid, type, etc.).
  - Extend the existing log in `FileStoreUtil::writeQueueCreationRecordImpl` to include PSN and file offsets.

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_filestore.cpp
+++ b/src/groups/mqb/mqbs/mqbs_filestore.cpp
@@ -3756,6 +3756,12 @@ void FileStore::writeQueueOpRecordImpl(DataStoreRecordHandle*  handle,
                                       bmqp::StorageMessageType::e_QUEUE_OP,
                                       RecordType::e_QUEUE_OP,
                                       recordOffset);
+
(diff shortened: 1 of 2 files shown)


chrisbeard:
These feel like they should be DEBUG/TRACE logs, not INFO. Said another way, I don't think these logs are what you want in production, so they shouldn't be visible in production.

kaikulimu:
Per discussion during team meeting: journal/queue op records will remain at the INFO level. All the message-related records will not be logged at all, hence I will remove all the `BMQ_LOGTHROTTLE_INFO` lines.

pniedzielski (approved):

PR #1588, opened 2026-07-07
Test: Fuzztest for AckMessageIterator

jakeski2015:


src/standalones/s_bmqfuzz/s_bmqfuzz_bmqp_ackmessageiterator.fuzz.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+using namespace BloombergLP;
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
+{
+    if (size < sizeof(bmqp::EventHeader)) {
+        return 0;
+    }
+
+    bslma::Allocator*              alloc = bslma::Default::defaultAllocator();
```
pniedzielski, line 33: It's definitely fine to call this out explicitly.  In allocator-aware types, not passing the allocator pointer defaults to using the default allocator.

I don't have strong opinions either way, there's no problem with this as is.

src/standalones/s_bmqfuzz/s_bmqfuzz_bmqp_ackmessageiterator.fuzz.cpp
```diff
@@ -0,0 +1,55 @@
...
+    bdlbb::PooledBlobBufferFactory bufferFactory(1024, alloc);
+
+    bdlbb::Blob blob(&bufferFactory, alloc);
+    bdlbb::BlobUtil::append(&blob,
+                            reinterpret_cast<const char*>(data),
+                            static_cast<int>(size));
```
pniedzielski, line 39: If `size` is too large, this can become negative, violating the precondition of `append`: ("The behavior is undefined unless the range `[source, source + length)` is valid memory, and `length <= dest->totalSize() - dest->length()` or `0 != dest->factory()`.")  I think we can exit early around line 28 if `size > bsl::numeric_limits<int>::max()`.

pniedzielski (changes requested):
Looks good except for the one real comment below.  This component is tricky, the beginning of the blob needs to be the same as the event header we pass in to `reset`.  Your test does that correctly.

src/standalones/s_bmqfuzz/s_bmqfuzz_bmqp_ackmessageiterator.fuzz.cpp
jakeski2015: At least in my personal use, libfuzzer defaults to 4096 as the max for size, which I'm assuming is significantly less than the bsl max int. With that said though I think you are right in terms of code correctness because if the default limit is removed by whoever runs the fuzz test then the concern you mentioned will occur.

pniedzielski (dismissed):
Merge conflict, probably caused by merging an other fuzz test.  Should be easy to solve.

pniedzielski (approved):
Will just let CI run, then I'll merge

PR #1589, opened 2026-07-07
Test: Fuzztest for PushMessageIterator

jakeski2015:
This fuzztest found a broker crash on safe builds through PUSH messages. I haven't tried to reproduce it yet because it's iterator logic for PUSH messages.


src/standalones/s_bmqfuzz/s_bmqfuzz_bmqu_stringutil_match.fuzz.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2025 Bloomberg Finance L.P.
+// Copyright 2026 Bloomberg Finance L.P.
```
pniedzielski, line 1: Generally these sorts of changes should be in separate PRs, since they don't relate to the actual PR.  Yeah, we missed it originally, but a 1 line PR is still fine.

pniedzielski (dismissed):

pniedzielski (approved):

pniedzielski:
This Python Formatting Check might be failing on `main` too...not going to block this unrelated PR on that.

PR #1590, opened 2026-07-07
Test: Fuzz PutMessageIterator

jakeski2015:
This test found a memory leak in PutMessageIterator


src/standalones/s_bmqfuzz/s_bmqfuzz_bmqp_putmessageiterator.fuzz.cpp
```diff
@@ -0,0 +1,67 @@
+// Copyright 2025 Bloomberg Finance L.P.
```
pniedzielski: 2026

src/standalones/s_bmqfuzz/s_bmqfuzz_bmqp_putmessageiterator.fuzz.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include <bdlbb_blobutil.h>
+#include <bdlbb_pooledblobbufferfactory.h>
+#include <bsl_cstring.h>
+
+using namespace BloombergLP;
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
+{
+    if (size < sizeof(bmqp::EventHeader) + 1) {
```
pniedzielski, line 26: https://github.com/bloomberg/blazingmq/pull/1588/changes/32da7c6b1f445e0cfbc0befab023f7f3d4481d28#diff-2c1e2373ef5a65b65b17e63145b62c2e6ca64fa07714fd009132f389bfcff6daR28
These two fuzz tests differ here.  Should they be the same, or is there a good reason?

jakeski2015: This is because we have a decompress flag as the first byte. The flag gets pulled out of the input in the next lines.

pniedzielski (approved):

PR #1592, opened 2026-07-07
CI: Make clang-tidy checks not fail the lint pipeline

hallfox:
This makes the noisy clang-tidy checks not cause the CI pipeline to turn red while we're still figuring out how to use it effectively.


pniedzielski (dismissed):
I think this is a good thing to do, but I'm not sure it will fix the red cross on failing PRs?

pniedzielski (dismissed):

chrisbeard (approved):
lgtm!

PR #1596, opened 2026-07-08
Refactor: Remove unused async bind code

hallfox:
bmqio_ntcchannel has this old #ifdef macro that seemed to be in place wishfully for a world where listen was a completely asynchronous operation. We don't have any plans to make this happen soon, and it's just kind of noisy as it is. We can add this back later if we choose to redesign this part of the channel stack.


pniedzielski (approved):
Linter checks for using `constexpr` rather than macros are aspirational at best in a C++03 world.  LGTM.

PR #1597, opened 2026-07-08
Feat: Add TLS parameters to bmqtool schema

hallfox:
This change adds a new set of schema types for bmqtool


pniedzielski (approved):
It's okay to add elements here rather than at the end, since we don't need compatibility over the wire; this XSD serves just to generate vocabulary types for bmqtool.

PR #1601, opened 2026-07-09
Fix: Widen contract of `mqba::ClientSession::processEvent`

pniedzielski:
This patch widens the contract of `mqba::ClientSession::processEvent` to accept authentication events and log an error if they are received. Currently, we check the existing precondition at the call site, but this is somewhat brittle, and a simple code refactoring could open up the opportunity for a broker crash here on protocol input.  Instead, we choose to make this a little more robust.


chrisbeard (dismissed):

pniedzielski:
(Fixed the failing clang-format check)

678098 (approved):

PR #1608, opened 2026-07-10
Fix[bmqp]: assert on zero msgPropsAreaSize

678098:


678098:
TSAN issue is fixed here https://github.com/bloomberg/blazingmq/pull/1609

hallfox (approved):

PR #1610, opened 2026-07-10
Fix[mqbs::FileStore]: Conform journal to cluster state (extra queues)

kaikulimu:
## Summary
  - During Partition-FSM recovery the primary validates its journal against the cluster state. If the journal referenced a queue **absent** from the cluster state — an "extra" queue, e.g. a `QueueOp.CREATION` whose matching deletion never reached this journal — recovery previously failed and the broker terminated (`RECOVERY_FAILURE`), so the primary could not heal.
  - This PR makes the primary treat the cluster state as the source of truth and **conform** its journal instead of erroring: during `open()` it writes a corrective `QueueOp.DELETION` for each extra queue. If the near-full journal can't fit the correctives, conform rolls over once (which compacts the extra queues away, so no DELETION is needed).
  - Corrective records are stamped under the **new primary leaseId**, not the old lease. Continuing the old lease's sequence (e.g. write head `(2,5)` → `(2,6),(2,7)`) risks a cluster-wide PSN collision: a replica that advanced the old lease in a prior term may hold *different* records at those same PSNs. Stamping under the new lease (`(3,1),(3,2)`) makes the primary unambiguously ahead, so its corrections supersede the replica's stale records when pushed. The leaseId is threaded through `DataStore::open` / `FileStore::openInRecoveryMode` (sentinel `0` = replica/legacy, no conform).
  - The corrective writes advance the PSN and are captured by `do_storeSelfSeq` after `open()`, so the existing FSM push path replicates them to replicas automatically — **no FSM state-table changes**. Builds on #1594 (refresh cached self-PSN after opening storage).
  - Scope is limited to **extra** queues (journal has it, cluster state doesn't). The reverse "missing" case (cluster state has it, journal doesn't) is benign at recovery and left for a follow-up; the new-lease bump is structured (a `needsConform` guard) so that follow-up inherits it.

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_filestore.cpp
+++ b/src/groups/mqb/mqbs/mqbs_filestore.cpp
@@ -343,7 +343,8 @@ int FileStore::openInNonRecoveryMode()
 }
 
 int FileStore::openInRecoveryMode(bsl::ostream&    errorDescription,
-                                  QueueKeyInfoMap* queueKeyInfoMap)
+                                  QueueKeyInfoMap* queueKeyInfoMap,
+                                  unsigned int     primaryLeaseId)
 {
     // executed by the *DISPATCHER* thread
 
@@ -1032,11 +1033,21 @@ int FileStore::openInRecoveryMode(bsl::ostream&    errorDescription,
     BSLS_ASSERT_SAFE(d_config.recoveredQueuesCb());
     d_config.recoveredQueuesCb()(d_config.partitionId(), queueKeyInfoMap);
 
-    // Conform the journal to the cluster state.  Each "extra" queueKey found
-    // during recovery (present in the journal but absent from the cluster
-    // state) must be removed.
-    if (!extraQueueKeys.empty()) {
+    // Conform the journal to the cluster state.
+    const bool needsConform = !extraQueueKeys.empty();
+    if (needsConform) {
         BSLS_ASSERT_SAFE(asPrimary);
+        BSLS_ASSERT_SAFE(0 != primaryLeaseId);
+        BSLS_ASSERT_SAFE(primaryLeaseId >= d_writeHeadLeaseId);
+
+        if (primaryLeaseId > d_writeHeadLeaseId) {
+            setWriteHead(primaryLeaseId, 0);
+        }
+    }
+
+    // Each "extra" queueKey found during recovery (present in the journal but
+    // absent from the cluster state) must be removed.
+    if (!extraQueueKeys.empty()) {
         const bsls::Types::Uint64 timestamp =
             bdlt::EpochUtil::convertToTimeT64(bdlt::CurrentTime::utc());
 
@@ -5565,7 +5576,8 @@ FileStore::~FileStore()
 }
 
 // MANIPULATORS
-int FileStore::open(QueueKeyInfoMap* queueKeyInfoMap)
+int FileStore::open(QueueKeyInfoMap* queueKeyInfoMap,
+                    unsigned int     primaryLeaseId)
 {
     // executed by the *DISPATCHER* thread
 
@@ -5603,7 +5615,9 @@ int FileStore::open(QueueKeyInfoMap* queueKeyInfoMap)
     }
 
     bmqu::MemOutStream errorDescription;
-    int rc = openInRecoveryMode(errorDescription, queueKeyInfoMap);
+    int                rc = openInRecoveryMode(errorDescription,
+                                queueKeyInfoMap,
+                                primaryLeaseId);
     if (rc == 0) {
         d_isOpen = true;
     }
--- a/src/groups/mqb/mqbs/mqbs_filestore.cpp
+++ b/src/groups/mqb/mqbs/mqbs_filestore.cpp
@@ -364,7 +364,8 @@ int FileStore::openInRecoveryMode(bsl::ostream&    errorDescription,
         rc_PARTITION_FULL                      = -8,
         rc_CLOSE_FAILURE                       = -9,
         rc_OPEN_FAILURE                        = -10,
-        rc_SYNC_POINT_FAILURE                  = -11
+        rc_SYNC_POINT_FAILURE                  = -11,
+        rc_CONFORM_FAILURE                     = -12
     };
 
     MappedFileDescriptor journalFd;
@@ -816,6 +817,10 @@ int FileStore::openInRecoveryMode(bsl::ostream&    errorDescription,
         asPrimary       = false;
         queueKeyInfoMap = &temp;
     }
+
+    // As primary, collect extra queueKeys (present in the journal but absent
+    // from cluster state).
+    StorageKeys extraQueueKeys(d_allocator_p);
     BALL_LOG_INFO << partitionDesc()
                   << "Attempting to recover messages from the local storage "
                   << (asPrimary ? "as primary." : "as replica.");
@@ -829,6 +834,7 @@ int FileStore::openInRecoveryMode(bsl::ostream&    errorDescription,
                          &jit,
                          &qit,
                          &dit,
+                         asPrimary ? &extraQueueKeys : 0,
                          asPrimary);
     if (0 != rc) {
         BALL_LOG_ERROR << partitionDesc() << "Failed to recover messages from"
@@ -1021,23 +1027,80 @@ int FileStore::openInRecoveryMode(bsl::ostream&    errorDescription,
         }
     }
 
-    if (d_qListAware) {
-        BALL_LOG_INFO << partitionDesc()
-                      << "JOURNAL, QLIST and DATA files will be"
-                      << " written to at these offsets respectively: "
-                      << journalFileOffset << ", " << qlistFileOffset << ", "
-                      << dataFileOffset;
-    }
-    else {
-        BALL_LOG_INFO << partitionDesc() << "JOURNAL and DATA files will be"
-                      << " written to at these offsets respectively: "
-                      << journalFileOffset << ", " << dataFileOffset;
-    }
-
     // Hand over recovered queues to the storage manager *after* files have
     // been successfully opened.
     BSLS_ASSERT_SAFE(d_config.recoveredQueuesCb());
     d_config.recoveredQueuesCb()(d_config.partitionId(), queueKeyInfoMap);
+
+    // Conform the journal to the cluster state.  Each "extra" queueKey found
+    // during recovery (present in the journal but absent from the cluster
+    // state) must be removed.
+    if (!extraQueueKeys.empty()) {
+        BSLS_ASSERT_SAFE(asPrimary);
+        const bsls::Types::Uint64 timestamp =
+            bdlt::EpochUtil::convertToTimeT64(bdlt::CurrentTime::utc());
+
+        FileSet* activeFileSet = d_fileSets[0].get();
+
+        // Space to append one corrective DELETION per extra queue.
+        const bsls::Types::Uint64 requiredSpace =
+            extraQueueKeys.size() * FileStoreProtocol::k_JOURNAL_RECORD_SIZE +
+            k_RESERVED2_PURGE_SIZE;
+
+        if (needRollover(activeFileSet->d_journal.d_file,
+                         activeFileSet->d_journal.d_filePosition,
+                         requiredSpace)) {
+            BALL_LOG_INFO
+                << partitionDesc()
+                << "Conforming journal to cluster state: rolling over "
+                << "to compact away " << extraQueueKeys.size()
+                << " extra queue(s).";
+            rc = rolloverDuringRecovery(timestamp);
+            if (0 != rc) {
+                BALL_LOG_ERROR << partitionDesc()
+                               << "Failed to roll over while conforming "
+                                  "journal to cluster "
+                               << "state, rc: " << rc;
+                return 100 * rc + rc_CONFORM_FAILURE;  // RETURN
+            }
+        }
+        else {
+            for (StorageKeys::const_iterator it = extraQueueKeys.begin();
+                 it != extraQueueKeys.end();
+                 ++it) {
+                BALL_LOG_WARN
+                    << partitionDesc()
+                    << "Conforming journal to cluster state: writing a "
+                    << "corrective QueueOp.DELETION for extra queueKey "
+                    << "[" << *it << "].";
+                rc = writeCorrectiveQueueDeletionDuringRecovery(*it,
+                                                                timestamp);
+                if (0 != rc) {
+                    BALL_LOG_ERROR
+                        << partitionDesc()
+                        << "Failed to write corrective QueueOp.DELETION for "
+                        << "extra queueKey [" << *it << "], rc: " << rc;
+                    return 100 * rc + rc_CONFORM_FAILURE;  // RETURN
+                }
+            }
+        }
+    }
+
+    const FileSet* activeFileSet = d_fileSets[0].get();
+    if (d_qListAware) {
+        BALL_LOG_INFO << partitionDesc()
+                      << "JOURNAL, QLIST and DATA files will be"
+                      << " written to at these offsets respectively: "
+                      << activeFileSet->d_journal.d_filePosition << ", "
+                      << activeFileSet->d_qlist.d_filePosition << ", "
+                      << activeFileSet->d_data.d_filePosition;
+    }
+    else {
+        BALL_LOG_INFO << partitionDesc() << "JOURNAL and DATA files will be"
+                      << " written to at these offsets respectively: "
+                      << activeFileSet->d_journal.d_filePosition << ", "
+                      << activeFileSet->d_data.d_filePosition;
+    }
 
     return rc_SUCCESS;
 }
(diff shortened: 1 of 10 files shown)


kaikulimu:
@678098 This is ready for your review.

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -1021,24 +1028,91 @@ int FileStore::openInRecoveryMode(bsl::ostream&    errorDescription,
...
+
+        if (needRollover(activeFileSet->d_journal.d_file,
+                         activeFileSet->d_journal.d_filePosition,
+                         requiredSpace)) {
+            BALL_LOG_INFO
+                << partitionDesc()
+                << "Conforming journal to cluster state: rolling over "
+                << "to compact away " << extraQueueKeys.size()
+                << " extra queue(s).";
+            rc = rolloverDuringRecovery(timestamp);
```
678098, line 1074: This rollover should remove these extra queues without deletion records, right?
If so, may be possible to simplify the algorithm a bit by just rolling over unconditionally. It will take a bit more time on recovery, but is this extra queues case frequent enough to care about this?

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -1021,24 +1028,91 @@ int FileStore::openInRecoveryMode(bsl::ostream&    errorDescription,
         }
     }
 
+    // Hand over recovered queues to the storage manager *after* files have
+    // been successfully opened.
+    BSLS_ASSERT_SAFE(d_config.recoveredQueuesCb());
+    d_config.recoveredQueuesCb()(d_config.partitionId(), queueKeyInfoMap);
```
678098, line 1035: Are you sure that this callback needs to be called before the following queue deletion work is performed?

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -1049,11 +1123,15 @@ int FileStore::recoverMessages(QueueKeyInfoMap*     queueKeyInfoMap,
                                JournalFileIterator* jit,
                                QlistFileIterator*   qit,
                                DataFileIterator*    dit,
-                               bool                 withCSL)
+                               StorageKeys*         extraQueueKeys,
+                               bool                 withClusterState)
```
678098, line 1132: I think we need to evaluate if this methods requires splitting into 2 separate ones. A strong reason for this is a `bool withClusterState` flag that changes the behavior entirely

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -1324,20 +1406,26 @@ int FileStore::recoverMessages(QueueKeyInfoMap*     queueKeyInfoMap,
...
+                    BALL_LOG_WARN
                         << partitionDesc()
-                        << "Encountered a QueueOp.ADDITION record "
-                        << "for queueKey [" << queueKey
-                        << "] during 1st pass reverse iteration, "
-                        << "but the queueKey is not present in "
-                        << "cluster state.  Record offset: "
-                        << journalIt.recordOffset()
+                        << "Encountered "
+                           "a "
```
678098:
```suggestion
                        << "Encountered a "
```

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -1324,20 +1406,26 @@ int FileStore::recoverMessages(QueueKeyInfoMap*     queueKeyInfoMap,
...
+                        << "QueueOp.ADDITION record for queueKey [" << queueKey
+                        << "] during 1st pass reverse iteration, but the "
+                        << "queueKey is not present in cluster state.  Will "
+                        << "remember this extra queue.  "
+                           "Record "
```
678098:
```suggestion
                        << "remember this extra queue.  Record "
```

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -1385,20 +1473,26 @@ int FileStore::recoverMessages(QueueKeyInfoMap*     queueKeyInfoMap,
...
+                    BALL_LOG_WARN
                         << partitionDesc()
-                        << "Encountered a QueueOp.CREATION record "
-                        << "for queueKey [" << queueKey
-                        << "] during 1st pass reverse iteration, "
-                        << "but the queueKey is not present in "
-                        << "cluster state.  Record offset: "
-                        << journalIt.recordOffset()
+                        << "Encountered "
+                           "a "
```
678098:
```suggestion
                        << "Encountered a "
```

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -1385,20 +1473,26 @@ int FileStore::recoverMessages(QueueKeyInfoMap*     queueKeyInfoMap,
...
+                        << "QueueOp.CREATION record for queueKey [" << queueKey
+                        << "] during 1st pass reverse iteration, but the "
+                        << "queueKey is not present in cluster state.  Will "
+                        << "remember this extra queue.  "
+                           "Record "
```
678098:
```suggestion
                        << "remember this extra queue.  Record "
```
The formatting is odd here

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -1791,15 +1885,24 @@ int FileStore::recoverMessages(QueueKeyInfoMap*     queueKeyInfoMap,
...
+                        // 1st pass and skipped above.  Reaching here means an
+                        // orphan record: the queueKey is absent from the
+                        // cluster state and had no QueueOp.CREATION record in
+                        // the 1st pass.
+                        BMQTSK_ALARMLOG_ALARM("RECOVERY")
                             << partitionDesc()
-                            << "Encountered a QueueOp.PURGE record for "
+                            << "Encountered an orphan QueueOp.PURGE record "
+                               "for "
                             << "queueKey [" << queueKey
```
678098:
```suggestion
                            << "for queueKey [" << queueKey
```

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -1791,15 +1885,24 @@ int FileStore::recoverMessages(QueueKeyInfoMap*     queueKeyInfoMap,
...
                             << "], offset: " << jit->recordOffset()
                             << ", index: " << jit->recordIndex()
-                            << ", but the queueKey is not present in cluster "
-                            << "state.";
+                            << ": the queueKey is not present in the cluster "
+                            << "state and had no QueueOp.CREATION record in "
+                               "the "
+                            << "first pass." << BMQTSK_ALARMLOG_END;
```
678098:
```suggestion
                            << "the first pass." << BMQTSK_ALARMLOG_END;
```

src/groups/mqb/mqbs/mqbs_filestore.t.cpp
```diff
@@ -163,6 +167,81 @@ void recoveredQueuesCb(
...
+
+    mqbs::MappedFileDescriptor journalFd;
+    bmqu::MemOutStream         errDesc(bmqtst::TestHelperUtil::allocator());
+    int rc = mqbs::FileStoreUtil::openFileSetReadMode(errDesc,
+                                                      fileSet,
+                                                      &journalFd);
+    BMQTST_ASSERT_EQ(0, rc);
+    if (0 != rc) {
+        return false;  // RETURN
+    }
```
678098:
```suggestion
    BMQTST_ASSERT_EQ(0, rc);
```
I am not sure if this graceful return is needed.
The test should just fail if rc is not 0

src/groups/mqb/mqbs/mqbs_filestore.t.cpp
```diff
@@ -163,6 +167,81 @@ void recoveredQueuesCb(
...
+
+    mqbs::MappedFileDescriptor journalFd;
+    bmqu::MemOutStream         errDesc(bmqtst::TestHelperUtil::allocator());
+    int rc = mqbs::FileStoreUtil::openFileSetReadMode(errDesc,
+                                                      fileSet,
+                                                      &journalFd);
+    BMQTST_ASSERT_EQ(0, rc);
+    if (0 != rc) {
+        return -1;  // RETURN
+    }
```
678098:
```suggestion
```
Same here: if the test failed, no need to process gracefully
If `BMQTST_ASSERT` continues execution, can replace it with `BSLS_ASSERT_OPT`

src/groups/mqb/mqbs/mqbs_filestore.t.cpp
```diff
@@ -1208,7 +1287,7 @@ static void test3_partitionFullAlarm()
     fs.close();
     BMQTST_ASSERT_EQ(false, fs.isOpen());
 
-    rc = fs.open(0);
+    rc = fs.open(0, 2);
```
678098: This is the only place where `primaryLeaseId == 2` is used as a magic constant

src/groups/mqb/mqbs/mqbs_filestore.t.cpp
```diff
@@ -998,13 +1078,12 @@ static void test2_printTest()
 
     bmqtst::TestHelperUtil::ignoreCheckDefAlloc() = true;
 
-    Tester           tester("./test-cluster123-2");
-    mqbs::FileStore& fs = tester.fileStore();
-    BSLS_ASSERT_OPT(fs.open(0) == 0);
+    Tester             tester("./test-cluster123-2");
+    mqbs::FileStore&   fs             = tester.fileStore();
+    const unsigned int primaryLeaseId = 1;
```
678098: Might reuse the initial value by declaring this constant per file
```suggestion
    const unsigned int k_INITIAL_LEASE_ID = 1;
```

src/groups/mqb/mqbs/mqbs_filestore.t.cpp
```diff
@@ -1394,38 +1473,39 @@ static void test5_writeHeadFollowsAppliedLease()
...
                              journalOffsetWords,
                              dataOffsetDwords,
                              qlistOffsetWords,
                              bmqtst::TestHelperUtil::allocator());
     journalOffsetWords += k_RECORD_WORDS;
-    BMQTST_ASSERT_EQ(1U, fs.writeHeadLeaseId());
+    BMQTST_ASSERT_EQ(primaryLeaseId, fs.writeHeadLeaseId());
     BMQTST_ASSERT_EQ(2ULL, fs.writeHeadSeqNum());
 
+    primaryLeaseId = 2;
```
678098:
```suggestion
    primaryLeaseId = fs.writeHeadLeaseId() + 1;
```
Might use the current lease id to increment

src/groups/mqb/mqbs/mqbs_filestore.t.cpp
```diff
@@ -1543,6 +1625,254 @@ static void test6_leaseTransitionWithoutSeal()
...
+//
+// Testing:
+//   Corrective QueueOp.DELETION during conform without rollover
+// ------------------------------------------------------------------------
+{
+    bmqtst::TestHelper::printTestName("CONFORM EXTRA QUEUE WITHOUT ROLLOVER");
+
+    bmqtst::TestHelperUtil::ignoreCheckDefAlloc() = true;
+
+    const char k_FILE_STORE_LOCATION[] = "./test-cluster123-6";
```
678098:
```suggestion
    const char k_FILE_STORE_LOCATION[] = "./test-cluster123-7";
```

src/groups/mqb/mqbs/mqbs_filestore.t.cpp
```diff
@@ -1543,6 +1625,254 @@ static void test6_leaseTransitionWithoutSeal()
...
+//
+// Testing:
+//   Corrective QueueOp.DELETION during conform causes rollover
+// ------------------------------------------------------------------------
+{
+    bmqtst::TestHelper::printTestName("CONFORM EXTRA QUEUE CAUSES ROLLOVER");
+
+    bmqtst::TestHelperUtil::ignoreCheckDefAlloc() = true;
+
+    const char k_FILE_STORE_LOCATION[] = "./test-cluster123-4";
```
678098:
```suggestion
    const char k_FILE_STORE_LOCATION[] = "./test-cluster123-8";
```

src/groups/mqb/mqbs/mqbs_filestore.t.cpp
kaikulimu: Fixed, now it says `++primaryLeaseId` instead of `2`.

src/groups/mqb/mqbs/mqbs_filestore.cpp
kaikulimu: Yes, this rollover will remove the extra queues. Keep in mind that we will handle other situations which will conform journal to cluster state. For example, in the missing queue case, we need to write new record(s) to the journal and simply rolling over won't solve it. I wrote the code in a more generic way such that it is easy to extend to other situations (missing queue, extra app, missing app, etc.) And it does avoid unnecessary rollover which is nice.

src/groups/mqb/mqbs/mqbs_filestore.cpp
kaikulimu: I agree. `FileStore::recoverMessages` is 1670 lines long, which is one of the longest functions we have, if not the longest. Refactoring will lead to many lines of code change, and should be done in a separate PR. Also, @dorjesinpo might have reservations about us making big refactoring while he is working on `mqbraft`. Therefore, I will delay this effort for now.

kaikulimu: Created https://github.com/bloomberg/blazingmq/issues/1727

src/groups/mqb/mqbs/mqbs_filestore.cpp
kaikulimu: Yes — the rollover path depends on it. Conform can call `rolloverDuringRecovery` → `rolloverImpl`, whose rollover summary does `d_storages.find(queueKey)` + `sit->second->queueUri()` for every queue with outstanding records (`mqbs_filestore.cpp:2880-2890`). Those storages are registered by `recoveredQueuesCb` → `StorageUtil::recoveredQueuesCb` → `fs->registerStorage` (`mqbc_storageutil.cpp:1913`), so running conform first would hit the `BSLS_ASSERT_SAFE` (and dereference an end iterator in opt builds). The in-place deletion branch has no such dependency, but since `needRollover` picks the branch at runtime, the ordering has to hold either way.

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -2275,15 +2375,23 @@ int FileStore::recoverMessages(QueueKeyInfoMap*     queueKeyInfoMap,
...
+                    // An "extra" queue (present in the journal but absent from
+                    // the cluster state) is tracked as deleted in the 1st pass
+                    // and skipped above.  Reaching here means an orphan
+                    // record: the queueKey is absent from the cluster state
+                    // and had no QueueOp.CREATION record in the 1st pass.
+                    BMQTSK_ALARMLOG_ALARM("RECOVERY")
                         << partitionDesc()
-                        << "Encountered a DELETION record for queueKey ["
+                        << "Encountered an orphan DELETION record for "
+                           "queueKey ["
```
678098: Formatting issue

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -2275,15 +2375,23 @@ int FileStore::recoverMessages(QueueKeyInfoMap*     queueKeyInfoMap,
...
                         << rec.queueKey()
                         << "], offset: " << jit->recordOffset()
                         << ", index: " << jit->recordIndex()
-                        << ", but the queueKey is not present in cluster "
-                        << "state.";
+                        << ": the queueKey is not present in the cluster "
+                           "state "
```
678098: Formatting

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -2354,15 +2462,23 @@ int FileStore::recoverMessages(QueueKeyInfoMap*     queueKeyInfoMap,
...
+                    // An "extra" queue (present in the journal but absent from
+                    // the cluster state) is tracked as deleted in the 1st pass
+                    // and skipped above.  Reaching here means an orphan
+                    // record: the queueKey is absent from the cluster state
+                    // and had no QueueOp.CREATION record in the 1st pass.
+                    BMQTSK_ALARMLOG_ALARM("RECOVERY")
                         << partitionDesc()
-                        << "Encountered a CONFIRM record for queueKey ["
+                        << "Encountered an orphan CONFIRM record for queueKey "
+                           "["
```
678098: Formatting

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -2354,15 +2462,23 @@ int FileStore::recoverMessages(QueueKeyInfoMap*     queueKeyInfoMap,
...
                         << rec.queueKey()
                         << "], offset: " << jit->recordOffset()
                         << ", index: " << jit->recordIndex()
-                        << ", but the queueKey is not "
-                        << "present in cluster state.";
+                        << ": the queueKey is not present in the cluster "
+                           "state "
```
678098: Formatting

678098 (dismissed):

kaikulimu:
Fixed all log formatting

678098 (dismissed):

kaikulimu:
@678098 The two latest commits address additional AI findings. They are both small commits. Could you please review again?

678098 (approved):

PR #1613, opened 2026-07-14
Test: Fuzz test for ProtocolUtil::DecodeMessage

jakeski2015:


src/standalones/s_bmqfuzz/s_bmqfuzz_bmqp_protocolutil_decodemessage.fuzz.cpp
```diff
@@ -0,0 +1,65 @@
...
+    data += 2;
+    size -= 2;
+
+    bslma::Allocator*              alloc = bslma::Default::defaultAllocator();
+    bdlbb::PooledBlobBufferFactory bufferFactory(bufferSize, alloc);
+
+    bdlbb::Blob blob(&bufferFactory, alloc);
+    bdlbb::BlobUtil::append(&blob,
+                            reinterpret_cast<const char*>(data),
+                            static_cast<int>(size));
```
pniedzielski, line 48: I think we need a similar check here as in other fuzz tests to prevent `size > numeric_limits<int>`

pniedzielski (changes requested):

src/standalones/s_bmqfuzz/s_bmqfuzz_bmqp_protocolutil_decodemessage.fuzz.cpp
jakeski2015: Good catch

pniedzielski (dismissed):

pniedzielski (approved):

PR #1617, opened 2026-07-15
Refactor[mqbnet]: void* -> NegotiationUserData protocol

678098:
Replace the opaque `void*` negotiation-user-data cookie in the initial-connection/negotiation
path with a typed `mqbnet::NegotiationUserData` protocol, held end-to-end as a `bsl::shared_ptr`.

- `InitialConnectionContext`, `TransportManager`, `TCPSessionFactory: void*/ManagedPtr<void>` -> `shared_ptr<NegotiationUserData>`.
- `mqbblp::ClusterCatalog::NegotiationUserData` implements the protocol.
- Consumer in `mqba::SessionNegotiator` uses a checked `dynamic_cast`.
- Remove dead `d_myNodeId `field.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
@@ -51,6 +51,21 @@
 namespace BloombergLP {
 namespace mqbblp {
 
+// -----------------------------------------
+// class ClusterCatalog::NegotiationUserData
+// -----------------------------------------
+
+ClusterCatalog::NegotiationUserData::NegotiationUserData(
+    bsl::string_view  clusterName,
+    bslma::Allocator* allocator)
+: d_clusterName(clusterName, allocator)
+{
+}
+
+ClusterCatalog::NegotiationUserData::~NegotiationUserData()
+{
+}
+
 // --------------------
 // class ClusterCatalog
 // --------------------
@@ -73,18 +88,15 @@ int ClusterCatalog::createNetCluster(
         connectionMode = mqbnet::TransportManager::e_CONNECT_ALL;
     }
 
-    NegotiationUserData* userData = new (*d_allocator_p) NegotiationUserData;
-    userData->d_clusterName       = name;
-    userData->d_myNodeId          = -1;  // Unused when not reversed connection
-
-    bslma::ManagedPtr<void> userDataMp(userData, d_allocator_p);
+    bsl::shared_ptr<mqbnet::NegotiationUserData> userData =
+        bsl::allocate_shared<NegotiationUserData>(d_allocator_p, name);
 
     return d_transportManager_p->createCluster(errorDescription,
                                                out,
                                                name,
                                                nodes,
                                                connectionMode,
-                                               &userDataMp);
+                                               userData);
 }
 
 struct Named {
--- a/src/groups/mqb/mqbblp/mqbblp_clustercatalog.h
+++ b/src/groups/mqb/mqbblp/mqbblp_clustercatalog.h
@@ -36,6 +36,7 @@
 #include <mqbcfg_messages.h>
 #include <mqbi_cluster.h>
 #include <mqbi_domain.h>
+#include <mqbnet_initialconnectioncontext.h>
 #include <mqbnet_multirequestmanager.h>
 #include <mqbnet_session.h>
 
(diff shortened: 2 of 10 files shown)


src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.h
```diff
@@ -247,21 +264,18 @@ class InitialConnectionContext {
...
+    /// the session.  This may or may not be set by the
+    /// caller, during
+    /// 'TcpSessionFactory::handleInitialConnection()'.
+    /// This is used to bind high level data (from
+    /// application layer) to the application layer (the
+    /// negotiator concrete implementation) (typically for
+    /// the case of 'connect' sessions to provide
+    /// information to use for negotiating the session with
+    /// the remote peer).
+    bsl::shared_ptr<NegotiationUserData> d_userData_sp;
```
678098, line 278: It can be much simpler if we just save `bsl::string d_clusterName` here.
However, this would make `mqbnet` know something about clusters which might be bad for concerns separation.

src/groups/mqb/mqbblp/mqbblp_clustercatalog.h
```diff
@@ -104,15 +107,40 @@ class ClusterCatalog {
...
+        explicit NegotiationUserData(bsl::string_view  clusterName,
+                                     bslma::Allocator* allocator = 0);
+
+        /// Destroy this object.
+        ~NegotiationUserData() BSLS_KEYWORD_OVERRIDE;
+
+        // ACCESSORS
+
+        /// Return the name of the cluster to negotiate the connection with.
+        const bsl::string& clusterName() const;
```
678098, line 143: So much code for `bsl::string` wrapper

src/groups/mqb/mqbblp/mqbblp_clustercatalog.h
```diff
@@ -104,15 +107,40 @@ class ClusterCatalog {
...
-    /// @bbref{ mqba::SessionNegotiator} for usage of that struct.  This
+    /// Class holding some context state used during negotiation: refer to the
+    /// @bbref{ mqba::SessionNegotiator} for usage of that class.  This
     /// `userData` is created here, passed to `mqbnet` to hold on to it and
-    /// deliver it back to the `Negotiator` ~ this hackery mechanism is needed
-    /// in order to avoid dependency cycles and keep `mqbnet` layer abstracted
-    /// away from this logic.
-    struct NegotiationUserData {
-        bsl::string d_clusterName;
-        int         d_myNodeId;
```
678098, line 115: `d_myNodeId` is not used which allows to simplify.
However, `NegotiationUserData` class might still be good if we want to extend it one day

src/groups/mqb/mqbblp/mqbblp_clustercatalog.cpp
```diff
@@ -73,18 +88,15 @@ int ClusterCatalog::createNetCluster(
         connectionMode = mqbnet::TransportManager::e_CONNECT_ALL;
     }
 
-    NegotiationUserData* userData = new (*d_allocator_p) NegotiationUserData;
-    userData->d_clusterName       = name;
-    userData->d_myNodeId          = -1;  // Unused when not reversed connection
-
-    bslma::ManagedPtr<void> userDataMp(userData, d_allocator_p);
```
678098, line 80: The idea was to get rid of direct `new` call, intermediate `ManagedPtr` and `reinterpret_cast`

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.h
```diff
@@ -196,6 +196,23 @@ struct InitialConnectionEvent {
...
+// =========================
+// class NegotiationUserData
+// =========================
+
+/// Protocol for a caller-supplied, opaque piece of data carried through the
+/// negotiation of a session and handed back to the negotiator implementation.
+/// The `mqbnet` layer holds an instance purely as an owned lifetime handle and
+/// never interprets its contents; the concrete type is defined and recovered
+/// by the application layer.
+class NegotiationUserData {
```
hallfox, line 208: Should we have a `clusterName()` method on this class so we don't even need the downcast?

678098: If we ever mention `clusterName` on mqbnet layer, it breaks separation of concerns (it is my upper comment).

If we want to do this, we can simply cache `string d_clusterName` and remove intermediate class.

hallfox (approved):

PR #1618, opened 2026-07-15
Test: Added fuzz test for check sum class

jakeski2015:


pniedzielski (dismissed):
Just to double check, this test runs both variants of the Crc32c calculate, and makes sure that they always match, right?

jakeski2015:
> Just to double check, this test runs both variants of the Crc32c calculate, and makes sure that they always match, right?

Yea basically, it just runs both of the overloads and asserts that they are the same.

pniedzielski (approved):

PR #1621, opened 2026-07-15
Feat: Add authorization plugin controller

hallfox:
This commit adds a controller object for the authorization plugin. The controller is initialized by the broker and loads the appropriate plugins as configured by a new setting in the broker config schema. If no plugin is configured, we load a default allow-all authorizer.

An interface for authorization was also added that simply adapts the controller object and mimics the plugin interface. We then pass a handle to that interface into SessionNegotiator, which can use it to authorize new sessions and further pass that handle into the created sessions for them to authorize operations within the session.

Only authorization on admin session creation has been implemented, mostly as an example. This may be moved into a separate commit.

### DIFF
--- a/src/groups/mqb/mqba/mqba_application.cpp
+++ b/src/groups/mqb/mqba/mqba_application.cpp
@@ -18,11 +18,13 @@
 #include <mqbscm_version.h>
 // MQB
 #include <mqba_authenticator.h>
+#include <mqba_authorizer.h>
 #include <mqba_configprovider.h>
 #include <mqba_dispatcher.h>
 #include <mqba_domainmanager.h>
 #include <mqba_sessionnegotiator.h>
 #include <mqbauthn_authenticationcontroller.h>
+#include <mqbauthz_authorizationcontroller.h>
 #include <mqbblp_clustercatalog.h>
 #include <mqbblp_relayqueueengine.h>
 #include <mqbcfg_brokerconfig.h>
@@ -164,6 +166,7 @@ Application::Application(bdlmt::EventScheduler* scheduler,
 , d_pluginManager_mp()
 , d_statController_mp()
 , d_authenticationController_mp()
+, d_authorizationController_mp()
 , d_configProvider_mp()
 , d_dispatcher_mp()
 , d_transportManager_mp()
@@ -253,6 +256,7 @@ int Application::start(bsl::ostream& errorDescription)
         rc_ADMIN_POOL_START_FAILURE          = -10,
         rc_PLUGINMANAGER                     = -11,
         rc_AUTHENTICATIONCONTROLLER          = -12,
+        rc_AUTHORIZATIONCONTROLLER           = -13,
     };
 
     int rc = rc_SUCCESS;
@@ -305,6 +309,18 @@ int Application::start(bsl::ostream& errorDescription)
         return (rc * 100) + rc_AUTHENTICATIONCONTROLLER;  // RETURN
     }
 
+    // Start the AuthorizationController
+    const mqbcfg::AppConfig& brokerConfig = mqbcfg::BrokerConfig::get();
+    rc = mqbauthz::AuthorizationController::allocateManaged(
+        &d_authorizationController_mp,
+        errorDescription,
+        brokerConfig.authorization(),
+        *d_pluginManager_mp,
+        d_allocators.get("AuthorizationController"));
+    if (rc != 0) {
+        return (rc * 100) + rc_AUTHORIZATIONCONTROLLER;  // RETURN
+    }
+
     // Start the config provider
     d_configProvider_mp.load(new (*d_allocator_p) ConfigProvider(
                                  d_allocators.get("ConfigProvider")),
@@ -334,6 +350,11 @@ int Application::start(bsl::ostream& errorDescription)
                                            d_allocators.get("Authenticator")),
         d_allocator_p);
 
+    bslma::ManagedPtr<mqbi::Authorizer> authorizer_mp(
+        bslma::ManagedPtrUtil::allocateManaged<mqba::Authorizer>(
+            d_allocators.get("Authorizer"),
+            d_authorizationController_mp.get()));
+
     SessionNegotiator* sessionNegotiator = new (*d_allocator_p)
         SessionNegotiator(&d_bufferFactory,
                           d_dispatcher_mp.get(),
@@ -346,10 +367,11 @@ int Application::start(bsl::ostream& errorDescription)
         .setAdminCommandEnqueueCallback(
             bdlf::BindUtil::bind(&Application::enqueueCommand,
                                  this,
-                                 bdlf::PlaceHolders::_1,    // source
-                                 bdlf::PlaceHolders::_2,    // cmd
-                                 bdlf::PlaceHolders::_3,    // onProcessedCb
-                                 bdlf::PlaceHolders::_4));  // fromReroute
+                                 bdlf::PlaceHolders::_1,   // source
+                                 bdlf::PlaceHolders::_2,   // cmd
+                                 bdlf::PlaceHolders::_3,   // onProcessedCb
+                                 bdlf::PlaceHolders::_4))  // fromReroute
+        .setAuthorizer(bslmf::MovableRefUtil::move(authorizer_mp));
 
     bslma::ManagedPtr<mqbnet::Negotiator> negotiatorMp(sessionNegotiator,
                                                        d_allocator_p);
@@ -547,6 +569,7 @@ void Application::stop()
     DESTROY_OBJ(d_transportManager_mp, "TransportManager");
     DESTROY_OBJ(d_dispatcher_mp, "Dispatcher");
     DESTROY_OBJ(d_configProvider_mp, "ConfigProvider");
+    DESTROY_OBJ(d_authorizationController_mp, "AuthorizationController");
     DESTROY_OBJ(d_authenticationController_mp, "AuthenticationController");
     DESTROY_OBJ(d_statController_mp, "StatController");
     DESTROY_OBJ(d_pluginManager_mp, "PluginManager");
--- a/src/groups/mqb/mqba/mqba_authorizer.cpp
+++ b/src/groups/mqb/mqba/mqba_authorizer.cpp
@@ -1,0 +1,57 @@
+// Copyright 2026 Bloomberg Finance L.P.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <mqba_authorizer.h>
+
+#include <mqbscm_version.h>
+
+// MQB
+#include <mqbauthz_authorizationcontroller.h>
+#include <mqbplug_authorizer.h>
+
+// BDE
+#include <ball_log.h>
+#include <ball_scopedattribute.h>
+#include <bsl_memory.h>
+
+namespace BloombergLP {
+namespace mqba {
+
+// -------------------
+// class Authenticator
+// -------------------
+
+// CREATORS
+Authorizer::Authorizer(mqbauthz::AuthorizationController* authzController)
+: d_authzController_p(authzController)
+{
+    // PRECONDITIONS
+    BSLS_ASSERT(d_authzController_p);
+}
+
+/// Destructor
+Authorizer::~Authorizer()
+{
+}
+
+bool Authorizer::authorize(const mqbact::Action&                action,
+                           const mqbplug::AuthenticationResult& authnResult)
+{
+    ball::ScopedAttribute principalAttr("principal", authnResult.principal());
+    return d_authzController_p->authorizer().authorize(action, authnResult);
+}
+
+}  // close package namespace
+}  // close enterprise namespace
--- a/src/groups/mqb/mqba/mqba_authorizer.h
+++ b/src/groups/mqb/mqba/mqba_authorizer.h
@@ -1,0 +1,81 @@
+// Copyright 2026 Bloomberg Finance L.P.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef INCLUDED_MQBA_AUTHORIZER
+#define INCLUDED_MQBA_AUTHORIZER
+
+/// @file mqba_authorizer.h
+///
+/// @brief Provide an authorizer for authenticating a connection.
+
+// MQB
+#include <mqbi_authorizer.h>
+
+// BDE
+#include <bsl_ostream.h>
+#include <bsls_keyword.h>
+
+namespace BloombergLP {
+
+namespace mqbauthz {
+class AuthorizationController;
+}
+
+namespace mqba {
+
+// ===================
+// class Authorizer
+// ===================
+
+/// authorizer for a BlazingMQ session with client or broker
+class Authorizer : public mqbi::Authorizer {
+  private:
+    // DATA
+
+    /// Authorization Controller.
+    mqbauthz::AuthorizationController* d_authzController_p;
+
+  private:
+    // NOT IMPLEMENTED
+
+    /// Copy constructor and assignment operator not implemented.
+    Authorizer(const Authorizer&) BSLS_KEYWORD_DELETED;
+    Authorizer& operator=(const Authorizer&) BSLS_KEYWORD_DELETED;
+
+  public:
+    // CREATORS
+
+    /// Create a new `authorizer` adapting the specified `authnController`.
+    explicit Authorizer(mqbauthz::AuthorizationController* authnController);
+
+    /// Destructor
+    ~Authorizer() BSLS_KEYWORD_OVERRIDE;
+
+    // MANIPULATORS
+
+    /// Check if the supplied action is allowed based on the result of
+    /// authentication.
+    ///
+    /// @param action The action being authorized
+    /// @param authnResult The result of an authenticated connection
+    bool authorize(const mqbact::Action&                action,
+                   const mqbplug::AuthenticationResult& authnResult)
+        BSLS_KEYWORD_OVERRIDE;
+};
+
+}  // close package namespace
+}  // close enterprise namespace
+
+#endif
--- a/src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
+++ b/src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
@@ -1,0 +1,206 @@
+// Copyright 2026 Bloomberg Finance L.P.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <mqbauthz_authorizationcontroller.h>
+
+#include <mqbscm_version.h>
+
+// MQB
+#include <mqbauthz_basicauthorizer.h>
+#include <mqbauthz_pluginlibrary.h>
+#include <mqbcfg_messages.h>
+#include <mqbplug_authorizer.h>
+#include <mqbplug_pluginfactory.h>
+#include <mqbplug_plugintype.h>
+
+// BDE
+#include <ball_log.h>
+#include <bdlb_nullablevalue.h>
+#include <bsl_ostream.h>
+#include <bsl_string.h>
+#include <bsl_unordered_set.h>
+#include <bsl_vector.h>
+#include <bslma_managedptr.h>
+#include <bslmf_movableref.h>
+
+namespace BloombergLP {
+namespace mqbauthz {
+
+namespace {
+
+}  // close unnamed namespace
+
+// ------------------------------
+// class AuthorizationController
+// ------------------------------
+
+// CREATORS
+
+AuthorizationController::AuthorizationController(
+    bslmf::MovableRef<AuthorizerMp> authorizer)
+: d_authorizer_mp(bslmf::MovableRefUtil::move(authorizer))
+{
+    // PRECONDITIONS
+    BSLS_ASSERT(d_authorizer_mp);
+}
+
+void AuthorizationController::collectAvailablePluginFactories(
+    PluginFactories*              pluginFactories,
+    const mqbplug::PluginManager& pluginManager,
+    mqbplug::PluginType::Enum     type)
+{
+    // PRECONDITIONS
+    BSLS_ASSERT(pluginFactories);
+
+    // Collect external plugin factories loaded by PluginManager
+    pluginManager.get(pluginFactories, type);
+
+    // Add built-in plugin factories
+    PluginLibrary                                    builtInPlugins;
+    bsl::vector<mqbplug::PluginInfo>::const_iterator it =
+        builtInPlugins.plugins().cbegin();
+    for (; it != builtInPlugins.plugins().cend(); ++it) {
+        if (it->type() == type) {
+            pluginFactories->insert(it->factory().get());
+        }
+    }
+}
+
+int AuthorizationController::createConfiguredAuthorizer(
+    AuthorizerMp*                   result,
+    bsl::ostream&                   errorDescription,
+    const PluginFactories&          pluginFactories,
+    const mqbcfg::AuthorizerConfig& authorizerConfig,
+    bsl::allocator<>                allocator)
+{
+    // PRECONDITIONS
+    BSLS_ASSERT(result);
+
+    enum RcEnum { rc_SUCCESS = 0, rc_PLUGIN_NOT_FOUND = -1 };
+
+    // If authorizer is explicitly configured, create it
+    const bsl::optional<mqbcfg::AuthorizerPluginConfig>& pluginConfigOpt =
+        authorizerConfig.authorizer();
+
+    if (!pluginConfigOpt.has_value()) {
+        BALL_LOG_INFO << "No configured authorizer";
+        return rc_SUCCESS;
+    }
+
+    const mqbcfg::AuthorizerPluginConfig& authorizerCfg =
+        pluginConfigOpt.value();
+
+    const bsl::string& pluginName = authorizerCfg.name();
+    BALL_LOG_INFO << "Attempting to create authorizer plugin '" << pluginName
+                  << "'";
+
+    // Try to create the authorizer for this plugin name
+    bsl::optional<mqbplug::AuthorizerPluginFactory*> factoryOpt =
+        mqbplug::PluginFactoryUtil::findType<mqbplug::AuthorizerPluginFactory>(
+            pluginFactories.cbegin(),
+            pluginFactories.cend());
+    if (!factoryOpt.has_value()) {
+        errorDescription
+            << "authorizer plugin '" << pluginName
+            << "' not found. Ensure the plugin is either built-in or listed "
+               "in plugins.enabled[], or its configuration is correct.";
+        return rc_PLUGIN_NOT_FOUND;  // RETURN
+    }
+    *result = (*factoryOpt)->create(allocator.mechanism());
+
+    return rc_SUCCESS;
+}
+
+void AuthorizationController::ensureDefaultAuthorizer(
+    AuthorizerMp*    result,
+    bsl::allocator<> allocator)
+{
+    // PRECONDITIONS
+    BSLS_ASSERT(result);
+
+    // Have authenticators, nothing to do
+    if (*result) {
+        return;
+    }
+
+    // If no authenticators are configured, use BasicAuthorizer for allow-all
+    // authorization
+    BALL_LOG_INFO << "No authorizer configured, using "
+                     "BasicAuthorizer as default";
+
+    BasicAuthorizerPluginFactory basicAuthzFactory;
+    *result = basicAuthzFactory.create(allocator.mechanism());
+}
+
+int AuthorizationController::allocateManaged(
+    bslma::ManagedPtr<AuthorizationController>* result,
+    bsl::ostream&                               errorDescription,
+    const mqbcfg::AuthorizerConfig&             authorizerConfig,
+    const mqbplug::PluginManager&               pluginManager,
+    const bsl::allocator<>&                     allocator)
+{
+    // PRECONDITIONS
+    BSLS_ASSERT(result);
+
+    enum RcEnum {
+        // Enum for the various RC error categories
+        rc_SUCCESS                   = 0,
+        rc_CREATE_AUTHORIZERS_FAILED = -1,
+    };
+
+    BALL_LOG_INFO << "Starting AuthorizationController";
+
+    // Step 1: Collect all available plugin factories (built-in + external)
+    PluginFactories pluginFactories;
+    collectAvailablePluginFactories(&pluginFactories,
+                                    pluginManager,
+                                    mqbplug::PluginType::e_AUTHORIZER);
+
+    // Step 2: Create authorizers based on configuration
+    AuthorizerMp authorizer;
+    if (0 != createConfiguredAuthorizer(&authorizer,
+                                        errorDescription,
+                                        pluginFactories,
+                                        authorizerConfig,
+                                        allocator)) {
+        return rc_CREATE_AUTHORIZERS_FAILED;  // RETURN
+    }
+
+    // Step 3: Ensure at least one authorizer is available
+    ensureDefaultAuthorizer(&authorizer, allocator);
+
+    BSLS_ASSERT(authorizer);
+
+    BALL_LOG_INFO << "Created authorizer '" << authorizer->name() << "'";
+
+    *result = bslma::ManagedPtrUtil::allocateManaged<AuthorizationController>(
+        allocator,
+        bslmf::MovableRefUtil::move(authorizer));
+
+    BALL_LOG_INFO << "Successfully initialized authorizer controller";
+
+    return rc_SUCCESS;
+}
+
+mqbplug::Authorizer& AuthorizationController::authorizer()
+{
+    // PRECONDITIONS
+    BSLS_ASSERT(d_authorizer_mp);
+
+    return *d_authorizer_mp;
+}
+
+}  // close package namespace
+}  // close enterprise namespace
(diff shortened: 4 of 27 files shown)


src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.h
```diff
@@ -0,0 +1,160 @@
...
+    // PRIVATE TYPES
+    typedef bslma::ManagedPtr<mqbplug::Authorizer> AuthorizerMp;
+
+    // DATA
+
+    /// Registered authorizer plugins.
+    /// Mapping an Authorization mechanism to a mqbplug::authorizer.
+    /// The mechanisms are stored in upper case to make the lookup
+    /// case-insensitive.
+    bsl::optional<AuthorizerMp> d_authorizer;
```
chrisbeard:
```suggestion
    /// Registered authorizer plugin.
    bsl::optional<AuthorizerMp> d_authorizer;
```

On first glance, I would expect this to not be optional, there will always be an authorizer, right? It just may be a built-in, permissive one?

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.h
```diff
@@ -0,0 +1,160 @@
...
+    /// `pluginFactories` collection.  Return 0 on success, non-zero on
+    /// error.
+    int collectAvailablePluginFactories(
+        bsl::unordered_set<mqbplug::PluginFactory*>* pluginFactories,
+        mqbplug::PluginType::Enum                    type);
+
+    /// Create authorizers based on configuration from the specified
+    /// `authorizerConfig` using the specified `pluginFactories`.
+    /// Return 0 on success, non-zero on error.
+    int createConfiguredAuthorizers(
```
chrisbeard: nit: should be singular?
```suggestion
    int createConfiguredAuthorizer(
```

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.h
```diff
@@ -0,0 +1,160 @@
...
+
+    // MANIPULATORS
+
+    /// Start the AuthorizationController.  Return 0 on success, or a non-zero
+    /// return code on error and fill in the specified `errorDescription`
+    /// stream with the description of the error.
+    int start(bsl::ostream& errorDescription);
+
+    /// Stop the AuthorizationController.  The caller must ensure that no
+    /// `authenticate()` calls are in flight before invoking this method.
```
chrisbeard:
```suggestion
    /// `authorize()` calls are in flight before invoking this method.
```

src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -538,6 +557,7 @@ void Application::stop()
     STOP_OBJ(d_dispatcher_mp, "Dispatcher");
     STOP_OBJ(d_configProvider_mp, "ConfigProvider");
     STOP_OBJ(d_authenticationController_mp, "AuthenticationController");
+    STOP_OBJ(d_authorizationController_mp, "AuthorizationController");
```
chrisbeard: double check the init/stop/destroy ordering here, authz is created after authn, should be stopped/destroyed before authn?

src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -334,6 +348,10 @@ int Application::start(bsl::ostream& errorDescription)
                                            d_allocators.get("Authenticator")),
         d_allocator_p);
 
+    bslma::ManagedPtr<mqbi::Authorizer> authorizerMp(
+        new (*d_allocator_p) Authorizer(d_authorizationController_mp.get()),
+        d_allocator_p);
```
chrisbeard: What's the lifetime of the `Authorizer`? I assume it's supposed to outlive to `AuthorizationController`.

chrisbeard (commented):
First pass looking good

src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -305,6 +309,16 @@ int Application::start(bsl::ostream& errorDescription)
         return (rc * 100) + rc_AUTHENTICATIONCONTROLLER;  // RETURN
     }
 
+    // Start the AuthorizationController
+    d_authorizationController_mp.load(
+        new (*d_allocator_p) mqbauthz::AuthorizationController(
+            d_pluginManager_mp.get(),
+            d_allocators.get("AuthorizationControlller")));
```
678098:
```suggestion
    d_authorizationController_mp = bslma::ManagedPtrUtil::allocateManaged<mqbauthz::AuthorizationController>(
            d_allocators.get("AuthorizationController"),
            d_pluginManager_mp.get());
```

678098:
```suggestion
    d_authorizationController_mp.load(
        new (*d_allocator_p) mqbauthz::AuthorizationController(
            d_pluginManager_mp.get(),
            d_allocators.get("AuthorizationController")));
```
3 `L`s

src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -334,6 +348,10 @@ int Application::start(bsl::ostream& errorDescription)
                                            d_allocators.get("Authenticator")),
         d_allocator_p);
 
+    bslma::ManagedPtr<mqbi::Authorizer> authorizerMp(
```
678098: Let's also use `allocateManaged`

src/groups/mqb/mqba/mqba_application.cpp
```diff
@@ -346,10 +364,11 @@ int Application::start(bsl::ostream& errorDescription)
...
                                  this,
-                                 bdlf::PlaceHolders::_1,    // source
-                                 bdlf::PlaceHolders::_2,    // cmd
-                                 bdlf::PlaceHolders::_3,    // onProcessedCb
-                                 bdlf::PlaceHolders::_4));  // fromReroute
+                                 bdlf::PlaceHolders::_1,   // source
+                                 bdlf::PlaceHolders::_2,   // cmd
+                                 bdlf::PlaceHolders::_3,   // onProcessedCb
+                                 bdlf::PlaceHolders::_4))  // fromReroute
+        .setAuthorizer(authorizerMp.get());
```
678098: A general question, should we use `ManagedPtr + raw ptrs` or use `shared_ptr`.
`ManagedPtr` allows to destruct the object in `mqba::Application`: this ensures that the object is destroyed here. However, any raw pointers that we stored start pointing on a memory block without valid object.
`shared_ptr` might be more suitable for shared ownership. We have a guarantee that a pointer is pointing at an existing object, we just have to ensure the destruction order.

src/groups/mqb/mqba/mqba_authorizer.cpp
```diff
@@ -0,0 +1,57 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <mqba_authorizer.h>
+
+#include <mqbscm_version.h>
+
+// MQB
+#include <mqbcfg_messages.h>
```
678098:
```suggestion
```
Is it used here?

src/groups/mqb/mqba/mqba_authorizer.h
```diff
@@ -0,0 +1,82 @@
...
+
+#ifndef INCLUDED_MQBA_AUTHORIZER
+#define INCLUDED_MQBA_AUTHORIZER
+
+/// @file mqba_authorizer.h
+///
+/// @brief Provide an authorizer for authenticating a connection.
+
+// MQB
+#include <mqbauthz_authorizationcontroller.h>
```
678098:
```suggestion
```
Can be replaced with a forward declaration in this header

src/groups/mqb/mqba/mqba_authorizer.h
```diff
@@ -0,0 +1,82 @@
...
+// ===================
+// class Authorizer
+// ===================
+
+/// authorizer for a BlazingMQ session with client or broker
+class Authorizer : public mqbi::Authorizer {
+  private:
+    // CLASS-SCOPE CATEGORY
+    BALL_LOG_SET_CLASS_CATEGORY("MQBA.AUTHORIZER");
+
```
678098:
```suggestion
```

Nothing is logged here, can remove ball_log related code

src/groups/mqb/mqbplug/mqbplug_pluginfactory.h
```diff
@@ -26,6 +26,10 @@
 
 // MQB
 
+#include <bsl_cstddef.h>
+#include <bsl_optional.h>
+#include <bslstl_optional.h>
```
678098:
```suggestion
```

Should we include just `bsl_optional.h`?

src/groups/mqb/mqbi/mqbi_authorizer.h
```diff
@@ -0,0 +1,71 @@
...
+#define INCLUDED_MQBI_AUTHORIZER
+
+/// @file mqbi_authorizer.h
+///
+/// @brief Provide a protocol for an authorizer.
+
+// MQB
+
+// BMQ
+
```
678098:
```suggestion
```
Less lines

src/groups/mqb/mqbi/mqbi_authorizer.h
```diff
@@ -0,0 +1,71 @@
...
+// BDE
+#include <bsl_memory.h>
```
678098:
```suggestion
```
Not used here

src/groups/mqb/mqbcfg/mqbcfg.xsd
```diff
@@ -451,6 +452,37 @@
...
+      <documentation>
+        Top level type for the broker's authorization configurations.
+
+        authorizer...........:
+            Configuration entries for the authorizer plugin (built-in or external).
+      </documentation>
+    </annotation>
+    <sequence>
+      <element name='authorizer' type='tns:AuthorizerPluginConfig' minOccurs='0' />
+     </sequence>
```
678098: 5 spaces -> 4

```suggestion
    </sequence>
```

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
```diff
@@ -0,0 +1,265 @@
...
+                                             authorizerConfig)) {
+        return rc * 10 + rc_CREATE_AUTHORIZERS_FAILED;  // RETURN
+    }
+
+    // Step 3: Ensure at least one authorizer is available
+    if (int rc = ensureDefaultAuthorizer(errorDescription)) {
+        return rc * 10 + rc_DEFAULT_AUTHORIZER_FAILED;  // RETURN
+    }
+
+    BALL_LOG_INFO << "Successfully initialized authorizer plugin";
```
678098:
```suggestion
    BALL_LOG_INFO << "Successfully initialized authorizer controller";
```

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
```diff
@@ -0,0 +1,265 @@
...
+    // Add built-in plugin factories
+    bsl::vector<mqbplug::PluginInfo>::const_iterator it =
+        d_builtInPluginLibrary->plugins().cbegin();
+    for (; it != d_builtInPluginLibrary->plugins().cend(); ++it) {
+        if (it->type() == type) {
+            pluginFactories->insert(it->factory().get());
+        }
+    }
+
+    return 0;
```
678098: Can it fail?
It always returns 0, might simplify initialization checks.

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.h
```diff
@@ -0,0 +1,160 @@
...
+#include <bslmf_movableref.h>
+#include <mqbact_actions.h>
+#include <mqbplug_authorizer.h>
+#include <mqbplug_pluginmanager.h>
+
+// BDE
+#include <ball_log.h>
+#include <bsl_memory.h>
+#include <bsl_optional.h>
+#include <bsl_ostream.h>
```
678098:
```suggestion
#include <bsl_iosfwd.h>
```
Lightweight header with forward declarations to `bsl::ostream` etc

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
```diff
@@ -0,0 +1,265 @@
...
+    const mqbact::Action&                action,
+    const mqbplug::AuthenticationResult& authnResult)
+{
+    ball::ScopedAttribute principalAttr("mqbauthn.principal",
+                                        authnResult.principal());
+    bmqu::MemOutStream    errorStream;
+    bool isAllowed = authorizer().authorize(action, authnResult);
+    if (!isAllowed) {
+        BALL_LOG_INFO << "Failed to authorize for '" << authnResult.principal()
+                      << "'";
```
678098: Might be useful to know what `action` failed authorization

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
```diff
@@ -0,0 +1,265 @@
...
+    d_authorizer.reset();
+}
+
+bool AuthorizationController::authorize(
+    const mqbact::Action&                action,
+    const mqbplug::AuthenticationResult& authnResult)
+{
+    ball::ScopedAttribute principalAttr("mqbauthn.principal",
+                                        authnResult.principal());
+    bmqu::MemOutStream    errorStream;
```
678098:
```suggestion
    bmqu::MemOutStream    errorStream(get_allocator());
```

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    BALL_LOG_INFO << "Stopping AuthorizationController";
+
+    d_isStarted = false;
+
+    d_authorizer.reset();
+}
+
+bool AuthorizationController::authorize(
```
678098: What if there is a late connection or a dispatcher event that causes authorization attempt (`authorize` call) on an `AuthorizationController` that is already stopped?

The broker will either crash (if assert level is sufficient) or there will be an UB. Also, `authorize` might be called in the middle of `d_authorizer.reset();`.

I think the controller should have thread safe stop, might use `AtomicGate` for this

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+typedef bsl::unordered_set<mqbplug::PluginFactory*> PluginFactories;
+
+}  // close unnamed namespace
+
+// ------------------------------
+// class AuthorizationController
+// ------------------------------
+
+int AuthorizationController::initializeAuthorizers(
```
678098: Might remove `int AuthorizationController::initializeAuthorizers` symbol by placing the implementation to `start`

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.h
```diff
@@ -0,0 +1,160 @@
...
+    // DATA
+
+    /// Registered authorizer plugins.
+    /// Mapping an Authorization mechanism to a mqbplug::authorizer.
+    /// The mechanisms are stored in upper case to make the lookup
+    /// case-insensitive.
+    bsl::optional<AuthorizerMp> d_authorizer;
+
+    /// Hold the PluginLibrary for all the built-in Authorization plugins.
+    bslma::ManagedPtr<mqbplug::PluginLibrary> d_builtInPluginLibrary;
```
678098:
```suggestion
    bslma::ManagedPtr<mqbplug::PluginLibrary> d_builtInPluginLibrary_mp;
```

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.h
678098: I think it can/should be simpler `optional<ManagedPtr<...>>` -> `ManagedPtr<...>`

This allows to simplify other places such as
```
mqbplug::Authorizer& AuthorizationController::authorizer()
{
    BSLS_ASSERT(d_authorizer && *d_authorizer);
    return **d_authorizer;
}
```

If we guarantee that `d_authorizer_mp` always exists, we can remove this private accessor `authorizer()` fully and just use `d_authorizer_mp`.
Less code, less symbols

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.h
```diff
@@ -0,0 +1,160 @@
...
+#include <mqbplug_pluginmanager.h>
+
+// BDE
+#include <ball_log.h>
+#include <bsl_memory.h>
+#include <bsl_optional.h>
+#include <bsl_ostream.h>
+#include <bsl_string_view.h>
+#include <bsl_unordered_map.h>
+#include <bsl_unordered_set.h>
```
678098, line 35:
```suggestion
#include <bsl_unordered_set.h>
```
Are these 2 headers used?

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.h
```diff
@@ -0,0 +1,160 @@
...
+/// @file mqbauthn_authorizationcontroller.h
+///
+/// @brief Provide an utility class that orchestrates authorization
+/// plugins.
+
+// MQB
+#include <bslmf_movableref.h>
+#include <mqbact_actions.h>
+#include <mqbplug_authorizer.h>
+#include <mqbplug_pluginmanager.h>
```
678098:
```suggestion
```
All 3 can be forward-declared in the header

src/groups/mqb/mqbauthz/mqbauthz_basicauthorizer.h
```diff
@@ -55,6 +55,7 @@ namespace mqbauthz {
 class BasicAuthorizer : public mqbplug::Authorizer {
   public:
     // CLASS DATA
+    static const char* k_NAME;
```
678098: Do we need this public constant?
`const char*` doesn't contain length compared to `bsl::string_view` constants

src/groups/mqb/mqbauthz/mqbauthz_basicauthorizer.cpp
```diff
@@ -37,13 +43,15 @@ BasicAuthorizer::~BasicAuthorizer()
...
 }
 
-bool BasicAuthorizer::authorize(const mqbact::Action&,
-                                const mqbplug::AuthenticationResult&)
+bool BasicAuthorizer::authorize(
+    const mqbact::Action&                action,
+    const mqbplug::AuthenticationResult& authnResult)
 
 {
+    BALL_LOG_INFO << "Authorize check on " << action;
```
678098:
```suggestion
    BALL_LOG_INFO << "Authorize [true] for " << action;
```
Might just log that we authorized something right here

src/groups/mqb/mqbauthz/mqbauthz_basicauthorizer.cpp
```diff
@@ -15,17 +15,23 @@
...
 #include <mqbauthz_basicauthorizer.h>
 
+#include <mqbscm_version.h>
+
 // MQB
+#include <mqbact_actions.h>
 #include <mqbplug_authorizer.h>
 
 // BDE
+#include <ball_log.h>
```
678098, line 25:
```suggestion
```
Already included in the header

src/groups/mqb/mqbauthz/mqbauthz_pluginlibrary.h
```diff
@@ -0,0 +1,92 @@
...
+// MQB
+#include <mqbplug_plugininfo.h>
+#include <mqbplug_pluginlibrary.h>
+
+// BDE
+#include <bsl_vector.h>
+#include <bslma_allocator.h>
+#include <bslma_usesbslmaallocator.h>
+#include <bslmf_movableref.h>
+#include <bslmf_nestedtraitdeclaration.h>
```
678098: I think nested trait for allocator is missing for this class

678098 (changes requested):
Have some comments

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.h
hallfox: Yeah this is my Rust-brain kicking in. There is a brief period of time where `d_authorizer` must be `NULL`, because we don't load anything until `start`/`stop` kicks in. I really don't like this non-RAII compatible approach to the way this class is designed, but it's mostly just coming from what the design in mqba_application.

src/groups/mqb/mqba/mqba_application.cpp
pniedzielski: This should probably be done to every instance of this pattern in this function.

678098: Should be, but I don't want to extend the scope of the PR.
Will be good if the new code at least uses `allocate_shared` or `allocateManaged`

src/groups/mqb/mqbauthz/mqbauthz_pluginlibrary.h
hallfox: `bsl::allocator` style allocator-aware classes only require the `allocator_type` member to be recognized as allocator aware in the other type traits.

678098: Does it work with construction utils? So it automatically passes the allocator as the last arg during construction?
And if it works like this, can we remove trait headers?

src/groups/mqb/mqba/mqba_application.cpp
hallfox: It seems like `mqba_application` really wants to be the sole owner of all of these pointers, and these objects it creates are supposed to borrow from it under the assumption that `mqba_application` will outlive everything it passes these pointers down to. This is maybe fine, but in experience we know that the line between borrower and owner are blurred really easily.

I've changed this so that this particular pointer's ownership is transferred to the `SessionNegotiator`, and so now it is an owner and a unique owner. This does not really solve an even more underlying problem, which is that the type behind this pointer itself borrows from `AuthorizationController`. Now we need to be sure that `SessionNegotiator` or whatever it passes it on to is aware that it's borrowing from `Application`. I suspect as we implement the actual authorization checks it will need to be converted to a `shared_ptr`.

src/groups/mqb/mqba/mqba_authorizer.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <mqba_authorizer.h>
+
+#include <mqbscm_version.h>
+
+// MQB
+#include <mqbauthz_authorizationcontroller.h>
+#include <mqbplug_authorizer.h>
+
+// BMQ
+
```
678098:
```suggestion
```

src/groups/mqb/mqba/mqba_authorizer.h
```diff
@@ -0,0 +1,83 @@
...
+/// @file mqba_authorizer.h
+///
+/// @brief Provide an authorizer for authenticating a connection.
+
+// MQB
+#include <mqbi_authorizer.h>
+
+// BMQ
+
```
678098:
```suggestion
```

src/groups/mqb/mqba/mqba_authorizer.h
```diff
@@ -0,0 +1,83 @@
...
+    /// Copy constructor and assignment operator not implemented.
+    Authorizer(const Authorizer&) BSLS_KEYWORD_DELETED;
+    Authorizer& operator=(const Authorizer&) BSLS_KEYWORD_DELETED;
+
+  public:
+    // CREATORS
+
+    /// Create a new `authorizer` adapting the specified `authnController`.
+    Authorizer(mqbauthz::AuthorizationController* authnController);
```
678098:
```suggestion
    explicit Authorizer(mqbauthz::AuthorizationController* authnController);
```

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
```diff
@@ -0,0 +1,199 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <mqbauthz_authorizationcontroller.h>
+
+#include <mqbscm_version.h>
+
+// BMQ
+
```
678098:
```suggestion
```

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
```diff
@@ -0,0 +1,199 @@
...
+}
+
+int AuthorizationController::createConfiguredAuthorizer(
+    AuthorizerMp*                   result,
+    bsl::ostream&                   errorDescription,
+    const PluginFactories&          pluginFactories,
+    const mqbcfg::AuthorizerConfig& authorizerConfig,
+    bsl::allocator<>                allocator)
+{
+    enum RcEnum { rc_SUCCESS = 0, rc_PLUGIN_NOT_FOUND = -1 };
```
678098, line 91:
```suggestion
    // PRECONDITIONS
    BSLS_ASSERT(result);

    enum RcEnum { rc_SUCCESS = 0, rc_PLUGIN_NOT_FOUND = -1 };
```

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    const mqbcfg::AuthorizerPluginConfig& authorizerCfg =
+        pluginConfigOpt.value();
+
+    BALL_LOG_INFO << "Creating explicitly configured authorizer";
+
+    const bsl::string& pluginName = authorizerCfg.name();
+    BALL_LOG_INFO << "Attempting to create authorizer plugin '" << pluginName
+                  << "'";
```
678098:
```suggestion
    const bsl::string& pluginName = authorizerCfg.name();
    BALL_LOG_INFO << "Attempting to create explicitly configured authorizer plugin '"
                  << pluginName << "'";
```
Possible to log just 1 line

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
```diff
@@ -0,0 +1,199 @@
...
+    bsl::optional<mqbplug::AuthorizerPluginFactory*> factoryOpt =
+        mqbplug::PluginFactoryUtil::findType<mqbplug::AuthorizerPluginFactory>(
+            pluginFactories.cbegin(),
+            pluginFactories.cend());
+    if (!factoryOpt.has_value()) {
+        errorDescription
+            << "authorizer plugin '" << pluginName
+            << "' not found or could not be created."
+            << " Ensure the plugin is either built-in or listed in "
+            << "plugins.enabled[], or its configuration is correct.";
```
678098:
```suggestion
            << "Authorizer plugin '" << pluginName << "' not found. "
            << "Ensure the plugin is either built-in or listed in "
            << "plugins.enabled[]";
```

We do not attempt to create a plugin during `findType` call, so creation attempt failure should not be listed in the error message.

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
```diff
@@ -0,0 +1,199 @@
...
+    }
+    *result = (*factoryOpt)->create(allocator.mechanism());
+
+    return rc_SUCCESS;
+}
+
+void AuthorizationController::ensureDefaultAuthorizer(
+    AuthorizerMp*    result,
+    bsl::allocator<> allocator)
+{
```
678098, line 129:
```suggestion
{
    // PRECONDITIONS
    BSLS_ASSERT(result);

```

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
```diff
@@ -0,0 +1,199 @@
...
+    *result = basicAuthzFactory.create(allocator.mechanism());
+}
+
+int AuthorizationController::allocateManaged(
+    bslma::ManagedPtr<AuthorizationController>* result,
+    bsl::ostream&                               errorDescription,
+    const mqbcfg::AuthorizerConfig&             authorizerConfig,
+    const mqbplug::PluginManager&               pluginManager,
+    const bsl::allocator<>&                     allocator)
+{
```
678098, line 153:
```suggestion
{
    // PRECONDITIONS
    BSLS_ASSERT(result);

```

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
```diff
@@ -0,0 +1,199 @@
...
+    enum RcEnum {
+        // Enum for the various RC error categories
+        rc_SUCCESS                   = 0,
+        rc_CREATE_AUTHORIZERS_FAILED = -1,
+    };
+
+    BALL_LOG_INFO << "Starting AuthorizationController";
+
+    // Step 1: Collect all available plugin factories (built-in + external)
+    PluginFactories pluginFactories;
```
678098, line 166:
```suggestion
    PluginFactories pluginFactories(allocator);
```

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
```diff
@@ -0,0 +1,199 @@
...
+        bslmf::MovableRefUtil::move(authorizer));
+
+    BALL_LOG_INFO << "Successfully initialized authorizer controller";
+
+    return rc_SUCCESS;
+}
+
+mqbplug::Authorizer& AuthorizationController::authorizer()
+{
+    return *d_authorizer_mp;
```
678098, line 202:
```suggestion
    // PRECONDITIONS
    BSLS_ASSERT(d_authorizer_mp);

    return *d_authorizer_mp;
```

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.g.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+// gtest
+#include <gtest/gtest.h>
+
+using namespace BloombergLP;
+
+namespace {
+
+}
+
```
678098:
```suggestion
```

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.h
```diff
@@ -0,0 +1,145 @@
...
+// class AuthorizationController
+// ==============================
+
+class AuthorizationController {
+  private:
+    // CLASS-SCOPE CATEGORY
+    BALL_LOG_SET_CLASS_CATEGORY("MQBAUTHZ.AUTHORIZATIONCONTROLLER");
+
+    // PRIVATE TYPES
+    typedef bslma::ManagedPtr<mqbplug::Authorizer>      AuthorizerMp;
```
678098: `AuthorizerMp` is used in public interface, the type should be public

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.h
```diff
@@ -0,0 +1,145 @@
...
+    /// Return 0 on success, non-zero on error.
+    static int createConfiguredAuthorizer(
+        AuthorizerMp*                   result,
+        bsl::ostream&                   errorDescription,
+        const PluginFactories&          pluginFactories,
+        const mqbcfg::AuthorizerConfig& authorizerConfig,
+        bsl::allocator<>                allocator);
+
+    /// Ensure at least one authorizer is available for default
+    /// Authorization, adding default Anonauthorizer if none are
```
678098:
```suggestion
    /// Authorization, adding default anonymous BasicAuthorizer if none are
```
There is no `AnonAuthorizer` class so we should mention the existing type

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.h
```diff
@@ -0,0 +1,145 @@
...
+
+// FORWARD DECLARATION
+namespace mqbplug {
+class PluginManager;
+class PluginFactory;
+class AuthenticationResult;
+}
+
+namespace mqbact {
+class Action;
```
678098: This class is forward-declared now which is good, however, heavy includes are still there and might be removed

src/groups/mqb/mqbi/mqbi_authorizer.h
```diff
@@ -0,0 +1,64 @@
...
+
+namespace mqbplug {
+class AuthenticationResult;
+}
+
+namespace mqbi {
+
+// ===================
+// class Authorizer
+// ===================
```
678098:
```suggestion
// ================
// class Authorizer
// ================
```

Alternatively it might be good to remove these comments at all.
Are there any benefits from having it?

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.h
```diff
@@ -0,0 +1,145 @@
...
+    // PRIVATE TYPES
+    typedef bslma::ManagedPtr<mqbplug::Authorizer>      AuthorizerMp;
+    typedef bsl::unordered_set<mqbplug::PluginFactory*> PluginFactories;
+
+    // DATA
+
+    /// Registered authorizer plugins.
+    /// Mapping an Authorization mechanism to a mqbplug::authorizer.
+    /// The mechanisms are stored in upper case to make the lookup
+    /// case-insensitive.
```
678098: This comment mentions plugins in plural, and it mentions lookup by name.
The field is a managed pointer to a single object, not a map.
So the comment seems not relevant

src/groups/mqb/mqbauthz/mqbauthz_pluginlibrary.h
```diff
@@ -0,0 +1,90 @@
...
+    // DATA
+    bsl::vector<mqbplug::PluginInfo> d_plugins;
+
+  private:
+    // NOT IMPLEMENTED
+    PluginLibrary(const PluginLibrary& other) BSLS_KEYWORD_DELETED;
+    PluginLibrary& operator=(const PluginLibrary& other) BSLS_KEYWORD_DELETED;
+    PluginLibrary(bslmf::MovableRef<PluginLibrary> other) BSLS_KEYWORD_DELETED;
+    PluginLibrary&
+    operator=(bslmf::MovableRef<PluginLibrary> other) BSLS_KEYWORD_DELETED;
```
678098: Do we need to delete movable ref constructor/assignment?
It makes 4 deleted methods when typically we only delete 2.

src/groups/mqb/mqbauthz/mqbauthz_pluginlibrary.h
```diff
@@ -0,0 +1,90 @@
...
+
+  public:
+    // TYPES
+    typedef bsl::allocator<> allocator_type;
+
+    // CREATORS
+    explicit PluginLibrary(const allocator_type& allocator = allocator_type());
```
678098, line 56: Here we use `typedef bsl::allocator<> allocator_type;`
and in other places in this PR `bsl::allocator<>` is used directly without a typedef

src/groups/mqb/mqbauthz/mqbauthz_pluginlibrary.h
hallfox: Since this type is allocator-aware in order to support move constructors we technically need to implement both the move constructor, move assignment operator, and an allocator-extended move constructor in order for move operations to be correct. ~~Since C++20 targets will implicitly implement the move operators on this class, I marked them explicitly deleted since the ones that the compiler will generate will not be necessarily be correct. However, I believe for this specific class the compiler will generate the special move functions correctly since the allocator is owned by the vector (thus most of the relevant operations will be delegated.) However, we still would need to implement the move-extended constructor (`PluginLibrary(PluginLibrary&&, allocator_type alloc = {}`) to be correct, so I can quickly implement that instead of deleting move constructors.~~ I'm totally wrong about this, declaring the copy constructors as deleted like we do prevents the special move functions from being generated by the compiler. We can just keep the deleted copy constructors, deleting the move constructors is just being more explicit.

src/groups/mqb/mqbauthz/mqbauthz_pluginlibrary.h
hallfox: This is because allocator-aware types require this typedef to be here, (it is equivalent to the `UsesBslmaAllocator` nested trait) and `PluginLibrary` is AA. Contrast that with the `AuthorizationController`, which is not itself an AA type (and thus has no `allocator_type` typedef) but has function which needs to allocate, and so we accept `bsl::allocator<>` directly.

src/groups/mqb/mqbauthz/mqbauthz_authorizationcontroller.cpp
hallfox: This is a transient allocation, I don't think we should pass an allocator here.

678098 (approved):

chrisbeard (approved):

PR #1623, opened 2026-07-16
Fix[mqbc::ClusterStateLedgerUtil]: Prevent recordSize overflow

kaikulimu:
## Summary
  - Hardens CSL record header handling against integer overflow when a record's `leaderAdvisoryWords` field holds an unexpectedly large value (e.g. from a corrupt or malformed record).
  - `ClusterStateLedgerUtil::recordSize()` computed `(headerWords + leaderAdvisoryWords) * k_WORD_SIZE` in 32-bit arithmetic. For certain `leaderAdvisoryWords` values the product wrapped: it could wrap to `0` (the incore ledger iterator then calls `incrementOffset(0)` and stops advancing), or to a small value (causing CRC to be validated over an incorrectly sized window).
  - Additionally, a secondary overflow in `loadClusterMessage()` yields a negative length: `msgLen = leaderAdvisoryWords * k_WORD_SIZE` is computed into a signed `int` and goes negative for large values. `validateRecordHeader()` only rejects `leaderAdvisoryWords == 0`, so other out-of-range values were not caught before reaching these paths.
  - Fix: structurally bound `leaderAdvisoryWords` so downstream size arithmetic cannot overflow, rather than relying on callers to validate first.

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_clusterstateledgerprotocol.h
+++ b/src/groups/mqb/mqbc/mqbc_clusterstateledgerprotocol.h
@@ -235,7 +235,7 @@ bsl::ostream& operator<<(bsl::ostream&                stream,
 /// +---------------+---------------+---------------+---------------+
 /// |   HW  |  RT   |                   Reserved                    |
 /// +---------------+---------------+---------------+---------------+
-/// |                       LeaderAdvisoryWords                     |
+/// |Reservd|                  LeaderAdvisoryWords                  |
 /// +---------------+---------------+---------------+---------------+
 /// |                    Elector Term Upper Bits                    |
 /// +---------------+---------------+---------------+---------------+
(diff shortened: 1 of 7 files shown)


src/groups/mqb/mqbc/mqbc_clusterstateledgerprotocol.h
```diff
@@ -497,6 +503,9 @@ ClusterStateRecordHeader::setRecordType(ClusterStateRecordType::Enum value)
 inline ClusterStateRecordHeader&
 ClusterStateRecordHeader::setLeaderAdvisoryWords(unsigned int value)
 {
+    // PRECONDITIONS: protect against overflow
```
pniedzielski, line 508: This should be documented, not just in the implementation.

pniedzielski (changes requested):

src/groups/mqb/mqbc/mqbc_clusterstateledgerprotocol.h
kaikulimu: Added a commit to document setter preconditions

pniedzielski (approved):

PR #1625, opened 2026-07-17
Feat: allow downgrading primary

678098:


678098:
On hold while we are evaluating raft

PR #1629, opened 2026-07-20
Feat: startAsync and StopAsync callbacks

dorjesinpo:
Add completion-callback overloads to bmqa async session start/stop

  `bmqa::Session::startAsync/stopAsync` previously reported completion only via
  session events, unlike the rest of the async API (`openQueueAsync` et al.) which
  take a completion callback. Add callback-taking overloads that deliver a typed
  status to a user callback, mirroring the openQueueAsync pattern.

  - New value types `bmqa::StartStatus` / `bmqa::StopStatus` (result + error
    description), registered in package/bmqa.mem.
  - New virtuals in `AbstractSession`; implemented in `Session` and `MockSession`.
  - `bmqimp: BrokerSession` caches the callback on entering `STARTING`
    (`handleStartRequest`) / on accepting a stop (`handleStopRequest`), and attaches
    it to the terminal event (`CONNECTED`, `CONNECTION_TIMEOUT`, `DISCONNECTED`),
    suppressing that session event and delivering it to the callback instead
    (as `openQueueAsync` does). Callback runs on the event-delivery thread, never
    the FSM thread. resetState clears any callback that never rode a terminal
    event, preventing leaks across start/stop cycles.
  - Non-callback path is unchanged.

  Tests: value-semantics tests for `Start`/`StopStatus`; `MockSession` callback case;
  `BrokerSession` real-FSM case asserting the callback fires and the terminal
  session event is suppressed.
  
  UPDATE:
1. Replaced `bmqa::StartStatus` + `bmqa::StopStatus` with one `bmqa::SessionStatus` (new `bmqa_sessionstatus.{h,cpp,t.cpp`}
2. Replaced `StartCallback` + `StopCallback` with one `SessionCallback` typedef.
3. Moved the two new virtuals to the end of `AbstractSession`.
4. Merged the two result wrappers into `sessionResultCallbackWrapper`; it forwards `event->result()` except for `CONNECTED`/`DISCONNECTED`/`CONNECTION_TIMEOUT`.
5. `Session::stopAsync(cb)` with no application: completes the callback inline instead of no-opping.
6. `setStopped`: emits a terminal event for a pending start callback when there's no timeout (`e_CANCELED`, or `e_ERROR/e_NOT_CONNECTED`).
7. `setStopped`: emits `DISCONNECTED` when `!d_onceConnected` if a stop callback is pending.
9. `handleStartRequest`, `e_STARTED`: enqueues `CONNECTED` carrying the callback.
10. `handleStopRequest`, `CLOSING_*`: WARN log when a redundant stop's callback is dropped.
11. `eventHandlerCbWrapper` bugfix: removed the early return, callback invocations now counted.


src/groups/bmq/bmqimp/bmqimp_brokersession.t.cpp
```diff
@@ -9763,6 +9877,7 @@ int main(int argc, char* argv[])
 
     switch (_testCase) {
     case 0:
+    case 71: test71_startStopAsyncCallback(); break;
```
678098: There are multiple removed tests, can reuse a number. The lowest one here is:
```
case 27: /* removed test */ break;
```

PR #1633, opened 2026-07-21
Fix[mqbc::ClusterStateTable]: Prevent double-heal (e_FOL_WAITING)

kaikulimu:
## Summary
  - Introduce a new Cluster FSM follower state, `e_FOL_WAITING`, sitting between `UNKNOWN` and `FOL_HEALING`: a follower that has sent a `RegistrationRequest` now waits in `FOL_WAITING` for the `RegistrationResponse` before entering `FOL_HEALING`. This is analogous to the `e_REPLICA_WAITING ` introduced in the Partition FSM.
  - This prevents a **follower double-heal**: the follower's LSN reaches the leader via the `RegistrationRequest` (which carries `sequenceNumber`), so while waiting the follower must reject `FollowerLSNRequest`s — otherwise the leader would receive the same LSN twice (once via `RegistrationRequest`, once via `FollowerLSNResponse`) and heal the follower twice, doing duplicate work.
  - Split out of #1505 (follower-side piece); independent of the leader-side STG3 work. Depends only on already-merged pieces.

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_clusterstatemanager.cpp
+++ b/src/groups/mqb/mqbc/mqbc_clusterstatemanager.cpp
@@ -500,20 +500,40 @@ void ClusterStateManager::do_sendFailureFollowerLSNResponse(
     const InputMessage&            inputMessage = metadata.inputMessage();
 
     bmqp_ctrlmsg::ControlMessage controlMsg;
-    controlMsg.rId() = inputMessage.requestId();
-
+    controlMsg.rId()               = inputMessage.requestId();
     bmqp_ctrlmsg::Status& response = controlMsg.choice().makeStatus();
-    response.category()            = bmqp_ctrlmsg::StatusCategory::E_REFUSED;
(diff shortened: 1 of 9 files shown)


src/groups/mqb/mqbc/mqbc_clusterstatemanager.cpp
```diff
@@ -500,20 +500,40 @@ void ClusterStateManager::do_sendFailureFollowerLSNResponse(
...
-    response.message()             = "Not a follower";
+
+    if (!d_clusterFSM.isSelfFollower()) {
+        response.category() = bmqp_ctrlmsg::StatusCategory::E_REFUSED;
+        response.code()     = mqbi::ClusterErrorCode::e_NOT_FOLLOWER;
+        response.message()  = "Not a follower";
+
+        BALL_LOG_INFO
+            << d_clusterData_p->identity().description()
+            << ": Self not follower! Sent failure response " << controlMsg
```
678098, line 513:
```suggestion
            << ": Self is not a follower! Sent failure response " << controlMsg
```

678098 (approved):

PR #1634, opened 2026-07-22
Fix[mqbc::ClusterStateTable]: e_LDR_HEALING_STG3 late-joiner re-entry

kaikulimu:
## Summary
  - Introduce the leader-side Cluster FSM state `e_LDR_HEALING_STG3`, splitting the old `LDR_HEALING_STG2` into **STG2** (request the CSL snapshot from the most up-to-date follower) and **STG3** (self holds the most up-to-date CSL and applies/commits a leader advisory to heal everyone).
  - Handle a late-joining follower: in STG3 a follower's `RegistrationRequest` / `FollowerLSNResponse` re-enters `do_applyCSLSelf` to apply a *second* leader advisory, **reusing** the partition assignment already chosen for this term instead of recomputing it (which could reshuffle primaries).
  - Split out of #1505; independent of the `e_FOL_WAITING` follower-side PR. Depends on `uncommittedAdvisories()` from #1632 (already merged).


678098 (approved):
Looks good

PR #1639, opened 2026-07-24
Fix[mqbnet::InitialConnectionContext]: propagate close to authn

678098:
The channel might be in the middle of opening when the broker stops.
In this case, `InitialConnectionContext` and `AuthenticationContext` both exist for this channel, and we might have scheduled reuthn callback.
Bug: `InitialConnectionContext::onClose` doesn't propagate close to `AuthenticationContext`. As a result, we don't remove scheduled event for reauthn from scheduler, and this leads to shutdown crash.

Fixes:
https://github.com/bloomberg/blazingmq/actions/runs/30048572316/job/89347302563

### DIFF
--- a/src/groups/mqb/mqbnet/mqbnet_authenticationcontext.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_authenticationcontext.cpp
@@ -100,6 +100,7 @@ bool AuthenticationState::fromAscii(AuthenticationState::Enum* out,
 // ---------------------------
 
 AuthenticationContext::AuthenticationContext(
+    bdlmt::EventScheduler*                     scheduler,
     InitialConnectionContext*                  initialConnectionContext,
     bsl::string_view                           mechanism,
     const bmqp_ctrlmsg::AuthenticationMessage& authenticationMessage,
@@ -107,6 +108,7 @@ AuthenticationContext::AuthenticationContext(
     AuthenticationState::Enum                  state,
     bslma::Allocator*                          allocator)
 : d_allocator_p(allocator)
+, d_scheduler_p(scheduler)
 , d_self(this)  // use default allocator
 , d_mutex()
 , d_authenticationResultSp()
@@ -117,7 +119,8 @@ AuthenticationContext::AuthenticationContext(
 , d_authenticationMessage(authenticationMessage)
 , d_encodingType(authenticationEncodingType)
 {
-    // NOTHING
+    // PRECONDITION
+    BSLS_ASSERT_SAFE(d_scheduler_p);
 }
 
 void AuthenticationContext::setAuthenticationResult(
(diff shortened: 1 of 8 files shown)


src/groups/mqb/mqbnet/mqbnet_authenticationcontext.cpp
```diff
@@ -250,21 +251,18 @@ void AuthenticationContext::onReauthenticationError(
     channel_sp->close(status);
 }
 
-void AuthenticationContext::onClose(bdlmt::EventScheduler* scheduler_p)
+void AuthenticationContext::onClose()
 {
     // executed by *ANY* thread
 
-    // PRECONDITIONS
-    BSLS_ASSERT_SAFE(scheduler_p);
```
678098, line 258: Not null due to invariant check on construction

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -951,7 +949,7 @@ void TCPSessionFactory::onClose(const bsl::shared_ptr<bmqio::Channel>& channel,
 
         // Disable reauthentication timer if there's any
         if (channelInfo->d_authenticationCtx_sp) {
-            channelInfo->d_authenticationCtx_sp->onClose(d_scheduler_p);
+            channelInfo->d_authenticationCtx_sp->onClose();
```
678098, line 952: This `onClose` handled the channels that were already moved to the internal collection `d_channels`.
It doesn't cover the case when channel hasn't arrived in `d_channels` yet.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -896,12 +896,10 @@ void TCPSessionFactory::onClose(const bsl::shared_ptr<bmqio::Channel>& channel,
 
         // set the 'isClosed' flag under lock to be checked under lock in
         // 'initialConnectionComplete'.
-        bsl::shared_ptr<InitialConnectionContext> initialConnectionContext_sp;
         InitialConnectionContextMp::iterator      iter =
             d_initialConnectionContextCache.find(channel.get());
         if (iter != d_initialConnectionContextCache.end()) {
-            initialConnectionContext_sp = iter->second;
-            initialConnectionContext_sp->onClose();
+            iter->second->onClose();
```
678098, line 902: `InitialConnectionContext::onClose` now propagates close to `AuthenticationContext`, if it exists

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.h
```diff
@@ -278,14 +278,14 @@ class InitialConnectionContext {
...
+    bsl::shared_ptr<bmqio::Channel> d_channel_sp;
 
     /// The AuthenticationContext updated upon receiving an
     /// authentication message.
-    bsl::shared_ptr<AuthenticationContext> d_authenticationCtxSp;
+    bsl::shared_ptr<AuthenticationContext> d_authenticationCtx_sp;
 
     /// The NegotiationContext updated upon receiving a negotiation message.
-    bsl::shared_ptr<NegotiationContext> d_negotiationCtxSp;
+    bsl::shared_ptr<NegotiationContext> d_negotiationCtx_sp;
```
678098, line 288: BDE convention to have `_sp`

src/groups/mqb/mqbnet/mqbnet_authenticationcontext.cpp
```diff
@@ -117,7 +119,8 @@ AuthenticationContext::AuthenticationContext(
 , d_authenticationMessage(authenticationMessage)
 , d_encodingType(authenticationEncodingType)
 {
-    // NOTHING
+    // PRECONDITION
+    BSLS_ASSERT_SAFE(d_scheduler_p);
```
678098, line 123: Cached single scheduler `d_scheduler_p` so we have a guarantee that the same scheduler is used for both scheduling/cancelling events.
Before, the API has weakness, allowing providing different schedulers

chrisbeard (approved):

PR #1641, opened 2026-07-24
Fix[bmqio]: return peerUri by value to eliminate data race

### LINKED ISSUE #1611: Fix[bmqio]: data races on d_streamSocket_sp and d_peerUri in NtcChannel
`processClose()` resets `d_streamSocket_sp` under `d_mutex`, but `setWriteQueueLowWatermark()`, `setWriteQueueHighWatermark()`, `peerEndpoint()`, and `sourceEndpoint()` read and dereference it without acquiring the lock. A concurrent `close()` resets the `shared_ptr` mid-read, corrupting the control block refcount.

`peerUri()` returns `d_peerUri` by const reference without the lock while `processConnect()` writes it under `d_mutex`, the same race fixed in `Channel::d_description` (#1566).

SrinathhSatuluri:
Closes #1611

Return peerUri by value instead of by const reference so the caller is not left holding an unprotected reference after the lock drops. This changes the base Channel interface and all four implementations. In ResolvingChannelFactory_Channel, the constructor previously took the address of the return value for the AtomicPointer pattern, so a stored copy is used instead. Also fixes two latent StringRef dangles that only worked because peerUri returned a reference to a stable member.

@678098

### DIFF
--- a/src/groups/bmq/bmqio/bmqio_ntcchannel.cpp
+++ b/src/groups/bmq/bmqio/bmqio_ntcchannel.cpp
@@ -1342,8 +1342,9 @@ ntsa::Endpoint NtcChannel::sourceEndpoint() const
                              : ntsa::Endpoint();
 }
 
-const bsl::string& NtcChannel::peerUri() const
+bsl::string NtcChannel::peerUri() const
 {
+    bslmt::LockGuard<bslmt::Mutex> lock(&d_mutex);
     return d_peerUri;
 }
 
--- a/src/groups/bmq/bmqio/bmqio_resolvingchannelfactory.cpp
+++ b/src/groups/bmq/bmqio/bmqio_resolvingchannelfactory.cpp
@@ -95,12 +95,13 @@ ResolvingChannelFactory_Channel::ResolvingChannelFactory_Channel(
     bslma::Allocator*               basicAllocator)
 : DecoratingChannelPartialImp(channel, basicAllocator)
 , d_resolvedPeerUri(basicAllocator)
+, d_basePeerUri(channel->peerUri(), basicAllocator)
 , d_peerUri()
 {
     // PRECONDITIONS
     BSLS_ASSERT(channel);
 
-    d_peerUri = &channel->peerUri();
+    d_peerUri = &d_basePeerUri;
 }
 
 // MANIPULATORS
@@ -117,7 +118,7 @@ void ResolvingChannelFactory_Channel::updatePeerUri()
 }
 
 // ACCESSORS
-const bsl::string& ResolvingChannelFactory_Channel::peerUri() const
+bsl::string ResolvingChannelFactory_Channel::peerUri() const
 {
     return *d_peerUri;
 }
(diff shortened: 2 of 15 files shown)


src/groups/bmq/bmqio/bmqio_resolvingchannelfactory.cpp
```diff
@@ -95,12 +95,13 @@ ResolvingChannelFactory_Channel::ResolvingChannelFactory_Channel(
     bslma::Allocator*               basicAllocator)
 : DecoratingChannelPartialImp(channel, basicAllocator)
 , d_resolvedPeerUri(basicAllocator)
+, d_basePeerUri(channel->peerUri(), basicAllocator)
```
SrinathhSatuluri, line 98: peerUri() returns a temporary now, so d_peerUri needs a stable copy to point at. d_basePeerUri owns that copy until resolution replaces it.

src/groups/bmq/bmqio/bmqio_ntcchannel.cpp
```diff
@@ -1342,8 +1342,9 @@ ntsa::Endpoint NtcChannel::sourceEndpoint() const
                              : ntsa::Endpoint();
 }
 
-const bsl::string& NtcChannel::peerUri() const
+bsl::string NtcChannel::peerUri() const
```
SrinathhSatuluri, line 1345: Return by value so the caller gets a snapshot. A reference would be unprotected after the lock drops.

src/groups/bmq/bmqio/bmqio_resolvingchannelfactory.cpp
```diff
@@ -248,7 +249,7 @@ void ResolvingChannelFactoryUtil::defaultResolutionFn(
 
     bdlma::LocalSequentialAllocator<128> arena;
 
-    bsl::string     ipAddrStr(peerUri.data(), colon.data(), &arena);
+    bsl::string     ipAddrStr(peerUri.c_str(), colon.data(), &arena);
```
SrinathhSatuluri, line 252: bsl::string::data() returns char* in C++17, StringRef::data() returns const char*. Iterator pair deduction fails on the mismatch. c_str() is always const char*.

src/groups/mqb/mqbnet/mqbnet_session.h
```diff
@@ -99,7 +99,7 @@ class Session : public SessionEventProcessor {
     /// executed.  The execution result (structured, non-structured text or
     /// possible error message) is expected to be passed to the specified
     /// `onProcessed` callback.
-    typedef bsl::function<void(const bslstl::StringRef&       source,
+    typedef bsl::function<void(const bsl::string&             source,
```
SrinathhSatuluri, line 102: StringRef bound into a thread pool job captures a pointer, not the data. With peerUri() returning a temporary, the data is freed before the job runs.

678098 (approved):
Looks good

678098:

PR #1644, opened 2026-07-27
Fix[mqbnet]: keep a weak ptr to a channel for reauthn

678098:
It seems wrong to own a copy of a shared_ptr to a channel in timeout callback.

channel_lifetime


chrisbeard (approved):
I think this makes sense, lgtm

PR #1656, opened 2026-07-28
Fix: Add accessor calls to the uri fuzztest

jakeski2015:
There was no values being read in the initial fuzztest. Even though a lot of these accessor calls are likely not going to find anything due to their simplicity, it's still worth fuzz testing the class because the URI comes directly from the client.


pniedzielski (approved):
Good catch

PR #1663, opened 2026-07-29
Fix[mqbnet]: active-node PANIC alarms during shutdown

chrisbeard:
**Describe your changes**
During shutdown, proxies often emit a false cluster availability alarm.

> ERROR mqbnet_clusteractivenodemanager.cpp PANIC [CLUSTER_ACTIVE_NODE] ClusterProxy (foo): no node available !!!

The root cause is shutdown ordering:

1. `initiateShutdown` runs and each `ClusterProxy` sets `d_isStopping`.
2. The broker stops `TransportManager` *before* the `ClusterCatalog`, so every cluster node channel gets closed while the proxy is still registered as a node state observer.
3. Each channel close triggers `ClusterActiveNodeManager::onNodeDown`. When the current active node drops it re-selects another node. The remaining nodes' down events are still queued and therefore still appear available.
4. This repeats until the last node drops, `onNewActiveNode(0)` finds no candidates and raises the PANIC alarm.
5. `ClusterProxy`'s `d_isStopping` guard logic sit below the alarm in `onActiveNodeUp` and `onActiveNodeDown`, so they suppress the reaction to the event but not the alarm itself.

To fix, make `ClusterActiveNodeManager` shutdown-aware.

- `ClusterProxy` tells the `ClusterActiveNodeManager` when it enters the stopping state (in `initiateShutdownDispatched`) so that no PANIC alarm is logged when the last node drops.
- Add unit test covering the shutdown case.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
@@ -138,8 +138,9 @@ void ClusterProxy::initiateShutdownDispatched(const VoidFunctor& callback)
 
     BALL_LOG_INFO << "Shutting down ClusterProxy: [name: '" << name() << "']";
 
-    // Mark self as stopping.
+    // Mark self as stopping and notify active node manager.
     d_isStopping = true;
+    d_activeNodeManager.onClusterStopping();
 
     d_queueHelper.requestToStopQueues();
     // 'checkUnconfirmedV2' serves as synchronization.
--- a/src/groups/mqb/mqbnet/mqbnet_clusteractivenodemanager.t.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_clusteractivenodemanager.t.cpp
@@ -316,6 +316,88 @@ static void test4_panicInExtendedMode()
     }
 }
 
+static void test5_noPanicWhenStopping()
+// Validate that once the cluster is marked as stopping, failing to select an
+// active node does NOT emit a PANIC log (all nodes are expected to go down
+// during shutdown).
+{
+    bmqtst::TestHelper::printTestName("NO PANIC WHEN STOPPING");
+
+    bmqtst::ScopedLogObserver logObserver(ball::Severity::e_ERROR,
+                                          bmqtst::TestHelperUtil::allocator());
+
+    // Set up mock cluster
+    mqbcfg::ClusterDefinition clusterConfig(
+        bmqtst::TestHelperUtil::allocator());
+    bdlbb::PooledBlobBufferFactory bufferFactory(
+        1024,
+        bmqtst::TestHelperUtil::allocator());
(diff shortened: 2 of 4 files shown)


src/groups/mqb/mqbnet/mqbnet_clusteractivenodemanager.t.cpp
```diff
@@ -316,6 +316,88 @@ static void test4_panicInExtendedMode()
...
+    mqbnet::ClusterActiveNodeManager mgr =
+        mqbnet::ClusterActiveNodeManager(nodes, description, dataCenter);
+    BMQTST_ASSERT(!mgr.activeNode());
+
+    bmqp_ctrlmsg::NegotiationMessage negotiationMessage(
+        bmqtst::TestHelperUtil::allocator());
+    negotiationMessage.makeClientIdentity().hostName() = "dummyIdentity";
+
+    // Mark the cluster as stopping.
+    mgr.onClusterStopping();
```
678098, line 373: The UT uses the newly added `onClusterStopping`.
Does this exact code fail in `main` if `onClusterNodeStopping` is removed?

src/groups/mqb/mqbblp/mqbblp_clusterproxy.cpp
```diff
@@ -138,8 +138,11 @@ void ClusterProxy::initiateShutdownDispatched(const VoidFunctor& callback)
 
     BALL_LOG_INFO << "Shutting down ClusterProxy: [name: '" << name() << "']";
 
-    // Mark self as stopping.
+    // Mark self as stopping.  Also notify the active node manager: during the
+    // subsequent teardown all nodes go down, and losing the last one must not
+    // raise a spurious "no node available" PANIC.
```
678098:
```suggestion
    // Mark self as stopping.
```
I think the comment should not mention the bug that is already fixed. The comments should mostly tell about the current state of the code, not the entire multiverse of code changes, because we have to read it and we have limited attention.

`onClusterStopping` docs tell exactly the current state: it suppresses the alarm.

src/groups/mqb/mqbnet/mqbnet_clusteractivenodemanager.t.cpp
chrisbeard: Yes, it does fail.

```
TEST /blazingmq/src/groups/mqb/mqbnet/mqbnet_clusteractivenodemanager.t.cpp CASE 5

29JUL2026_18:16:53.623 71872:8519671168 ERROR /blazingmq/src/groups/mqb/mqbnet/mqbnet_clusteractivenodemanager.cpp:218 MQBNET.CLUSTERACTIVENODEMANAGER PANIC [CLUSTER_ACTIVE_NODE] dummy: no node available !!! PANIC
```

678098 (approved):

PR #1669, opened 2026-07-29
Fix[bmqc]: first clean and block queues, then unset d_started

678098:
`MQTP::stop` sets `d_started = false` first before stopping each dispatcher queue.
This leads to a thread race where dispatcher events being processed in its thread might try to enqueue another event through a dispatcher that is going to close. Since we've set `d_started = false`, enqueue attempt fails on assertion.

The fix is to set `d_started` flag only after we've disabled push to each dispatcher queue and cleared its contents. During stop, we just drop any events in queues, so disabling push is equivalent to just dropping a dispatcher event, which is fine at shutdown.

mqtp_crash_sequence

### DIFF
--- a/src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
+++ b/src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
@@ -856,8 +856,6 @@ inline void MultiQueueThreadPool<TYPE>::stop()
 {
     BSLS_ASSERT_SAFE(isStarted() && "MQTP has not been started");
 
-    d_started = false;
-
     if (d_config.d_eventScheduler_p) {
         d_config.d_eventScheduler_p->cancelEventAndWait(
             &d_alarmMonitorEventHandle);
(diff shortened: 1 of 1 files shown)


src/groups/bmq/bmqc/bmqc_multiqueuethreadpool.h
```diff
@@ -856,8 +856,6 @@ inline void MultiQueueThreadPool<TYPE>::stop()
 {
     BSLS_ASSERT_SAFE(isStarted() && "MQTP has not been started");
 
-    d_started = false;
```
chrisbeard, line 859: It feels like this is a logical bug because we use `d_started` to indicate the started _and_ stopping states. If the stopping state were distinct, we would be able to test that condition and avoid asserting?

678098: If we have one variable `d_state = {stopped, starting, stopping}`, it will still assert, unless we change the condition to `BSLS_ASSERT(d_state != stopped)` everywhere.
If we have a second independent bool flag `d_stopped` that we set to true during `stop()` without changing `d_started`, the assertion will be gone; however, `d_stopped` flag is useless on its own.
We can achieve the same behavior without introducing a new variable or state.

PR #1673, opened 2026-07-30
Refactor[mqbs::FileStore]: Encapsulate write-head cursor accessors

kaikulimu:
## Summary
  - Consolidates access to the FileStore "write head" — the `(leaseId, seqNum)` cursor identifying the next record this store writes or applies — behind a small, intention-revealing accessor family, instead of scattered raw field/map manipulation.
  - Follows up the earlier write-head renames (member `d_primaryLeaseId` → `d_writeHeadLeaseId` in #1660, accessor `primaryLeaseId()` → `writeHeadLeaseId()` in #1666): the two halves of the cursor now have symmetric names and the leaky mutable-reference accessor is gone.
  - Pure refactoring — **no behavior change**.


src/groups/mqb/mqbs/mqbs_filestore.h
```diff
@@ -1160,9 +1161,16 @@ inline FileStore::NodeContext::NodeContext(BlobSpPool* blobSpPool_p,
...
+{
+    return ++d_highestSeqNums.at(d_writeHeadLeaseId);
+}
+
+inline void FileStore::setWriteHead(unsigned int        leaseId,
+                                    bsls::Types::Uint64 seqNum)
 {
-    return d_highestSeqNums[d_writeHeadLeaseId];
+    d_writeHeadLeaseId        = leaseId;
+    d_highestSeqNums[leaseId] = seqNum;
```
678098, line 1176: Should this method has any preconditions such as `d_writeHeadLeaseId <= leaseId`?

src/groups/mqb/mqbs/mqbs_filestore.h
```diff
@@ -1160,9 +1161,16 @@ inline FileStore::NodeContext::NodeContext(BlobSpPool* blobSpPool_p,
 // ---------------
 
 // PRIVATE MANIPULATORS
-inline bsls::Types::Uint64& FileStore::currentSeqNumRef()
+inline bsls::Types::Uint64 FileStore::incrementWriteHeadSeqNum()
+{
+    return ++d_highestSeqNums.at(d_writeHeadLeaseId);
```
678098: `.at` might throw an exception and the current design choice is not to throw exceptions in the code:
```
    at(const LOOKUP_KEY& key) {
        HashTableLink *node = d_impl.find(key);

        if (!node) {
            BloombergLP::bslstl::StdExceptUtil::throwOutOfRange(
                      "unordered_map<...>::at(LOOKUP_KEY): invalid key value");
        }
        return static_cast<HashTableNode *>(node)->value().second;
    }
```

What about changing it to
```
auto iter = d_highestSeqNums.find(d_writeHeadLeaseId);
BSLS_ASSRT_SAFE(iter != d_highestSeqNums.end());
++iter.second;
```

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -6052,7 +6051,7 @@ int FileStore::writeQueueCreationRecord(DataStoreRecordHandle*  handle,
                     qlistOffset,
                     qlistRecTotalLength);
 
-    DataStoreRecordKey key(sequenceNumber(), d_writeHeadLeaseId);
+    DataStoreRecordKey key(writeHeadSeqNum(), d_writeHeadLeaseId);
```
678098: Found it weird that `DataStoreRecordKey` has inverted order of args compared to basically everything in this file, for example:
```
            setWriteHead(recHeader->primaryLeaseId(),
                         recHeader->sequenceNumber());
```

src/groups/mqb/mqbs/mqbs_filestore.h
kaikulimu: No `auto` in C++03, but I will do the assert instead.

src/groups/mqb/mqbs/mqbs_filestore.cpp
kaikulimu: I can also fix it in this PR. Still pure refactoring so won't be too much.

src/groups/mqb/mqbs/mqbs_filestore.h
kaikulimu: Not yet, in legacy mode in [`FileStore::clearPrimary()`](https://github.com/bloomberg/blazingmq/blob/e7d19e8940ba5378341a947bbdca92b7c1de2f6b/src/groups/mqb/mqbs/mqbs_filestore.cpp#L7159) there is a write head rollbacks mechanism. I am trying to eliminate rollbacks but will do it across a series of PRs to make the changes gradual.

src/groups/mqb/mqbs/mqbs_datastore.h
```diff
@@ -179,10 +179,10 @@ struct DataStoreRecord {
 struct DataStoreRecordKey {
   public:
     // PUBLIC DATA
-    bsls::Types::Uint64 d_sequenceNum;
-
     unsigned int d_primaryLeaseId;
 
+    bsls::Types::Uint64 d_sequenceNum;
```
678098, line 182: Now member alignment is not ideal:
```
d_primaryLeaseId - 4 bytes
[empty space] - 4 bytes
d_sequenceNum - 8 bytes
```

If only we could use 64bit type for `leaseId` too...

src/groups/mqb/mqbs/mqbs_datastore.h
```diff
@@ -192,8 +192,8 @@ struct DataStoreRecordKey {
     // CREATORS
     DataStoreRecordKey();
 
-    DataStoreRecordKey(const bsls::Types::Uint64 sequenceNum,
-                       unsigned int              primaryLeaseId);
+    DataStoreRecordKey(unsigned int              primaryLeaseId,
+                       const bsls::Types::Uint64 sequenceNum);
```
678098:
```suggestion
                      bsls::Types::Uint64 sequenceNum);
```
This is a basic type, no need to have `const`

src/groups/mqb/mqbs/mqbs_filestore.h
```diff
@@ -1160,9 +1161,20 @@ inline FileStore::NodeContext::NodeContext(BlobSpPool* blobSpPool_p,
...
 
 // PRIVATE MANIPULATORS
-inline bsls::Types::Uint64& FileStore::currentSeqNumRef()
+inline bsls::Types::Uint64 FileStore::incrementWriteHeadSeqNum()
+{
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(d_highestSeqNums.find(d_writeHeadLeaseId) !=
+                     d_highestSeqNums.end());
+
+    return ++d_highestSeqNums[d_writeHeadLeaseId];
```
678098: There are 2 lookups now when we can have 1 for better performance.

```
BSLS_ASSERT_SAFE(d_highestSeqNums.find(d_writeHeadLeaseId) !=
                     d_highestSeqNums.end());
```
This check searches for iterator and immediately drops it, when it can be reused to increment the value

src/groups/mqb/mqbs/mqbs_datastore.h
kaikulimu: Switched them back only in the data member declarations.

678098 (approved):
Looks good

PR #1675, opened 2026-07-31
Fix[mqb]: improve non-printable handling in negotiation/session logs

chrisbeard:
Restrict negotiation messages (and derived fields) to printable characters in logs.

Additionally, add a BSLS_REVIEW check for negotiation messages containing non-printable characters in anticipation of enforcing that string fields are limited to printable ASCII characters.

### DIFF
--- a/src/groups/bmq/bmqu/bmqu_stringutil.cpp
+++ b/src/groups/bmq/bmqu/bmqu_stringutil.cpp
@@ -17,12 +17,14 @@
 
 #include <bmqscm_version.h>
 // BDE
+#include <bdlb_chartype.h>
 #include <bsl_algorithm.h>
 #include <bsl_bitset.h>
 #include <bsl_cctype.h>
 #include <bsl_climits.h>
 #include <bsl_functional.h>
 #include <bsls_assert.h>
+#include <bsls_performancehint.h>
 
 namespace BloombergLP {
 namespace bmqu {
@@ -110,6 +112,20 @@ bool StringUtil::endsWith(const bslstl::StringRef& str,
     int j = static_cast<int>(suffix.length() - 1);
     while ((i >= 0) && (j >= 0)) {
         if (str[i--] != suffix[j--]) {
+            return false;  // RETURN
+        }
+    }
+
+    return true;
+}
+
+bool StringUtil::isPrintable(const bslstl::StringRef& str)
+{
+    for (bslstl::StringRef::const_iterator it = str.begin(); it != str.end();
+         ++it) {
+        if (BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY(
+                !bdlb::CharType::isPrint(*it))) {
+            BSLS_PERFORMANCEHINT_UNLIKELY_HINT;
             return false;  // RETURN
         }
     }
--- a/src/groups/bmq/bmqu/bmqu_stringutil.h
+++ b/src/groups/bmq/bmqu/bmqu_stringutil.h
@@ -57,6 +57,10 @@ struct StringUtil {
     static bool endsWith(const bslstl::StringRef& str,
                          const bslstl::StringRef& suffix);
 
+    /// Return `true` if every character in the specified string `str` is
+    /// printable, and `false` otherwise.  An empty string is printable.
+    static bool isPrintable(const bslstl::StringRef& str);
+
     /// Perform an in-place white spaces trimming at the beginning and the
     /// end of the specified string `str` and return a reference offering
     /// modifiable access to it.
--- a/src/groups/bmq/bmqu/bmqu_stringutil.t.cpp
+++ b/src/groups/bmq/bmqu/bmqu_stringutil.t.cpp
@@ -524,6 +524,60 @@ static void test9_squeeze()
     }
 }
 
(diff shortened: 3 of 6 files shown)


src/groups/bmq/bmqu/bmqu_stringutil.cpp
```diff
@@ -117,6 +118,18 @@ bool StringUtil::endsWith(const bslstl::StringRef& str,
     return true;
 }
 
+bool StringUtil::isPrintable(const bslstl::StringRef& str)
+{
+    for (bslstl::StringRef::const_iterator it = str.begin(); it != str.end();
+         ++it) {
+        if (!bdlb::CharType::isPrint(*it)) {
```
678098: If we don't expect this to happen in most cases, we can put this in `BSLS_PERFORMANCEHINT_PREDICT_UNLIKELY` block with `BSLS_PERFORMANCEHINT_UNLIKELY_HINT`

src/groups/bmq/bmqu/bmqu_stringutil.t.cpp
```diff
@@ -524,6 +524,61 @@ static void test9_squeeze()
...
+                  {L_, "\x7f", false},   // DEL
+                  {L_, "\x80", false},   // high-bit byte
+                  {L_, "\xff", false}};  // high-bit byte
+
+    const size_t k_NUM_DATA = sizeof(k_DATA) / sizeof(*k_DATA);
+
+    for (size_t idx = 0; idx < k_NUM_DATA; ++idx) {
+        const Test& test = k_DATA[idx];
+
+        PVV(test.d_line << ": checking printability of line " << test.d_line);
```
678098:
```suggestion
        PVV(test.d_line << ": checking printability");
```
the same printed twice

src/groups/bmq/bmqu/bmqu_stringutil.h
```diff
@@ -57,6 +57,10 @@ struct StringUtil {
     static bool endsWith(const bslstl::StringRef& str,
                          const bslstl::StringRef& suffix);
 
+    /// Return `true` if every character in the specified string `str` is
+    /// printable, and `false` otherwise.  An empty string is printable.
+    static bool isPrintable(const bslstl::StringRef& str);
```
678098, line 62: I thought we were going to use `bsl::string_view` from now on.
Worth changing in the future

src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
```diff
@@ -196,19 +199,58 @@ void loadBrokerIdentity(bmqp_ctrlmsg::ClientIdentity* identity,
...
+        }
+    }
+}
+
+/// Return the streamed representation of `obj`, safe to log (non-printable
+/// characters replaced with `?`).
+template <class TYPE>
+bsl::string logSafe(const TYPE& obj)
+{
+    bmqu::MemOutStream os;
```
678098, line 223: Might use local sequential allocator to ensure no heap allocations in most cases

src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
```diff
@@ -196,19 +199,58 @@ void loadBrokerIdentity(bmqp_ctrlmsg::ClientIdentity* identity,
...
+    sanitize(&result);
+    return result;
+}
+
+/// Return `true` if the streamed representation of `obj` contains only
+/// printable characters, and `false` otherwise.
+template <class TYPE>
+bool isPrintable(const TYPE& obj)
+{
+    bmqu::MemOutStream os;
```
678098, line 236: Local sequential allocator might be used here too

src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
```diff
@@ -275,6 +318,10 @@ int SessionNegotiator::createSessionOnMsgType(
     const NegotiationContextSp& negotiationContext =
         context_p->negotiationContext();
 
+    // Detect non-printable characters in negotiation messages.
+    // TODO: fail negotiation if message contains non-printable characters
+    BSLS_REVIEW_OPT(isPrintable(negotiationContext->negotiationMessage()));
```
678098, line 326: If `negotiationMessage` is printable, it automatically means that its client identity is also printable. This means that the following safe logs are not needed: `logSafe(clientIdentity)`

chrisbeard: We don't know if it's printable, that's why we have to sanitize until we reach the end goal of rejecting negotiation if not printable.

src/groups/bmq/bmqu/bmqu_stringutil.h
chrisbeard: All methods in this class using StringRef, I'll save that migration for another PR.

src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
chrisbeard: I followed convention in this file, most places don't use it for logging.

678098 (approved):

PR #1681, opened 2026-07-31
Fix[mqbs::FileStore]: Track write-head leaseId on applied records 

kaikulimu:
## Summary
  - Makes a FileStore replica advance its **write head** (the `(leaseId, seqNum)` cursor) to the leaseId of every record it applies in `processStorageEvent`, not just during partition-sync. Previously the leaseId was bumped only for partition-sync events, so a record carrying a higher leaseId applied through any other path left the write head — and therefore `writeHeadSeqNum()` — lagging the record just written.
  - A lagging write head desyncs the reported PSN and lets a later `setActivePrimary` for the new leaseId reset the (now stale) sequence number to zero, which can cascade into a spurious out-of-sync/abort.
  - In the FSM workflow this can arise when a record of a new lease is applied before the primary's active-status advisory (`setActivePrimary`) has been processed. The write head now follows the record so the PSN stays correct regardless of ordering.

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_filestore.cpp
+++ b/src/groups/mqb/mqbs/mqbs_filestore.cpp
@@ -548,7 +548,7 @@ int FileStore::openInRecoveryMode(bsl::ostream&    errorDescription,
             if (0 == jit.lastSyncPointPosition()) {
                 // Scenario (3) from above.
 
-                needTruncation   = true;
+                needTruncation     = true;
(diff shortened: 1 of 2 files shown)


678098 (approved):

kaikulimu:
Accidentally closed. Now the changes are on https://github.com/bloomberg/blazingmq/pull/1687

PR #1690, opened 2026-08-04
Feat: Add authorization checks to session creation

hallfox:
This PR adds authorization checks for session negotiation. Client, admin, and cluster peer session creation now needs to be authorized in order to be established. In an upcoming PR both session objects will use the authorizer handle to authorize access to certain resources while the session is active.

### DIFF
--- a/src/groups/mqb/mqba/mqba_adminsession.cpp
+++ b/src/groups/mqb/mqba/mqba_adminsession.cpp
@@ -301,8 +301,8 @@ AdminSession::AdminSession(
               d_clientIdentity_p->features()),
           allocator)
 , d_scheduler_p(scheduler)
-, d_adminCb(adminCb)
 , d_authorizer_sp(authorizer)
+, d_adminCb(adminCb)
 {
     // Register this client to the dispatcher
     mqbi::Dispatcher::ProcessorHandle processor = dispatcher->registerClient(
--- a/src/groups/mqb/mqba/mqba_adminsession.cpp
+++ b/src/groups/mqb/mqba/mqba_adminsession.cpp
@@ -304,6 +304,14 @@ AdminSession::AdminSession(
 , d_authorizer_sp(authorizer)
 , d_adminCb(adminCb)
 {
+    // PRECONDITIONS
+    BSLS_ASSERT(channel);
+    BSLS_ASSERT(dispatcher);
+    BSLS_ASSERT(blobSpPool);
+    BSLS_ASSERT(scheduler);
+    BSLS_ASSERT(adminCb);
+    BSLS_ASSERT(authorizer);
+
     // Register this client to the dispatcher
     mqbi::Dispatcher::ProcessorHandle processor = dispatcher->registerClient(
         this,
--- a/src/groups/mqb/mqba/mqba_adminsession.cpp
+++ b/src/groups/mqb/mqba/mqba_adminsession.cpp
@@ -288,6 +288,7 @@ AdminSession::AdminSession(
     AdminSessionState::BlobSpPool*                blobSpPool,
     bdlmt::EventScheduler*                        scheduler,
     const mqbnet::Session::AdminCommandEnqueueCb& adminCb,
+    const bsl::shared_ptr<mqbi::Authorizer>&      authorizer,
     bslma::Allocator*                             allocator)
 : d_self(this)  // use default allocator
 , d_running(true)
@@ -301,6 +302,7 @@ AdminSession::AdminSession(
           allocator)
 , d_scheduler_p(scheduler)
 , d_adminCb(adminCb)
+, d_authorizer_sp(authorizer)
 {
     // Register this client to the dispatcher
     mqbi::Dispatcher::ProcessorHandle processor = dispatcher->registerClient(
--- a/src/groups/mqb/mqba/mqba_adminsession.h
+++ b/src/groups/mqb/mqba/mqba_adminsession.h
@@ -158,11 +158,11 @@ class AdminSession : public mqbnet::Session, public mqbi::DispatcherClient {
     /// Pointer to the event scheduler to use (held, not owned).
     bdlmt::EventScheduler* d_scheduler_p;
 
-    /// The callback to invoke on received admin command.
-    mqbnet::Session::AdminCommandEnqueueCb d_adminCb;
-
     /// The authorizer to use for this session
     bsl::shared_ptr<mqbi::Authorizer> d_authorizer_sp;
+
+    /// The callback to invoke on received admin command.
+    mqbnet::Session::AdminCommandEnqueueCb d_adminCb;
 
   private:
     // NOT IMPLEMENTED
--- a/src/groups/mqb/mqba/mqba_adminsession.h
+++ b/src/groups/mqb/mqba/mqba_adminsession.h
@@ -61,6 +61,9 @@ class Semaphore;
 namespace bmqp {
 class Event;
 }
+namespace mqbi {
+class Authorizer;
+}
 
 namespace mqba {
 
@@ -158,6 +161,9 @@ class AdminSession : public mqbnet::Session, public mqbi::DispatcherClient {
     /// The callback to invoke on received admin command.
     mqbnet::Session::AdminCommandEnqueueCb d_adminCb;
 
+    /// The authorizer to use for this session
+    bsl::shared_ptr<mqbi::Authorizer> d_authorizer_sp;
+
   private:
     // NOT IMPLEMENTED
 
@@ -224,6 +230,7 @@ class AdminSession : public mqbnet::Session, public mqbi::DispatcherClient {
                  AdminSessionState::BlobSpPool*          blobSpPool,
                  bdlmt::EventScheduler*                  scheduler,
                  const mqbnet::Session::AdminCommandEnqueueCb& adminEnqueueCb,
+                 const bsl::shared_ptr<mqbi::Authorizer>&      authorizer,
                  bslma::Allocator*                             allocator);
 
     /// Destructor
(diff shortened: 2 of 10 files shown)


src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
```diff
@@ -565,7 +563,11 @@ bsl::shared_ptr<mqbnet::Session> SessionNegotiator::onClientIdentityMessage(
                            clientIdentity,
                            *(context_p->channel().get()));
 
-    createSession(&session, context_p, description);
+    bsl::shared_ptr<mqbnet::Session> session;
+    rc = createSession(&session, context_p, description);
```
chrisbeard: We could do authorization here, right? In effect, all of the validation would be done here, and `createSession` would simply just pass the authorizer down?

In this method, we know for sure that we have an incoming connection, which might simplify things a bit.

src/groups/mqb/mqba/mqba_authenticator.cpp
```diff
@@ -305,6 +305,10 @@ void Authenticator::authenticate(
         else {
             event.value() = InitialConnectionEvent::e_AUTHN_SUCCESS;
         }
+
+        // Set authentication result
+        context_sp->setAuthenticationResult(result);
```
chrisbeard, line 310: 👍

chrisbeard (commented):
Overall looks very reasonable to me. Would be good to get some more eyes, but I think this is usable.

src/groups/mqb/mqba/mqba_clientsession.t.cpp
```diff
@@ -643,20 +644,35 @@ T assertFail()
     return T();
 }
 
+/// Authorizer for testing that allows all
```
pniedzielski, line 647: It is probably worth pulling out this authorizer that's duplicated in each test harness.  It seems generally useful (and like the good default for open-source).

pniedzielski: And, I suppose for the test, it might be good to have it accept a `bool` and return that, so we can test the failure case too.

src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
```diff
@@ -801,6 +810,11 @@ void SessionNegotiator::createSession(
                           ->negotiationMessage()
                           .isUndefinedValue());
 
+    enum {
+        e_OK                    = 0,
+        e_AUTHORIZATION_FAILURE = 1,
```
pniedzielski: Our convention for these awful anonymous `enum` return codes is usually negative for errors (see, e.g., https://github.com/bloomberg/blazingmq/blob/main/src/groups/mqb/mqbc/mqbc_storageutil.cpp#L1074).

pniedzielski: This change shouldn't affect your callsites, I think.

src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
pniedzielski: I think this would make the semantics a little more clear on `createSession`, rather than needing to know that `createSession` does authorization here and not in `onBrokerResponseMessage` below.

pniedzielski (changes requested):

src/groups/mqb/mqba/mqba_adminsession.cpp
```diff
@@ -301,6 +302,7 @@ AdminSession::AdminSession(
           allocator)
 , d_scheduler_p(scheduler)
 , d_adminCb(adminCb)
+, d_authorizer_sp(authorizer)
```
678098, line 304: Worth adding precondition checks in constructors, there are 0 checks there now

src/groups/mqb/mqba/mqba_adminsession.h
```diff
@@ -158,6 +161,9 @@ class AdminSession : public mqbnet::Session, public mqbi::DispatcherClient {
     /// The callback to invoke on received admin command.
     mqbnet::Session::AdminCommandEnqueueCb d_adminCb;
 
+    /// The authorizer to use for this session
+    bsl::shared_ptr<mqbi::Authorizer> d_authorizer_sp;
```
678098, line 162:
```suggestion
    /// The authorizer to use for this session
    bsl::shared_ptr<mqbi::Authorizer> d_authorizer_sp;

/// The callback to invoke on received admin command.
mqbnet::Session::AdminCommandEnqueueCb d_adminCb;
```

I would rather reorder these fields because `sizeof(shared_ptr)` is much smaller than sizeof of a callback. This gives more chances that `d_authorizer_sp` will be in a cache line due to better memory layout

src/groups/mqb/mqba/mqba_adminsession.t.cpp
```diff
@@ -132,6 +134,20 @@ struct TestAdminRetranslator {
...
 };
 
+/// Authorizer for testing that allows all
+class TestAuthorizer : public mqbi::Authorizer {
+  public:
+    TestAuthorizer() {}
+
+    bool
+    authorize(BSLA_UNUSED const mqbact::Action& action,
+              BSLA_UNUSED const mqbplug::AuthenticationResult& authnResult)
```
678098, line 144:
```suggestion
    authorize(const mqbact::Action&,
              const mqbplug::AuthenticationResult&)
```
Can use a shorter form and remove include to BSLA_UNUSED, less text to read

src/groups/mqb/mqba/mqba_adminsession.t.cpp
```diff
@@ -142,6 +158,7 @@ class TestBench {
     mqbmock::Dispatcher                 d_mockDispatcher;
     bdlmt::EventScheduler               d_scheduler;
     TestClock                           d_testClock;
+    bsl::shared_ptr<TestAuthorizer>     d_authorizer;
```
678098:
```suggestion
    bsl::shared_ptr<TestAuthorizer>     d_authorizer_sp;
```

src/groups/mqb/mqba/mqba_clientsession.cpp
```diff
@@ -2455,6 +2458,7 @@ ClientSession::ClientSession(
 , d_scheduler_p(scheduler)
 , d_periodicUnconfirmedCheckHandler()
 , d_shutdownChain(allocator)
+, d_authorizer_sp(authorizer)
```
678098, line 2461: No precondition checks here too

src/groups/mqb/mqba/mqba_clientsession.t.cpp
```diff
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "mqbi_authorizer.h"
```
678098:
```suggestion
#include <mqbi_authorizer.h>
```

src/groups/mqb/mqba/mqba_clientsession.t.cpp
```diff
@@ -643,20 +644,35 @@ T assertFail()
...
+};
+
 /// The `TestBench` holds system components together.
 class TestBench {
   public:
     // DATA
     bdlbb::PooledBlobBufferFactory            d_bufferFactory;
     BlobSpPool                                d_blobSpPool;
-    bsl::shared_ptr<bmqio::TestChannel>   d_channel;
+    bsl::shared_ptr<bmqio::TestChannel>       d_channel;
```
678098:
```suggestion
    bsl::shared_ptr<bmqio::TestChannel>       d_channel_sp;
```

src/groups/mqb/mqba/mqba_clientsession.t.cpp
```diff
@@ -643,20 +644,35 @@ T assertFail()
...
     mqbmock::Cluster                          d_cluster;
     mqbmock::Dispatcher                       d_mockDispatcher;
     MyMockDomain                              d_domain;
     mqbmock::DomainFactory                    d_mockDomainFactory;
     const bsl::shared_ptr<bmqst::StatContext> d_clientStatContext_sp;
     bdlmt::EventScheduler                     d_scheduler;
     TestClock                                 d_testClock;
+    bsl::shared_ptr<TestAuthorizer>           d_authorizer;
```
678098:
```suggestion
    bsl::shared_ptr<TestAuthorizer>           d_authorizer_sp;
```

src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
```diff
@@ -611,10 +612,14 @@ bsl::shared_ptr<mqbnet::Session> SessionNegotiator::onBrokerResponseMessage(
...
-        return session;  // RETURN
+    {
+        const int rc = populateNegotiationContext(errorDescription, context_p);
+        if (rc != 0) {
+            return NULL;  // RETURN
+        }
     }
+
+    bsl::shared_ptr<mqbnet::Session> session;
     createSession(&session, context_p, description);
```
678098, line 623:
```suggestion
    bsl::shared_ptr<mqbnet::Session> session = createSession(context_p, description);
```
I am thinking why don't we have such API

src/groups/mqb/mqba/mqba_sessionnegotiator.h
```diff
@@ -294,9 +301,9 @@ SessionNegotiator::setDomainFactory(mqbi::DomainFactory* value)
 }
 
 inline SessionNegotiator&
-SessionNegotiator::setAuthorizer(bslmf::MovableRef<AuthorizerMp> authorizer)
+SessionNegotiator::setAuthorizer(const AuthorizerSp& authorizer)
 {
```
678098, line 312: Might also add a precondition check to support non-null invariant in all lifetime of SessionNegotiator

src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
chrisbeard: It looks like at some point in the past, the function used to have a failure condition of some kind, but that is no longer the case. It can change (maybe we'll do it separately).

678098 (dismissed):
Looks good

678098 (approved):

PR #1695, opened 2026-08-05
CI: enable MSAN memory origins, sanitize all ITs

678098:
- Run all ITs in sanitizers CI
- Run sanitized ITs in parallel
- Enable MSAN memory origins

### DIFF
--- a/.github/workflows/sanitize.yaml
+++ b/.github/workflows/sanitize.yaml
@@ -39,4 +39,4 @@ jobs:
       - name: IT [c++,${{ github.job }}]
         if: ${{ matrix.fuzz == 'off' }}
         run: |
-          docker run --rm sanitizer-${{ github.job }} /blazingmq/cmake.bld/Linux/run-it.sh -k breathing --bmq-tolerate-dirty-shutdown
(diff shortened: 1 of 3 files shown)


.github/workflows/sanitize.yaml
```diff
@@ -39,4 +39,4 @@ jobs:
       - name: IT [c++,${{ github.job }}]
         if: ${{ matrix.fuzz == 'off' }}
         run: |
-          docker run --rm sanitizer-${{ github.job }} /blazingmq/cmake.bld/Linux/run-it.sh -k breathing --bmq-tolerate-dirty-shutdown
+          docker run --rm sanitizer-${{ github.job }} /blazingmq/cmake.bld/Linux/run-it.sh --bmq-tolerate-dirty-shutdown
```
chrisbeard: If I recall, this will take longer than the 6 hour github actions timeout.

678098: My point is just to run it and see if I have more crashes. Breathing tests path doesn't cover everything.
Maybe, there is enough time with matrix parallelization

PR #1696, opened 2026-08-05
Refactor: StringUtil::endsWith to avoid int casts

JJC3321:
*Issue number of the reported bug or feature request: #87*

**Describe your changes**
Refactor StringUtil::endsWith to use size_t indexing instead of int casts, avoiding -Wconversion warnings.

**Testing performed**
Ran bmqu_stringutil  tests with all cases passing. 

**Additional context**


JJC3321:
Hello @678098  

Can you review my PR.

chrisbeard:
Better yet, we should remove these helpers entirely and use [`bdlb::StringViewUtil`](https://bloomberg.github.io/bde-resources/doxygen/bde_api_prod/structbdlb_1_1StringViewUtil.html) from BDE.

https://github.com/bloomberg/blazingmq/pull/1697

678098:
Hi @JJC3321, thank you for showing this problem.
It turned out there are existing primitives to do the same in BDE library, so there is no need to support a custom implementation of these methods

PR #1700, opened 2026-08-10
Refactor[mqba]: authorize session negotiation earlier

chrisbeard:
Authorization should be performed as early as possible during session negotiation for incoming connections.

On authorization failure, send a minimal refusal BrokerResponse to the client indicating authorization failure.

### DIFF
--- a/src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
+++ b/src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
@@ -443,6 +443,24 @@ bsl::shared_ptr<mqbnet::Session> SessionNegotiator::onClientIdentityMessage(
     }
     }
 
+    // Authorize the incoming connection before sending the normal
+    // negotiation response.  If authorization fails, send a minimal refusal
+    // response and terminate negotiation.
+    if (!authorizeIncomingConnection(errorDescription, context_p)) {
+        bmqp_ctrlmsg::NegotiationMessage refusedMessage;
+        bmqp_ctrlmsg::BrokerResponse&    refused =
+            refusedMessage.makeBrokerResponse();
+        refused.result().category() = bmqp_ctrlmsg::StatusCategory::E_REFUSED;
+        refused.result().code()     = -1;
+        refused.result().message()  = "Connection not authorized";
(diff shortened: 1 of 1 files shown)


src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
```diff
@@ -443,6 +443,24 @@ bsl::shared_ptr<mqbnet::Session> SessionNegotiator::onClientIdentityMessage(
...
+            refusedMessage.makeBrokerResponse();
+        refused.result().category() = bmqp_ctrlmsg::StatusCategory::E_REFUSED;
+        refused.result().code()     = -1;
+        refused.result().message()  = "Connection not authorized";
+        refused.protocolVersion()   = bmqp::Protocol::k_VERSION;
+        refused.brokerVersion() = mqbcfg::BrokerConfig::get().brokerVersion();
+
+        bmqu::MemOutStream sendError;
+        sendNegotiationMessage(sendError, refusedMessage, negotiationContext);
+        return NULL;  // RETURN
```
678098, line 461: The connection is then dropped from the broker side, right?
Can we add an IT to challenge this behavior?

chrisbeard: Correct, all of the failure scenarios that return `NULL` here result in a disconnect.

678098: The fuzz test found something

chrisbeard: I think fuzz tests are showing an unrelated, existing race condition:

> 2026-08-10T18:30:44.2516667Z 10AUG2026_18:30:31.774 (140532517570432) INFO mqbnet_tcpsessionfactory.cpp:1419 TCPSessionFactory 'TCPInterface': waiting up to 20s for 1 clients to close
2026-08-10T18:30:44.2516774Z Broker hasn't exited after the timeout
2026-08-10T18:30:44.2516866Z Sending SIGQUIT to the broker...

678098 (approved):

PR #1701, opened 2026-08-10
Cleanup[mqba::SessionNegotiator]: require authorizer at construction

chrisbeard:
Provide Authorizer via SessionNegotiator constructor instead of via `setAuthorizer` as there is no ordering constraint for this field (unlike clusterCatalog/domainFactor).

Also add precondition checks to SessionNegotiator constructor for all mandatory fields.

### DIFF
--- a/src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
+++ b/src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
@@ -1001,12 +1001,12 @@ SessionNegotiator::SessionNegotiator(bdlbb::BlobBufferFactory* bufferFactory,
 , d_authorizer_sp(authorizer)
 {
     // PRECONDITIONS
-    BSLS_ASSERT(bufferFactory);
-    BSLS_ASSERT(dispatcher);
-    BSLS_ASSERT(statContext);
-    BSLS_ASSERT(blobSpPool);
-    BSLS_ASSERT(scheduler);
-    BSLS_ASSERT(authorizer);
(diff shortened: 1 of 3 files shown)


src/groups/mqb/mqba/mqba_sessionnegotiator.cpp
```diff
@@ -997,8 +998,15 @@ SessionNegotiator::SessionNegotiator(bdlbb::BlobBufferFactory* bufferFactory,
...
+, d_authorizer_sp(authorizer)
 {
-    // NOTHING
+    // PRECONDITIONS
+    BSLS_ASSERT(bufferFactory);
+    BSLS_ASSERT(dispatcher);
+    BSLS_ASSERT(statContext);
+    BSLS_ASSERT(blobSpPool);
+    BSLS_ASSERT(scheduler);
+    BSLS_ASSERT(authorizer);
```
678098:
```suggestion
    BSLS_ASSERT(d_bufferFactory_p);
    BSLS_ASSERT(d_dispatcher_p);
    BSLS_ASSERT(d_statContext_p);
    BSLS_ASSERT(d_blobSpPool_p);
    BSLS_ASSERT(d_scheduler_p);
    BSLS_ASSERT(d_authorizer_sp);
```
I think it is more useful to check each field itself, it gives guarantee that the fields were set in this constructor. Checking args only doesn't provide such a guarantee and might allow some bugs

678098 (dismissed):
Have one small comment, otherwise, good to merge

678098 (approved):

PR #1703, opened 2026-08-10
Fix[mqbs]: disable gc by TTL for broadcast queues

678098:
- Add an IT that checks broadcast correctness with TTL = 0
- Fix: disable gc for broadcast queues

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_inmemorystorage.cpp
+++ b/src/groups/mqb/mqbs/mqbs_inmemorystorage.cpp
@@ -82,6 +82,7 @@ InMemoryStorage::InMemoryStorage(DataStore*              dataStore_p,
 , d_isEmpty(1)
 , d_currentlyAutoConfirming()
 , d_autoConfirms(d_allocator_p)
+, d_isBroadcast(config.mode().isBroadcastValue())
 {
     BSLS_ASSERT_SAFE(0 <= d_ttlSeconds);  // Broadcast queues can use 0 for TTL
 
@@ -444,6 +445,11 @@ int InMemoryStorage::gcExpiredMessages(const bdlt::Datetime& currentTimeUtc,
                                        bsls::Types::Uint64   secondsFromEpoch,
                                        int                   limit)
 {
+    if (d_isBroadcast) {
+        // Broadcast queues never expire messages via TTL.
+        return 0;  // RETURN
+    }
+
     bsls::Types::Uint64      latestMsgTimestampEpoch = 0;
     int                      numMsgsDeleted = 0;
     const bsls::Types::Int64 now = bmqu::Time::highResolutionTimer();
--- a/src/integration-tests/test_broadcast.py
+++ b/src/integration-tests/test_broadcast.py
@@ -13,12 +13,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import time
 from itertools import islice
 
 import blazingmq.dev.it.testconstants as tc
 from blazingmq.dev.it.fixtures import (
     Cluster,
     order,
+    tweak,
 )
 from blazingmq.dev.it.process.client import Client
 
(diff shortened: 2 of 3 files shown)


src/integration-tests/test_broadcast.py
```diff
@@ -430,3 +432,92 @@ def test_add_variable_priority_consumers(
...
+        assert not node.erases_messages(uri, timeout=1), (
+            f"node '{node.name}' garbage-collected broadcast messages due to "
+            "TTL expiration while a consumer was subscribed"
+        )
+
+    missing = [payload for payload in posted if payload not in received]
+    assert not missing, (
+        f"{len(missing)} of {len(posted)} broadcast messages were never "
+        f"delivered to a subscribed consumer (e.g. {missing[:5]})"
+    )
```
678098, line 535: This IT fails in `main` differently each time

```
FAILED test_broadcast.py::test_ttl_zero_delivers_to_subscribed_consumer[multi_node-strong_consistency] - AssertionError: 1 of 2250 broadcast messages were never delivered to a subscribed consumer (e.g. ['m-1199'])
assert not ['m-1199']
```

```
FAILED test_broadcast.py::test_ttl_zero_delivers_to_subscribed_consumer[multi_node_fsm-eventual_consistency] - AssertionError: 2 of 2250 broadcast messages were never delivered to a subscribed consumer (e.g. ['m-249', 'm-2249'])
assert not ['m-249', 'm-2249']
```

dorjesinpo (commented):
this looks good.
one minor comment, maybe we can avoid `time.sleep(3)` if we wait for exact number of messages posted? Just trying to shave test time

src/groups/mqb/mqbs/mqbs_inmemorystorage.cpp
```diff
@@ -444,6 +445,11 @@ int InMemoryStorage::gcExpiredMessages(const bdlt::Datetime& currentTimeUtc,
                                        bsls::Types::Uint64   secondsFromEpoch,
                                        int                   limit)
 {
+    if (d_isBroadcast) {
```
chrisbeard: When first looking at the issue, I convinced myself that `queue()->isAtMostOnce()` was the right condition to check here

678098: What do you also think about renaming `isAtMostOnce` -> `isBroadcast`?
Will have less words for the same thing.
UPD: it's not the same thing.

src/integration-tests/test_broadcast.py
678098: Fails like this in main now
```
FAILED test_broadcast.py::test_ttl_zero_delivers_to_subscribed_consumer[single_node-strong_consistency] - AssertionError: node 'single' garbage-collected broadcast messages due to TTL expiration while a consumer was subscribed
assert not <re.Match object; span=(156, 296), match='queue [bmq://bmq.test.mem.broadcast/qqq1] and que>
 +  where <re.Match object; span=(156, 296), match='queue [bmq://bmq.test.mem.broadcast/qqq1] and que> = erases_messages('bmq://bmq.test.mem.broadcast/qqq1', timeout=1)
 +    where erases_messages = Broker(single).erases_messages
```

src/groups/mqb/mqbs/mqbs_inmemorystorage.cpp
678098: Also, queue is not set before we call `InMemoryStorage::setQueue`.
`config` is available on `InMemoryStorage` construction.
I think that having construction-time constraints is a safer practice than a dynamic values that might be not set in some time windows.

678098:
> this looks good.
> one minor comment, maybe we can avoid `time.sleep(3)` if we wait for exact number of messages posted? Just trying to shave test time

I updated the test to have less sleeps. As a result, it became less flaky: I observed only 1 IT failure compared to 4 in the previous one.

chrisbeard (approved):

PR #1708, opened 2026-08-12
Authentication docs

chrisbeard:
The Security page will cover authn, authz, and TLS. The initial authn docs cover the basics: the feature set, the protocol, client configuration, and broker plugin/configuration.

Broker-to-broker authn, authz, and TLS docs will be added when the features are fully ready.

Authenticator plugin essentials are also documented in the Plugins page, which we may want to revisit as we expand with additional plugin types, built-in vs external plugins, etc.

### DIFF
--- a/docs/features/plugins.md
+++ b/docs/features/plugins.md
@@ -11,6 +11,109 @@ nav_order: 9
 * toc
 {:toc}
 
+## Plugin Types and Loading
+
+BlazingMQ brokers support a pluggable extension system for running custom code
+inside the broker process.
+
+### Stats and Metrics Consumers
+
+Plugins in this category export broker metrics and statistics (for example, the
+[Prometheus plugin](#prometheus-plugin) below).
+
+### Authenticators
+
+Plugins in this category authenticate client credentials during connection setup.
+
+- **Built-in authenticators** are provided by the broker task and do not
+  require external plugin libraries (see
+  [Built-in Authenticators](security.md#built-in-authenticators)).
+- **Custom authenticators** are loaded from external shared libraries via
+  plugin configuration (see [Writing a Custom Authenticator
+  Plugin](#writing-a-custom-authenticator-plugin)).
+
+### Loading and Enabling Plugins
+
+Plugin libraries are loaded as shared libraries and enabled in the broker
+configuration file (`bmqbrkcfg.json`):
+
+```json
+"appConfig": {
+    "plugins": {
+        "libraries": ["<path-to-plugin-library-directory>"],
+        "enabled": ["<PluginName>"]
+    }
+}
+```
+
+Every plugin library exports an `instantiatePluginLibrary` C symbol so the
+broker can load it via `dlopen`.
+
+## Writing a Custom Authenticator Plugin
+
+Custom authenticators are dynamically loaded shared libraries.  A plugin
+provides three pieces:
+
+1. An **`Authenticator`** -- subclass of `mqbplug::Authenticator` that
+   implements `authenticate()`, `name()`, `mechanism()`, `start()`, and
+   `stop()`.
+2. An **`AuthenticatorPluginFactory`** -- creates instances of your
+   authenticator.
+3. A **`PluginLibrary`** -- exports the `instantiatePluginLibrary` C symbol
+   so the broker can load the plugin via `dlopen`.
+
+See
+[`mqbplug_authenticator.h`](https://github.com/bloomberg/blazingmq/blob/main/src/groups/mqb/mqbplug/mqbplug_authenticator.h)
+for the full interface documentation.  The built-in
+[`BasicAuthenticator`](https://github.com/bloomberg/blazingmq/blob/main/src/groups/mqb/mqbauthn/mqbauthn_basicauthenticator.cpp)
+and
+[`AnonAuthenticator`](https://github.com/bloomberg/blazingmq/blob/main/src/groups/mqb/mqbauthn/mqbauthn_anonauthenticator.cpp)
+serve as working reference implementations.
+
+A few things to keep in mind:
+
+{: .important }
+> `authenticate()` is called from the broker's authentication thread pool.
+> Your implementation **must be thread-safe** (the method is `const`).
+
+- The `authenticate()` method receives the raw credential bytes and the
+  client's IP address.  On success, return 0 and populate the result with a
+  `principal` and an optional `lifetimeMs`.  On failure, return non-zero and
+  write a reason to `errorDescription`.
+
+- Plugin settings are passed as key-value pairs from the broker config.  Use
+  `AuthenticatorUtil::findAuthenticatorConfig()` to look up your plugin's
+  config by name.
+
+### Deploying an authenticator plugin
+
+Compile the plugin as a shared library (`.so` / `.dylib`) and configure the
+broker to load it:
+
+```json
+{
+  "appConfig": {
+    "plugins": {
+      "libraries": ["/path/to/plugin/directory/"],
+      "enabled": ["MyJwtAuthenticator"]
+    },
+    "authentication": {
+      "authenticators": [
+        {
+          "name": "MyJwtAuthenticator",
+          "settings": []
+        }
+      ]
+    }
+  }
+}
+```
+
+For background on how authentication works in BlazingMQ, see
+[Security](security.md).
+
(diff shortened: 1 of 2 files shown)


docs/features/plugins.md
```diff
@@ -11,6 +11,109 @@ nav_order: 9
...
+
+## Writing a Custom Authenticator Plugin
+
+Custom authenticators are dynamically loaded shared libraries.  A plugin
+provides three pieces:
+
+1. An **`Authenticator`** -- subclass of `mqbplug::Authenticator` that
+   implements `authenticate()`, `name()`, `mechanism()`, `start()`, and
+   `stop()`.
+2. An **`AuthenticatorPluginFactory`** -- creates instances of your
```
678098:
```suggestion
2. An **`AuthenticatorPluginFactory`** -- creates instances of the defined
```

Possible to make docs less personal

docs/features/plugins.md
```diff
@@ -11,6 +11,109 @@ nav_order: 9
...
+A few things to keep in mind:
+
+{: .important }
+> `authenticate()` is called from the broker's authentication thread pool.
+> Your implementation **must be thread-safe** (the method is `const`).
+
+- The `authenticate()` method receives the raw credential bytes and the
+  client's IP address.  On success, return 0 and populate the result with a
+  `principal` and an optional `lifetimeMs`.  On failure, return non-zero and
+  write a reason to `errorDescription`.
```
678098: I think it is worth adding the function signature here if its args are mentioned one by one

docs/features/security.md
```diff
@@ -0,0 +1,309 @@
...
+
+3. The plugin returns an **`AuthenticationResult`** with a human-readable
+   `principal` and an optional `lifetimeMs`.
+
+4. The broker sends an **`AuthenticationResponse`** back to the client.  On
+   success, session negotiation proceeds.  On failure, the broker closes the
+   connection (see [Failure Handling](#failure-handling) below).
+
+5. If `lifetimeMs` is present, the client SDK schedules reauthentication at
+  **90 % of the lifetime** so authentication is renewed before expiry.
```
678098, line 95: I am now curious do we have a good reason to have 90% in SDK and 80% in broker-broker

https://github.com/bloomberg/blazingmq/blob/e2c0f29531a82bee1ab91064cbbb5e783a178303/src/groups/mqb/mqba/mqba_authenticationclient.cpp#L229-L230

docs/features/security.md
```diff
@@ -0,0 +1,309 @@
...
+      "minThreads": 1,
+      "maxThreads": 8
+    }
+  }
+}
+```
+
+| Field | Description |
+|-------|-------------|
+| `authenticators` | List of authenticator plugin configurations.  Each entry names a plugin and provides its settings.  All plugins must have unique mechanisms. |
```
678098, line 141:
```suggestion
| `authenticators` | List of authenticator plugin configurations.  Each entry names a plugin and provides its settings.  All plugins must have unique mechanisms, or the broker will refuse to start. |
```

Might mention consequences.

docs/features/plugins.md
```diff
@@ -11,6 +11,109 @@ nav_order: 9
...
+
+Plugin libraries are loaded as shared libraries and enabled in the broker
+configuration file (`bmqbrkcfg.json`):
+
+```json
+"appConfig": {
+    "plugins": {
+        "libraries": ["<path-to-plugin-library-directory>"],
+        "enabled": ["<PluginName>"]
+    }
```
678098: 4 spaces in this json, 2 spaces in others below

docs/features/plugins.md
```diff
@@ -11,6 +11,109 @@ nav_order: 9
...
+
+A few things to keep in mind:
+
+{: .important }
+> `authenticate()` is called from the broker's authentication thread pool.
+> Your implementation **must be thread-safe** (the method is `const`).
+
+- The `authenticate()` method receives the raw credential bytes and the
+  client's IP address.  On success, return 0 and populate the result with a
+  `principal` and an optional `lifetimeMs`.  On failure, return non-zero and
```
678098: Do you think it's worth adding a note that `lifetimeMs` should not be too low or otherwise SDK or other brokers might not have enough time to process event queues and perform reauthn?

docs/features/security.md
chrisbeard: It doesn't really matter, but we can normalize to 90%.

docs/features/plugins.md
chrisbeard: I think that's obvious enough, I don't think we need to state it explicitly. It's broadly true, not BlazingMQ-specific.

docs/features/plugins.md
chrisbeard: Reworked this a bit, I think the main point is to look at the header for the interface.

docs/features/security.md
chrisbeard: Added this to the top, it's generally true for all plugins.

docs/features/plugins.md
```diff
@@ -11,6 +11,108 @@ nav_order: 9
...
+### Stats and Metrics Consumers
+
+Plugins in this category export broker metrics and statistics (for example, the
+[Prometheus plugin](#prometheus-plugin) below).
+
+### Authenticators
+
+Plugins in this category authenticate client credentials during connection setup.
+
+- **Built-in authenticators** are provided by the broker task and do not
```
pniedzielski: The built-in authenticators aren't really plugins, but we conflate them here and elsewhere.  I'm not sure this discussion belongs here, but rather early on in the security page.

For here,

> Plugins in this category authenticate client credentials during connection setup.  While the BlazingMQ broker task provides a few [built-in authenticators](security.md#built-in-authenticators) that do not require external plugin libraries, but an external plugin library can provide custom authenticators to support additional authentication mechanisms, or to implement the built-in mechanisms differently.  See [Writing a Custom Authenticator Plugin](#writing-a-custom-authenticator-plugin) for more details.

docs/features/security.md
```diff
@@ -0,0 +1,309 @@
...
+### How Authentication Works in BlazingMQ
+
+The following sequence shows the authentication and negotiation flow when a
+client connects and provides credentials:
+
+```mermaid
+sequenceDiagram
+    participant Client
+    participant Broker
+    participant Plugin as Authenticator Plugin
```
pniedzielski: It doesn't really matter whether the authenticator came from a plugin or not.  I think we should just call it Authenticator.

docs/features/security.md
```diff
@@ -0,0 +1,309 @@
...
+        Plugin-->>Broker: AuthenticationResult
+        Broker-->>Client: AuthenticationResponse
+    end
+```
+
+1. The client sends an **`AuthenticationRequest`** containing the mechanism
+   name (e.g. `"BASIC"`) and credential data (mechanism-specific binary
+   payload).
+
+2. The broker looks up the **authenticator plugin** registered for that
```
pniedzielski: "The broker looks up the Authenticator registered for that mechanism...", here and below.

docs/features/security.md
```diff
@@ -0,0 +1,309 @@
...
+- **Non-blocking** -- broker authentication runs asynchronously on a dedicated
+  thread pool and does not block other clients.
+
+{: .important }
+> Authentication is not enforced by default.  Without explicit configuration,
+> unauthenticated clients are accepted as anonymous.  To require credentials,
+> configure an authenticator and set `anonymousCredential` to `disallow`
+> (see [Configuration](#configuration)).
+
+### How Authentication Works in BlazingMQ
```
pniedzielski, line 55: In terms of organization, this section feels out of place.  We haven't introduced the cast of characters yet.  Some section that talks about:

> Authentication is BlazingMQ managed by the *authenticators* configured on the broker.  Each authenticator implements a specific mechanism (e.g., `BASIC`, `JWT`), and the set of authenticators configured for the broker determines the mechanisms that a client can use to authenticate.  The broker provides a set of built-in authenticators, for supporting basic authentication workflows, and additional authenticators can be provided via the broker's [plugin mechanism](plugins.md).

would help a lot.

Something like this naturally leads to an example of how to configure authentication on the broker.  Then, the reference doc style sections below make sense.

pniedzielski (changes requested):

pniedzielski (approved):

PR #1709, opened 2026-08-12
Fix[mqbnet]: return identity by value with mutex guard

### LINKED ISSUE #1652: ClusterNodeImp::identity() returns by reference to unsynchronized mutable state
ClusterNode::identity() returns const bmqp_ctrlmsg::ClientIdentity& to d_identity, which is modified without synchronization in setChannel() at [mqbnet_clusterimp.cpp:77](https://github.com/bloomberg/blazingmq/blob/285e86c61f0abe9bdc91876de311186ac9781581/src/groups/mqb/mqbnet/mqbnet_clusterimp.cpp#L77) and resetChannel() at [mqbnet_clusterimp.cpp:132](https://github.com/bloomberg/blazingmq/blob/285e86c61f0abe9bdc91876de311186ac9781581/src/groups/mqb/mqbnet/mqbnet_clusterimp.cpp#L132). Both are called from IO threads. ClusterNodeImp has no mutex, and d_identity has no guard of any kind. Current callers happen to copy by value, so this may not manifest today, but the unprotected reference means any future const auto& binding would be a live data race.

Also worth noting that setChannel() and resetChannel() also write other members like d_description without synchronization. I suppose we should consider whether a fix should cover identity() alone or the broader class.

SrinathhSatuluri:
Closes #1652

`identity()` returned a const reference to `d_identity`, but `setChannel()` and `resetChannel()` write that member from the IO thread, so a caller on another thread could read a partially-written object. 

Return by value and guard the three access sites with a `mutable bslmt::Mutex`. The lock is scoped narrowly in `setChannel()` and `resetChannel()` and released before `notifyObserversOfNodeStateChange()` to avoid deadlock through the observer chain.

### DIFF
--- a/src/groups/mqb/mqbnet/mqbnet_clusterimp.cpp
+++ b/src/groups/mqb/mqbnet/mqbnet_clusterimp.cpp
@@ -66,6 +66,12 @@ ClusterNodeImp::~ClusterNodeImp()
     // NOTHING: Needed due to inheritance
 }
 
+bmqp_ctrlmsg::ClientIdentity ClusterNodeImp::identity() const
+{
+    bslmt::LockGuard<bslmt::Mutex> guard(&d_identityMutex);
+    return d_identity;
+}
+
 ClusterNode*
 ClusterNodeImp::setChannel(const bsl::weak_ptr<bmqio::Channel>& value,
                            const bmqp_ctrlmsg::ClientIdentity&  identity,
@@ -74,7 +80,10 @@ ClusterNodeImp::setChannel(const bsl::weak_ptr<bmqio::Channel>& value,
     // Save the value
     d_readCb    = readCb;
     d_isReading = false;
-    d_identity  = identity;
+    {
+        bslmt::LockGuard<bslmt::Mutex> guard(&d_identityMutex);
+        d_identity = identity;
+    }
 
     d_channel.setChannel(value);
 
@@ -129,7 +138,10 @@ ClusterNode* ClusterNodeImp::resetChannel(
     }
 
     d_isReading = false;
-    d_identity.reset();
+    {
+        bslmt::LockGuard<bslmt::Mutex> guard(&d_identityMutex);
+        d_identity.reset();
+    }
     d_readCb = bmqio::Channel::ReadCallback();
 
     // Notify the cluster of changes to this node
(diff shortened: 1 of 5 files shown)


src/groups/mqb/mqbnet/mqbnet_clusterimp.h
```diff
@@ -93,6 +93,8 @@ class ClusterNodeImp : public ClusterNode {
 
     bmqp_ctrlmsg::ClientIdentity d_identity;
 
+    mutable bslmt::Mutex d_identityMutex;
```
SrinathhSatuluri, line 96: mutable because identity() is const. The mutex is synchronization, not logical state.

src/groups/mqb/mqbnet/mqbnet_clusterimp.cpp
```diff
@@ -66,6 +66,12 @@ ClusterNodeImp::~ClusterNodeImp()
     // NOTHING: Needed due to inheritance
 }
 
+bmqp_ctrlmsg::ClientIdentity ClusterNodeImp::identity() const
```
SrinathhSatuluri, line 69: Moved out of the header to avoid adding bslmt_lockguard.h to the include set

src/groups/mqb/mqbnet/mqbnet_clusterimp.cpp
```diff
@@ -74,7 +80,10 @@ ClusterNodeImp::setChannel(const bsl::weak_ptr<bmqio::Channel>& value,
     // Save the value
     d_readCb    = readCb;
     d_isReading = false;
-    d_identity  = identity;
+    {
+        bslmt::LockGuard<bslmt::Mutex> guard(&d_identityMutex);
+        d_identity = identity;
+    }
```
SrinathhSatuluri, line 86: Lock released before notifyObserversOfNodeStateChange(), the observer chain calls back into identity(), which would deadlock on the non-recursive mutex

src/groups/mqb/mqbnet/mqbnet_mockcluster.h
```diff
@@ -143,7 +143,7 @@ class MockClusterNode : public ClusterNode {
     //   (virtual mqbnet::ClusterNode)
 
     /// Return identity from the last received negotiation message.
-    const bmqp_ctrlmsg::ClientIdentity& identity() const BSLS_KEYWORD_OVERRIDE;
+    bmqp_ctrlmsg::ClientIdentity identity() const BSLS_KEYWORD_OVERRIDE;
```
SrinathhSatuluri, line 146: No mutex in the mock, tests are single-threaded

678098 (approved):
Looks good

678098:

PR #1713, opened 2026-08-14
Fix[bmqp::RequestManager]: unsafe read without mutex

678098:
- Remove `reinterpret_cast`
- Add a UT to verify request cancellation order
- Fix unsafe field access without mutex
- Use `vector`, not the `map`, to iterate through the requests that has to be cancelled

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_requestmanager.cpp
+++ b/src/groups/bmq/bmqp/bmqp_requestmanager.cpp
@@ -33,6 +33,7 @@
 #include <bdlf_placeholder.h>
 #include <bdlmt_eventscheduler.h>
 #include <bsl_utility.h>
+#include <bsl_vector.h>
 #include <bsla_annotations.h>
 #include <bslma_managedptr.h>
 #include <bslmt_lockguard.h>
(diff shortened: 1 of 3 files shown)


src/groups/bmq/bmqp/bmqp_requestmanager.cpp
```diff
@@ -494,10 +496,7 @@ void RequestManager::cancelAllRequestsImpl(
             if (matchGroup && matchComponent) {
                 // Do not notify about timed out requests.
                 if (!it->second->d_haveTimeout) {
-                    BSLA_MAYBE_UNUSED bsl::pair<RequestMapIter, bool>
-                                      insertRC = requestsCopy.insert(
-                            bsl::make_pair(it->first, it->second));
-                    BSLS_ASSERT_SAFE(insertRC.second);
```
678098, line 500: We iterate through `bmqc::OrderedHashMap` so we have guarantee that keys in this map are unique. Due to this, `insert` return codes checks are redundant. Moreover, we don't need a second map here and can use a `vector`

src/groups/bmq/bmqp/bmqp_requestmanager.cpp
```diff
@@ -472,16 +470,20 @@ void RequestManager::cancelAllRequestsImpl(
...
+    typedef bsl::vector<RequestSp> RequestVec;
+
     // Note that requests must be cancelled in the same order in which they
-    // were sent.
+    // were sent.  Since 'd_requests' is traversed in insertion order, simply
+    // appending to a sequence container preserves that order.
 
-    // Create a new map so we can work on it outside the mutex
-    RequestMap requestsCopy(d_requests.bucket_count(),
-                            d_requests.get_allocator());
```
678098, line 480: `d_requests.bucket_count()` is not thread-safe: this call here was a bug

src/groups/bmq/bmqp/bmqp_requestmanager.cpp
```diff
@@ -472,16 +470,20 @@ void RequestManager::cancelAllRequestsImpl(
...
+    // Collect the requests to cancel so we can work on them outside the mutex
+    RequestVec requestsCopy(d_allocator_p);
 
     {
         bslmt::LockGuard<bslmt::Mutex> guard(&d_mutex);  // MUTEX LOCKED
 
+        requestsCopy.reserve(d_requests.size());
```
678098, line 485: This `reserve` might over-reserve memory, if we are cancelling only a small part of all requests. This raises a question what is better: dynamic resizes or one resize for a possibly big chunk of memory.
I think one allocation is better here

PR #1714, opened 2026-08-14
Fix[bmqp::RequestManager]: handle unset `rId` in response

678098:
Check for `rId` being not null before trying to perform lookup
`d_requests.find(response.rId().value());`.

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_requestmanager.cpp
+++ b/src/groups/bmq/bmqp/bmqp_requestmanager.cpp
@@ -404,6 +404,14 @@ int RequestManager::processResponse(
         rc_NOT_FOUND = -1
     };
 
+    if (response.rId().isNull()) {
+        // The response carries no correlator, so it cannot be matched to an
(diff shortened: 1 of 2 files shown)


src/groups/bmq/bmqp/bmqp_requestmanager.cpp
```diff
@@ -404,6 +404,14 @@ int RequestManager::processResponse(
         rc_NOT_FOUND = -1
     };
 
+    if (response.rId().isNull()) {
+        // The response carries no correlator, so it cannot be matched to an
+        // outstanding request.
+        BALL_LOG_DEBUG << "Received a response without a request id"
+                       << ", dropping it";
+        return rc_NOT_FOUND;  // RETURN
+    }
```
678098, line 413: A few lines below
```
RequestMapIter it = d_requests.find(response.rId().value());
```

pniedzielski (approved):

PR #1715, opened 2026-08-14
Fix[bmqp::RequestManager]: handle int32 signed overflow

678098:
- Wrap to 1 on reaching int32::max() when selecting the next requestId
- Attempt to find the next unused requestId in a map

### DIFF
--- a/src/groups/bmq/bmqp/bmqp_requestmanager.cpp
+++ b/src/groups/bmq/bmqp/bmqp_requestmanager.cpp
@@ -43,6 +43,18 @@
 namespace BloombergLP {
 namespace bmqp {
 
+namespace {
+
+/// Lowest identifier which can be assigned to a request; zero is reserved
+/// to indicate that no request has been sent yet.
+const int k_FIRST_REQUEST_ID = 1;
+
+/// Highest identifier which can be assigned to a request, as dictated by
+/// the type of the identifier field on the wire.
+const int k_LAST_REQUEST_ID = bsl::numeric_limits<int>::max();
+
+}  // close unnamed namespace
+
 // --------------------
 // class RequestManager
 // --------------------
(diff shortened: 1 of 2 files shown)


src/groups/bmq/bmqp/bmqp_requestmanager.cpp
```diff
@@ -69,6 +81,22 @@ RequestManager::sendHelper(bmqio::Channel*                     channel,
...
+    // executed with 'd_mutex' locked
+
+    // Each iteration considers a distinct identifier, and there are more
+    // identifiers than there can be entries in 'd_requests', hence this loop
+    // terminates.
+    do {
+        d_nextRequestId = (d_nextRequestId == k_LAST_REQUEST_ID)
+                              ? k_FIRST_REQUEST_ID
+                              : d_nextRequestId + 1;
+    } while (d_requests.find(d_nextRequestId) != d_requests.end());
```
678098, line 95: This lookup should be performed just once every time we call this function, before we reach 2.147kk requests.
If we reached it, we might theoretically have multiple iterations of this loop.
What do you think? Is it worth attempting to find the next unused `rId`?

src/groups/bmq/bmqp/bmqp_requestmanager.cpp
```diff
@@ -390,7 +418,7 @@ RequestManager::sendRequest(const RequestSp&          request,
     // Insert the request in the map
     BSLA_MAYBE_UNUSED bsl::pair<RequestMapIter, bool> insertRC =
         d_requests.insert(bsl::make_pair(requestId, request));
-    BSLS_ASSERT_SAFE(insertRC.second);
```
678098, line 393: This code might have failed on `int32::max` wrap without an attempt to find next; with this PR, we are guaranteed to find the next unused.
Also, changing this to `ASSERT_OPT`

src/groups/bmq/bmqp/bmqp_requestmanager.cpp
```diff
@@ -69,6 +81,22 @@ RequestManager::sendHelper(bmqio::Channel*                     channel,
...
+    // Each iteration considers a distinct identifier, and there are more
+    // identifiers than there can be entries in 'd_requests', hence this loop
+    // terminates.
+    do {
+        d_nextRequestId = (d_nextRequestId == k_LAST_REQUEST_ID)
+                              ? k_FIRST_REQUEST_ID
+                              : d_nextRequestId + 1;
+    } while (d_requests.find(d_nextRequestId) != d_requests.end());
+
+    return d_nextRequestId;
```
678098, line 97: This code is also difficult to test.
UT that calls `bmqp::RequestManager::sendRequest` 2kkk times might take hours to execute. The only way to write a good test is to expose `bmqp::RequestManager` internal state to fast-forward `d_nextRequestId`

hallfox: Yeah I might try to separate this request id assigner into a separate component-private class and then you can expose constructors and what have you all you like for unit test purposes, e.g.

```c++
class RequestManager_IdGenerator {
private:
  int d_nextRequestId;
public:
  RequestManager_IdGenerator(int startRequestId);
  int nextRequestId(const RequestMap& map);
};
```

Then you can rig up tests to pass in whatever state you want for a test.

src/groups/bmq/bmqp/bmqp_requestmanager.cpp
hallfox: Seems fine? If we're worried about a linear search having potentially bad performance profiles in the case where we've run out of identifiers that would probably depend on actual request identifier density.

PR #1721, opened 2026-08-19
Fix[mqbc::StorageMgr]: Validate E_PUSH sender and partitionId

kaikulimu:
## Summary
  - `StorageManager::processReplicaDataRequestPush` used `BSLS_ASSERT_SAFE` to enforce two conditions that depend on the contents of the received request: that `partitionId` is in range, and that the sender is the partition's current primary. A stale or out-of-date peer can legitimately send a request that fails either check — for example after a primary changeover — so the broker would abort in assert-enabled builds instead of handling it. Both are now runtime checks that return `E_REFUSED` with `e_NO_PARTITION` or `e_SOURCE_NOT_PRIMARY`.
  - The new checks mirror the adjacent `processReplicaDataRequestDrop`, which already validated both conditions this way; the two handlers now read identically. The "partition has no primary" case falls out of the same comparison, since `primaryNodeId()` is `k_INVALID_NODE_ID` while a partition is unassigned.

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_storagemanager.cpp
+++ b/src/groups/mqb/mqbc/mqbc_storagemanager.cpp
@@ -1120,11 +1120,51 @@ void StorageManager::processReplicaDataRequestPush(
                      bmqp_ctrlmsg::ReplicaDataType::E_PUSH);
 
     const int partitionId = replicaDataRequest.partitionId();
-    BSLS_ASSERT_SAFE(0 <= partitionId &&
-                     partitionId < static_cast<int>(d_fileStores.size()));
-    BSLS_ASSERT_SAFE(
-        source->nodeId() ==
-        d_clusterState_p->partitionsInfo().at(partitionId).primaryNodeId());
+    if (partitionId < 0 ||
+        partitionId >= static_cast<int>(d_fileStores.size())) {
+        BALL_LOG_ERROR
+            << d_clusterData_p->identity().description()
+            << " Received ReplicaDataRequestPush: " << message << " from "
+            << source->nodeDescription()
+            << " with invalid partitionId.  Sending failure response.";
+
+        bmqp_ctrlmsg::ControlMessage controlMsg;
+        controlMsg.rId() = message.rId();
+
+        bmqp_ctrlmsg::Status& status = controlMsg.choice().makeStatus();
+        status.category()            = bmqp_ctrlmsg::StatusCategory::E_REFUSED;
+        status.code()                = mqbi::ClusterErrorCode::e_NO_PARTITION;
+        status.message()             = "Invalid partitionId";
+
+        d_clusterData_p->messageTransmitter().sendMessageSafe(controlMsg,
+                                                              source);
+        return;  // RETURN
+    }
+    if (source->nodeId() !=
+        d_clusterState_p->partitionsInfo().at(partitionId).primaryNodeId()) {
+        const mqbnet::ClusterNode* primaryNode =
+            d_clusterState_p->partitionsInfo().at(partitionId).primaryNode();
+        BALL_LOG_ERROR << d_clusterData_p->identity().description()
+                       << " Partition [" << partitionId << "]: "
+                       << " Received ReplicaDataRequestPush: " << message
+                       << " from " << source->nodeDescription()
+                       << " but self's perceived primary is "
+                       << (primaryNode ? primaryNode->nodeDescription()
+                                       : " ** NULL **")
+                       << ".  Sending failure response.";
+
+        bmqp_ctrlmsg::ControlMessage controlMsg;
+        controlMsg.rId() = message.rId();
+
+        bmqp_ctrlmsg::Status& status = controlMsg.choice().makeStatus();
+        status.category()            = bmqp_ctrlmsg::StatusCategory::E_REFUSED;
+        status.code()    = mqbi::ClusterErrorCode::e_SOURCE_NOT_PRIMARY;
+        status.message() = "Source node is not recognized as the primary";
+
+        d_clusterData_p->messageTransmitter().sendMessageSafe(controlMsg,
+                                                              source);
+        return;  // RETURN
+    }
 
     BALL_LOG_INFO << d_clusterData_p->identity().description()
                   << " Partition [" << partitionId << "]: "
@@ -1175,15 +1215,6 @@ void StorageManager::processReplicaDataRequestDrop(
 
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(d_cluster_p->inDispatcherThread());
-    BSLS_ASSERT_SAFE(message.choice().isClusterMessageValue());
-    BSLS_ASSERT_SAFE(
-        message.choice().clusterMessage().choice().isPartitionMessageValue());
-    BSLS_ASSERT_SAFE(message.choice()
-                         .clusterMessage()
-                         .choice()
-                         .partitionMessage()
-                         .choice()
-                         .isReplicaDataRequestValue());
 
     const bmqp_ctrlmsg::ReplicaDataRequest& replicaDataRequest =
         message.choice()
--- a/src/groups/mqb/mqbc/mqbc_storagemanager.t.cpp
+++ b/src/groups/mqb/mqbc/mqbc_storagemanager.t.cpp
@@ -1105,13 +1105,13 @@ struct TestHelper {
                 .choice()
                 .makePrimaryStateResponse();
 
-        bmqp_ctrlmsg::PartitionSequenceNumber PSN;
-        PSN.sequenceNumber() = 1U;
-        PSN.primaryLeaseId() = 1U;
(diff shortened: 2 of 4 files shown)


src/groups/mqb/mqbc/mqbc_storagemanager.cpp
hallfox, line 1228: Is this another protocol controllable assert?

src/groups/mqb/mqbc/mqbc_storagemanager.t.cpp
```diff
@@ -727,6 +727,93 @@ struct TestHelper {
...
+                                      bmqtst::TestHelperUtil::allocator());
+
+        char journalBuf[mqbs::FileStoreProtocol::k_JOURNAL_RECORD_SIZE];
+        bsl::memset(journalBuf, 0, sizeof(journalBuf));
+        bsl::shared_ptr<char> journalSp(journalBuf,
+                                        bslstl::SharedPtrNilDeleter(),
+                                        bmqtst::TestHelperUtil::allocator());
+        bdlbb::BlobBuffer     journalBlobBuf(journalSp, sizeof(journalBuf));
+
+        char dataBuf[bmqp::Protocol::k_WORD_SIZE];
```
hallfox, line 788: How do we know this buffer is big enough for the packed message?

src/groups/mqb/mqbc/mqbc_storagemanager.t.cpp
```diff
@@ -989,6 +1076,51 @@ struct TestHelper {
...
+        pstMessage.rId() = k_PRIMARY_REQUEST_ID;
+        bmqp_ctrlmsg::PrimaryStateResponse& primaryStateResponse =
+            pstMessage.choice()
+                .makeClusterMessage()
+                .choice()
+                .makePartitionMessage()
+                .choice()
+                .makePrimaryStateResponse();
+
+        bmqp_ctrlmsg::PartitionSequenceNumber PSN;
```
hallfox:
```suggestion
        bmqp_ctrlmsg::PartitionSequenceNumber psn;
```

src/groups/mqb/mqbc/mqbc_storagemanager.t.cpp
```diff
@@ -3402,208 +3535,228 @@ static void test22_rstUnknownCancelsInFlightRequests()
...
 
-    helper.clearChannels();
+        replicaDataRequest.replicaDataType() = test.d_dataType;
+        replicaDataRequest.partitionId()     = partitionId;
+        replicaDataRequest.primaryLeaseId()  = 1U;
 
-    // 5. Send a storage event (PUT) and verify it is buffered, not processed
-    const bsls::Types::Uint64 seqNumBefore =
-        storageManager.fileStore(k_PARTITION_ID).writeHeadSeqNum();
+        mqbnet::ClusterNode* const source = test.d_fromPrimary ? primaryNode
```
hallfox, line 3728:
```suggestion
        const mqbnet::ClusterNode* source = test.d_fromPrimary ? primaryNode
```

src/integration-tests/test_replica_data_request.py
```diff
@@ -0,0 +1,268 @@
...
+
+from blazingmq.dev.it.fixtures import (
+    Cluster,
+    order,
+    start_cluster,
+)
+from blazingmq.dev.it.process.admin import AdminClient
+from blazingmq.dev.it.process.rawclient import RawClient
+
+pytestmark = order(99)
```
hallfox: Why?

src/integration-tests/test_replica_data_request.py
```diff
@@ -0,0 +1,268 @@
...
+
+def _wait_for_primary(
+    admin: AdminClient, cluster_name: str, partition_id: int, exclude_node_id=None
+):
+    """
+    Wait for 'partition_id' to have a primary other than the optionally
+    specified 'exclude_node_id', and return it as a (name, id) tuple.
+    """
+
+    for _ in range(30):
```
hallfox: Why 30?

src/integration-tests/test_replica_data_request.py
```diff
@@ -0,0 +1,268 @@
...
+pytestmark = order(99)
+
+PARTITION_ID = 0
+NEGOTIATION_ATTEMPTS = 30
+
+LOG_NAME = {"E_PUSH": "Push", "E_DROP": "Drop"}
+DATA_TYPES = ["E_PUSH", "E_DROP"]
+
+
+def _partition_primary(admin: AdminClient, cluster_name: str, partition_id: int):
```
hallfox:
```suggestion
def _partition_primary(admin: AdminClient, cluster_name: str, partition_id: int) -> Optional[(str, int)]:
```

src/integration-tests/test_replica_data_request.py
```diff
@@ -0,0 +1,268 @@
...
+    specified 'exclude_node_id', and return it as a (name, id) tuple.
+    """
+
+    for _ in range(30):
+        primary = _partition_primary(admin, cluster_name, partition_id)
+        if primary is not None and primary[1] != exclude_node_id:
+            return primary
+        time.sleep(1)
+
+    raise AssertionError(f"partition {partition_id} never got a usable primary")
```
hallfox: Does `AssertionError` make the most sense here? If so maybe this should just be
```python
assert False, f"partition {partition_id} never got a usable primary"
```

src/integration-tests/test_replica_data_request.py
```diff
@@ -0,0 +1,268 @@
...
+    order,
+    start_cluster,
+)
+from blazingmq.dev.it.process.admin import AdminClient
+from blazingmq.dev.it.process.rawclient import RawClient
+
+pytestmark = order(99)
+
+PARTITION_ID = 0
+NEGOTIATION_ATTEMPTS = 30
```
hallfox: Why pick partition 0 and negotiation attempts this way?

src/groups/mqb/mqbc/mqbc_storagemanager.t.cpp
kaikulimu: The signature of `processReplicaDataRequest` a few lines below requires a non-const `mqbnet::ClusterNode*`, so your suggestion won't compile. Changing the signature of `processReplicaDataRequest` would require changing multiple places. Do you still want me to change it?

src/integration-tests/test_replica_data_request.py
kaikulimu: Changed to `order(6)`, matching the other tests that kill a node and wait for failover `(test_cluster_node_shutdown, test_heartbeat, test_leader_node_delay)`.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: Not protocol controllable. `bmqp_ctrlmsg::ReplicaDataRequest` itself has no way to enforce whether a request is of type `PUSH`, `PULL`, or `DROP`.

src/integration-tests/test_replica_data_request.py
kaikulimu: I renamed `30` as `PRIMARY_FAILOVER_SECONDS`, reduced from `30` to `20`, and refactored the `for` loop into a `wait_until` call. Please review if this flows better.

src/integration-tests/test_replica_data_request.py
kaikulimu: The partition id does not matter. `0` is as good as any.

kaikulimu: I updated the code such that we will only re-attempt negotiation for 10 seconds. 30 seconds was overkill.

src/groups/mqb/mqbc/mqbc_storagemanager.t.cpp
kaikulimu: The buffer is an input, not an output — `packMessage` takes it as `const bdlbb::BlobBuffer&` and appends it to the builder's own blob, which the builder grows itself, so nothing is ever written back into `dataBuf`. The only precondition documented is that both buffer lengths be a multiple of `bmqp::Protocol::k_WORD_SIZE`.

src/groups/mqb/mqbc/mqbc_storagemanager.t.cpp
hallfox: Sorry, I misread the ordering of that `const`. All good here 👍🏻.

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
hallfox: Sorry, I think what I was asking was is it possible for someone to craft a message that can make the broker execute this code and set something other than `E_DROP` as the replica data type?

hallfox (approved):

src/groups/mqb/mqbc/mqbc_storagemanager.cpp
kaikulimu: No, should not be possible.

PR #1723, opened 2026-08-20
Fix[mqbs::FileStore]: terminate on replica write failure

### LINKED ISSUE #1702: Discussion: processStorageEvent error handling after replica write failure
processStorageEvent continues iterating after a replica write failure, logging "Ignoring this message." Because the PSN is not bumped for the failed record, the next record fails the sequence number continuity check and calls terminate(e_STORAGE_OUT_OF_SYNC).

The code:

https://github.com/bloomberg/blazingmq/blob/ec10f62b5/src/groups/mqb/mqbs/mqbs_filestore.cpp#L6484-L6518

The next iteration hits:

https://github.com/bloomberg/blazingmq/blob/ec10f62b5/src/groups/mqb/mqbs/mqbs_filestore.cpp#L6438-L6454

We terminate either way, but the log blames a PSN gap rather than the write failure that caused it. If the failed record is the last in the event, there is also a window where we run with an un-bumped PSN before the next event triggers the terminate. The write failure paths have had logging improvements and one trigger was worked around at the recovery level, but the error handling in the loop itself wasn't changed.

SrinathhSatuluri:
Closes #1702

`processStorageEvent` continued iterating after a replica write failure, logging "Ignoring this message." Because the PSN was not bumped for the failed record, the next record failed the sequence number continuity check and terminated with `e_STORAGE_OUT_OF_SYNC`, attributing the failure to a PSN gap rather than the write error that caused it. If the failed record was the last in the event, the replica ran with a stale PSN until the next event arrived. We terminate directly on write failure, matching the severity of the existing PSN gap handler.

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_filestore.cpp
+++ b/src/groups/mqb/mqbs/mqbs_filestore.cpp
@@ -6505,8 +6505,10 @@ void FileStore::processStorageEvent(const bsl::shared_ptr<bdlbb::Blob>& blob,
                 << printPSN(recHeader->primaryLeaseId(),
                             recHeader->sequenceNumber())
                 << "), journal offset words: " << header.journalOffsetWords()
-                << ", rc: " << rc << ". Ignoring this message."
+                << ", rc: " << rc << ". Aborting broker."
                 << BMQTSK_ALARMLOG_END;
+            mqbu::ExitUtil::terminate(mqbu::ExitCode::e_STORAGE_OUT_OF_SYNC);
+            // EXIT
         }
     } while (1 == iter.next());
 


kaikulimu:
Are the changes to `ClientIdentity` related?

SrinathhSatuluri:
> Are the changes to `ClientIdentity` related?

hi @kaikulimu No, that's unrelated. I branched off the wrong base and it carried over. Fixing now.

SrinathhSatuluri:
@kaikulimu Fixed the branch issue. this PR now has the related change only.

kaikulimu (approved):
LGTM

PR #1724, opened 2026-08-20
Feat: Add simple policy description VSTs

hallfox:
This PR adds VSTs for describing policies that an authorizer plugin may choose to use to describe permissioning of roles in BlazingMQ. The intent for this is to use these pieces as a building block to describe policies for the `DefaultAuthorizer`. We could describe the intended behavior of an anonymous user in a JSON config like so:

```json
{
	"roles": [
		{
			"id": "anonymous",
			"permissions": [
				{
					"action": "CONNECT_CLIENT",
				},
				{
					"action": "QUEUE_READ",
					"resources": [
						{
							"id": "*", // allow read on all queues
						}
					]
				},
				{
					"action": "QUEUE_WRITE",
					"resources": [
						{
							"id": "*", // allow write on all queues
						}
					]
				}
			]
		}
	]
}
```

### DIFF
--- a/src/groups/mqb/mqbpoly/mqbpoly.xsd
+++ b/src/groups/mqb/mqbpoly/mqbpoly.xsd
@@ -13,7 +13,7 @@
       </documentation>
     </annotation>
     <sequence>
-      <element name='roles'   type='tns:Role' maxOccurrs='unbounded' />
+      <element name='roles'   type='tns:Role' maxOccurs='unbounded' />
     </sequence>
   </complexType>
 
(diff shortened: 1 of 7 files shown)


src/groups/mqb/mqbpoly/mqbpoly.xsd
```diff
@@ -0,0 +1,81 @@
...
+  <complexType name='Permission'>
+    <annotation>
+      <documentation>
+      This type represents a permission to access a resource. A permission is a simple
+      association of an action identifier with a set of resources related to that
+      action.
+
+      The list of supported actions are:
+      - connectClient
+      - connectProxy
```
chrisbeard, line 55: We should remove this, forgot to do it in the action xsd, otherwise lgtm

chrisbeard (approved):

PR #1725, opened 2026-08-20
Fix[mqbblp]: remove ConfigureContext, report configure result from queue thread

678098:
- Remove ref-counted shared `ConfigureContext` (no need to share it)
- Introduce stack-scoped `ConfigureNotifier`. It allows to execute the stored callback at most once, and ensures that the callback is either called or passed next in the end.
- Build routing close to the place it is used first.
- Add an IT that fails in main with configure context destructed from a wrong thread.

### DIFF
--- a/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
+++ b/src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
@@ -55,6 +55,7 @@
 #include <bsl_limits.h>
 #include <bsl_sstream.h>
 #include <bsl_string.h>
+#include <bsl_string_view.h>
 #include <bsl_utility.h>
 #include <bsl_vector.h>
 #include <bsla_annotations.h>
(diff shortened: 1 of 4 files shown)


src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -371,15 +369,22 @@ void RelayQueueEngine::onHandleConfiguredDispatched(
...
     // executed by the *QUEUE DISPATCHER* thread
 
+    // PRECONDITIONS
+    BSLS_ASSERT_SAFE(context);
+
+    // Notify the requester from this thread, on every exit path.
+    // Make sure the stored callback is called from this thread and cleared.
+    bdlb::ScopeExitAny invokeCallbackGuard(
+        bdlf::BindUtil::bind(&ConfigureContext::invokeCallback,
+                             context.get()));
```
678098: There are several return paths here that set the status in this context but do not the stored callback right away. Due to this, the callback might be called on the context destruction from a cluster thread, leading to `inDispatcherThread` check failure.

dorjesinpo (changes requested):
I think the original reason for `RelayQueueEngine::ConfigureContext::~ConfigureContext` was support for wildcards which got removed.  Over there, the same context could be shared across multiple `configureApp`.
Now this seems to be always one `ConfigureContext` per configure request and it makes sense to call the callback explicitly, as in this PR.
Then, let's remove the `~ConfigureContext` so that there is no chance the callback is called from a wrong thread?

678098:
> Now this seems to be always one `ConfigureContext` per configure request and it makes sense to call the callback

Thanks for the details, removing it completely simplifies it greatly. Will look into it

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -848,7 +967,14 @@ void RelayQueueEngine::configureApp(
         << "', about to rebuild upstream state [current stream parameters: "
         << previousParameters << "]";
 
-    rebuildUpstreamState(context->d_routing_sp.get(),
+    // Routing data to advertise upstream (including Subscription Ids).  It
+    // becomes effective either right below, or once upstream has responded.
+    bsl::shared_ptr<Routers::AppContext> routing_sp =
+        bsl::allocate_shared<Routers::AppContext>(
+            d_allocator_p,
+            &d_queueState_p->routingContext());
```
678098: This is the closest place where we can allocate the routing.
Don't need a context to keep it.

src/groups/mqb/mqbblp/mqbblp_relayqueueengine.cpp
```diff
@@ -952,17 +1083,18 @@ void RelayQueueEngine::rebuildUpstreamState(Routers::AppContext* context,
...
     // executed by the *QUEUE DISPATCHER* thread
 
     // PRECONDITIONS
     BSLS_ASSERT_SAFE(d_queueState_p->queue()->inDispatcherThread());
+    BSLS_ASSERT_SAFE(routing_sp);
 
     app.undoRouting();
-
-    app.routing() = context.d_routing_sp;
+    app.routing() = routing_sp;
```
678098: `RelayQueueEngine::applyConfiguration` never used the whole `context`, it only needs `routing_sp`

dorjesinpo (approved):
one minor note - use the allocator for `ConfigureNotifier`

PR #1728, opened 2026-08-24
Refactor[mqba]: simplify AdminSession, SessionNegotiator

678098:
- Remove `mqba::AdminSession::d_scheduler_p` field
- `mqba::SessionNegotiator::createSession`: return `bsl::shared_ptr` directly, not as an output arg


chrisbeard (approved):
nice 👍

PR #1731, opened 2026-08-25
Fix[mqbc::StorageMgr]: validate E_PULL sender and partitionId

kaikulimu:
Follow-up to #1721 , this time validation for ReplicaDataRequestPull. Also did some refactoring to reduce code duplication.


src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -1004,6 +1021,86 @@ void StorageManager::startSendDataChunksAsPrimary(
...
+    BSLS_ASSERT_SAFE(d_cluster_p->inDispatcherThread());
+    BSLS_ASSERT_SAFE(source);
+
+    const bmqp_ctrlmsg::ReplicaDataRequest& replicaDataRequest =
+        message.choice()
+            .clusterMessage()
+            .choice()
+            .partitionMessage()
+            .choice()
+            .replicaDataRequest();
```
hallfox: Is this chain of choice selection operators guaranteed to be safe?

kaikulimu: I changed `validateReplicaDataRequest` to take a `bmqp_ctrlmsg::ReplicaDataRequest` as param instead of a `bmqp_ctrlmsg::ControlMessage`. So the chain of choice selection operators is no longer needed.

To make this happen, I also simplified `StorageManager::sendFailureResponse` to only take request Id instead of the entire request as input param.

hallfox (dismissed):

hallfox (approved):

PR #1738, opened 2026-08-26
Fix[mqbc::StorageMgr]: validate PSN range and report end PSN mismatch

kaikulimu:
## Summary

  - `processReplicaDataRequestPull` compared the requested end PSN against self's by reading `d_nodeToPSNCtxMapVec` from the cluster dispatcher thread, but `mqbc_storagemanager.h` documents that member as readable only from the queue dispatcher thread owning the partition, and the `.at()` lookup throws when self has no entry. The comparison moves to `do_sendDataToPrimary`, which already runs on the owning thread.

  - `do_sendDataToPrimary` asserted that self has a stored PSN. It now raises the Partition FSM defect alarm and re-heals, which restores the entry.

  - `beginSequenceNumber`/`endSequenceNumber` were left to `BSLS_ASSERT_SAFE`, so an unusable range aborted assert-enabled builds. PULL and PUSH now validate the range on arrival and refuse with `E_INVALID_ARGUMENT`.

  - A mismatch reached the primary as `e_STORAGE_FAILURE` / "Failed to send data chunks", which is inaccurate — no send is attempted — and indistinguishable from a real transfer failure. Adds `e_END_PSN_MISMATCH` and a Partition FSM event that keeps the replica in `e_REPLICA_HEALING`.

- Follows #1721 and #1731

### DIFF
--- a/src/groups/mqb/mqbc/mqbc_storagemanager.cpp
+++ b/src/groups/mqb/mqbc/mqbc_storagemanager.cpp
@@ -1129,7 +1129,7 @@ void StorageManager::processReplicaDataRequestPull(
     BSLS_ASSERT_SAFE(replicaDataRequest.replicaDataType() ==
                      bmqp_ctrlmsg::ReplicaDataType::E_PULL);
 
-    int requestId;
+    int requestId = 0;
     if (!validateReplicaDataRequest(&requestId,
                                     replicaDataRequest,
                                     message.rId(),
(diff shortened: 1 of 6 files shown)


src/groups/mqb/mqbc/mqbc_storagemanager.cpp
```diff
@@ -2814,6 +2827,37 @@ void StorageManager::do_sendDataToPrimary(
         1,
         eventData.partitionSeqNumDataRange());
 
+    const NodeToPSNCtxMap& nodeToPSNCtxMap = d_nodeToPSNCtxMapVec.at(
+        partitionId);
```
hallfox: We should not be open to a potentially throwing operation here. Please check that this `partitionId` is in range and if not, return with an error.

kaikulimu: We already asserted that `partitionId` will be in range at [the beginning of `do_sendDataToPrimary`](https://github.com/kaikulimu/blazingmq/blob/0158509572c57538786cc111a05672e82c223a5e/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L2800-L2801). `do_sendDataToPrimary` is internally [called by `processReplicaDataRequestPull`](https://github.com/kaikulimu/blazingmq/blob/0158509572c57538786cc111a05672e82c223a5e/src/groups/mqb/mqbc/mqbc_storagemanager.cpp#L1194-L1195) which has already validated the `partitionId`. Therefore, I don't feel the need to check range again.

hallfox: Then please change it to use `operator[]` instead of `at()`

kaikulimu: Done

hallfox (approved):

PR #1747, opened 2026-08-27
Feat policy evaluator

hallfox:
This is the implementation of a parser and evaluator for the policy rules described by the types in #1724


src/groups/mqb/mqbauthz/mqbauthz_policy.h
```diff
@@ -0,0 +1,335 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef INCLUDED_MQBAUTHZ_POLICY
+#define INCLUDED_MQBAUTHZ_POLICY
+
+/// @file mqbauthz_policy.h
+///
+/// @brief Provide the built-in default authorizer plugin.
```
chrisbeard, line 21: nit: update the brief for Policy here

src/groups/mqb/mqbauthz/mqbauthz_policy.cpp
```diff
@@ -0,0 +1,571 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <mqbauthz_policy.h>
+
+// MQB
+#include <mqbcmd_messages.h>
+#include <mqbcmd_parseutil.h>
```
chrisbeard: Doesn't look like these includes are actually used? If not needed, we can remove mqbcmd from the dep fo;e.

src/groups/mqb/mqbauthz/mqbauthz_policy.h
```diff
@@ -0,0 +1,335 @@
...
+/// | executeAdminCommand | This role can execute admin commands.    | None                      |
+// clang-format on
+///
+/// ### Queue Resource Syntax
+/// The syntax for specifing queue resources is:
+/// - "*" => Allow all
+/// - "valid-domain-name" => Allow opening any queue in the domain
+/// "valid-domain-name"
+/// - "valid-domain-name/valid-queue-name" => Allow opening the queue
+/// "valid-queue-name" in the domain "valid-domain-name"
```
chrisbeard, line 142: I think my only question is should this be a full URI syntax? E.g. `bmq://valid-domain-name` for a domain.

I don't have a strong preference, we can always evolve this later if need be.

hallfox: I tried to think of a reason that we'd need the protocol part of the URI for pattern matching. Since permissions are granted on a per-domain or per-queue basis, that was the only important part of the syntax I wanted to focus on, and the protocol didn't seem that relevant.

chrisbeard (approved):

PR #1749, opened 2026-08-28
CI[Sanitizers]: fix build by installing existing python

678098:
Ubuntu 26.04 provides `python3.14` and our attempt to install `python3.12` fails.
Make the sanitizers build script more robust by installing `python3` that auto-resolves to the existing python version.

### DIFF
--- a/docker/sanitizers/build_sanitizer.sh
+++ b/docker/sanitizers/build_sanitizer.sh
@@ -59,8 +59,8 @@ FUZZER="${2}"
 
 # Install prerequisites
 # Set up CA certificates first before installing other dependencies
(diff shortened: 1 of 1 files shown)


docker/sanitizers/build_sanitizer.sh
```diff
@@ -59,8 +59,8 @@ FUZZER="${2}"
 
 # Install prerequisites
 # Set up CA certificates first before installing other dependencies
-apt-get update && \
```
678098, line 62: No need to concat these commands because it is just a script, not a docker layer.

hallfox (approved):

PR #1751, opened 2026-08-28
Fix[mqbs::FileStore]: validate DataHeader in writeMessageRecord

### LINKED ISSUE #1719: DataHeader validation in replica writeMessageRecord
The recovery path in `recoverMessages` validates the DataHeader fields before computing `d_appDataUnpaddedLen`. However, the replica replication in `writeMessageRecord` performs the same arithmetic with no validation, as the only guard is a `BSLS_ASSERT_SAFE` stripped in release builds.

Since `d_appDataUnpaddedLen` is `unsigned int`, I suppose a corrupt DataHeader where `headerWords` exceeds `messageWords` causes the subtraction to underflow, wrapping the value and propagating it into `processMessageRecord` and `aliasMessage`.

SrinathhSatuluri:
Closes #1719 

`writeMessageRecord` computes `d_appDataUnpaddedLen` from DataHeader fields with no validation, while `recoverMessages` checks all five fields before the same arithmetic. These checks are ported into the replication path, returning `rc_INVALID_DATA_HEADER` on failure.

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_filestore.cpp
+++ b/src/groups/mqb/mqbs/mqbs_filestore.cpp
@@ -4454,6 +4454,7 @@ int FileStore::writeMessageRecord(const bmqp::StorageHeader& header,
         rc_WRITE_MESSAGE_RECORD_ERROR = -3,
         rc_UNKNOWN_QUEUE_KEY          = -4,
         rc_INCOMPATIBLE_STORAGE       = -5,
+        rc_INVALID_DATA_HEADER        = -6,
     };
 
     FileSet* activeFileSet = d_fileSets[0].get();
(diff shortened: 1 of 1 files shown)


src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -4371,6 +4399,29 @@ int FileStore::writeMessageRecord(const bmqp::StorageHeader& header,
     const char lastByte =
```
SrinathhSatuluri, line 4399: Kept `lastByte` as `const char` to match the existing declaration here. But `recoverMessages` uses `const int` for the same field. We can change it to int if that is the preferred convention.

SrinathhSatuluri:
Hi @678098, just rebased this PR to rerun CI after the earlier sanitizer failures.

678098:
> Hi @678098, just rebased this PR to rerun CI after the earlier sanitizer failures.

@SrinathhSatuluri, cool! I will review the PR soon after I fix a few more issues in our repos.

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -4550,6 +4578,29 @@ int FileStore::writeMessageRecord(const bmqp::StorageHeader& header,
...
+        BALL_LOG_ERROR << partitionDesc()
+                       << "Received DATA record with invalid padding: "
+                       << static_cast<int>(lastByte)
+                       << " [queueKey: " << queueKey
+                       << ", GUID: " << messageGuid << "].";
+        return rc_INVALID_DATA_HEADER;  // RETURN
+    }
+
+    if (messageSize <
+        (headerSize + optionsSize + static_cast<unsigned char>(lastByte))) {
```
678098: A few lines below:
```
    record.d_appDataUnpaddedLen = messageSize - headerSize - optionsSize -
                                  static_cast<unsigned char>(lastByte);
```

The condition on these lines doesn't return on equality, this means that the current condition allows `d_appDataUnpaddedLen == 0`.
Zero length appData is not supported, so we should do:

```suggestion
    if (messageSize <=
        (headerSize + optionsSize + static_cast<unsigned char>(lastByte))) {
```

SrinathhSatuluri: @678098 should we apply same change to recoverMessages as well?https://github.com/bloomberg/blazingmq/blob/fcb86968314628f4855c03d8cf24d7edb9916432/src/groups/mqb/mqbs/mqbs_filestore.cpp#L2669

678098: It looks correct, you can either update this PR or open another one with recoverMessages fix

678098 (approved):
Looks good

PR #1756, opened 2026-09-03
Feat[mqb]: outbound broker authentication

678098:
# Support outbound authentication

A broker now authenticates itself when it dials another broker, provided a
credential provider is configured.  Without one, outbound connections negotiate
directly as before.

## Changes

Outbound authentication

* `mqbnet::Authenticator::createAuthenticationClient()` replaces the
  unimplemented `authenticationOutbound()`; an empty pointer means "no
  credential provider".
* `InitialConnectionContext` creates its `AuthenticationClient` in the
  constructor, so the context is complete before the channel can report a
  close.
* New state `e_AUTHENTICATING_OUTBOUND`, new events
  `e_OUTBOUND_AUTHENTICATION` and `e_AUTHN_RESPONSE`.
* Authentication messages are classified at decode time by type, not by state.
* `ChannelInfo` carries the client so reauthentication works after the
  handshake.
* `mqbnet::AuthenticationClient::onClose()` added to the protocol; called from
  `InitialConnectionContext::onClose()` and `TCPSessionFactory::onClose()`.
  The destructor delegates to it.

Fixes

* `tryStartReauthentication()` now takes the message and encoding and adopts
  them only if the transition happens.  Previously they were stored first, so a
  request arriving mid-reauthentication overwrote the one in flight before
  being dropped.  `setAuthenticationMessage()` /
  `setAuthenticationEncodingType()` are gone.
* An authentication event on a channel with no matching context or client is
  dropped with a log instead of hitting an assert.
* A failed reauthentication on an outbound channel closes the channel, matching
  the inbound path.
* `AuthenticationContext` guards its reauthentication-timeout callback with a
  `bmqu::GateKeeper` instead of a state check, so the callback cannot run past
  `onClose()`.
* `mqbnet_tcpsessionfactory.cpp`: reauthentication error stream used the
  default allocator.

Cleanup

* `reauthnOnAuthenticationEvent()` -> `onAuthenticationEvent()`, now routing
  both directions.
* `startOutboundNegotiation()` extracted; the negotiate + schedule-read block
  is no longer duplicated.
* Seven copies of the "Unexpected event received" block collapsed into
  `MQBNET_LOG_UNEXPECTED`.
* Dead `onAuthenticationResponse()` handler and its
  `SELECTION_ID_AUTHENTICATION_RESPONSE` case removed.
* Both `InitialConnectionContext` enums renumbered into logical order, every
  enumerator documented, all four handshake paths drawn in the component
  header.
* Malformed `/// ACCESSORS` section comment in `mqba_authenticator.h`.

Tests

* `mqbnet_initialconnectioncontext.t.cpp`: outbound success, rejection, send
  failure, no credential provider, unexpected response.
* `mqba_authenticationclient.t.cpp`: `onClose()` cancels a pending
  reauthentication timer.
* `test_authn.py`: broker-to-broker success (leader elected, post/consume), no
  credential provider (no request sent), rejected credentials (no leader
  elected).

## Known gap

`TODO(678098)`: a short `lifetimeMs` schedules reauthentication before the
session is established; the response then arrives while the context is
negotiating and fails the connection.


src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -249,5 +248,20 @@ int AuthenticationClient::handleResponse(
     return rc_SUCCESS;
 }
 
+void AuthenticationClient::onClose()
+{
+    // executed by *ANY* thread
```
678098, line 253: Both `AuthenticationClient::onClose` and `AuthenticationContext::onClose` can be executed from any thread, however, these `onClose` cannot be called concurrently from different threads at the same time.

This is guaranteed by `TCPSessionFactory` mutex.

src/groups/mqb/mqbnet/mqbnet_authenticationcontext.cpp
```diff
@@ -121,6 +122,8 @@ AuthenticationContext::AuthenticationContext(
 {
     // PRECONDITION
     BSLS_ASSERT_SAFE(d_scheduler_p);
+
+    d_gateKeeper.open();
```
678098, line 126: We had `d_gateKeeper` to ensure shutdown safety in `AuthenticationClient`.
I added it to this component to prevent possible deadlock.

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -81,6 +81,7 @@ const char* InitialConnectionState::toAscii(InitialConnectionState::Enum value)
         CASE(AUTHENTICATING)
```
678098, line 81: Reordered enums to make it intuitive

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -316,8 +328,13 @@ int InitialConnectionContext::processBlob(bsl::ostream&      errorDescription,
...
                  message)) {
+        const bool isResponse = bsl::get<bmqp_ctrlmsg::AuthenticationMessage>(
+                                    message)
+                                    .isAuthenticationResponseValue();
+
         handleEvent(bsl::string(),
-                    InitialConnectionEvent::e_AUTHN_REQUEST,
+                    isResponse ? InitialConnectionEvent::e_AUTHN_RESPONSE
+                               : InitialConnectionEvent::e_AUTHN_REQUEST,
                     message);
```
678098, line 338: @chrisbeard to your earlier question, if we have an unexpected event, we will go to FAILED state in this `handleEvent`

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -548,39 +581,49 @@ void InitialConnectionContext::handleEvent(
 
     InitialConnectionState::Enum oldState = d_state;
 
+#define MQBNET_LOG_UNEXPECTED(DETAILS)                                        \
+    do {                                                                      \
+        errStream << "Unexpected event received: " << oldState << " -> "      \
+                  << event << DETAILS;                                        \
```
678098, line 587: no macro brackets `<< (DETAILS)` because we want to inject multiple items to stream

src/groups/mqb/mqbnet/mqbnet_initialconnectioncontext.cpp
```diff
@@ -548,39 +581,49 @@ void InitialConnectionContext::handleEvent(
...
         }
         else {
-            errStream << "Unexpected event received: " << oldState << " -> "
-                      << event;
-            BALL_LOG_ERROR << "#UNEXPECTED_STATE " << errStream.str()
-                           << " [peer: " << channel().get() << "]";
+            MQBNET_LOG_UNEXPECTED("");
         }
-        break;
-    }
```
678098, line 583:
```
case N: {
    ...
} break;
```
Should be like this (BDE code style)

src/groups/mqb/mqba/mqba_authenticationclient.cpp
```diff
@@ -222,6 +217,10 @@ int AuthenticationClient::handleResponse(
 
     // If a lifetime is provided, schedule a reauthentication request to be
     // sent at 80% of the lifetime before the connection expires.
+    //
+    // TODO(678098): handle the case when lifetimeMs is short enough that we
+    // trigger a reauthentication event before we finished establishing a
+    // session.
```
678098, line 223: This can be fixed later, I don't want to increase the scope of this PR.
Basically we can safely rely on `lifetimeMs` not being too low to cause problems

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -1100,25 +1102,61 @@ void TCPSessionFactory::reauthnOnAuthenticationEvent(
...
+        rc = client->handleResponse(errorStream, authenticationMessage);
+        if (rc != 0) {
+            BALL_LOG_ERROR << "#AUTHENTICATION_FAILED " << description
+                           << ": Reauthentication failed [reason: '"
+                           << errorStream.str() << "', rc: " << rc << "]";
+
+            bmqio::Status status(bmqio::StatusCategory::e_GENERIC_ERROR,
+                                 "reauthenticationError",
+                                 rc,
+                                 d_allocator_p);
```
chrisbeard: This error should match the values that the authenticator returns to clients, check the error enum and message returned there.

678098: https://github.com/bloomberg/blazingmq/blob/35e1dd483a3d15f083110d9b340ea6405c289a1a/src/groups/mqb/mqbnet/mqbnet_authenticationcontext.cpp#L250-L253

Updated to `e_CANCELED`

chrisbeard (approved):
lgtm, looks clean

Issue #1727, opened 2026-08-21
Refactor excessively long functions


kaikulimu:
Top 10 longest functions in BlazingMQ in number of lines (C++ only):

```
  ┌─────┬───────┬──────────────────────────────────────────────────┬────────────────────────────────────────────┐
  │  #  │ Lines │                     Function                     │                  Location                  │
  ├─────┼───────┼──────────────────────────────────────────────────┼────────────────────────────────────────────┤
  │ 1   │ 1670  │ FileStore::recoverMessages                       │ mqbs/mqbs_filestore.cpp:1119               │
  ├─────┼───────┼──────────────────────────────────────────────────┼────────────────────────────────────────────┤
  │ 2   │ 773   │ FileStore::openInRecoveryMode                    │ mqbs/mqbs_filestore.cpp:345                │
  ├─────┼───────┼──────────────────────────────────────────────────┼────────────────────────────────────────────┤
  │ 3   │ 654   │ StorageUtil::recoveredQueuesCb                   │ mqbc/mqbc_storageutil.cpp:1555             │
  ├─────┼───────┼──────────────────────────────────────────────────┼────────────────────────────────────────────┤
  │ 4   │ 612   │ RecoveryManager::processStorageSyncRequest       │ mqbblp/mqbblp_recoverymanager.cpp:3718     │
  ├─────┼───────┼──────────────────────────────────────────────────┼────────────────────────────────────────────┤
  │ 5   │ 474   │ RecoveryManager::onStorageSyncResponseDispatched │ mqbblp/mqbblp_recoverymanager.cpp:892      │
  ├─────┼───────┼──────────────────────────────────────────────────┼────────────────────────────────────────────┤
  │ 6   │ 440   │ FileStore::writeJournalRecord                    │ mqbs/mqbs_filestore.cpp:4740               │
  ├─────┼───────┼──────────────────────────────────────────────────┼────────────────────────────────────────────┤
  │ 7   │ 439   │ RecoveryManager::processPartitionSyncDataRequest │ mqbblp/mqbblp_recoverymanager.cpp:4483     │
  ├─────┼───────┼──────────────────────────────────────────────────┼────────────────────────────────────────────┤
  │ 8   │ 383   │ RecoveryManager::openRecoveryFileSet             │ mqbc/mqbc_recoverymanager.cpp:941          │
  ├─────┼───────┼──────────────────────────────────────────────────┼────────────────────────────────────────────┤
[...]

PR #1759, opened 2026-09-10
Refactor[mqbs::FileStore]: split rolloverImpl into phase helpers

rmukondiwa:
Splits the long `FileStore::rolloverImpl` (#1727) into private helpers, reducing it from ~325 to ~250 lines:

* Extract `copyOutstandingRecords` for the outstanding-record copy loop
* Extract `logQueueRolloverSummary` (const) for the per-queue summary logging
* Extract `logCompactionMetrics` (const) for the file-size metrics logging

Behavior-preserving; blocks moved as-is with locals passed as parameters. The order-sensitive sync-point write and final file-set swap are left inline. Verified against `mqbs_filestore.t` case 8.

### DIFF
--- a/src/groups/mqb/mqbs/mqbs_filestore.cpp
+++ b/src/groups/mqb/mqbs/mqbs_filestore.cpp
@@ -2820,6 +2820,47 @@ void FileStore::logQueueRolloverSummary(
     BALL_LOG_INFO << outStream.str();
 }
 
+void FileStore::logCompactionMetrics(const FileSet& activeFileSet,
+                                     const FileSet& newActiveFileSet) const
+{
+    // Print some rollover-related metrics.
+
+    bmqu::MemOutStream out;
+    out << partitionDesc() << "Compaction metrics: \n"
+        << "    DATA file size    (new/old): "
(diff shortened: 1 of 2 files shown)


rmukondiwa:
Hi @kaikulimu ,
Opened this for #1727 (rolloverImpl). Would love your feedback when you get a chance.

kaikulimu:
@rmukondiwa Thanks so much for your contribution! I am on vacation right now, but perhaps @678098 or @dorjesinpo can take a look?

src/groups/mqb/mqbs/mqbs_filestore.cpp
```diff
@@ -1088,8 +1088,8 @@ int FileStore::openInRecoveryMode(bsl::ostream&    errorDescription,
                 BALL_LOG_WARN
                     << partitionDesc()
                     << "Conforming journal to cluster state: writing a "
-                    << "corrective QueueOp.DELETION for extra queueKey "
-                    << "[" << *it << "].";
+                    << "corrective QueueOp.DELETION for extra queueKey " << "["
```
678098:
```suggestion
                    << "corrective QueueOp.DELETION for extra queueKey ["
```

PR #1760, opened 2026-09-10
Fix[bmq]: Count the user callback invocation

dorjesinpo:
User callback invocation should be counted as `++d_inProgressEventHandlerCount`.
Otherwise, one event handler thread can call the callback, while another can call `d_stopSemaphore.post()` without waiting for the callback to return.  As a result, the callback may find the session already destroyed.


pniedzielski (approved):
👍🏻

PR #1761, opened 2026-09-11
Fix[mqbnet]: gracefully terminate broker if local ip resolution failed

678098:
Local host ip resolution failure is close to IO/filesystem failures that we handle gracefully by terminating a broker. There is no need to assert and generate a core dump when we can just stop the broker gracefully in this case.

Should get rid of the following crashes:
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f30912b7ee5 in __GI_abort () at abort.c:79
#2  0x0000000000c8b6c5 in BloombergLP::bsls::AssertImpUtil::failByAbort ()
    at /groups/bsl/bsls/bsls_assertimputil.cpp:91
#3  0x0000000000cf3683 in bmqAssertHandler (
    comment=0x2052470 "false && \"Failed to get IP address of the host.\"", 
    file=0x2052020 
```


src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -1233,24 +1212,49 @@ int TCPSessionFactory::start(bsl::ostream& errorDescription)
...
+        return rc_HOSTNAME_FAILURE;  // RETURN
+    }
+
+    ntsa::Ipv4Address defaultIP;
+    error = bmqio::ResolveUtil::getIpAddress(&defaultIP, hostname);
+    if (error.code() != ntsa::Error::e_OK) {
+        errorDescription << d_name << ": failed to get IP address of the host "
+                         << "'" << hostname << "' [error: " << error << "]";
+        return rc_IP_RESOLUTION_FAILURE;  // RETURN
+    }
```
678098, line 1242: Interestingly, `defaultIp` / `hostname` are not even used except for logging here.
We might even try to continue broker start, however, resolution failure tells us that something is probably wrong.
I think that terminating gracefully (not asserting) is safe at least.

src/groups/mqb/mqbnet/mqbnet_tcpsessionfactory.cpp
```diff
@@ -1233,25 +1212,54 @@ int TCPSessionFactory::start(bsl::ostream& errorDescription)
...
+        return (rc * 10) + rc_INVALID_CONFIG;  // RETURN
     }
 
     ntca::InterfaceConfig interfaceConfig = ntcCreateInterfaceConfig(
         *d_config_mp);
 
     bslmt::ThreadAttributes attributes;
     attributes.setThreadName("bmqDNSResolver");
-    rc = d_resolutionContext_sp->start(attributes);
-    BSLS_ASSERT_SAFE(rc == 0);
```
678098, line 1254: `d_resolutionContext_sp` might return non-zero code if it fails to spawn a thread.
We can still terminate the app gracefully because it is a host resources problem, not a logics problem.

hallfox (approved):

PR #1762, opened 2026-09-11
CI: reuse sanitizers common env

678098:
Build a base sanitizers image with the required dependencies and LLVM to avoid CI failures for too many requests.

Changes:
- Extract common packages installation from `build_sanitizers.sh` to `install_common.sh`.
- Build a base sanitizers image once per workflow and reuse it to build specific sanitizer instrumented builds.
- Simplify sanitizers build by using packaged `cmake` and installing only the required llvm packages.
- Start sanitizers CI early.

Avoid random CI failures like
```
#17 361.4 + printf '[error] %s\n' 'GPG key not reachable at https://apt.llvm.org/llvm-snapshot.gpg.key'
#17 361.4 [error] GPG key not reachable at https://apt.llvm.org/llvm-snapshot.gpg.key
#17 361.4 + exit 2
#17 ERROR: process "/bin/sh -c docker/sanitizers/build_sanitizer.sh ${SANITIZER_NAME} ${FUZZER}" did not complete successfully: exit code: 2
------
 > [9/9] RUN docker/sanitizers/build_sanitizer.sh tsan off:
361.4 + check_url https://apt.llvm.org/llvm-snapshot.gpg.key
361.4 + local url=https://apt.llvm.org/llvm-snapshot.gpg.key
361.4 + command -v wget
361.4 + wget -q --method=HEAD --timeout=15 --tries=3 https://apt.llvm.org/llvm-snapshot.gpg.key
361.4 + error 'GPG key not reachable at https://apt.llvm.org/llvm-snapshot.gpg.key' 2
361.4 + local 'msg=GPG key not reachable at https://apt.llvm.org/llvm-snapshot.gpg.key'
361.4 + local code=2
361.4 + printf '[error] %s\n' 'GPG key not reachable at https://apt.llvm.org/llvm-snapshot.gpg.key'
361.4 [error] GPG key not reachable at https://apt.llvm.org/llvm-snapshot.gpg.key
361.4 + exit 2
```


.github/workflows/sanitize.yaml
```diff
@@ -36,15 +55,23 @@ jobs:
...
+      - name: UT [c++,${{ matrix.mode }}]
         if: ${{ matrix.fuzz == 'off' }}
-        run: docker run --rm sanitizer-${{ github.job }}
-      - name: IT [c++,${{ github.job }}]
+        run: docker exec sanitizer /blazingmq/cmake.bld/Linux/run-unittests.sh
+      - name: IT [c++,${{ matrix.mode }}]
         if: ${{ matrix.fuzz == 'off' }}
         run: |
-          docker run --rm sanitizer-${{ github.job }} /blazingmq/cmake.bld/Linux/run-it.sh -k breathing --bmq-tolerate-dirty-shutdown
+          docker exec sanitizer /blazingmq/cmake.bld/Linux/run-it.sh -k breathing --bmq-tolerate-dirty-shutdown
```
678098, line 77: `run/exec/exec` is not graceful because the container is the same, but it works and I am fine with this as long as we don't have CI failures

docker/sanitizers/build_sanitizer.sh
```diff
@@ -199,6 +148,8 @@ if [ "${FUZZER}" == "off" ]; then
             -DCMAKE_CXX_FLAGS="-Wno-reserved-macro-identifier" \
             -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi;libunwind" \
             -DLLVM_USE_SANITIZER="${LLVM_SANITIZER_NAME}" \
+            -DLLVM_INCLUDE_TESTS=OFF \
+            -DLIBCXX_INCLUDE_TESTS=OFF \
```
678098, line 152: These test targets require more dependencies from llvm. So I decided to remove them to install less from llvm.

docker/sanitizers/build_sanitizer.sh
```diff
@@ -57,45 +61,6 @@ fi
...
-    jq \
-    ninja-build \
-    bison \
-    libfl-dev \
-    pkg-config \
-    python3 \
-    python3-venv
-rm -rf /var/lib/apt/lists/*
-
-# Install prerequisites for LLVM: latest cmake version, Ubuntu apt repository contains stale version
```
678098, line 80: We used `ubuntu:24.04` when this script was created, now we use `ubuntu:26.04` and it provides the required cmake version to build everything. No need to install from sources.

.github/workflows/sanitize.yaml
```diff
@@ -36,15 +55,23 @@ jobs:
           sudo docker image prune --all --force
           echo "=== Disk usage after purge ==="
           df -h /
-      - name: Build [${{ github.job }}]
```
678098, line 39: `github.job` is not known in advance and it produces incorrect labels like `Build []`

docker/sanitizers/install_common.sh
```diff
@@ -0,0 +1,94 @@
...
+    make \
+    ninja-build \
+    bison \
+    libfl-dev \
+    pkg-config \
+    python3 \
+    python3-venv
+rm -rf /var/lib/apt/lists/*
+
+# Install LLVM: clang, lld, lldb and clangd
```
pniedzielski: Could we clarify this comment?  llvm.sh looks like it sets up an apt source as well, so llvm-${LLVM_VERSION} and libclang-rt-${LLVM_VERSION}-dev below are presumably coming from there too.  Wasn't clear from reading this file, had to dive into llvm.sh to figure that out.

.github/workflows/sanitize.yaml
```diff
@@ -17,13 +40,9 @@ jobs:
           - mode: "asan"
             fuzz: "on"
       fail-fast: false
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-latest
```
pniedzielski: should we pin to `ubuntu-26.04`?

pniedzielski (changes requested):
Minor change for the comment, otherwise very clear.

.github/workflows/sanitize.yaml
678098: `latest` might eventually break our builds on ubuntu LTS update. Maybe it is a good thing because we will have motivation to update the build.
Don't have a strong opinion, 26.04 is fine too.

docker/sanitizers/install_common.sh
678098: These are the last lines of llvm.sh
```
apt-get update
PKG="clang-$LLVM_VERSION lldb-$LLVM_VERSION lld-$LLVM_VERSION clangd-$LLVM_VERSION"
if [[ $ALL -eq 1 ]]; then
    # same as in test-install.sh
    # No worries if we have dups
    PKG="$PKG clang-tidy-$LLVM_VERSION clang-format-$LLVM_VERSION clang-tools-$LLVM_VERSION llvm-$LLVM_VERSION-dev lld-$LLVM_VERSION lldb-$LLVM_VERSION llvm-$LLVM_VERSION-tools libomp-$LLVM_VERSION-dev libc++-$LLVM_VERSION-dev libc++abi-$LLVM_VERSION-dev libclang-common-$LLVM_VERSION-dev libclang-$LLVM_VERSION-dev libclang-cpp$LLVM_VERSION-dev liblldb-$LLVM_VERSION-dev libunwind-$LLVM_VERSION-dev"
    if test $LLVM_VERSION -gt 14; then
        PKG="$PKG libclang-rt-$LLVM_VERSION-dev libpolly-$LLVM_VERSION-dev"
    fi
fi
apt-get install -y $PKG
```

We don't pass `all` flag so we only install `clang`, `lldb`, `lld`, `clangd`.
But this is not enough so we have to install extras `llvm` and `libclang-rt`.
`llvm.sh` only installs these extra packages if `all` is passed.

678098: Updated comments

pniedzielski (dismissed):

pniedzielski (approved):

hallfox (approved):

PR #1765, opened 2026-09-15
Fix[mqb]: improve non-printable handling in broker negotiation/session logs

chrisbeard:
This is a minor refactor of previous work that also restricts broker negotiation messages to printable characters in logs.

Follows from: https://github.com/bloomberg/blazingmq/pull/1675


src/groups/bmq/bmqu/bmqu_stringutil.cpp
```diff
@@ -92,6 +92,22 @@ bool StringUtil::isPrintable(const bslstl::StringRef& str)
     return true;
 }
 
+bsl::string& StringUtil::sanitize(bsl::string* str)
```
678098, line 95: What concerns me is that the API allows to pass pointers to strings that are not alive when we try to use the returned reference. For example:
```
bsl::cout << StringUtil::sanitize(&TempObject().name()) << bsl::endl;
```

I think a good way to deal with this is to use a stream manipulator.
For example, we have `PrintUtil::PrettyIntManipulator::PrettyIntManipulator` and similar and we define stream functions:
```
/// Print the specified `manipulator` into the specified output `stream` and
/// return `stream`.
bsl::ostream& operator<<(bsl::ostream&               stream,
                         const PrettyIntManipulator& manipulator);
```

The manipulator itself is not enough to ensure API safety, and we will have to make a string copy on constructing it.

src/groups/bmq/bmqu/bmqu_stringutil.t.cpp
```diff
@@ -480,6 +480,66 @@ static void test8_isPrintable()
...
+        int         d_line;
+        const char* d_input;
+        const char* d_expected;
+    } k_DATA[] = {{L_, "", ""},
+                  {L_, "hello world", "hello world"},
+                  {L_, " ~", " ~"},  // first and last printable
+                  {L_, "abc\ndef", "abc?def"},
+                  {L_, "\r\t\x1f", "???"},
+                  {L_, "\x7f", "?"},          // DEL
+                  {L_, "a\x80\xff", "a??"}};  // high-bit bytes
```
678098: I think we can afford checking the full byte range
https://facweb.cs.depaul.edu/sjost/it212/documents/ascii-pr.htm

What about making a string on length 256 x 2 (or even 256 x 3), filling it with sequential values.

```
std::string makeAllBytes()
{
    std::string result; 
    result.reserve(256);
    for (int i = 0; i < 256; ++i) {
        result.push_back(static_cast<char>(i));
    }
    return result;
}

const bsl::string k_SAFE_BYTES =
    "????????????????"                    // 0x00
    "????????????????"                    // 0x10
    " !\"#$%&'()*+,-./"                   // 0x20
    "0123456789:;<=>?"                    // 0x30
    "@ABCDEFGHIJKLMNO"                    // 0x40
    "PQRSTUVWXYZ[\\]^_"                   // 0x50
    "`abcdefghijklmno"                    // 0x60
    "pqrstuvwxyz{|}~?"                    // 0x70
    "????????????????"                    // 0x80
    "????????????????"                    // 0x90
    "????????????????"                    // 0xA0
    "????????????????"                    // 0xB0
    "????????????????"                    // 0xC0
    "????????????????"                    // 0xD0
    "????????????????"                    // 0xE0
    "????????????????";                   // 0xF0

bsl::string unsafe = makeAllBytes() + makeAllBytes();
bsl::string expected = k_SAFE_BYTES + k_SAFE_BYTES;
```

src/groups/bmq/bmqu/bmqu_stringutil.cpp
chrisbeard: This API matches the existing convention of `trim` and friends in this class. I debated this for a bit, but decided I'm fine with it as-is, these aren't used heavily.

678098 (approved):
Looks good

PR #1778, opened 2026-09-21
IT[test_admin_client]: fix flaky too-often-stats

678098:


src/integration-tests/test_admin_client.py
```diff
@@ -413,14 +417,6 @@ def test_app_id_stats(multi_node: Cluster, domain_urls: tc.DomainUrls) -> None:
     consumer_bar.close(f"{task.uri}?id=bar")
     consumer_baz.close(f"{task.uri}?id=baz")
 
-    # Stage 4: check too-often stats safeguard
-    for i in range(5):
-        admin.send_admin("encoding json_pretty stat show")
-    res = admin.send_admin("encoding json_pretty stat show")
-    obj = json.loads(res)
-
-    expect_same_structure(obj, dt.TEST_QUEUE_STATS_TOO_OFTEN_SNAPSHOTS, "too-often")
```
678098, line 422: No need to test the same code path in app id stats. It was another source of flakiness. Only check it for a single node to make sure it is fast.

src/integration-tests/test_admin_client.py
```diff
@@ -270,12 +270,21 @@ def test_queue_stats(single_node: Cluster, domain_urls: tc.DomainUrls) -> None:
     consumer_baz.close(f"{task.uri}?id=baz")
 
     # Stage 4: check too-often stats safeguard
-    for _ in range(5):
-        admin.send_admin("encoding json_pretty stat show")
```
678098, line 274: This test is flaky in this scenario:
4 initial stat requests are valid
5th request (the last one in this loop) fails with rejection

then we are lucky that a leaky bucket refills

6th request is valid (and we fail on assert)

chrisbeard (approved):

